PR #76, opened 2013-08-31
Add support for a user's guide

### LINKED ISSUE #75: Adding user's guides
We'd like to be able to add user's guides to our documentation, like Erlang has.

The current plan is to allow passing a list of markdown files to ex_doc which will be used as the user's guide. Directories can be passed as well, those are expanded to the files in them (alphabetically).

pminten:
Pass -g/--guide args with files/dirs on the command line or guide_files to the
compile_docs functions. Files must be in markdown and if a dir is passed the
files in it will be put in the documentation alphabetically, but only for files
ending in ".md"

Fixes #75


milmazz:
This idea is great! :+1:

@pminten @josevalim Do you need some help with this? Just let me know if there some way I can help.

josevalim:
I would wait for the new layout. Then we can discuss how we can add guides to the menu and how things will look like as it is the trickiest part.  :)

PR #177, opened 2015-01-03
use <input type="search" /> in left sidebar

eksperimental:
I find input type=search a better option for the job.
It gives a wider search area, and in Chrome, you can reset the box clicking on the X,

See how it looks in
***CHROME***
**Empty**
search_box_chrome

**Loading:**
search_box_chrome_loading

**Loaded:**
search_box_chrome_loaded

***FIREFOX***
**Empty:**
search_ff

**Loading:**
search_box_ff_loading

**Loaded:**
search_box_ff_loaded

My only concern is that erlier version of IE will not recognize "input type=search", but it will degrade to input type=text,

By the way what is the lowest IE browser we want to support with ex_doc,
same goes for Chrome and Firefox


josevalim:
There is no minimum version, but we can expect developers to be using recent ones.

Issue #173, opened 2014-12-19
Always provide uncompressed/original version of libraries, to facilitate debugging.


eksperimental:
_I came across this issue trying to debug #172_
It's kind of hard to know what compressed into, specially when it's a customized version of the library, like in the case of highlight-js. So I think it's advisable to also have the uncompressed version. Same should go for JQuery.

Also, in the case of taylored versions, a comment explaining what options were included in would be greatly appreciated.

whatyouhide:
The jQuery version at https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc/formatter/html/templates/js/jquery.js already has a version tag at the top of the file. As for highlight.js, I have no idea what version https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc/formatter/html/templates/js/highlight.pack.js is. If I get that, I'll be happy to send a quick PR for this!

eksperimental:
hi @whatyouhide 
I had packed already a new version that supports a few more formats that will be common in elixir projects.
I will probably update this tomorrow as I will have some free time, and also submit an issue to hightlighthjs to include a comment at the top of the file and a link to license.

eksperimental:
we can close this one, if it's all good with the new changes.

PR #182, opened 2015-01-14
update highlightjs to support more languages (v8.4)  + add header info

eksperimental:
solves #173 
supported languages that I found will be more likely to be on an elixir project:
* Bash
* CoffeeScript
* CSS
* Diff
* Elixir
* Erlang
* Erlang-REPL
* HTML/XML
* HTTP
* JavaScript
* JSON
* Markdown
* SQL


josevalim:
I had to revert because adding more languages caused Elixir to no longer be recognised properly. Many times it was tagged as Ruby or even SQL. :(

eksperimental:
then we do need to add the html class elixir. I will submit a fix soon, as well as the issue with iex code being detected as erlang.

josevalim:
If we are adding the html class though, then everything will be marked as Elixir, even those that are not, so it is not better than not shipping with them. We would need to support other languages in our markdown markup first.

PR #184, opened 2015-01-30
Fix the HTML escaping broken by 5b520c4

whatyouhide:
Following the discussion here: https://github.com/elixir-lang/ex_doc/commit/5b520c4cad67c3a4b55042ca588ecfb5bea0b41f#commitcomment-9521480.

This PR reverts a change introduced in 5b520c4 which made a lot of `ex_doc` output look like this:

```
iex\> ...
```

### DIFF
--- a/lib/ex_doc/formatter/html/templates.ex
+++ b/lib/ex_doc/formatter/html/templates.ex
@@ -69,7 +69,12 @@ defmodule ExDoc.Formatter.HTML.Templates do
   defp presence(other), do: other
 
   defp h(binary) do
-    escape_map = [{"&", "&amp;"}, {"<", "&lt;"}, {">", "&gt;"}, {"\"", "&quot;"}]
+    escape_map = [
+      {"&", "\\&amp;"},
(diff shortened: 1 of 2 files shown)


josevalim:
Can we add a test please? :)

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Lead Developer

eksperimental:
I think the test is within Elixir, it has nothing to do with ex_doc, becase ex_doc master is running fine with elixir v1.0.2,
even when at this stage a week ago: the commit bf587d408568dff24402638dd25b6b528c24fa1a 
exdoc is running fine. i'm trying to run `git bisect` to track down the problem

josevalim:
Ok, i found out the reason. Elixir fixed a bug but earmark is relying on the previous behaviour. I am sending patches to earmark and reverting the change from v1.0 branch. Thank you.

whatyouhide:
Great, thank you!

tony612:
@josevalim Great job! :+1:

Issue #172, opened 2014-12-19
hightlight-js is inconsistent identifying IEX and Elixir code


eksperimental:
/cc @tonini
hightlight-js is inconsistent identifying IEX and Elixir code. Sometimes it identifies it as elixir, and other times as erlang when it's an IEX snippet.

Output from my browser inspector in Kernel.SpecialForms.html
    <code class="hljs elixir">
      iex&gt; map = %{<span class="hljs-symbol">a:</span> <span class="hljs-symbol">:b</span>}
      iex&gt; map.a
      <span class="hljs-symbol">:b</span>
    </code>
and others erroneously as erlang

    <pre><code class="hljs erlang"><span class="hljs-comment">%User{}</span></code></pre>

or this IEX sample code:
    <pre><code class="hljs erlang">iex&gt; <span class="hljs-keyword">fun</span> = &amp;<span class="hljs-variable">Kernel</span>.is_atom/<span class="hljs-number">1</span>
    iex&gt; <span class="hljs-keyword">fun</span>.(:atom)
    <span class="hljs-literal">true</span>
    iex&gt; <span class="hljs-keyword">fun</span>.(<span class="hljs-string">"string"</span>)
    <span class="hljs-literal">false</span></code></pre>

eksperimental:
true!, there are lexers for all of them
http://pygments.org/docs/lexers/#lexers-for-erlang

I will update this soon

PR #185, opened 2015-01-31
update highlightjs to support more languages

eksperimental:
highlight.js has been updated
* hightlight.js updated to latest version (v8.4)
* Highlighting support for all these languages: Bash, CoffeeScript, CSS, Diff, Elixir, Erlang, Erlang-REPL, HTML/XML, HTTP, JavaScript, JSON, Markdown, SQL.
* Header added to the hightligh.js library, showing the options were used to compile it, including license, copyright notice, and the date it was downloaded  (solving issue #173)
*  highlighting autodetect has been disabled
* any language in the code blocks can be defined, even though the ones not supported, but when no language is specified, the default is elixir (this solves the issue why this previous PR had to be reverted:  https://github.com/elixir-lang/ex_doc/pull/182#issuecomment-70820109 )
* IEX code is detected and a class is added ( solving https://github.com/elixir-lang/ex_doc/issues/172#issuecomment-67610083 )
* Dependencies have been updated

Pending Improvements:
1. Elixir markup have some issues, It needs to be reported to the maintainers. I have used [this test file](https://bitbucket.org/birkenfeld/pygments-main/src/tip/tests/examplefiles/example_elixir.ex)
2. We are lacking Iex support, if anybody wants to tackle down on it. [There is already support for it in pygments](https://bitbucket.org/birkenfeld/pygments-main/commits/45af979829ad0bad180224ceb6d30f13d0d2bc6e), I'm not sure if that helps when porting it to highlight.js
3. Support for markdown is very poor and there is not much highlighting
4. I have added to the header of the file html files the languages supported, and their alias, just because I didn't know where to put it in the ex_doc documentation, so if anybody knows a good place, we can add it, and provide examples of how to include code from other languages.
```
    // Languages supported and their aliases
    bash, sh, zsh,
    coffeescript, coffee, cson, iced,
    css,
    diff, patch,
    elixir, iex,
    erlang, erl,
    erlang-repl,
    http,
    javascript, js,
    json,
    markdown, md, mkdown, mkd,
    sql,
    xml, html, xhtml, rss, atom, xsl, plist,
```


eksperimental:
Pending: I still need to add test for set_codeblock_language/1

josevalim:
Thank you but let's not support other languages yet. :) People complained a lot about invalid markup the last time. If people want custom markup, they can inject their own highlight.js versoin.

eksperimental:
Sorry about the last bug Jose, I understand your concerns and how bad code might have been looking.  But I think with the new changes, not including these languages doesn't solve anything.
The root mistake here was to let the javascript library detect the languages by itself. In this PR, if the codeblock language is missing, the class is set in html as "elixir". Alternatively, if they language specified is not a supported one, highlight.js doesn't try to autodetect it, meaning, that it won't be formatted at all.

You can see the results in this test page, an the different test cases (no highlight, no language, language not supported):
http://eksperimental.github.io/elixir/highlight_sample/FooBar2.html
and the .ex file
https://github.com/eksperimental/eksperimental.github.io/blob/master/highlight_sample/foobar.ex

Please let me know what you think, so I can make the necessary changes if you still consider removing language support is the best option.
Thanks

josevalim:
Ok. I see, thanks! I think we should move this code somewhere else so we can test it easily. Also, I don't think we need to specially handle erlang or html/xml. :)

eksperimental:
I thought would be common, but .. it wouldn't be common to have erlang code in elixir (sorry maybe it is a silly question), 
or in projects developed with phoenix to include html code in the source, or maybe dealing with templating functions..

I was thinking of putting that code somewhere, but I have no idea where.
Should it go into a fixture, or where do you think it is the best place.

josevalim:
I think we could include it the ExDoc.Markdown.to_html function directly. What do you think?

eksperimental:
Sorry, i think i misundertood completely your previous two messages. that's why the random answer.
I moved it into `ExDoc.Markdown.to_html`,
now i see what you meant by removing erlang and xhtml
I think we are good to go.

josevalim:
Thank you! We just need a test now!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

lib/ex_doc/formatter/html/templates/module_template.eex
```diff
@@ -13,7 +13,27 @@
     <script type="text/javascript" charset="utf-8" src="js/jquery.js"></script>
     <script type="text/javascript" charset="utf-8" src="js/app.js"></script>
     <script type="text/javascript" charset="utf-8" src="js/highlight.pack.js"></script>
-    <script>hljs.initHighlightingOnLoad();</script>
+    <script type="text/javascript" charset="utf-8">
+      hljs.initHighlightingOnLoad();
+      // languages supported: 'bash', 'coffeescript', 'css', 'diff', 'elixir', 'erlang', 'erlang-repl', 'http', 'javascript', 'json', 'markdown', 'sql', 'xml'
```
josevalim: We don't need to include the languages once again here and down below since we include them in the highlight.js script!

eksperimental: It's because i don't know where to put the information about the aliases.
We should add it to the documentation, or README.md, ... I 'm clueless where to add it to

      /* languages supported and their aliases
        bash, sh, zsh,
        coffeescript, coffee, cson, iced,
        css,
        diff, patch,
        elixir, iex,
        erlang, erl,
        erlang-repl,
        http,
        javascript, js,
        json,
        markdown, md, mkdown, mkd,
        sql,
        xml, html, xhtml, rss, atom, xsl, plist,
      */

josevalim: We could add to the README. Actually, a section about syntax highlighting and mentioning that we use highlight.js would be really welcome!

eksperimental:
yes. please test as much as possible
update docs are here: http://eksperimental.github.io/elixir/doc/elixir/ 

@tonini @whatyouhide @knewter

whatyouhide:
Everything looks nice at first sight. I got all excited thinking this was a fix to isagalaev/highlight.js#730 or isagalaev/highlight.js#707 :disappointed: Boy we need those fixes! :)

josevalim:
Oops, sorry. This conflicted with the other pull request from you I just merged. Please rebase this one and I will get it in right away!

eksperimental:
there you go

josevalim:

eksperimental:
Thank you José!

PR #188, opened 2015-02-03
Link to pages and load frame

eksperimental:
I noticed in the elixir website, when trying to link to a page in the docs, the frames are missing.
well, this solves this issue.

Let's say you want to link to: 
http://eksperimental.github.io/elixir/doc/Kernel.SpecialForms.html

then if you want to load the frames you link to:
http://eksperimental.github.io/elixir/doc//#!Kernel.SpecialForms.html
or even to special function in particular
http://eksperimental.github.io/elixir/doc/#!Kernel.SpecialForms.html#__DIR__/0

Let's say you are running the file in your computer, so you have to call `index.html`
http://eksperimental.github.io/elixir/doc/index.html#!Kernel.SpecialForms.html#__DIR__/0
it does still work.

Too bad I cannot update the url when you click, (so you can directly copy the url from the address bar) this is due to security restrictions dealing with frames... hopefully when we get rid of frames, we won't need this anymore.

### DIFF
--- a/lib/ex_doc/formatter/html/templates/index_template.eex
+++ b/lib/ex_doc/formatter/html/templates/index_template.eex
@@ -4,9 +4,18 @@
 <head>
 	<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
 	<title><%= config.project %> v<%= config.version %> Documentation</title>
+	<script type="text/javascript" charset="utf-8">
+		window.onload = function frame_redirect() {
+			if ( window.location.href.match(/\/#!/, window.location.hash)  ) {
+				document.querySelector('#main').src = window.location.href.replace(/\/#!/, '/');
+			} else if ( window.location.href.match(/\/<%= config.main %>.html#!/, window.location.hash)  ) {
+				document.querySelector('#main').src = window.location.href.replace(/\/<%= config.main %>.html#!/, '/');
+			}
+		}
+	</script>	
 </head>
 <frameset cols="20%,*">
-  <frame name="list" src="modules_list.html" />
-  <frame name="main" src="<%= config.main %>.html" />
+  <frame name="list" id="list" src="modules_list.html" />
+  <frame name="main" id="main" src="<%= config.main %>.html" />
 </frameset>
 </html>
--- a/mix.lock
+++ b/mix.lock
@@ -1,3 +1,3 @@
-%{"earmark": {:package, "0.1.10"},
+%{"earmark": {:hex, :earmark, "0.1.10"},
   "hoedown": {:git, "git://github.com/hoedown/hoedown.git", "0610117f44b173a4e2112afb2f510156a32355b5", []},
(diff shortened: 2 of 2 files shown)


lib/ex_doc/formatter/html/templates/index_template.eex
```diff
@@ -4,9 +4,18 @@
 <head>
 	<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
 	<title><%= config.project %> v<%= config.version %> Documentation</title>
+	<script type="text/javascript" charset="utf-8">
```
josevalim: This is awesome! Should we embed it in the page or can we add it to any of the javascript files?

josevalim:
> Too bad I cannot update the url when you click, (so you can directly copy the url from the address bar) this is due to security restrictions dealing with frames... hopefully when we get rid of frames, we won't need this anymore.

Should we merge this then given we plan to get rid of frames? Because we would have to change links... or do you plan to support this URL schema too once we remove frames?

lib/ex_doc/formatter/html/templates/index_template.eex
eksperimental: It only works with the top frame (ie. index.html)

eksperimental:
I think it will be pretty straightforward to support it, but if we find some incompatibilities, then  we'd better wait.

right now we have this situation:

Calling index.html and loading this, in the main frame.
```
/docs/#!Kernel.SpecialForms.html#__DIR__/0   --> /docs/Kernel.SpecialForms.html#__DIR__/0
or 
/docs/index.html#!Kernel.SpecialForms.html#__DIR__/0   --> /docs/Kernel.SpecialForms.html#__DIR__/0
```

and once we get rid of frames, i'm pretty sure we will have this one:
Calling index, and we will redirect directly (no frames) to the same page we are redirecting now.
```
/docs/#!Kernel.SpecialForms.html#__DIR__/0   --> /docs/Kernel.SpecialForms.html#__DIR__/0
or 
/docs/index.html#!Kernel.SpecialForms.html#__DIR__/0   --> /docs/Kernel.SpecialForms.html#__DIR__/0
```
the difference is the target, 

what do you think?

josevalim:
Perfect, great work!

josevalim:

eksperimental:
sweet!

eksperimental:
@josevalim please let me know when elixir-lang.org/docs/ will reflect these changes, so I can make the proper changes to elixir-lang.github.io

josevalim:
likely when v1.0.3 is released which will be at maximum by the beginning of next week!

eksperimental:
:+1:

eksperimental:
hi @josevalim,
I checked the docs, and I see it working
http://elixir-lang.org/docs/master/elixir/#!Exception.html#message/1

so I corrected all the link in elixir-lang.org
but.. then I realized, this new feature is reflected only on master (which is btw not linked from anywhere in the site), and not in **stable** and **v1.0** have not been created with the new version
http://elixir-lang.org/docs/stable/elixir/#!Exception.html#message/1
http://elixir-lang.org/docs/v1.0/elixir/#!Exception.html#message/1

Issue #181, opened 2015-01-11
Issues with functions that use the `do:` keyword


whatyouhide:
`ex_doc` messes whitespace up when functions are defined with

```elixir
def f(args), do: # code
```

instead of

```elixir
def f(args) do
  # code
end
```

Here's a minimal working example for which this issue comes up:

```elixir
defmodule ExDocIssue do
  @moduledoc """
      defmodule Foo do
        def foo, do: :foo
        def bar, do: :bar
      end
  """
end
```

The generated output:

![screen shot 2015-01-11 at 21 48 10](https://cloud.githubusercontent.com/assets/3890250/5696756/8d534e2a-99db-11e4-86d9-bb4b2b0e3451.png)

I noticed this happens in quite a few stdlib modules too.

I generated this output with:

```bash
ex_doc ExDocIssue 0.0.1 _build/...
```

but it behaves the same when generating docs with `mix docs`. Thanks so much!

josevalim:
It may be an issue of the syntax highlighter. Have you checked if the generated source is correct or what happens when you remove the highlighter command directly from source before loading the page?

whatyouhide:
The HTML source is kind of weird actually, I think it nests a `<span>` into another `<span>`, which causes some troubles.

The `.html`-only source in `ExDocIssue.html` seems to be fine:

```html
        <section id="moduledoc" class="docstring">
          <pre><code>defmodule Foo do
  def foo, do: :foo
  def bar, do: :bar
end</code></pre>
```

so it's probably the highlighter. You're using highlight.js right?

josevalim:
Yes.

whatyouhide:
Ok, I can confirm it's an highlight.js issue. For further reference, this HTML code can reproduce the bug:

```html
<html>
  <head>
    <link rel="stylesheet" href="highlight/styles/default.css">
    <script src="highlight/highlight.pack.js"></script>
    <script>hljs.initHighlightingOnLoad();</script>
  </head>

  <body>
      <pre><code class="elixir">defmodule Foo do
  def foo, do: :foo
  def bar, do: :bar
end</code></pre>
  </body>
</html>
```

I'm deeply sorry I wasted your time by pointing at the wrong repository! :disappointed:

josevalim:
It is completely ok. If it doesn't  work , it is an ex_doc bug too.

-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Lead Developer*

[...]

PR #208, opened 2015-04-17
Update highlight to v8.5

lidashuang:
Single expression functions def foo, do: ... now work in Elixir.

#181


josevalim:
!!!!!!!!!!!!!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

eksperimental:
great, just to mention that support for Java was included, but i'm wondering if it's relevant to have include that by default

whatyouhide:
Whoohooo, this is GREAT! :D

PR #213, opened 2015-05-23
Fixed HTML5 markup validation errors.

milmazz:
This PR include some improvements related with the markup, according to the specification for HTML5:

* The `type` attribute for the `script` element by default is `text/javascript` and now is optional.
* The `type` attribute for the `style` element by default is `text/css`, also the `charset` parameter must not be specified.
* Exactly one of the `name`, `http-equiv`, and `charset` attributes must be specified in the `meta` element.
* If either `name` or `http-equiv` is specified, then the `content` attribute must also be specified. Otherwise, it must be omitted.

I also put the scripts at the bottom of the page to avoid blocking parallel downloads, this might improve the overall rendering time. Finally, jQuery now can be obtained from a _Content Delivery Network_, if this fails we can obtain the local version.


josevalim:

PR #216, opened 2015-05-27
Delete output directory before generate docs.

milmazz:
Reference: elixir-lang/ex_doc#215

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -11,6 +11,7 @@ defmodule ExDoc.Formatter.HTML do
   """
   def run(modules, config)  do
     output = Path.expand(config.output)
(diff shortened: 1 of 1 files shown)


lib/ex_doc/formatter/html.ex
```diff
@@ -11,6 +11,7 @@ defmodule ExDoc.Formatter.HTML do
   """
   def run(modules, config)  do
     output = Path.expand(config.output)
+    {:ok, _} = File.rm_rf output
```
josevalim: I would use `File.rm_rf!(output)`. If it fails we will get a better reason than badmatch. :)

milmazz: Thanks for your feedback @josevalim

josevalim:

Issue #214, opened 2015-05-26
function `pretty_codeblock` leaves an end tag for `code` element which is not open


milmazz:
The current [`pretty_codeblock`](https://github.com/elixir-lang/ex_doc/blob/7c7365c18f3763482695bf3440fd2c10baa7bc7f/lib/ex_doc/formatter/html.ex#L80) function in `html.ex` generates invalid HTML code, this function leaves an end tag for the `code` element which is not open.

```elixir
iex> bin = "<pre><code>Hello world</code></pre><pre><code class=\"\">Hello world</code></pre>"
"<pre><code>Hello world</code></pre><pre><code class=\"\">Hello world</code></pre>"
iex> Regex.replace(~r/<pre><code\s*(class=\"\")?>/, bin, "<pre class=\"codeblock\">")
"<pre class=\"codeblock\">Hello world</code></pre><pre class=\"codeblock\">Hello world</code></pre>"
```

One way to solve this is updating the *regex*:

```elixir
iex> Regex.replace(~r/<pre><code\s*(class=\"\")?>(.+?)<\/code><\/pre>/, bin, "<pre class=\"codeblock\">\\2</pre>")
"<pre class=\"codeblock\">Hello world</pre><pre class=\"codeblock\">Hello world</pre>"
```

But, if we look further, right now there is no rule for the `.codeblock` CSS class. So, I suggest to delete this `pretty_codeblock` and try to control the look of the `<pre>` and `<code>` tags in the CSS files, because there is where the rules that are in control of the style of HTML blocks belongs.

But, I really want to know your opinion about this matter before I propose a Pull Request.

milmazz:
Hello @tonini, I really want to know your opinion about this issue. Do you remember the [`pretty_codeblock`](https://github.com/elixir-lang/ex_doc/commit/7c7365c18f3763482695bf3440fd2c10baa7bc7f#diff-fc9775865b82e656523d7cbbe16526d2) function purpose?

josevalim:
@milmazz the idea about `pretty_codeblocks` is that "\`\`\`" by defaults adds `<code class="">` and that is not going to be formatted as Elixir. Now that we are enabling "\`\`\`" everywhere, we should fix it by somehow setting the default language for such code blocks to Elixir (it can be with a regex) or by changing highlight.js to use elixir by default when another language was not specified.

Does it make sense?

milmazz:
@josevalim Got it!, I'll try to work on this.

[...]

PR #217, opened 2015-05-29
Handle code blocks with and without language spec

milmazz:
This commit includes the following:

 * Improve `pretty_codeblocks` function to work with Hoedown and Pandoc
 * Add more tests for `pretty_codeblocks`
 * Delete duplicate code in `ExDoc.Markdown.to_html`

This is a possible fix for #214

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -75,11 +75,17 @@ defmodule ExDoc.Formatter.HTML do
   end
 
   @doc false
-  # Helper to handle plain code blocks (```...```) without
+  # Helper to handle plain code blocks (```...```) with and without
   # language specification and indentation code blocks
   def pretty_codeblocks(bin) do
-    Regex.replace(~r/<pre><code\s*(class=\"\")?>/,
-                  bin, "<pre class=\"codeblock\">")
+    bin = Regex.replace(~r/<pre><code(\s+class=\"\")?>\s*iex&gt;/,
+                        # Add "elixir" class for now, until we have support for
+                        # "iex" in highlight.js
+                        bin, "<pre><code class=\"iex elixir\">iex&gt;")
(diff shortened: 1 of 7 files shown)


lib/ex_doc/formatter/html.ex
```diff
@@ -75,11 +75,28 @@ defmodule ExDoc.Formatter.HTML do
...
-                  bin, "<pre class=\"codeblock\">")
+    bin = Regex.replace(~r/<pre><code(\s+class=\"\")?>\s*iex&gt;/,
+                        # Add "elixir" class for now, until we have support for
+                        # "iex" in highlight.js
+                        bin, "<pre><code class=\"iex elixir\">iex&gt;")
+    bin = Regex.replace(~r/<pre><code(\s+class=\"\")?>/,
+                        bin, "<pre><code class=\"elixir\">")
+
+    case Application.fetch_env(:ex_doc, :markdown_processor) do
+      {:ok, ExDoc.Markdown.Hoedown} ->
```
josevalim: I think we should move this inside each processor, no? :)

josevalim:
This is perfect. I think we should just move the logic specific to each markdown processor inside the markdown processor module and we will be good to go. :)

lib/ex_doc/formatter/html.ex
milmazz: Oh, you're definitely right, that makes more sense. I'll work on that! Thanks for your feedback :+1:

milmazz:
@josevalim Do I need to do something else to complete this Pull Request? are you too busy right now to review this?

I'm not in a hurry, I'm just curious, that's all! :+1:

josevalim:
@milmazz that's ok. please do ping me after changes :) otherwise I am not notified.

josevalim:

josevalim:
@milmazz ex doc needs love so your work is very appreciated, thank you!

milmazz:
@josevalim Thanks! :smiley:

Issue #224, opened 2015-06-11
New ExDoc version is not functional on mobile


josevalim:
That's because the sidebar takes all the space. @milmazz, thoughts?

milmazz:
@josevalim I can add certain CSS media queries to solve this inconvenience.
The first thought that comes to my mind is that we can offer something like
a "hamburger button" to toggle the sidebar, do you agree in something like
this?

Dignifiedquire:
How about using this for the navigation http://mmenu.frebsite.nl/ and setting it to be always open on desktop.

Dignifiedquire:
Here is a demo of what I mean: http://mmenu.frebsite.nl/demo/index.html?demo=widescreen

josevalim:
That sounds great! We don't want to take your attention away from the
redesign though. :)

-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Lead Developer*

Dignifiedquire:
Maybe I can squeeze that in, that way we don't have to change the menu twice ;)

milmazz:
@Dignifiedquire That plugin is really nice, please let me know if you
can/want to work on this. I have some free time in just 30 minutes, so, I
can work on this issue if you prefer to focus on the new design.

Dignifiedquire:
Sure go ahead, probably won't have a lot of  time today myself.

milmazz:
@Dignifiedquire @josevalim Guys, I must be honest, I prefer the current search field and its current behaviour. The search field offered by [mmenu](http://mmenu.frebsite.nl/) plugin does not substitute the current features (e.g. highlight, search find inner nodes) that we offer in the sidebar. Also, I think that `mmenu` plugin adds more complexity when the user wants to navigate through the nodes (modules, exceptions, protocols).

 Please, see the following example: http://output.jsbin.com/fojodo/2 (please click in the upper-left link)

I'll try to find another way and I'll share with you the results, from there, we can compare both solutions and take a decision. Please let me know what you think.

josevalim:
@milmazz good points. I don't think we need to use their whole menu though. Maybe we should just copy the hamburger menu functionality?

@lucasmazza any ideas on how we could improve this? The latest versions of the docs can be found here: http://elixir-lang.org/docs/master/elixir/

Dignifiedquire:
[...]

PR #226, opened 2015-06-13
WIP: off-canvas navigation for ExDoc

milmazz:
This PR includes the following:

* A way to offer an off-canvas navigation for mobile devices using [Slidebars](http://plugins.adchsm.me/slidebars/)
* Get rid of the absolute positioning of the sidebar and content elements

@josevalim @Dignifiedquire @lucasmazza this is a possible fix for #224, comments are more than welcome :smiley:


milmazz:
Please, put this PR in hold on, I'll try to drop `Slidebars` as dependency, instead I'll use Bootstrap as @Dignifiedquire suggested.

Dignifiedquire:
I just looked and the idea is there, but I'm seeing a lot of flickering and sudden collapsing/vanishing of the menu and its background.

josevalim:
@milmazz is this good to go?

milmazz:
@josevalim Finally got some free time to work on this issue, please review it and let me know your opinion.

As a side note, I made a few tests on: Chrome (Desktop & Android), Safari, Firefox and seems to be working correctly :smiley:

milmazz:
@eksperimental @josevalim @ericmj Can you please review this PR? If this PR gets merged into the master branch I can continue trying to fix the new issues that some of you reported. Also @Dignifiedquire can continue with his design proposal.

eksperimental:
I'm not too sure if you are inheriting this from SlideBars but element ids and classes in ex_doc use underscores as a joiner, not hyphens. So "row-offcanvas-left" should be "row_offcanvas_left"

josevalim:
@milmazz I was doing it right now. :) There is a minor annoyance that the stylesheets for search and sidebar are a bit off. Here is a screenshot:

screen shot 2015-06-21 at 18 47 00

There is more space than it should both on the top and left side and the search button is no longer stylised.

Also this solution is bringing the whole bootstrap css. :( Do we have alternatives?

Dignifiedquire:
> Also this solution is bringing the whole bootstrap css. :( Do we have alternatives?

It would be much easier using parts of bootstrap if we would use less as a css preprocessor, that way we could pull in only the needed files without having to rely on copy paste of parts of it. (See https://github.com/twbs/bootstrap/tree/master/less) But that would introduce the dependency of having to run a less compiler in here. Which I would very much welcome as it makes writing styles a lot easier, but I don't know how you feel about adding something like that.

josevalim:
@Dignifiedquire i am ok with that if that is going to make your job easier. We can have the less source files and then have a command that compiles them. We will always ship with the precompiled css files and, everytime you update the less files, you just need to precompile them again. /cc @ericmj are you ok too? To be clear, only folks editing the css would need less.

josevalim:
@Dignifiedquire in any case, for now, the ship has sailed, I would rather just to try handroll the button styles if possible.

milmazz:
@eksperimental No, that class is used with Bootstrap, but I'll change the name to maintain consistency as you suggested.

@josevalim Let me review that, precisely I'm working right now in the search field, @eksperimental suggested that the _clear button_ is useful, So, I'm trying to recover it :)

About the CSS from Bootstrap, I'm :100:  percent agree with you and @Dignifiedquire but I didn't want to make too much noise including a bunch of LESS files without a confirmation. TBH I prefer to include right now the complete CSS and then in another PR focus on include the LESS files.

Also, I want to include in another PR some JS Task Manager to check the JS style code, concatenate files and minify, etc. I have previous experience with Grunt and Gulp, but just let me know if you're agree with this.

josevalim:
@milmazz about the LESS files... certainly they wouldn't be for now. For now, let's just get the minimum to close the remaining issues and, if possible, avoid the bootstrap.css file.

milmazz:
@josevalim Can you review the latest changes that I made on the sidebar? Thanks!

milmazz:
@josevalim Comments?

josevalim:
Both your PRs are on my queue for ASAP reviewing. :) The problem is that there are a bunch of stuff in this queue. :S Sorry!

milmazz:
@josevalim Hi, there's no hurry at all. So, just let me know when you can.

josevalim:
I am finally back!

josevalim:
It looks great, thank you!

josevalim:
@milmazz i have pushed another commit that inlines the breadcrumbs and the button!

screen shot 2015-07-20 at 14 20 58

milmazz:
@josevalim Great news!, I was waiting for this to add new features and fix other things. Thanks! :grinning:

@Dignifiedquire I'll try to add support for LESS in another PR, but first, please let me know if there is something else that I can help with. My idea is to help you get closer to your design refresh proposal #150

Dignifiedquire:
@milmazz the menu is a great step, and adding less is pretty much the most needed thing, to progress so :+1: from my side.
If you want to help more on the design I can try to document it a bit more (colors, margins, fonts, etc)

josevalim:
@Dignifiedquire, curiosity: do you have a fork or the design is so far a separate repository?

Dignifiedquire:
@josevalim https://github.com/dignifiedquire/ex_doc/tree/new-design and the old version https://github.com/dignifiedquire/ex_doc/tree/design

josevalim:
@Dignifiedquire what would be the best action then? for @milmazz to add less support directly on your fork? Or add it to master and you rebase? it feels like doing that on the fork would be less cumbersome?

Dignifiedquire:
I don't think it's needed to do his work on the fork, he can just add it to master and I'll rebase onto that

milmazz:
@josevalim I think it is better to add support for LESS directly to the `ExDoc` master instead of the @Dignifiedquire fork, why? because in that process we can reduce the size of the current CSS files (and concatenate all of them). Please remember that right now we're using the minified version of the CSS file that comes with the Bootstrap distribution (~120 kB).

After reading this [comment](https://github.com/elixir-lang/ex_doc/issues/150#issuecomment-122909273) from @Dignifiedquire I can do the following (only if you both agree with this proposal):

* Create a PR that adds support for LESS (`ExDoc` master)
* Fork @Dignifiedquire repository.
* Try to update/merge @Dignifiedquire work in [new-design](https://github.com/dignifiedquire/ex_doc/tree/new-design) with the latest ExDoc's master.
* Push all the changes to @Dignifiedquire or if he can assist me in the design I can continue his work.

Please, whatever option that work for you works for me as well. I just want to help :smile:

josevalim:
Ok, I was just wondering! Whatever you folks decide. :) It seems doing it straight to master is better.

Dignifiedquire:
@milmazz sounds exactly right to me, just let me know if you have any questions?

PR #234, opened 2015-06-21
quote commands in bin file

eksperimental:


josevalim:
I think we should quote just the command and not the modules. :)

eksperimental:
done! ;-)

josevalim:
Merged manually, thank you!

mix.lock
```diff
@@ -1,3 +1,3 @@
 %{"earmark": {:hex, :earmark, "0.1.17"},
-  "hoedown": {:git, "git://github.com/hoedown/hoedown.git", "2a4cf17c70e6572ed42cdca3ea7ca3e768ed17be", []},
```
josevalim: You are using elixir master which now defaults to https. So you will need to be careful to not send the update mix.lock.

eksperimental: is there a way to avoid updating this file with a git config?

josevalim: If you want #git to ignore ONLY local changes to a file:

    $ git update-index --assume-unchanged path/to/some/file

[Via @hashnuke](https://twitter.com/HashNuke/status/612560420021256192).

eksperimental: :+1:

PR #245, opened 2015-07-27
Update local links to the current page

### LINKED ISSUE #200: Do not link to the current page
Take [HasDict](http://elixir-lang.org/docs/stable/elixir/HashDict.html) as an example:

**HashDict** is mentioned three times in the introduction, and every time it is linking to the very same current page.

Well, this applies to every Module page because the first word is a link to the module itself. This first word I think should be bold, and the rest of the links maybe italics.

Maybe after introduction, if we want to keep the links to the same page, it should be replaced with an internal link to the top of the page, so no new page is reloaded.

I would like to hear your opinions,
thanks

milmazz:
Each local link that points to the current page must link to the top of the page instead, this way we can avoid reload the page each time that somebody follow those links. Fix #200

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -22,19 +22,19 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     locals = Enum.map module.docs, &(doc_prefix(&1) <> &1.id)
 
     if moduledoc = module.moduledoc do
-      moduledoc = moduledoc |> local_doc(locals) |> project_doc(modules)
+      moduledoc = moduledoc |> local_doc(locals) |> project_doc(modules, module.id)
     end
 
     docs = for node <- module.docs do
       if doc = node.doc do
-        doc = doc |> local_doc(locals) |> project_doc(modules)
+        doc = doc |> local_doc(locals) |> project_doc(modules, module.id)
       end
       %{node | doc: doc}
     end
 
     typedocs = for node <- module.typespecs do
       if doc = node.doc do
-        doc = doc |> local_doc(locals) |> project_doc(modules)
+        doc = doc |> local_doc(locals) |> project_doc(modules, module.id)
       end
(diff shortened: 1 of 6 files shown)


josevalim:
Excellet! I think this broke some tests though. What if instead of redirecting to a dynamic ID though, we always redirect to something we expect to be there like `#content`?

josevalim:
Alternatively, we could fix this in ExDoc itself: https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc/formatter/html/autolink.ex It would probably be cleaner than changing the generated stuff later with JS.

eksperimental:
I am more inclined to not to rely on javascript to solve this and other issues that can be solved with elixir.
I think it will be an issues if we try to use other formats such as pdf, epub or other simply templates.

milmazz:
@josevalim Yes, I already fix the tests. Let me look into the `Autolink` module, for sure it's better to solve the problem there. Although, the `updateLocalLinks` function can be useful when we build the sidebar links with [fillSidebarWithNodes](https://github.com/milmazz/ex_doc/blob/issue_200/lib/ex_doc/formatter/html/templates/js/app.js#L346-L353).

milmazz:
@eksperimental As @josevalim mentioned before, in this case maybe is better to fix the `Autolink` module instead of using JS to update the links, I'll look into that. This issue in particular is related with the HTML format and I don't think this will apply to the PDF or the epub format in the future. So, it can be fix via JS or ExDoc itself. Remember that if you are not agree with some proposal your comments and pull requests are always welcome.

milmazz:
@josevalim Could you please review this?, any feedback is more than welcome :smile: 

@eksperimental Could you please create a new ticket explaining which bugs or features in your opinion are better to tackle with Elixir instead of JavaScript?. After finish with the fix of some bugs and the redesign I would love to review that ticket. Thanks!

josevalim:
Do we need to include `#content` on almost all links?

PR #247, opened 2015-07-28
Disable nav link when module type list is empty

### LINKED ISSUE #246: Search process never ends if the module type list is empty
This issue happens with both, the new and the old menu.

How to reproduce this issue:
- Visit [ExUnit](http://elixir-lang.org/docs/stable/ex_unit/) (stable) or [ExUnit](http://elixir-lang.org/docs/master/ex_unit/) (master) docs
- Select  _Protocols_ in the navigation area.
- Put something in the search bar.

**Result:** Search process never ends.
**Expected:** _No results were found_ message.

milmazz:
Fix #246


josevalim:

PR #253, opened 2015-08-07
deal with empty table cells

eksperimental:
fixes the bacground issue when no description is provided.
http://elixir-lang.org/docs/master/elixir/overview.html#exceptions_summary

Display empty cells, so tables look consistent.
Make use of the css property "empty-cells"
https://developer.mozilla.org/en/docs/Web/CSS/empty-cells


josevalim:

Issue #254, opened 2015-08-07
Breadcrumbs lock the scroll on menu


eksperimental:
~~This one happens only in firefox for linux (latest version)~~
Found in Firefox for Linux

visit any page, let's say home page.
click on the last item in the breadcrumb, the top div, goes a bit to the left.
overlaps the scroll bar in the menu.
and locks that scrollbar, without being able to scroll down the menu.
WEIRD!

Update:
It does happen if you visit directly the page including "#content"
for example: http://elixir-lang.org/docs/master/elixir/Version.html#content

josevalim:
Haha, since it is very specific, please try to come with a fix for this one if you can! Or open up a bug report on Firefox! :D

eksperimental:
Let me rephrase myself.
I have only tested on linux, and I found this in Firefox.

eksperimental:
It does not happen only when you click, but if you link directly to "#content",
for example: http://elixir-lang.org/docs/master/elixir/Version.html#content

josevalim:
@eksperimental I may have fixed this in master. We were setting a negative margin on the right side of the breadcrumbs and that could have been the issue.

eksperimental:
sorry @josevalim, but it did not fix it.
@milmazz is there anything that we are doing with javascript checking for "#content", because when i directly visit http://elixir-lang.org/docs/master/elixir/Version.html#content
and that may be the hint that there is a js function doing something to that div.
thank you.

milmazz:
@eksperimental Appears to be a problem related w/Firefox, I made a few tests with Chrome/Safari and everything appears to be working smoothly. But when I visit a page (w/Firefox) with any anchor link the `#content` section looks `-10px` to the left (but the computed CSS rules and the box model appears to be the same in both scenarios -with and without anchor link-, which is weird). I don't know why is this behaviour in Firefox, I can dig into this later.

@Dignifiedquire What do you think?

Dignifiedquire:
@milmazz fixed, see attached PR

eksperimental:
I am reopening this as #274 didn't fixit.

I have the feeling it is related to this line of code
https://github.com/elixir-lang/ex_doc/blob/master/assets/js/events.js#L344

and the offcanvas-navigation thing.
I tried replacing the links to #content  with #sidebar and it's not giving that issue (at least in firebug)

I'm trying to install but I cannot get it working, so i'm just reporting it here for now.

/cc @milmazz @Dignifiedquire

[...]

PR #255, opened 2015-08-07
Increase Margins

eksperimental:
- [x] Expand summary tables to the width of the page.
- [x] Increase margins in #content
- [x] And makes the top div cover 100% to the width

### DIFF
--- a/lib/ex_doc/formatter/html/templates/css/style.css
+++ b/lib/ex_doc/formatter/html/templates/css/style.css
@@ -70,6 +70,10 @@ body {
   line-height: 1.618;
 }
 
+#content {
+  padding: 0 28px 20px 28px;
+}
+
 #sidebar a:link,
 #sidebar a:visited {
   color: #05a;
@@ -404,9 +408,9 @@ li.r2 {
 }
 
 div.breadcrumbs {
-  padding: 3px 6px;
+  padding: 3px 15px;
   background-color: #EEE;
-  margin: 0 -15px 10px;
+  margin: 0 -28px 10px;
   line-height: 28px;
   display: block;
 }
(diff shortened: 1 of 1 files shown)


lib/ex_doc/formatter/html/templates/css/style.css
```diff
@@ -70,6 +70,12 @@ body {
   line-height: 1.618;
 }
 
+#content {
+  padding: 0 28px 20px 28px;
+}
+
```
josevalim: Too many new lines? :D

lib/ex_doc/formatter/html/templates/css/style.css
```diff
@@ -404,9 +410,9 @@ li.r2 {
 }
 
 div.breadcrumbs {
-  padding: 3px 6px;
+  padding: 3px 15px;
   background-color: #EEE;
-  margin: 0 -15px 10px;
+  margin: 0 -28px 10px;
```
milmazz: Why do you want to add more space? this change generate the following problem:

capture

eksperimental: yes, but in https://github.com/eksperimental/ex_doc/blob/fix2/lib/ex_doc/formatter/html/templates/css/style.css#L73-L75 I have increased the margin.
What broweser gave you that problem? Did you click on the breadcumb links for that to happen? if so, it's a bug that is unrelated to this change and it's happening even before these changes: https://github.com/elixir-lang/ex_doc/issues/254

lib/ex_doc/formatter/html/templates/css/style.css
eksperimental: ooops.. :-P
fixed

lib/ex_doc/formatter/html/templates/css/style.css
milmazz: Maybe was something in the cache, now I see it fine in Chrome. So, go ahead.

PR #256, opened 2015-08-07
Update Bootstrap to v3.3.5

milmazz:
Update Bootstrap to the latest version available.


josevalim:
Do we need to? :) we need to remember those changes make assets stale on
all docs and what not, so unless we them, we should avoid. I noticed this
with the latest highlight.js update.

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

milmazz:
@josevalim I was thinking to update all the assets (e.g. Bootstrap, Highlight.js) first and then include support for *Brunch* or *Gulp* in `ExDoc`. Why? because with those task managers we can concatenate & minify all the assets.

If we combine all the assets we can avoid the limitation in some browser regarding the [concurrent connections per hostname](http://www.browserscope.org/?category=network) (e.g. Chrome & Firefox: 6, IE 10: 8). The all idea behind this is to reduce the *Time To First Byte* in those requests.

screen shot 2015-08-07 at 6 33 13 pm

josevalim:
@milmazz that sounds great :+1:. I would vote for brunch but the choice is yours, let me know if you need help getting that setup.

I am merging this because I have realized that we still haven't done a release with bootstrap, so this is fine! I still believe we should generally avoid doing those changes though. :D

Dignifiedquire:
I vote for gulp I have made excellent ecperience with that in the past. @milmazz let me know if you need some help setting it up, happy to help

Dignifiedquire:
I would also suggest using nnpm to pull in the js dependencies rather than keeping them checked into source control, what do you think?

josevalim:
@Dignifiedquire we need to keep them into source control because we need to include them in the releases, tars and everything else anyway. I also don't want to push people to install node/npm if all they want is to fix a bug in the elixir part of the code. :)

Dignifiedquire:
@josevalim I meant doing this when the build system is in place, and then only keeping final output checked in, as when you are running the buildsystem you need node anyway and if not you can use it withe compiled versions. Preferrably there would be a minified and non minified version checked in, for easier debugging. But source files like boostrap.css could be pulled in through npm only when needing to update the bundle.

josevalim:
@Dignifiedquire ah, I see, that works too!

Dignifiedquire:
@milmazz @josevalim I've started a basic setup with gulp, I should be ready for a pr tomorrow, could you setup that branch, so I can PR against it?

milmazz:
@Dignifiedquire Ok, I'm glad that you can help us again! I'll wait for your PR. There is already a new branch called `build_system` to try this.

I hope that @josevalim is OK with this.

josevalim:
I am definitely OK! Go ahead!

-- 

PR #257, opened 2015-08-08
Load Jquery from CDN only when using a remote site

eksperimental:
A javascript function "should_request_from_CDN" has been created to determine
whether to load from the CDN or from a local copy.
The rules are. if the page is served via http or https, and the host is
not a private network (127.x.x.x, or 192.168.x.x., or others), and
the hostname is contains a "."  (to avoid loading from localhost and others
user may have configured), only then, it will load from the CDN.

Range of private networks I am using: https://en.wikipedia.org/wiki/Private_network#Private_IPv4_address_spaces

I have used parts of this code, to determine if it's a private network
https://gist.github.com/dperini/729294

The other change is that now it requests from "//" and not from "https://"
*****

A second commit removes something that was remaning from I don't know where
https://github.com/elixir-lang/ex_doc/commit/1bede82875d7a17bfb3642a4e44835f5ab831488

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -91,3 +91,13 @@ Copyright 2011, The Dojo Foundation
   https://github.com/jquery/sizzle/blob/master/LICENSE.txt
 
 ==========================================================================
+
+Javascript fuction "should_request_from_CDN", based the regular expression
+to detect the private networks on the works from
+"Regular Expression for URL validation"
+https://gist.github.com/dperini/729294
+
+Copyright (c) 2010-2013 Diego Perini (http://www.iport.it)
+Released under MIT License
+
+==========================================================================
--- a/lib/ex_doc/formatter/html/templates/footer_template.eex
+++ b/lib/ex_doc/formatter/html/templates/footer_template.eex
@@ -1,11 +1,6 @@
   </section> <!--/content -->
 </div><!--/row -->
   <!-- Scripts -->
-    <script>
-      relpath = '';
-      if (relpath != '') relpath += '/';
-    </script>
-
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js"></script>
     <script>window.jQuery || document.write('<script src="js/jquery.js"><\/script>')</script>
     <script src="js/highlight.pack.js"></script>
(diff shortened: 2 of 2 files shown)


josevalim:
I will leave this up to you and @milmazz!

Dignifiedquire:
I'm suggesting to load Jquery always from the local bundle when we are using a build system as this pulls all JavaScript into one file.

milmazz:
IMHO, this seems a little hack (that try to cover an exceptional case) and I don't really see the *benefit from the user point of view*. Maybe I'm missing something here.

Right now, if I'm offline, I cannot reach `ajax.googleappis.com`, in that case we load jQuery from the local file and with that we can provide all the features that the `app.js` file provides. Also, we need to consider that @Dignifiedquire is working on a build system than concatenates & minifies all the JS files. So, in that case jQuery will be always loaded from the local file.

eksperimental:
@milmazz what it does is when you are in an private network, or on your own computer (either http://localhost/, 127.0.0.1, or any non-qualified host name), or loading it from **file://**, it saves the the unnecessary request to load it from the CDN.
(and when using ladoing it from file://, the request doesn't get cached, at least I don't see it cached).

josevalim:
Given that we will be minifying everything on our own already, maybe we don't really need this as we will already be serving a single JS file?

eksperimental:
@josevalim It all depends on how long it will take for 0.9 to be released.
Personally I think we should not avoid changes because it will be obsolete in the new version.
I think at least 0.8.1 should be released with the fixes we are going to be added in the meantime 0.9 get ready.

josevalim:
@eksperimental I agree with that but given that we are not so sure about the provided function too, I think we can wait. :) Unless this was a clear 100% win, then I would have no objections. :)

eksperimental:
@josevalim We can leave it to catch only the scheme and if there is a "." in the hostname.

    ( protocol == 'http:' || protocol == 'https:' ) && host.indexOf('.') !== -1 )

or even this, if you want to still load from CDN when developing in localhost/

    ( protocol == 'http:' || protocol == 'https:' )

My point is, calling https when viewing the docs from file:// is a clear loss to me.
But we can wait :-)

PR #258, opened 2015-08-08
Avoid to use a function before its definition

milmazz:
This is known as JS Hoisting, more details:

http://www.adequatelygood.com/JavaScript-Scoping-and-Hoisting.html


josevalim:

PR #259, opened 2015-08-08
Improve events when ESC or ENTER keys are pressed

milmazz:
Recover both features from the old ExDoc version.


Dignifiedquire:
:+1:

PR #260, opened 2015-08-09
Build System

Dignifiedquire:
This is a work in progress, opening this for tracking purposes

- [x] Setup gulp
- [x] Setup linting via ESLint
- [x] Setup JavaScript compilation using webpack
- [x] Setup less compilation, including autoprefixer
- [x] Pull dependencies (bootstrap, highlightjs, jquery) via npm instead of keeping them checked in
- [x] Add minified version for JavaScript
- [x] Add minified version for CSS
- [x] Add watch task with recompilation on change for development convenience
- [x] Make ex_doc actually use the new files in `dist`
- [x] Setup browser tests using [karma](http://karma-runner.github.io/0.13/index.html) 
- [x] Setup travis to run linting and browser tests

cc @milmazz @josevalim


josevalim:
This looks great! I think we can move all the assets though to `assets` or `static` on root. We can leave just the precompiled files inside `lib` as those are the ones we effectively need during compilation. :) What do you think?

Dignifiedquire:
Sounds good, would be happy to not have the files in this deeply nested structure

Dignifiedquire:
@milmazz @josevalim The list is done, please review and let me know what you think

milmazz:
@Dignifiedquire Sorry for the delay, I'll review this ASAP, in the meantime, could you please resolve the conflicts?

Dignifiedquire:
@milmazz done

assets/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+In this directory reside all assets for ExDoc. The ready to use built versions are found in `dist`. To change any of them read on
+
+## Dependencies
+
+To work on these assets you need to install [Node.js] and [npm]. After that run
+
+```bash
+$ npm install -g gulp
+$ cd lib/ex_doc/formatter/html/templates/assets
```
milmazz: Please update this path.

Dignifiedquire: done

assets/less/code.less
```diff
@@ -123,3 +119,14 @@ Source: https://github.com/isagalaev/highlight.js/blob/master/src/styles/github.
 .hljs-chunk {
   color: #aaa;
 }
+
+
+// Customizations
+
+.hljs, pre {
+  background: #fffde8;
+}
```
milmazz: Oh!, you already fix this :)

Dignifiedquire:
@milmazz just pushed some more cleanup, and splitting files for better handling, so go update your local copy ;)

milmazz:
@Dignifiedquire Yes, I noticed your recent update, I was almost to comment about your `assets/less/app.less` file, but you already fix the background-color issue in the code samples :+1: 

Everything looks great!, great work!.

Maybe one thing that I feel it is missing is something like a *git hook*, the idea is that we need to guarantee that every time that somebody update a css/js file, the files under the `dist` directory also get updated. What do you think?

I'll proceed to update my local copy.

Dignifiedquire:
Sure I can add a pre-commit hook that will trigger a build if you want

milmazz:
@Dignifiedquire I'm OK with that. @josevalim are you Ok with this proposal?

Dignifiedquire:
@milmazz can you jump into IRC, that might be easier to hash out there

PR #264, opened 2015-08-10
Licenses updated

eksperimental:
* normalized.css added to LICENSE
 * Links to the respective projects added
 * Removed  "CSS Template: lib/ex_doc/formatter/html/templates/css/style.css",
     since not only that the default template itself (css, html and js) is
     based on YARD
 * Tab replace with 2 spaces in the YARD license
 * Link to license added in /lib/ex_doc/formatter/html/templates/css/elixir.css

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,6 @@
 ExDoc
 Copyright (c) 2012 Plataformatec
+https://github.com/elixir-lang/ex_doc/
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


eksperimental:
What is the wording we should have in README?
Currently we have.
> ExDoc source code is released under Apache 2 License with snippets under MIT-LICENSE.
> Check [LICENSE](LICENSE) file for more information.

1. There are many more licenses.
2. I think we should mention that the default theme was based on YARD.

josevalim:
Let's reword to something like:

> ExDoc source code is released under Apache 2 License. The generated contents, however, are under different licenses based on projects used to help render html, including css, js and other assets. Check the LICENSE file for more information.

I wouldn't worry about YARD right now because that's gping to change soon! :)

josevalim:

eksperimental:
;-D

PR #265, opened 2015-08-10
Redirect template added & Duplicate index page removed

eksperimental:
The current implementation of the index page made of a copy of
the "main" page in index.html. Creating an unncessary duplicate page.
This commit creates an HTML redirect page.

### DIFF
--- a/lib/ex_doc.ex
+++ b/lib/ex_doc.ex
@@ -21,12 +21,15 @@ defmodule ExDoc do
   Generates documentation for the given `project`, `version`
   and `options`.
   """
+  @spec generate_docs(String.t, String.t, Keyword.t) :: atom
   def generate_docs(project, version, options) when is_binary(project) and is_binary(version) and is_list(options) do
     config = build_config(project, version, options)
     docs = config.retriever.docs_from_dir(config.source_beam, config)
     find_formatter(config.formatter).run(docs, config)
   end
 
+  # Builds configuration by merging `options`, and normalizing the options.
+  # @spec build_config(String.t, String.t, Keyword.t) :: %ExDoc.Config{Keyword.t}
   defp build_config(project, version, options) do
     options = normalize_options(options)
     preconfig = %Config{
@@ -61,14 +64,16 @@ defmodule ExDoc do
   # Helpers
 
   defp normalize_options(options) do
+    options_norm = []
+
     pattern = options[:source_url_pattern] || guess_url(options[:source_url], options[:source_ref] || "master")
-    normalized_options = [source_url_pattern: pattern]
+    options_norm = options_norm ++ [source_url_pattern: pattern]
 
     if is_bitstring(options[:output]) do
-      normalized_options = normalized_options ++ [output: String.rstrip(options[:output], ?/)]
+      options_norm = options_norm ++ [output: String.rstrip(options[:output], ?/)]
     end
 
-    Keyword.merge(options, normalized_options)
+    Keyword.merge(options, options_norm)
   end
 
   defp guess_url(url = <<"https://github.com/", _ :: binary>>, ref) do
--- a/lib/ex_doc.ex
+++ b/lib/ex_doc.ex
@@ -29,7 +29,7 @@ defmodule ExDoc do
   end
 
   # Builds configuration by merging `options`, and normalizing the options.
-  # @spec build_config(String.t, String.t, Keyword.t) :: %ExDoc.Config{Keyword.t}
+  @spec build_config(String.t, String.t, Keyword.t) :: %ExDoc.Config{}
   defp build_config(project, version, options) do
     options = normalize_options(options)
     preconfig = %Config{
@@ -64,16 +64,14 @@ defmodule ExDoc do
   # Helpers
 
   defp normalize_options(options) do
-    options_norm = []
-
     pattern = options[:source_url_pattern] || guess_url(options[:source_url], options[:source_ref] || "master")
-    options_norm = options_norm ++ [source_url_pattern: pattern]
+    options = Keyword.put(options, :source_url_pattern, pattern)
 
     if is_bitstring(options[:output]) do
-      options_norm = options_norm ++ [output: String.rstrip(options[:output], ?/)]
+      options = Keyword.put(options, :output, String.rstrip(options[:output], ?/))
     end
 
-    Keyword.merge(options, options_norm)
+    options
   end
 
   defp guess_url(url = <<"https://github.com/", _ :: binary>>, ref) do
--- a/lib/ex_doc.ex
+++ b/lib/ex_doc.ex
@@ -62,7 +62,13 @@ defmodule ExDoc do
 
   defp normalize_options(options) do
     pattern = options[:source_url_pattern] || guess_url(options[:source_url], options[:source_ref] || "master")
-    Keyword.put(options, :source_url_pattern, pattern)
+    normalized_options = [source_url_pattern: pattern]
+
+    if is_bitstring(options[:output]) do
+      normalized_options = normalized_options ++ [output: String.rstrip(options[:output], ?/)]
+    end
+
+    Keyword.merge(options, normalized_options)
   end
 
   defp guess_url(url = <<"https://github.com/", _ :: binary>>, ref) do
--- a/lib/ex_doc/cli.ex
+++ b/lib/ex_doc/cli.ex
@@ -1,8 +1,8 @@
 defmodule ExDoc.CLI do
   def run(args, generator \\ &ExDoc.generate_docs/3) do
     {opts, args, _} = OptionParser.parse(args,
-               aliases: [o: :output, f: :formatter, u: :source_url, r: :source_root,
-                         m: :main, p: :homepage_url, c: :config])
+               aliases: [o: :output, f: :formatter, c: :config, r: :source_root,
+                         u: :source_url, m: :main, p: :homepage_url])
 
     [project, version, source_beam] = parse_args(args)
 
@@ -63,14 +63,15 @@ defmodule ExDoc.CLI do
       PROJECT            Project name
       VERSION            Version number
       BEAMS              Path to compiled beam files
-      -o, --output       Path to output docs, default: docs
-      --readme           Path to README.md file to generate a project README, default: nil
-      -f, --formatter    Docs formatter to use; default: html
+      -o, --output       Path to output docs, default: "doc"
+      --readme           Path to README.md file to generate a project README, default: `nil`
+      -f, --formatter    Docs formatter to use, default: "html"
       -c, --config       Path to the formatter's config file
-      -r, --source-root  Path to the source code root, default: .
+      -r, --source-root  Path to the source code root, default: "."
       -u, --source-url   URL to the source code
-      --source-ref       Branch/commit/tag used for source link inference, default: master
-      -m, --main         The main, entry-point module in docs
+      --source-ref       Branch/commit/tag used for source link inference, default: "master"
+      -m, --main         The main, entry-point module in docs,
+                           default: "overview" when --fomatter is "html"
       -p  --homepage-url URL to link to for the site name
 
     ## Source linking
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -6,10 +6,17 @@ defmodule ExDoc.Formatter.HTML do
   alias ExDoc.Formatter.HTML.Templates
   alias ExDoc.Formatter.HTML.Autolink
 
+  defmodule Config do
+    defstruct [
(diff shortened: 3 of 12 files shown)


lib/ex_doc/formatter/html/templates/redirect_template.eex
```diff
@@ -0,0 +1,11 @@
+<%= head_template(%{id: :redirect, content: config, redirect_to: redirect_to, }) %>
```
josevalim: I think that, instead of passing all those options, we should just generate the redirect template directly here, including the html head. It will make this and all other templates cleaner. What do you think?

eksperimental: absolutely!

eksperimental:
I have added a few tests as well. let me know if we should add a test for any other situation

josevalim:
Looks good to me! @milmazz?

milmazz:
@eksperimental Could you please update this section in the [sidebar](https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc/formatter/html/templates/sidebar_template.eex#L17-L30), [overview](https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc/formatter/html/templates/overview_template.eex#L8-L12) and [README](https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc/formatter/html/templates/readme_template.eex#L4-L9) templates?

After that, could please rebase your changes? Thanks in advance.

josevalim:
Good catch @milmazz!

eksperimental:
thank you @milmazz  for pointing that out.

Reading https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc/formatter/html/templates/sidebar_template.eex#L17-L30 makes me feel uncertain what is what when I need to link to index.html,
I propose a much simpler rationale.

These are the Files we can have:
- readme.html
- index.html
- overview.html
- MAIN.html

And this are the settings:
- has_main
- has_readme

_Note: MAIN.html is whatever file we pass to the option "main:"_

***********
**What I propose is the following rationale.**

* **has_readme: true**   => Link to README.html
* **has_readme: false**  => Do not show any link

* **has_main:   true**   => generate index.html that redirects to "MAIN".html
* **has_main:   false**  => generate index.html that redirects to either overview.html or readme.html (current behaviour is if has_readme is true, we redirect to readme.html, see proposal of change bellow)

With this, we can always count on "overview.html", "MAIN.html", and "index.html" to be there.
The only one that may not be is readme.html which can easily be checked with `has_readme`

And we never need to link to "index.html" or "./", only from external sources.
*******

Personally I would to simplify the the current behaviour regarding index.html.
The current one redirects to readme.html when there is a readme.html, but readme may not be present.
So I propose to change the default behaviour to redirect to overview.html unless main option given.
User can still set "main: readme"

josevalim:
We can calculate what is the proper "index" still in the Elixir code and have something like `config.index`. We just use it then instead of everywhere we use "index.html".

eksperimental:
@milmazz  &  @josevalim: all your suggestions have been implemented.

lib/ex_doc/cli.ex
```diff
@@ -63,15 +63,17 @@ defmodule ExDoc.CLI do
...
-      -r, --source-root  Path to the source code root, default: .
+      -r, --source-root  Path to the source code root, default: "."
       -u, --source-url   URL to the source code
-      --source-ref       Branch/commit/tag used for source link inference, default: master
-      -m, --main         The main, entry-point module in docs
+      --source-ref       Branch/commit/tag used for source link inference, default: "master"
+      -m, --main         The main, entry-point module in docs,
+                           default: "overview" when --fomatter is "html"
       -p  --homepage-url URL to link to for the site name
+      -i, --index        Index page, default: "index.html" when --fomatter is "html"
```
eksperimental: Should we list the default options for when --formatter is "html"?

test/ex_doc/formatter/html_test.exs
```diff
@@ -101,4 +119,56 @@ defmodule ExDoc.Formatter.HTMLTest do
...
   end
+
+  test "run generates the redirect index.html" do
+    config = %ExDoc.Config{output: "#{output_dir}/redirect", main: "RandomError", }
+    HTML.run(get_modules(config), config)
+
+    assert File.regular?("#{output_dir}/redirect/RandomError.html")
+    assert File.regular?("#{output_dir}/redirect/index.html")
+    content = File.read!("#{output_dir}/redirect/index.html")
+    assert content =~ ~r{<meta http-equiv="refresh" content="0; url=RandomError.html"\s*/>}
```
eksperimental: Does it make sense to do an assert on `File.regular?(file)` when right after we are doing an assert on the content of that file?

lib/ex_doc/cli.ex
josevalim: Sorry, I don't get it. Why do we need this option? How is it different from main?

test/ex_doc/formatter/html_test.exs
josevalim: No, it does not. :)

lib/ex_doc/cli.ex
eksperimental: I thought of  --index as a to store the address to the the index "index.htm", "", "./" or "/"
but --main as the path to what it should redirect to.
I think I broke my own rule of not linking.
I don't think we need need to ever link to index.html so i don't think we need config.index
the only thing we need is when people visit "test.com/doc/" it takes them to "overview.html" or whatever set by --main
What do you think

josevalim: Yes, we don't need this config I think. It is even a bit confusing. :)

eksperimental:
@josevalim @milmazz this is done. Let me know what you think.
I had to restructure the test and also separate what belongs to the HTML formatter and what belonged  to ExDoc module itself.

lib/ex_doc.ex
```diff
@@ -21,12 +21,15 @@ defmodule ExDoc do
...
   """
+  @spec generate_docs(String.t, String.t, Keyword.t) :: atom
   def generate_docs(project, version, options) when is_binary(project) and is_binary(version) and is_list(options) do
     config = build_config(project, version, options)
     docs = config.retriever.docs_from_dir(config.source_beam, config)
     find_formatter(config.formatter).run(docs, config)
   end
 
+  # Builds configuration by merging `options`, and normalizing the options.
+  # @spec build_config(String.t, String.t, Keyword.t) :: %ExDoc.Config{Keyword.t}
```
josevalim: Why is the `@spec` commented? :)

lib/ex_doc.ex
```diff
@@ -61,8 +64,16 @@ defmodule ExDoc do
   # Helpers
 
   defp normalize_options(options) do
+    options_norm = []
+
     pattern = options[:source_url_pattern] || guess_url(options[:source_url], options[:source_ref] || "master")
-    Keyword.put(options, :source_url_pattern, pattern)
+    options_norm = options_norm ++ [source_url_pattern: pattern]
```
josevalim: Why you are not using `Keyword.put`? It will be faster and semantically correct.

lib/ex_doc.ex
```diff
@@ -61,8 +64,16 @@ defmodule ExDoc do
...
+
+    if is_bitstring(options[:output]) do
+      options_norm = options_norm ++ [output: String.rstrip(options[:output], ?/)]
```
josevalim: Same here.

lib/ex_doc/formatter/html.ex
```diff
@@ -9,7 +9,9 @@ defmodule ExDoc.Formatter.HTML do
   @doc """
   Generate HTML documentation for the given modules
   """
-  def run(module_nodes, config)  do
+  # @spec run(list, %ExDoc.Config{Keyword.t}) :: String.t
```
josevalim: Why is this `@spec ` commented also?

lib/ex_doc.ex
eksperimental: %ExDoc.Config{Keyword.t} is not valid,
I just wanted you to review the specs if you could

lib/ex_doc.ex
josevalim: We should just `Keyword.put(options, :source_url_pattern, pattern)` as before. I don't think we should normalize to merge latter.

lib/ex_doc/formatter/html.ex
```diff
@@ -22,27 +24,36 @@ defmodule ExDoc.Formatter.HTML do
...
     generate_list(protocols, all, output, config, has_readme)
 
-    if config.main do
-      generate_main(output, config)
+    Path.join(config.output, "index.html")
+  end
+
+  defp build_config(config) when is_map(config) do
+    if config.main == "index" do
+      raise ArgumentError, message: "`main` cannot be set to \"index\", otherwise it will recursively link to itself"
```
josevalim: Perfect. We can drop the backtick ` from main and use quotes too.

lib/ex_doc/formatter/html.ex
```diff
@@ -22,27 +24,36 @@ defmodule ExDoc.Formatter.HTML do
...
     end
 
-    Path.join(config.output, "index.html")
+    preconfig = %{
+      main: "overview",
+    }
+    Map.merge(config, preconfig, fn(_k, v1, v2) ->
+      if is_nil(v1) do v2; else v1; end
+    end)
```
josevalim: This is equivalent to:

    Map.update(config, :main, "overview", fn v -> v || "overview" end)

Or:

    Map.put(config, :main, config[:main] || "overview")

Both shorter and clearer than the current merge approach. :)

test/ex_doc/formatter/html_test.exs
```diff
@@ -102,4 +124,34 @@ defmodule ExDoc.Formatter.HTMLTest do
...
+    config = doc_config([project: "Elixir", version: "1.0.1", output: "#{output_dir}/redirect", main: "RandomError", ])
+    generate_docs(config)
+
+    assert File.regular?("#{output_dir}/redirect/RandomError.html")
+    content = File.read!("#{output_dir}/redirect/index.html")
+    assert content =~ ~r{<meta http-equiv="refresh" content="0; url=RandomError.html"\s*/>}
+  end
+
+  test "run generates index.html and normalized options" do
+    for dir <- ["test/tmp/doc", "test/tmp/doc/", "test/tmp/doc//"] do
```
josevalim: This is being tested twice. Here and in the other test.

josevalim:
I have added some comments, thank you!

lib/ex_doc.ex
josevalim: It should just be `%ExDoc.Config{}`. :)

test/ex_doc/formatter/html_test.exs
eksperimental: this is the only suggestion that I didn't implement. I want your opinion first,
While it is double checking something we are checking in the ex_doc_test.exs, the inclusion of this  test was the one that made me realize that things in this tests were not running through ExDoc module, and that why the overhaul.
So that's why it makes me feel a bit uneasy to not to have something like this, at least to let us rest assure that we are generating the docs the right way (ie. not through HTML.run)
What do you think?

test/ex_doc/formatter/html_test.exs
```diff
@@ -102,4 +123,28 @@ defmodule ExDoc.Formatter.HTMLTest do
     assert HTML.pretty_codeblocks(iex_detected_without_class) ==
           "<pre><code class=\"iex elixir\">iex&gt; max(4, 5)"
   end
+
+  test "run generates index.html and normalized options" do
+    # 1. Check for default [main: "overview"]
+    # 2. Check for output dir having trailing "/" stripped
+    for dir <- ["#{output_dir}", "#{output_dir}/", "#{output_dir}//"] do
```
josevalim: I think we can test just only one of those. The trailing stuff is already checked on the unit test below. Generally speaking, we should avoid integration tests (that generates the docs and goes through everything) for every possible combination. The unit tests are enough. :)

eksperimental:
@josevalim thank you so much for all your suggestion =)  They are greatly appreciate and the certainly do improve the quality of my code.
I have implemented all of your suggestions (with the exception of one that I commented on), made a few improvements, removed a few duplicated things, please have a look at 15aad12bd9e11ef6ef64fc69d0e43f67399ae7a9

test/ex_doc/formatter/html_test.exs
eksperimental: true. it makes sense to only have the last one.

eksperimental:
@josevalim New commit pushed

josevalim:

eksperimental:
thanks once again @josevalim for following up ;-D

PR #266, opened 2015-08-10
Do not create doc test folder out of /test/tmp

eksperimental:
* docs/ renamed to doc/ to follow elixir convention

### DIFF
--- a/test/ex_doc/formatter/html_test.exs
+++ b/test/ex_doc/formatter/html_test.exs
@@ -4,14 +4,12 @@ defmodule ExDoc.Formatter.HTMLTest do
   alias ExDoc.Formatter.HTML
 
   setup_all do
+    # clean up from previous test
+    File.rm("test/tmp/README.md")
+    File.rm_rf("#{output_dir}")
+
     File.mkdir(output_dir)
     File.copy("test/fixtures/README.md", "test/tmp/README.md")
-
-    on_exit fn ->
-      File.rm("test/tmp/README.md")
-      System.cmd "rm", ["-rf", "#{output_dir}"]
-    end
-
     :ok
   end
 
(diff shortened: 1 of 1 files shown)


test/ex_doc/formatter/html_test.exs
```diff
@@ -5,10 +5,10 @@ defmodule ExDoc.Formatter.HTMLTest do
 
   setup_all do
     File.mkdir(output_dir)
-    File.copy("test/fixtures/README.md", "test/README.md")
+    File.copy("test/fixtures/README.md", "test/tmp/README.md")
 
     on_exit fn ->
-      File.rm("test/README.md")
+      File.rm("test/tmp/README.md")
       System.cmd "rm", ["-rf", "#{output_dir}"]
```
eksperimental: I'm  also in favour of getting rid of these two lines, so we can actually see the output when debugging exdoc.
Anyways they will be removed on `mix clean`

josevalim: We need to clean up at some point though, at most before the test starts. We can also use `File.rm_rf` now.

eksperimental: true, yeah, when the test starts sounds perfect to me

eksperimental:
let me know what you think @josevalim 
I have added a few tests as well. let me know if we should add a test for any other situation

eksperimental:
Any idea why the build is failing?
> ** (Mix) Expected :only in dependency to be an atom, got: [:test]
> The command "mix do local.hex --force, deps.get && mix test --include pandoc" exited with 1.
it's also happening with the other PR

josevalim:
It is a bug in Elixir I have just fixed.

eksperimental:
:+1: great
thank you

PR #267, opened 2015-08-10
Tidy up a few files

eksperimental:
Sort alphabetically long lists of options,
Indent mix.exs for easier reading.

### DIFF
--- a/lib/ex_doc.ex
+++ b/lib/ex_doc.ex
@@ -1,10 +1,19 @@
 defmodule ExDoc do
   defmodule Config do
     defstruct [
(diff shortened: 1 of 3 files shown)


mix.exs
```diff
@@ -2,21 +2,27 @@ defmodule ExDoc.Mixfile do
   use Mix.Project
 
   def project do
-    [app:        :ex_doc,
```
josevalim: Let's please keep this syntax as it is the one used in all `elixir-lang` projects.

josevalim: I mean, all changes are good, except the ones to this file.

PR #269, opened 2015-08-11
Update readme to v0.8

eksperimental:


README.md
```diff
@@ -10,8 +10,8 @@ To use ExDoc in your Mix projects, first add ExDoc as a dependency:
 
 ```elixir
 def deps do
-  [{:earmark, "~> 0.1", only: :dev},
-   {:ex_doc, "~> 0.7", only: :dev}]
+  [{:earmark, "~> 0.1.17 or ~> 0.2", only: :dev},
```
josevalim: It can be "~> 0.2". We should push people to the latest anyway. We just support older versions for convenience.

eksperimental: great! fixed

milmazz: @josevalim @eksperimental So, this means that we can also update the [mix.exs](https://github.com/elixir-lang/ex_doc/blob/master/mix.exs#L14) file.

josevalim:
Let's keep the mix.exs file the same so old users are fine. We shuold push new users to use the latest though.

josevalim:
Sorry, I am drunk. There is no earmark 0.2 yet :(

Issue #261, opened 2015-08-09
Release goals for v0.9


milmazz:
The idea is to define our list of goals for the next release.

* [x] New design #275
* [x] Include a build system to concatenate and reduce the size of our assets. Also this build system must provide a way to check the quality of our JS code
* [x] Include some tests for JS to avoid break previous features.
* [x] Include support for LESS w/Bootstrap

@Dignifiedquire  is already working on some of this tasks, for more details please see #260

@josevalim @ericmj @eksperimental @Dignifiedquire What do you think about this? We should include or remove some tasks for the next release?

josevalim:
I think this is great! :heart: I would move PDF/epub to another release though. A new design is already tons of work on its own. :)

Dignifiedquire:
I agree with @josevalim pdf/epub stuff is tricky to do from my experience so we should focus on the other things first and get that released and then one release focused on new formats

milmazz:
@josevalim I've been testing @Dignifiedquire's work in the `build_system` branch and let me say that it is impressive :+1:

For sure in near future I'll include include more tests for JS with Karma, but the all structure appears to be right, I just add a few things to the README file (below the `assets` directory). I just want to ask you when you think is the right moment to merge the `build_system` branch into `master` (I already merge the latest changes made in `master` into `build_system`). Do you prefer to do a release of the current status of `master` before the merge?

josevalim:
I am closing this since it is all done. :D

PR #271, opened 2015-08-11
Include gulp as a build system tool

milmazz:
This PR include Gulp as a build system to concatenate and reduce the size of our assets. Also this build system must provide a way to check the quality of our JS code. Now the CSS rules can be described as LESS files.

More details in #260 and #261


Dignifiedquire:
:+1:

milmazz:
@Dignifiedquire Again, great work!, 99.999% of this PR is yours :+1:

josevalim:
Folks, just go ahead with this. Don't wait for my input. :) I can take a look at it at all later!

PR #273, opened 2015-08-11
Redesign merged with build system

Dignifiedquire:
By far nothing perfect, but a start 

A small todo list

- [ ] Fixed header
- [ ] Sidebar styles
- [ ] Content styles
- [ ] Add footer
- [ ] Only include from bootstrap what we use, rather than bundling all this stuff we don't need
- [ ] Pull fonts not from `templates/fonts` but from `node_modules/font-awesome` instead
- [ ] Mobile stuff
- [ ] Pull fonts from google fonts into the repo so full offline usage is possible


assets/less/code.less
```diff
@@ -1,138 +1,127 @@
-/*
-
-github.com style (c) Vasily Polovnyov <vast@whiteants.net>
-  Released under the same license as highlight.js
-  https://github.com/isagalaev/highlight.js/blob/master/LICENSE
-
-Source: https://github.com/isagalaev/highlight.js/blob/master/src/styles/github.css
```
milmazz: Please, keep the header and update the reference to the original file. Is [this](https://raw.githubusercontent.com/isagalaev/highlight.js/master/src/styles/tomorrow.css) right?

Dignifiedquire: Fixed, the license was not correct anymore, but I added the source link back

milmazz:
@Dignifiedquire If you want, for now, I can work with the following tasks: 

* Header (we should add the option of an image logo?, remember #228)
* Footer
* Fix elixir tests.

Are you agree with this?

BTW, what do you mean with *mobile stuff*? Keep the current responsive design or something else?

Dignifiedquire:
@milmazz sounds good. I meant with "mobile stuff" making sure that everything looks and works well on smaller screen sizes.

milmazz:
@Dignifiedquire Already fix all tests.

Dignifiedquire:
@milmazz while it's nice that you merged this, the list is now in a closed issue, maybe we move the list to a new issue?

josevalim:
Moved to #275

PR #276, opened 2015-08-12
Safely reduce the size of the sidebar_items.js file

milmazz:


lib/ex_doc/formatter/html.ex
```diff
@@ -48,6 +48,8 @@ defmodule ExDoc.Formatter.HTML do
   defp generate_sidebar_items(modules, exceptions, protocols, output) do
     input = for node <- [%{id: "modules", value: modules}, %{id: "exceptions", value: exceptions}, %{id: "protocols", value: protocols}], !Enum.empty?(node.value), do: node
     content = Templates.sidebar_items_template(input)
+    # Safely reduce the size of the JS code
+    content = Regex.replace(~r/(\s|\n|\h|\v)/, content, "")
```
eksperimental: should \r be included?

milmazz: Actually I think I can use `\R` to replace all the newline sequences.

Dignifiedquire: Does is take into account whitespace between quotes?, e.g what happens when you have something like this:

```js
"hello world": ["we are some strings"]
```

milmazz: Yes, this Regex will replace those too, but in this particular case (`sidebar_items.js`) only include modules, exceptions and protocols names, I think is not valid an space in those, right?.

josevalim: We won't have space in those but this feels like such a hack. :D Do we have a really measurable gain? 

One alternative, just for sanity, is to be a bit more conservative and trim them only if the follow `(\:|\,)`. We can use one of the replace syntax to replace them but keep the leading character:

```iex
iex(2)> Regex.replace ~r/(\:|\,)(\s|\t|\R|\h|\v)+/, ~s("hello world": ["we are some strings"]), "\\1"
"\"hello world\":[\"we are some strings\"]"
```

josevalim: Just to be sure you see my comment. I think being a bit more conservative here wouldn't hurt. :)

eksperimental: Should we place this file under "dist/" dir, so we have all the js files in one same dir?

milmazz: @josevalim There is a gain on this hack ;D, (e.g. ~400% size reduction if we compare with the current `sidebar_items.js` in Elixir/master) and also this is to be consistent with the current `app.js` (also *uglify*). But I agree with you in be more conservative, can we add to the `:` and `,` list the following characters `[`, `]`, `{`, `}` and `;`?

josevalim: Sure, those are all fine too!

-- 


josevalim:
Can we try to get this in soon? I will release a new ex_doc version tomorrow. :)

milmazz:
Yes, just give me one hour, it's lunch time here :-)

milmazz:
@josevalim ping

josevalim:
Looks great! Is it still efficient? Ship it!

-- 


milmazz:
This is wrong, in my local tests I was generating the documentation for EEx without noticing (I have scripts for that). After fixing the script I notice that this task does not add a real gain. So, I'll close this because adds an unnecessary complexity in this case. Sorry for the inconvenience.

eksperimental:
I was wondering the same thing @milmazz , before you mentioned 400% I was only able to reduce it by 15%, that was 2kb in 15kb, that was by doing tricks and replacing more than just spaces.

PR #278, opened 2015-08-13
Better performance on fillSidebarWithNodes function

milmazz:
This is an alternative implementation of the `fillSidebarWithNodes` function using string concatenation instead of jQuery for DOM manipulation.

In http://jsperf.com/fillsidebarwithnodes you can find a comparison (and run your own tests) between both implementations.

screen shot 2015-08-12 at 9 16 55 pm


assets/js/events.js
```diff
@@ -243,13 +242,7 @@ function fillSidebarWithNodes (nodes, filter) {
...
 
       /* li.node */
-      li = $('<li>', {
-        'class': 'node'
-      })
-
-      if (element.hasOwnProperty('docs')) {
-        li.append($('<a/>').attr('class', 'toggle'))
-      }
+      li = '<li class="node">'
```
josevalim: maybe set `li_class = "node"` then down below do `li_class += " clicked"` and just then build the li?

josevalim:
This is awesome! I have added just one tiny comment but feel free to merge it.

Dignifiedquire:
In this form I'm :-1: right now I'm afraid, it just becomes too hard to read and brittle to change. I would suggest using a very simple templating language like [handlebars](http://handlebarsjs.com/) integrated into the build system, which gets precompiled in the build step and so is very fast as well but maintains readability.

josevalim:
@Dignifiedquire if it is precompiled and not a runtime dependency then let's definitely give it a try!

Dignifiedquire:
@josevalim there is a small runtime to reduce code duplication, but it's compiled down from it's text representation into pure function calls.

Dignifiedquire:
@milmazz this is what would be needed for integrating it into the build system https://github.com/altano/handlebars-loader

josevalim:
I am merging this today because it is a nice performance improvement and I want to ship a new ExDoc version. We can move to a template later if necessary. Do you think the new design is going to be more or less complex in this aspect?

Dignifiedquire:
At the moment it's very similar. The underlying html stays mostly the same

milmazz:
@Dignifiedquire Can you please help me setting the handlebars-loader? I have a preliminar version but I've been working with http://www.tryhandlebarsjs.com/ 

Don't know yet how to play with webpack, thanks in advance.

Dignifiedquire:
@milmazz sure, I'll take a look at it tomorrow

milmazz:
@Dignifiedquire After taking some time off, I gave it another try and I think the setup right now it's working :smiley: 

I'll keep you posted.

PR #281, opened 2015-08-13
Several Small fixes

eksperimental:
* align text in the ExDoc.CLI help
* ~~close void elements in head_template~~
* Add Project Name and Version to `<title>`
  - This information should be present to avoid confusion with different versions. Specially when bookmarking.
* Get rid of `~r/ /` in favour of `~r{ }`
    To avoid escaping slashes
* head_template standardized & titles improved
  * head_template : Now it takes 2 args. Stadardize with the rest of the templates,
      taking "config" as first argument
  * ~~head_template: titles improved, to not so duplicate information when a module
      has the same name as the project. ex: ExUnit,
      instead of reading: "ExUnit – ExUnit v1.1.0-dev", it now reads:
      "ExUnit v1.1.0-dev – Documentation"~~

### DIFF
--- a/lib/ex_doc/formatter/html/templates/head_template.eex
+++ b/lib/ex_doc/formatter/html/templates/head_template.eex
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="utf-8" />
+    <meta charset="utf-8">
     <meta http-equiv="x-ua-compatible" content="ie=edge">
     <%= cond do %>
     <%= page.id == :module -> %>
--- a/lib/ex_doc/formatter/html/templates/head_template.eex
+++ b/lib/ex_doc/formatter/html/templates/head_template.eex
@@ -5,11 +5,11 @@
     <meta http-equiv="x-ua-compatible" content="ie=edge">
     <%= cond do %>
       <%= page.id == :module -> %>
-        <title><%= page.module.id %> – <%= page.content.project %> v<%= page.content.version %></title>
+        <title><%= page.module %> – <%= config.project %> v<%= config.version %></title>
       <%= page.id == :readme -> %>
-         <title>README – <%= page.content.project %> v<%= page.content.version %></title>
+        <title>README – <%= config.project %> v<%= config.version %></title>
       <%= page.id == :overview -> %>
-        <title>Overview – <%= page.content.project %> v<%= page.content.version %></title>
+        <title>Overview – <%= config.project %> v<%= config.version %></title>
     <% end %>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta name="generator" content="ExDoc">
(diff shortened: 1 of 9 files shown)


lib/ex_doc/formatter/html/templates/head_template.eex
```diff
@@ -2,18 +2,22 @@
 <html>
   <head>
     <meta charset="utf-8" />
-    <meta http-equiv="x-ua-compatible" content="ie=edge">
+    <meta http-equiv="x-ua-compatible" content="ie=edge" />
```
milmazz: `meta` is a void element, according to the HTML5 spec the `/` character has no effect.

lib/ex_doc/formatter/html/templates/head_template.eex
```diff
@@ -2,18 +2,22 @@
...
+          <title><%= page.module %> – <%= config.project %> v<%= config.version %></title>
+        <% end %>
+      <%= page.id == :readme -> %>
+         <title>README – <%= config.project %> v<%= config.version %></title>
+      <%= page.id == :overview -> %>
+        <title>Overview – <%= config.project %> v<%= config.version %></title>
     <% end %>
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta name="generator" content="ExDoc">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
milmazz: `meta` is a void element, according to the HTML5 spec the `/` character has no effect.

lib/ex_doc/formatter/html/templates/head_template.eex
```diff
@@ -2,18 +2,22 @@
...
+    <meta name="generator" content="ExDoc" />
```
milmazz: `meta` is a void element, according to the HTML5 spec the `/` character has no effect.

lib/ex_doc/formatter/html/templates/head_template.eex
```diff
@@ -2,18 +2,22 @@
...
 
-    <link rel="stylesheet" href="dist/app.css">
+    <link rel="stylesheet" href="dist/app.css" />
```
milmazz: `link` is a void element, according to the HTML5 spec the `/` character has no effect.

eksperimental:
@milmazz then lets remove the one from meta charset

lib/ex_doc/formatter/html/templates/head_template.eex
```diff
@@ -2,18 +2,22 @@
...
+    <meta http-equiv="x-ua-compatible" content="ie=edge" />
     <%= cond do %>
-    <%= page.id == :module -> %>
-    <title><%= page.content %></title>
-    <%= page.id == :readme -> %>
-    <title><%= page.content.project %> v<%= page.content.version %> README</title>
-    <%= page.id == :overview -> %>
-    <title><%= page.content.project %> v<%= page.content.version %> Overview</title>
+      <%= page.id == :module -> %>
+        <%= if page.module == config.project do %>
```
josevalim: Actually, it is fine to drop this. I would rather be consistent with "Module - Project Version" then special case when they are equal. Typically speaking, if we can avoid special cases, we should avoid them. :)

eksperimental: yes. it can be an issue when writing tests.

eksperimental:
alright guys, both of your suggestions have been updated

milmazz:
:+1:

PR #293, opened 2015-08-15
Move template folder to /templates

eksperimental:
I think it will make our lives easier, but I don't now if I'm breaking any rules

Eventually in the future when we add support for different themes, and different outputs (epub, etc), 
it will look like

- templates/html/default/head_template.eex
- templates/epub/default/head_template.eex

for now, it is good to have.
- templates/head_template.eex


josevalim:
Templates should be in the lib directory so mix knows when to recompile
them!

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

eksperimental:
there it was the rule I was breaking.

PR #296, opened 2015-08-16
Fresh sidebar

Dignifiedquire:
screen shot 2015-08-17 at 00 27 11


josevalim:
@Dignifiedquire is this up for feedback? Typography and organization wise, it looks awesome, however I would love if we could keep the lighter colors of the previous sidebar in the new design. Thoughts?

josevalim:
Also, can we do anything about the scroll bars?

josevalim:
If this is still work in progress, please ignore my comments. :D

Dignifiedquire:
> Also, can we do anything about the scroll bars?

Do you mean styling wise or what do you mean??

Dignifiedquire:
@josevalim yes it's up for feedback, it's not perfectly implemented yet, but the general ideas are there. I have thought about keeping the lighter colors, but I feel that this way we gain a few things. First the contrast to the main content is increased, and making it easier for the user to focus on reading when they are reading. Second the previous color palette had a washed out feeling, making it less crisp which I tried to remedy with these new colors. Third, the purple for me feels much more fitting than the green/gray as this is after all the documentation generator for Elixir.

josevalim:
> Do you mean styling wise or what do you mean?

Yes. The white background mostly. It is calling a lot of attention.

>  Second the previous color palette had a washed out feeling, making it less crisp which I tried to remedy with these new colors. Third, the purple for me feels much more fitting than the green/gray as this is after all the documentation generator for Elixir.

Right. But this is the documentation for all Elixir projects too and they have their own identity. For example, Phoenix is all orange. The current approach pushes too much of the Elixir visual identify, that's why I preferred the more washed out one. It reminds me it is part of Elixir... but it is subtle. :) Does it make sense?

Dignifiedquire:
I will see what's possible on the sidebar styling, browsers sadly don't give you a lot of control there. 

It does make sense what you are saying. I will think about it and test with different documentations like the one from Phoenix.

eksperimental:
I think removing the elixir drop logo, will help us in seeing it more neutral

Dignifiedquire:
@eksperimental here is a screenshot of the docs for phoenix: 
screen shot 2015-08-17 at 13 47 32

eksperimental:
it looks good. maybe the darker color is too dark in the bar. (just don't count on my opinion, as I am terrible at this design things)
having another screenshot with a more washed out sidebar will help to make a decission
personally, It doesn't look that we are pushing the elixir identity.

The only thing i can tell you is that the current color theme for ExDoc, whenever i need to read `<code>` it kills my eyes. it is too light and that pinkish colour almost vanishes inside the white background...

Maybe having the ability for people to include their own stylesheets would be a bonus in future versions.

josevalim:
This is very incomplete but just to give an idea:

screen shot 2015-08-17 at 14 10 57

Btw, this new menu, regardless of the color, is a thousand times better than the one in the old design (currently deployed). More compact and more readable at the same time. :D

milmazz:
@Dignifiedquire Could you please update this PR? I already rebase all the changes from `master` and after that this PR has some conflicts.

Let me know if you need to help with this.

Dignifiedquire:
Will create a new branch, as the guidelines have changed anyway

PR #298, opened 2015-08-17
Test sets main as "readme"

eksperimental:
just to check we are redirecting to the proper reame.html page
https://github.com/elixir-lang/ex_doc/issues/297


eksperimental:
updated to deal with the changes introduced by https://github.com/elixir-lang/ex_doc/commit/1860547f3ffb0a64223fd8705337dba7bdd08b6d

PR #299, opened 2015-08-17
ExDoc version in generator <meta> + Thorough test checking tags in <head>

eksperimental:


josevalim:
We don't need to pass the ExDoc version in. We can just call `ExDoc.version` directly from the template.

eksperimental:
oh, I didn't know that

eksperimental:
@josevalim fixed!

test/ex_doc/formatter/html_test.exs
```diff
@@ -51,7 +51,49 @@ defmodule ExDoc.Formatter.HTMLTest do
...
+      module: [
+        ~r{<title>RandomError – Elixir v1.0.1</title>},
+        ~r{<meta name="viewport" content="width=device-width, initial-scale=1.0">},
+        ~r{<meta http-equiv="x-ua-compatible" content="ie=edge">},
+      ],
+    }
+
+    for {page, content} <- pages do
+      for {rule_page, rules} <- re, rule_page in [:shared, page] do
+        for rule <- rules, do:
```
josevalim: I have a feeling the straight-forward version of this test would be less lines of code than the current one. And more understandable. :) Can we give it a try?

eksperimental: @josevalim done ;-D

josevalim:
:shipit:

josevalim:
:shipit:

PR #300, opened 2015-08-17
Do no show diff in /lib/ex_doc/formatter/html/templates/dist/*

eksperimental:
Just treat them as binaries,
so it won't display the diff when we do `git log -p`
and run a diff.

Otherwise it spoils the output with such big files.

/cc @Dignifiedquire @milmazz

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/lib/ex_doc/formatter/html/templates/dist/* -diff


Dignifiedquire:
sweet, I didn't know about this :+1:

josevalim:
:+1:

PR #301, opened 2015-08-17
Enforce call to ExDoc.Formatter.HTML.run/3 to be done from ExDoc,

eksperimental:
Enforce call to ExDoc.Formatter.HTML.run/3 to be done from ExDoc, and not directly.

This way we can assure the configs are passed through ExDoc, doing
the preconfiguration and the normalization.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/lib/ex_doc/formatter/html/templates/dist/* -diff


lib/ex_doc/formatter/html.ex
```diff
@@ -10,10 +10,18 @@ defmodule ExDoc.Formatter.HTML do
   @main "overview"
 
   @doc """
-  Generate HTML documentation for the given modules
+  Generate HTML documentation for the given modules.
+
+  Please do not call this function directly. Use `ExDoc.generate_docs/3` instead.
   """
```
eksperimental: should we set this to `@docfalse` ?
I think it's good to have it, just to let the developers there is a formatter here,
but maybe we could just move that information to `@moduledoc`

josevalim:
yes, we don't need to do this. We should just set `@moduledoc false`. That is enough information that people should not fiddle with it. :)

PR #303, opened 2015-08-18
Enforce call to ExDoc.Formatter.HTML.run/3 to be done from ExDoc,

eksperimental:
[I have accidentally closed: https://github.com/elixir-lang/ex_doc/pull/301]

...and not directly.

This way we can assure the configs are passed through ExDoc, doing
the preconfiguration and the normalization.


josevalim:
This is not right. We should not do checks like these. :) Just mark it as `@moduledoc false` and it should be enough.

josevalim:
Reverted in master but kept `@moduledoc false`. :)

eksperimental:
sorry, when you said we don't need that, i thought you were referring to my question about removing `@doc` not about the whole commit. :-P

josevalim:
haha, no problem!

PR #304, opened 2015-08-18
Add a 404.html error page (not found)

milmazz:
A basic `404.html` page, useful for cases like [GitHub Pages](https://help.github.com/articles/custom-404-pages/)


josevalim:
Do you really think it is required? The less we generate the better and sites like hexdocs and even github pages already do a good job of showing a nice page when something goes wrong.

milmazz:
As a user, I really like when a website offers a custom 404 page, why? Regardless the reason of the error (mistyped URL in the address bar, clicked on an out-of-date link, broken link, etc), this custom page normally helps me (in a user-friendly way) find the page that I was looking for. For `ExDoc` we can provide the all side-bar and also a link to the *Overview*. With both elements I can easily find the information that I need. 

This 404 page is useful for projects like `Elixir`, `EEx`, `IEx`, `ExUnit`, `Logger`, `Mix`, etc. Instead of the useless page that GitHub Pages offers. Guys, don't get me wrong, for me, as a user, the default 404 page is useless because this page does not give me any alternative to find the content that I want, does not relate in any way with the Elixir project. Of course, I agree that this new 404 page could be useless in cases like hexdocs.pm. But please keep in mind that hexdocs.pm already offers a 404 custom page.

If you want, I can improve the message, For example, with JavaScript we can detect if the result of the 404 error was triggered by: a mistyped URL, an out-of-date search referral, a broken (in-site) link, etc.

@dignifiedquire @eksperimental @josevalim What do you think?

eksperimental:
I also believe that Error 404 can add value to the content, instead of just being the typical placeholder page that does nothing but tell you that what you are looking for is not there.

I love when a 404 page gives you a form, or give you results already of something similar that you are looking for. (I know these are static pages, but maybe something can be implemented).

Maybe something like Github's Quick jump betweenfiles feature:
https://github.com/elixir-lang/ex_doc/find/master

I think a good 404 page can add value to the documentation that ExDoc generates,
it's just a matter of finding how can this page add value to the content.

josevalim:
Go ahead then folks. :)

Dignifiedquire:
:+1: from my side, for the reasons @milmazz explained so nice :)

PR #308, opened 2015-08-20
Fix anchor links when node type is callback/type

### LINKED ISSUE #307: Deep links from sidebar are broken
Ex: http://hexdocs.pm/ecto/Ecto.Repo.html#update_all/3 is what you get when you click the sidebar
but http://hexdocs.pm/ecto/Ecto.Repo.html#c:update_all/3 is what it should be.

milmazz:
This commit fix the anchor links (in the sidebar area) when the node type is: `defmacrocallback`, `callback` or `type`.

As a side note, I think that we can do it better, I don't like the current implementation of the `sidebar_items_*_template.eex` templates, what do you think? any suggestion is more than welcome.

Closes #307


josevalim:
@milmazz moving to another template function seems to be too much. It feels like those functions generating JSON would be better written in pure Elixir than using `.eex` files. Their contents are mostly code instead of being mostly text.

milmazz:
@josevalim Better? Let me know if you have another tip :smiley:

josevalim:
Yes, I think so, way less indirection.  :shipit:

Dignifiedquire:
@milmazz why the addition of th eslintrc file?

milmazz:
@Dignifiedquire That was trying to fix the error reported by Travis-CI after the ESLint check. What do you suggest to do in this case?

Dignifiedquire:
@milmazz I see, but you don't need to create a new config, just move the one we have in `assets` https://github.com/elixir-lang/ex_doc/blob/master/assets/.eslintrc to the top level of the repo, that way all js files are covered.

milmazz:
@Dignifiedquire I'll fix this problem in another PR, even moving the `.eslintrc` file to the root directory is not enough, could be an issue with Travis-CI? I made some tests in other VM and everything is working OK.

Dignifiedquire:
@milmazz master looks good now, all fixed?

milmazz:
Yes sir :)

PR #315, opened 2015-08-23
More styles

Dignifiedquire:
* Toggle the sidebar whenever you want (shown by default on desktop, hidden on smaller screen sizes)
* Add blockquote styling
* Minor tweaks


assets/js/sidebar.js
```diff
@@ -0,0 +1,53 @@
...
+  var dur = duration
+  if (immediate) {
+    dur = 0
+  }
+  sidebar.animate({
+    '-webkit-flex-basis': width,
+    '-moz-flex-basis': width,
+    '-ms-flex-basis': width,
+    'flex-basis': width
+  }, 0, function () {
```
milmazz: I think that at this point you should reference the variable `dur` to indicate the duration and not a fixed value to perform the animation, right?

assets/js/sidebar.js
```diff
@@ -0,0 +1,53 @@
...
+    openSidebar()
+  }
+}
+
+function init () {
+  $('.sidebar-toggle').click(function () {
+    toggleSidebar()
+  })
+
+  if ($(window).width() > 768) {
```
milmazz: IMO, this kind of detection is more related with styling stuff... So, why not to use a *css media query* in this case?

milmazz:
@Dignifiedquire I made a few comments in your code. Also, could you please fix the ESLint error. Thanks!

assets/js/sidebar.js
Dignifiedquire: @milmazz not sure what you mean? I can't add a class using a `css media query`

Dignifiedquire:
@milmazz fixed

PR #317, opened 2015-08-26
More styling improvements

Dignifiedquire:


milmazz:
@Dignifiedquire Could you please fix the unit tests?

Dignifiedquire:
wtf they were passing locally :/

Dignifiedquire:
@milmazz no idea what to do, I ran `mix clean` deleted `_build´ and the tests are still passing locally

josevalim:
Let me take a look.

josevalim:
It is because it uses Elixir master and we changed the line reference in master. It is not your fault @Dignifiedquire. :D We should ExDoc to use a stable elixir version.

josevalim:
We should be good now in master. This is good to merge anyway.

milmazz:
@Dignifiedquire @josevalim Ok, I'll merge this into `new_design`, then later I'll apply a `rebase` with the latest changes from `master`.

PR #319, opened 2015-08-28
EPUB support

### LINKED ISSUE #305: EPUB support
Just to let you know that I've been doing some research and some work to include an EPUB formatter for `ExDoc`, also the idea of this message is to inform you about this task and avoid duplication of efforts. 

Of course, I'll give more priority to the issues related with the `new_design`, but I'll keep working whenever it's possible.

I'll keep you posted!

milmazz:
This is a work in progress, the idea is to know your opinion about the current progress.

Things to do:

* [x] Improve styles
* [x] Fix known issues

Tasks already done:

* EPUB document display correctly, at least on iBooks, [Calibre](http://calibre-ebook.com/) and Android [Aldiko](http://www.aldiko.com/)
* Basic templates (Table of Contents, Cover, Module documentation, stylesheets, assets, etc)
* Support for EPUB 2 and EPUB 3
* Reuse `Autolink` from the `HTML` formatter.
* Fix all issues reported by [Calibre](http://calibre-ebook.com/) (Tools :arrow_forward: Check Book)
* Add tests
* Support for *logo* option
* Support for *extras* option

Some of the known bugs are the following:

* Errors reported by [epubcheck](https://github.com/IDPF/epubcheck)
* ~~When `ExDoc` is executed from the command line (`../ExDoc/bin/ex_doc ... -m Kernel -f epub`) I'm getting an `UndefinedFunctionError` (the problem is related with the new `UUID` dependency), but everything runs OK via `MIX_ENV=docs mix docs -f epub`, any advice?~~ Fix: e4e5c65
* ~~According to the `epubcheck` tool the `mimetype` file contains a wrong type, but I don't know yet the reason for this (Am I using the `:zip.create` function correctly?). This file needs to comply with the following [rules](http://www.idpf.org/epub/30/spec/epub30-ocf.html#sec-zip-container-mime)~~ Fix: 0e8dab4
* ~~The other annoying error reported by `epubcheck` is related with the *ids* (anchor) generated by `Autolink`.~~ Fix: e680b00

```
ERROR: Ecto-v0.16.0.epub/OEBPS/modules/Ecto.Adapter.html(163,44): value of attribute "id" is invalid; must be an XML name without colons
WARNING: Ecto-v0.16.0.epub/OEBPS/modules/Ecto.Adapter.html(173,49): use of non-registered URI scheme type in href: Ecto.Type.html#t:t/0
```

One way to solve this is to post-process the results from the templates, something like this:

```elixir
  defp generate_module_page(output, config, node) do
    content = Templates.module_page(config, node) |> clean_ids_hrefs()
    File.write("#{output}/OEBPS/modules/#{node.id}.html", content)
  end
 
  defp clean_ids_hrefs(content) do
    id_pattern = ~r/[^A-Za-z0-9_.-]/
 
    result = Regex.replace(~r/href=\s*"(?!(http|https|ftp|mailto):\/\/)([^#]+)?#([^"]+)"/i, content, fn _, _, page, anchor -> 
"href=\"#{page}#" <> String.replace(anchor, id_pattern, "") <> "\"" end)
    Regex.replace(~r/id="([^\"]+)"/, result, fn _, id -> "id=\"" <> String.replace(id, id_pattern, "") <> "\"" end)
  end
```

But this to me seems like an ugly hack and maybe with this hack we can create a duplicate ID (I need to verify this, but each duplicate ID will be reported by `epubcheck` as an *error*), but I want to reuse `ExDoc.Formatter.HTML.Autolink` as much as possible. Any advice?

Example for Ecto:

```bash
$ export MIX_ENV=docs
$ mix deps.update ex_doc
$ mix deps.compile ex_doc
$ mix docs -f epub
$ cd doc
$ ls
Ecto-v1.0.0.epub
# How to use epubcheck
$ java -jar ../epubcheck-3.0.1/epubcheck-3.0.1.jar Ecto-v1.0.0.epub 
```

Closes #305


milmazz:
@josevalim ping

milmazz:
@josevalim Ping, any review is greatly appreciated.

At this stage, I think that this new feature is in *beta* and if you think it's right moment we can merge into *master*.

Thanks in advance :ok_hand:

josevalim:
Thank you @milmazz! I need to prepare for my keynote and release Elixir v1.1. So let's push ExDoc 0.10 first and then I will focus on this with you. :heart:

milmazz:
No problem @josevalim, in the meantime I'll try to improve the CSS styles,
add more documentation, tests and fix other things :)

I hope that everything goes well on the ElixirConf [image: :heart:]

henrik:
Would it make any sense to extract some of the epub generation to a general-purpose epub lib?

josevalim:
We haven't even landed epub support yet. Once we do and once it works, we can possibly start a discussion about extracting it.

henrik:
That's completely fair. I was mostly thinking that this could make this PR smaller and also give @milmazz something to do until this is back in focus :D @josevalim, best of luck with the keynote! Wish I could be there :)

milmazz:
@henrik In the meantime you can review the code if you want, any feedback is more than welcome :)

I'll try to update this PR as soon as possible.

lib/ex_doc/formatter/epub.ex
```diff
@@ -0,0 +1,231 @@
...
+    output = Path.expand(output)
+    target_path =
+      "#{output}/#{config.project}-v#{config.version}.epub"
+      |> String.to_char_list
+
+    {:ok, zip_path} = :zip.create(target_path,
+                                  files_to_add(output),
+                                  compress: ['.css', '.html', '.ncx', '.opf',
+                                             '.jpg', '.png', '.xml'])
+    {:ok, zip_path}
```
dukex: Here is possible, I'm new in elixir, only return `:zip.create...` eg:

``` elixir
defp generate_epub(output, config) do
    output = Path.expand(output)
    target_path =
      "#{output}/#{config.project}-v#{config.version}.epub"
      |> String.to_char_list

    :zip.create(target_path,
                files_to_add(output),
                compress: ['.css', '.html', '.ncx', '.opf',
                           '.jpg', '.png', '.xml'])
end
```

On line 42 you are doing the check if is `:ok`

lib/ex_doc/formatter/epub/templates/title_template.eex
```diff
@@ -0,0 +1,22 @@
...
+  <body>
+    <h1><%= config.project %></h1>
+    <h2>v<%= config.version %></h2>
+    <%= if config.logo do %>
+      <%= if Path.extname(config.logo) == ".png" do %>
+        <div><img src="assets/logo.png" alt="Logo"/></div>
+      <% end %>
+      <%= if Path.extname(config.logo) == ".jpg" do %>
+        <div><img src="assets/logo.jpg" alt="Logo"/></div>
+      <% end %>
```
dukex: Why not only concatenate the ext? 
``` elixir
<div><img src="assets/logo.<%= config.logo %>" alt="Logo"/></div>
```

Why the `logo`config is the logo extension?  I think it's a bit confused

josevalim: The logo config is not the extension, we are calling `Path.extname`.

lib/ex_doc/formatter/epub.ex
josevalim: That's a common Elixir practice. Somtimes it is preferable to match on the result when we call it so we get the proper cause on errors: http://blog.plataformatec.com.br/2014/09/writing-assertive-code-with-elixir/

eksperimental: but isn't the last time "{:ok, zip_path}" unnecessary? since that's what the previous line returns

lib/ex_doc/formatter/epub/templates/title_template.eex
eksperimental: how about this one?

```elixir
<%= if config.logo && Path.extname(config.logo) in [".jpg", ".png"] do %>
  <div><img src="assets/logo.png" alt="Logo"/></div>
<% end %>
```

dukex: my bad, sorry. But there is some reason to check the extension?

eksperimental: oops. i need to fix this.:  "assets/logo.png"

lib/ex_doc/formatter/epub.ex
```diff
@@ -0,0 +1,231 @@
...
+    |> struct([fragment: fragment])
+    |> to_string
+  end
+
+  defp id_replace(id) do
+    pattern = ~r/[^A-Za-z0-9_.-]/
+    replacement = "--"
+
+    String.replace(id, pattern, replacement)
+  end
```
eksperimental: @milmazz will not this create ID collisions?

lib/ex_doc/formatter/epub.ex
milmazz: Yes, the last line can be deleted, I'll fix that!

lib/ex_doc/formatter/epub/templates/title_template.eex
milmazz: @dukex `ExDoc` right now only support `.jpg` and `.png` (see: `ExDoc.Formatter.HTML.process_logo_metadata/1`)

@eksperimental We don't apply any transformation to the image format, we just the rename the file name to `logo`. So, in your alternative you're missing the `.jpg` extension with this:

```html
<div><img src="assets/logo.png" alt="Logo"/></div>
```

lib/ex_doc/formatter/epub.ex
milmazz: @eksperimental Yes, I think that an ID collision is possible with this workaround, as far as I remember I tried with `EEx` and there was no ID collisions at all, but I can review this tonight with the `Elixir`.

Thinking more about this ID transformation process, IMHO the best place to put this is in the `Autolink` module, we can have an `strict` mode (required by XHTML) to control the kind of IDs that we produce.

The only restrictions of the `id` element in [HTML5](http://www.w3.org/TR/html5/dom.html#the-id-attribute) are the following: Must be unique (duh!), must have at least one character and does not allow space characters. But for XHTML there are more restrictions and *epubchecker* see this kind of violations as errors.

A valid [XML name](http://www.w3.org/TR/REC-xml/#NT-Name) must comply with the following rules:

```
Name             ::=    NameStartChar (NameChar)*
NameStartChar    ::=    ":" | [A-Z] | "_" | [a-z] | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x2FF] | [#x370-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
NameChar         ::=    NameStartChar | "-" | "." | [0-9] | #xB7 | [#x0300-#x036F] | [#x203F-#x2040]
```

What do you think?

eksperimental: So epub accepts html5 ids?

milmazz: @eksperimental No, I don't think so. The `module_template.eex` is XHTML and as I mentioned before, if we use in the templates the IDs produced by `Autolink`, without any transformation, *epubcheck* catch those IDs as error.

eksperimental: Sorry im not acquainted with epub, but from what i read its like html in a comoressed gzip format. So if it accepts html5 maybe we should move to it since we csn reuse a lot of what we have for the html module. But if it doesnt please ignore this comment

eksperimental: Yes. Thats epub3 format only that supports  html5. Later i will look at the situation and try to coneup with something. For what we do now in html is to escape everything. So that should still give u a valid xhtml id guess

eksperimental: you will have collisions in kernel, for example between: `=/2` and `%/2` or `^/1` and `&/1`.

milmazz: Ok, thanks for the information, let me find a way to improve this!

milmazz:
@josevalim This weekend I finally resume working on this feature, I would like your review about this. If you approve this feature in `master`, then, we definitely will need some help in the design area, but I think the current behavior is a good start, wdyt?

screen shot 2016-09-11 at 5 31 28 pm

screen shot 2016-09-11 at 5 32 01 pm

screen shot 2016-09-11 at 5 31 48 pm

milmazz:
@josevalim ping?

lib/ex_doc/formatter/epub/templates.ex
```diff
@@ -0,0 +1,61 @@
...
+  """
+  def module_page(config, node) do
+    types = H.group_types(node)
+    module_template(config, node, types.types, types.functions, types.macros, types.callbacks)
+  end
+
+  defp extra_title(path), do: path |> String.upcase |> Path.basename(".MD")
+
+  templates = [
+    content_template: {[:config, :nodes, :uuid, :datetime], @content_template_doc},
```
josevalim: Since you want to attach docs, I would rather call EEx.function_from_file one by one instead of building this whole list. :)

josevalim:
@milmazz this looks great to me! I have added only one tiny comment. The other thing that is missing is the documentation aspect. How to use EPUB? I would add some information to the README as well to the `mix docs` task.

Another point of concern is code duplication between the formatters but I haven't taken a deep look yet. What do you think? Were you able to re-use most of the code? Is there any area you would like to avoid the duplication bit?

In any case, feel free to merge it to master any time you want.

gulpfile.js
```diff
@@ -65,16 +68,26 @@ gulp.task('buildHighlight', function (done) {
...
+  cb()
+}
+
+gulp.task('javascript', ['buildHighlight'], function () {
+  return javascript({src: 'assets/js/app.js', dest: distPath.html}, function () {});
+})
+
+gulp.task('epub-javascript', ['buildHighlight'], function () {
+  return javascript({src: 'assets/js/epub.js', dest: distPath.epub}, function () {});
 })
```
milmazz: @dignifiedquire Do you think this is a good way to reuse this task without affecting the current behavior?

milmazz:
@josevalim I just add more improvements, this is the current status:

screen shot 2016-09-21 at 11 15 28 pm

About your suggestions:

> The other thing that is missing is the documentation aspect. How to use EPUB? I would add some information to the README as well to the `mix docs` task.

You're right, I'll add more information about this new feature asap.

> Another point of concern is code duplication between the formatters but I haven't taken a deep look yet. What do you think? Were you able to re-use most of the code?

Yeah, I know, I tried to avoid the code duplication between both formats as much as possible. Actually, I use many functions from the `ExDoc.Formatter.HTML`,  `ExDoc.Formatter.HTML.Autolink` and some functions from `ExDoc.Formatter.HTML.Templates`. At least the `ExDoc.Formatter.EPUB.run/2` function follow almost the same workflow as `ExDoc.Formatter.HTML.run/2`, but, there are a lot of different templates between both formats. Although, I'll check this PR again and see if I miss something else :)

@elixir-lang/exdoc Any other suggestion is more than welcome.

josevalim:
It looks beautiful. I should do another full review pass today or tomorrow and add any pending feedback.

gulpfile.js
josevalim: It is fine to affect the current behaviour though since this is an internal tool.

assets/README.md
```diff
@@ -29,6 +29,10 @@ This will build a complete bundle, including JavaScript and CSS.
 Using the flag `--type production` will result in minified JavaScript and CSS
 bundles.
 
+### `epub-build`
```
josevalim: Would it possible to have `build` do both `epub` and `html` builds? This way we should still keep `build` but we can then use the `epub-build` and `html-build` only if I want explicitly too.

assets/README.md
```diff
@@ -38,6 +42,10 @@ for details.
 
 Build the JavaScript in `js` into a bundled file using [webpack].
 
+### `epub-javascript`
```
josevalim: Should we normalize all names to be `FORMATTER-TASK`? So we would have `epub-javascript` and `html-javascript`. Otherwise I am afraid it will be confusing. What do you think?

lib/ex_doc/formatter/epub/templates/content_template.eex
```diff
@@ -0,0 +1,43 @@
...
+    <item id="ncx" href="toc.ncx" media-type="application/x-dtbncx+xml"/>
+    <item id="nav" href="nav.xhtml" media-type="application/xhtml+xml" properties="nav scripted"/>
+    <item id="cover" href="title.xhtml" media-type="application/xhtml+xml" properties="scripted"/>
+    <%= for extra <- config.extras do %>
+      <item id="<%= extra_title(extra) %>" href="<%= extra_title(extra) %>.xhtml" media-type="application/xhtml+xml" properties="scripted"/>
+    <% end %>
+    <%= for node <- nodes do %>
+      <item id="<%= node.id %>" href="<%= node.id %>.xhtml" media-type="application/xhtml+xml" properties="scripted"/>
+    <% end %>
+    <% alias ExDoc.Formatter.HTML.Templates, as: T %>
```
josevalim: We should probably move those aliases to the top of the template OR to the module that defines the templates.

josevalim:
@milmazz i have added a couple more comments, then we are good to go!

assets/README.md
milmazz: Sure, I'll change that.

assets/README.md
milmazz: I agree with your suggestion.

lib/ex_doc/formatter/epub/templates/content_template.eex
milmazz: Better to move the alias to the module that uses the templates, that way we avoid a blank line at the beginning of this XML file (`epubchecker` emits a warning for this).

josevalim:
@milmazz I did another review and it looks amazing! :heart:

The only thing left I can think of is to add support for the `--formatter` flag in the mix task and document it here: https://github.com/elixir-lang/ex_doc/blob/master/lib/mix/tasks/docs.ex

milmazz:
@josevalim Done! 🎉 

```console
$ mix docs --formatter epub
Compiling 1 file (.ex)
Docs successfully generated.
View them at '/Users/milmazz/Development/elixir-lang/ex_doc/doc/ex_doc-v0.13.2.epub'.
$ mix docs -f epub
Docs successfully generated.
View them at '/Users/milmazz/Development/elixir-lang/ex_doc/doc/ex_doc-v0.13.2.epub'.
```

dignifiedquire:
@milmazz please fix the functions you created in https://github.com/elixir-lang/ex_doc/blob/master/gulpfile.js#L202 (`javascript` and `less`), they do not need to take a callback but rather `return` the gulp pipeline they crate, i.e. it should look like this

```js
var javascript = function (options) {
  return gulp.src(options.src)
    .pipe(webpack(isProduction ? config.production : config.development))
    .pipe($.if(isProduction, $.uglify()))
    .pipe($.if(isProduction, $.rev()))
    .pipe($.size({title: 'js'}))
    .pipe(gulp.dest(options.dest))
}
```

otherwise `gulp` will not handle these tasks correctly as the async nature of the operations is not take into account when running the tasks and you could end up with broken builds.

milmazz:
@dignifiedquire Thanks for your suggestion. I already made the [change](https://github.com/elixir-lang/ex_doc/commit/b4f794c84ee2bbd1d5209b3268087d90d7afaca4), please let me know what do you think.

dignifiedquire:
@milmazz LGTM :)

PR #322, opened 2015-09-04
Display search results in the content area.

Dignifiedquire:
cc @milmazz


milmazz:
@Dignifiedquire Wow, this is amazing, here a couple of suggestions:

* Provide an alternative way to start the search process, right now the user must press the `ENTER` key
* If you try to find something like `File.ls` does not work, I think must show the following functions: `ls/1`, `ls!/1`, `lstat/1` and `lstat!/1`.
* Provide a way to *close* the search result and go back to the previous page.

milmazz:
@Dignifiedquire Ping?

Dignifiedquire:
will fix later today

Dignifiedquire:
Fixed everything except the search in nested functions. Was that working before?

milmazz:
@Dignifiedquire are you referring to this?

> If you try to find something like `File.ls` does not work, I think must show the following functions: `ls/1`, `ls!/1`, `lstat/1` and `lstat!/1`.

Yes.

Dignifiedquire:
@milmazz all done :) let's get this shipped :ship:

milmazz:
@Dignifiedquire Awesome!

PR #323, opened 2015-09-05
[TravisCI] Migrate from legacy to container-based infrastructure

milmazz:
This PR includes certain improvements when we use Travis:

- Use the new container-based infrastructure
- Cache directories `node_modules` (for `npm install`) and `deps` (for `mix deps.get`)
- Use the latest release of Pandoc
- Fix a little ESLint error, with this the build result exit with 0 (OK)

@josevalim Let me know what do you think about this? Is it OK to cache `deps` directory?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,3 +22,7 @@ before_script:
 script:
   - "mix do local.hex --force, deps.get && mix test --include pandoc"
   - node_modules/.bin/gulp
+cache:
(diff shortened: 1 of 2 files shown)


josevalim:
@josevalim Let me know what do you think about this? Is it OK to cache deps directory?

I think so. Let's try it and see how it goes! :D

PR #325, opened 2015-09-06
.travis.yml: automatically download latest pandoc

eksperimental:
travis-before-install.sh created.
it checks the latest stable relesase
based on this redirect url: https://github.com/jgm/pandoc/releases/latest
then, we parse the page, and determine look for the *-amd64.deb file.
So long as there is a file that matches this naming convention it will be retrieved,
and if may are provided, the firt one will be chosen.

If no file could be determined, then an error message is displayed.

/cc @milmazz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,9 @@ notifications:
   recipients:
     - jose.valim@plataformatec.com.br
 before_install:
-  - wget -c https://github.com/jgm/pandoc/releases/download/1.15.0.6/pandoc-1.15.0.6-1-amd64.deb
-  - ar -x pandoc-1.15.0.6-1-amd64.deb
-  - tar xzf data.tar.gz
+  - ./travis-before-install.sh
   - export PATH=$PATH:$PWD/usr/bin/
+  - pandoc --version
 before_script:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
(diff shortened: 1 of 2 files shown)


josevalim:
I am :-1: because it is additional complexity without little benefits. We don't need to be running on latest pandoc and we can update it from time to time when required. :)

milmazz:
@eksperimental I agree with @josevalim, the only requirement for ExDoc is a Pandoc version greater or equal to 1.10, the package provided by Travis is to old. More details about this in #211 

> According to [this page](http://pandoc.org/releases.html#pandoc-1.10-2013-01-19) `pandoc` supports enable or disable specific syntax extensions by appending them (with + or -) to the writer or reader name (e.g. `markdown+fenced_code_blocks`) from its version 1.10.

> I've looked for more information about the [CI environment OS](http://docs.travis-ci.com/user/ci-environment/#CI-environment-OS) provided by Travis and I found that they use _Ubuntu 12.04 LTS Server Edition 64 bit_. Also, I found that the version of the pandoc package for [Ubuntu 12.04](http://packages.ubuntu.com/precise/pandoc) is 1.9.1.1-1. So, that's why this feature does not work in this environment.

HTH

eksperimental:
I thought we wanted to check on the latest pandoc, but i realize the tests that we do are too basic, so, yes, no point

PR #371, opened 2015-09-13
Make "View Source" to show actual definition

lexmag:
Closes #334.

PR #373, opened 2015-09-13
Fallback to doc line if there is no fun definition in abstract code

lexmag:
Follow-up on #371.
Docs building is successful.


josevalim:

PR #380, opened 2015-09-14
Post coverage to travis

milmazz:
This PR includes the Mix task `coveralls.travis`, which allows to submit the *coverage* results to the [coveralls.io](https://coveralls.io) servers when TravisCI build is executed.

Then we can show a *badge* in our README that shows the results.

@josevalim I think that to allow this, first [coveralls.io](https://coveralls.io) needs your authorization.

The coverage history is similar to this page:

https://coveralls.io/github/parroty/excoveralls

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,10 @@ before_script:
   - sh -e /etc/init.d/xvfb start
   - npm install -g npm
   - npm install
+env:
+  - MIX_ENV=test
 script:
-  - "mix do local.hex --force, deps.get && mix test --include pandoc"
+  - "mix do local.hex --force, deps.get, test --include pandoc, coveralls.travis"
   - node_modules/.bin/gulp
 cache:
   directories:


.travis.yml
```diff
@@ -19,8 +19,10 @@ before_script:
   - sh -e /etc/init.d/xvfb start
   - npm install -g npm
   - npm install
+env:
+  - MIX_ENV=test
 script:
-  - "mix do local.hex --force, deps.get && mix test --include pandoc"
+  - "mix do local.hex --force, deps.get, test --include pandoc, coveralls.travis"
```
josevalim: This is not going to work because the `mix test` will complain it runs on the wrong environment. Keep the && and do `mix do` twice.

josevalim:
We should be good on coveralls side!

.travis.yml
milmazz: But I already exported the `MIX_ENV=test` via `.travis.yml`, I don't see any complain [here](https://travis-ci.org/elixir-lang/ex_doc/builds/80292616).

.travis.yml
```diff
@@ -19,8 +19,10 @@ before_script:
   - sh -e /etc/init.d/xvfb start
   - npm install -g npm
   - npm install
+env:
+  - MIX_ENV=test
```
milmazz: Setting environment variables from `.travis.yml`

.travis.yml
josevalim: You are right.  Sorry :)

josevalim:
:shipit:

.travis.yml
milmazz: What a relieve, for a moment I was scratching my head :sweat_smile:

PR #387, opened 2015-09-16
[content] Improve footer text and display

### LINKED ISSUE #383: Create new footer
Originally posted here https://github.com/elixir-lang/ex_doc/issues/374 by me.

### FOOTER
1) I think it should have the font a bit smaller (right now its 16px as the rest of the page)

2) Optionally a tad bit greyer, not full black.

3) Can we do a preferred breaking point in between the two sentences. Right now, if i used a zoomed view (125%) in my PC broswer , or in landscape view in my phone,  I see it like:
> Designed by Friedel Ziegelmayer and the ExDoc team. Built using ExDoc
> (v0.9.1-dev).

and I think it's better if it will be 
> Designed by Friedel Ziegelmayer and the ExDoc team.
> Built using ExDoc (v0.9.1-dev).

The only concern is that this won't add a fourth line to the portrait view in mobiles, but, if we the font sized is decided to be reduced a bit, it's quiet possible it will fit into three lines (that's the number it takes now).
there are tricks for this http://stackoverflow.com/a/24357132 but there could be one for CSS3, since that post is a bit old.

4) I will change the order of the sentences. I will go with:
> Built using ExDoc (v0.9.1-dev). Designed by Friedel Ziegelmayer and the ExDoc team.

instead of 
> Designed by Friedel Ziegelmayer and the ExDoc team. Built using ExDoc (v0.9.1-dev). 

the reason is that when I'm reading the docs, usually i see the copyright notice at the bottom of the page, and what for latter gives me the impression is that the current project that i'm reading its documentation has been designed by "Friedel Ziegelmayer & friends". Placing ExDoc before removes the ambiguity.

### Relevant Comments:

@josevalim [commented](https://github.com/elixir-lang/ex_doc/issues/374#issuecomment-140036195):
> Regarding the footer, maybe: "Built using ExDoc (v0.9.1-dev), designed by Friedel Ziegelmayer." It is more concise and because ExDoc is coming first to avoid the ambiguity, I would keep the design focus only on @Dignifiedquire.

@eksperimental agrees 100% with @josevalim's comments
and
@Dignifiedquire "pretty much agree[s] with everything [@eksperimental] said here :)"

Dignifiedquire:
This improves the footer in three ways

* Reduce the contrast by using a gray font color
* Simplify the wording
* Specify preferred line breaks

Closes #383.

screen shot 2015-09-16 at 18 57 04

cc @eksperimental


eksperimental:
looks much better now ;-D

josevalim:
:shipit:

PR #394, opened 2015-09-17
[assets] Use custom highlight.js build

### LINKED ISSUE #386: Update highlight.js to v8.8.0

Dignifiedquire:
This provides a good size reduction of the JavaScript file to 164kb minified.

Closes #386


assets/less/app.less
```diff
@@ -1,5 +1,5 @@
 @import 'npm://font-awesome/less/font-awesome';
-@import 'npm://highlightjs/styles/tomorrow';
+@import 'npm://highlight.js/src/styles/tomorrow';
```
eksperimental: are we no longer using the previous github style?
if not, we should update the LICENSE.
https://github.com/elixir-lang/ex_doc/blob/master/LICENSE#L49-L56

Dignifiedquire: Since the new design we use `tomorrow` from the highlight.js repo

eksperimental: should we add them to the license? @josevalim ?

eksperimental: because now we are using a lot of projects. i think it will be hard to keep track of everything. what should we do?

josevalim: If we keep the license in the generated app.js file, we can remove them from license. Otherwise we will need to copy them all.

eksperimental:
@Dignifiedquire i'm running `gulp build --type production` and i'm getting an error saying tomorrow is not avaialble. how do I install all needed packages. o how to I update them if needed. thanks.

Dignifiedquire:
@eksperimental just run `npm install` that should fix it

eksperimental:
@Dignifiedquire that worked! thanks a lot.. I didn't want to play with commands trying to figure out which one it was... you already know why.

assets/less/app.less
eksperimental: @Dignifiedquire I don't think we are keeping any license in the app.js file. is there an option for this? or should i create an issue to update all the licenses to license.

Dignifiedquire: no licenses in there atm

PR #401, opened 2015-09-18
Update generate_assets to be a public function

milmazz:
This change will allow reuse the `generate_assets` function from others formatters (e.g. EPUB)

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -74,6 +74,11 @@ defmodule ExDoc.Formatter.HTML do
      {"fonts/*.{eot,svg,ttf,woff,woff2}", "fonts"}]
   end
 
+  # TODO: decouple EPUB/HTML
+  @doc """
+  Copy a list of assets into a given directory
+  """
+  @spec generate_assets(list, String.t) :: :ok
   def generate_assets(source, output) do
     Enum.each source, fn({ pattern, dir }) ->
       output = "#{output}/#{dir}"
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -84,7 +84,7 @@ defmodule ExDoc.Formatter.HTML do
   def generate_assets(source, output) do
     Enum.each source, fn({ pattern, dir }) ->
       output = "#{output}/#{dir}"
-      File.mkdir output
+      File.mkdir! output
 
       Enum.map Path.wildcard(pattern), fn(file) ->
         base = Path.basename(file)
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -1,5 +1,7 @@
 defmodule ExDoc.Formatter.HTML do
-  @moduledoc false
+  @moduledoc """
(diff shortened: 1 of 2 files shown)


josevalim:
We should probably create then a new module, called ExDoc.Assets and move the assets out of the HTML formatter. To put it in another way, the EPUB formatter should not call the HTML formatter. They should share facilities though. :)

josevalim:
Btw, the same logic should apply to autolink and friends... if you need anything in EPUB, let's move them out from HTML. :)

eksperimental:
actually now that @josevalim  mentions this.. 
don't we need an Markdown Formatter that we can call from other formatters?
I also sometimes when trying to find bugs in ExDoc find myself wanting to see the markdown representation of the site in the directory, separated by files.

josevalim:
The markdown formatter is already split and moved to ExDoc.Markdown, so this part at least we are done :D

milmazz:
> We should probably create then a new module, called ExDoc.Assets and move the assets out of the HTML formatter

I agree on this.

> the same logic should apply to autolink and friends

Also, I agree on this.

> To put it in another way, the EPUB formatter should not call the HTML formatter

Before I proceed with some changes, the following are the current dependencies from *EPUB* to *HTML*.

* `ExDoc.Formatter.EPUB` depends on:
 - `ExDoc.Formatter.HTML.Autolink`
 - `ExDoc.Formatter.HTML.generate_assets`
 - `ExDoc.Formatter.HTML.filter_list`
 - `ExDoc.Formatter.HTML.process_logo_metadata`
* `ExDoc.Formatter.EPUB.Templates` depends on:
 - `ExDoc.Formatter.HTML.Templates.group_types`
 - `ExDoc.Formatter.HTML.Templates.to_html`
 - `ExDoc.Formatter.HTML.Templates.summary_template`
 - `ExDoc.Formatter.HTML.Templates.type_detail_template`
 - `ExDoc.Formatter.HTML.Templates.detail_template`
 - `ExDoc.Formatter.HTML.Templates.get_specs`
 - `ExDoc.Formatter.HTML.Templates.link_id`
 - `ExDoc.Formatter.HTML.Templates.synopsis`
 - `ExDoc.Formatter.HTML.Templates.h`

I think that's all, but you get the idea :)

@josevalim What do you suggest to do with all this dependencies?

josevalim:
I need to look at the EPUB brunch with more detail but I can't do it now. Let's make generate_assets public and add a "# TODO: decouple EPUB/HTML" and I can take another pass later splitting everything and finding the proper boundary, right before or after merging the EPUB support. The main point is: if another module is calling it, the API is then public, and it must be documented, unit tested, etc. Otherwise coupling will bite us later. :)

josevalim:
So to make it clear, go ahead with coupling everything you need to couple, once all dependencies have been identified, let's discuss how to split it apart.

milmazz:
@josevalim Great!, in another PR I'll begin to document more functions and add specs to them, if you like this please merge into `master`.

josevalim:
Just to clarify, @milmazz, don't worry about specing and moving those functions around now. Let's get the EPUB work done, then we refactor.

PR #404, opened 2015-09-18
Load font css from https: when required

eksperimental:
This fixes the issue with loading a css file from unsecure
source when the docs are loaded from https://

### DIFF
--- a/lib/ex_doc/formatter/html/templates/head_template.eex
+++ b/lib/ex_doc/formatter/html/templates/head_template.eex
@@ -19,7 +19,7 @@
     <script>
       var cdn_protocol;
       cdn_protocol = ( window.location.protocol == 'https:' ) ? "https:" : "http:";
-      document.write('<link href="' + cdn_protocol + '\/\/fonts.googleapis.com\/css?family=Lato:400,300,700,900|Merriweather:300italic,300,700,700italic|Inconsolata:400,700" rel="stylesheet" type="text/css">');
+      document.write('<link href="' + cdn_protocol + '\/\/fonts.googleapis.com\/css?family=Lato:400,300,700,900%7CMerriweather:300italic,300,700,700italic%7CInconsolata:400,700" rel="stylesheet" type="text/css">');
     </script>
     <link rel="stylesheet" href="dist/app.css">
     <script src="dist/sidebar_items.js"></script>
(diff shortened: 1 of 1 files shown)


lib/ex_doc/formatter/html/templates/head_template.eex
```diff
@@ -16,8 +16,12 @@
     <% end %>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta name="generator" content="ExDoc v<%= ExDoc.version %>">
-    <link href="http://fonts.googleapis.com/css?family=Lato:400,300,700,900|Merriweather:300italic,300,700,700italic|Inconsolata:400,700" rel="stylesheet" type="text/css">
-    <link rel="stylesheet" href="dist/app.css" />
+    <script>
+      var cdn_protocol;
```
eksperimental: @milmazz  @Dignifiedquire is it alright creating this global var, if not just let me know what is the best way to do it.

josevalim:
:+1:! Let's wait for folks to validate the JS.

Dignifiedquire:
Sorry but this makes no sense. `//` already is a shortcut to say ise `http` or `https` depending on the currently used protocol.

Dignifiedquire:
so no need for any js here.

eksperimental:
yes, I thought of that @Dignifiedquire this deals with the special case of file://
sorry I forgot to mention it in the commit.

eksperimental:
@Dignifiedquire can you give me your input on the javascript?
I will rebase it cand check this one tomorrow.
I think we still need it, because I don't think it's necessary to download fonts via https when loading the page from http. (it's more traffic bc it's encrypted, right?)

josevalim:
It seems we don't need this because we ship the fonts?

Dignifiedquire:
We do not ship the fonts only the icons. But it is still not needed as we now always load via https and there is no good reason to not use https at any point.

eksperimental:
I don't agree with that, 
serving the fonts over https on an http:/ file:  page adds an overhead in time (encrypting/decrypting) and data transfer (an these fonts are a couple hundread kb) for no real benefit.

josevalim:
So please reopen and rebase so we can get this in. :D

PR #420, opened 2015-09-19
Makes ExDoc generate Valid HTML5 pages

eksperimental:
This commits fixes every single bug,
thus it makes the docs generated (checked only on the Elixir docs)
valid HTML5 documents.

I throughly tested everything,
we didn't have that many issues. but they were hard to fix,
I have sent a few a few PRs to earmark because some issues came from there.
there is on last PR that needs to be reviewed and merged.

I have validated the Elixir docs using html-proofer and html5validator,
no internal broken links found, only one issue and we are 100% valid.
I will may need help to fix it.

OK. trying to push this so we can test the longer the better..

### DIFF
--- a/lib/ex_doc/formatter/html/templates/sidebar_template.eex
+++ b/lib/ex_doc/formatter/html/templates/sidebar_template.eex
@@ -24,7 +24,7 @@
 
   <div class="sidebar-search">
     <i class="icon-search"></i>
-    <input type="text" class="sidebar-searchInput" placeholder="search" autocomplete="off" results="0" />
+    <input type="text" class="sidebar-searchInput" placeholder="search" autocomplete="off" />
   </div>
 
   <ul class="sidebar-listNav">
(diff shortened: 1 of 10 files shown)


lib/ex_doc/formatter/html/templates/sidebar_template.eex
```diff
@@ -24,7 +24,7 @@
 
   <div class="sidebar-search">
     <i class="icon-search"></i>
-    <input type="text" class="sidebar-searchInput" placeholder="search" autocomplete="off" results="0" />
+    <input type="text" class="sidebar-searchInput" placeholder="search" autocomplete="off" />
```
eksperimental: @Dignifiedquire I have removed "results" was it was not a valid attribute, I think it's webkit specific. Could you please add this via Jquery? So we don't need to worry about it when we run the validators? let me know if that's still ok.

Dignifiedquire: I don't think we use it anymore, so thanks for cleaning up :)

Dignifiedquire:
Thanks :octocat: lgtm

lib/ex_doc/formatter/html/templates/sidebar_template.eex
eksperimental: even better then

PR #421, opened 2015-09-20
Load font css from https: when required

eksperimental:
originally published in #404


josevalim:
@eksperimental weren't we using inline css for some other reason?

josevalim:
Because of this: https://github.com/elixir-lang/ex_doc/commit/f87c8f217dbd2862165d204dd4d023a2ca419942

eksperimental:
The blank page issue was an issue with my system configuration. It was just a work around, but it didnt work.

ericmj:
Why is javascript needed to load css? Why isn't `//` enough?

josevalim:
// won't work with file://

-- 


ericmj:
In which sense? Does it error or just use http?

Also, why not always use https? We are loading from an external CDN that we know supports https.

eksperimental:
@ericmj `//` will cause when loaded from `file://` to call 
`file://fonts.googleapis.com/css?family=Lato:400,300,700,900|Merriweather:300italic,300,700,700italic|Inconsolata:400,700`

my concern calling https when non needed was unnecessary overhead (for encrypting/decypting) and the higher data transfers due to encrypted data.
I have done some tests with wireshark,

its is __http: 217kb__ vs __https: 230kb__
it's only 13kb, which is only 5% of the data transfered via http.

I'll leave it up to you guys.

ex_doc_http_tcpdump

ex_doc_https_tcpdump

josevalim:
That's really cool @eksperimental! 5% is an OK price to pay though, so I think we are good. :)

PR #422, opened 2015-09-20
Add sort function by name/arity

eksperimental:
this fixes this bug that the lists in the summary were not sorted.
http://elixir-lang.org/docs/master/elixir/Kernel.html#summary

I created a `ExDoc.sort` function that sort functions by their name and their arities in case they share the same name.

### DIFF
--- a/lib/ex_doc.ex
+++ b/lib/ex_doc.ex
@@ -52,6 +52,20 @@ defmodule ExDoc do
     struct(preconfig, options)
   end
 
+  @doc """
+  Sorts ExDoc structs by name and arity.
+  """
+  @spec sort(Enum.t) :: Enum.t
+  def sort(collection) do
+    Enum.sort(collection, fn(x, y) ->
+      if x.name == y.name do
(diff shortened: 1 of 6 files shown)


eksperimental:
@milmazz and @josevalim I didn't have enough time to look into the function that sorts the sidebar lists,
could you point me to the line of code that takes care of th sorting. I would like to check that it is consistent with this new function.
from what i can see in the sidebar it is.

josevalim:
@eksperimental great catch! I have pushed a different solution for this to master that sorts the data already in the retriever: 922644c3b337a23c1302a275d652cf2c2ddff52e. This way we guarnatee everything ahead is using properly sorted data.

eksperimental:
yes, that is much better @josevalim 
I wrote a test today, that loads the docs, shuffles them and sort them with the option sort function I created. and the output is exactly the same in all the 160 something beam files in the elixir project.
so we are good.
I submitted some fixes to your commit in https://github.com/elixir-lang/ex_doc/pull/425

PR #424, opened 2015-09-21
fix handleAnchor

dukex:
I was seeing the [Ecto.Changeset doc](http://hexdocs.pm/ecto/Ecto.Changeset.html) and the anchors on sidebar doesn't works in Firefox. 

The problems is the global ```event``` is undefined in Firefox. I changed to use event from handler argument

Error: ```ReferenceError: event is not defined```
Firefox: 40.0.3 
OS: UbuntuGNOME linux


josevalim:

dukex:
:heartbeat:

PR #425, opened 2015-09-21
Validate anchor links in sidebar + fix summary sorting

eksperimental:
enc() helper added.

Checked by htmlproof and html5validator

Also removed the Enum.sort() that messed up the order in summary

### DIFF
--- a/lib/ex_doc/formatter/html/templates.ex
+++ b/lib/ex_doc/formatter/html/templates.ex
@@ -121,7 +121,7 @@ defmodule ExDoc.Formatter.HTML.Templates do
   end
 
   defp sidebar_items_object(id, anchor) do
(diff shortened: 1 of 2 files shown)


lib/ex_doc/formatter/html/templates.ex
```diff
@@ -121,7 +126,7 @@ defmodule ExDoc.Formatter.HTML.Templates do
   end
 
   defp sidebar_items_object(id, anchor) do
-    ~s/{"id":"#{id}","anchor":"#{anchor}"}/
+    ~s/{"id":"#{id}","anchor":"#{enc(anchor)}"}/
```
josevalim: Why not call `URI.encode` directly here and remove the indirection? :)

eksperimental: done ;-D

PR #426, opened 2015-09-21
[WIP] Create proper links in sidebar handlebar template

eksperimental:
This creates proper links, allowing me to view the generated html,
save it and run in against html link checker and validators

I created the isNotLocal helper, because i couldn't make isLocal
helper work with `unless`.


josevalim:
I have reviewed the commit and I couldn't see what really changed... why do we need the unless in all those nodes? If we generate `href="/Module#function"` we are still fine... no?

eksperimental:
i tried that, but for the local page, it was reloading it.

eksperimental:
it was not that, i wouldn't work at all.
It is not straight what's happening there. there is a function that removes the hash, so `Module.html#function` will take you to `#Module.htmlfunction`

this is needed by `scrollTo` helper.

josevalim:
@eksperimental maybe we can push the local logic to scroll_to then? If it is local there, we just move to `#...`. This way, the link generation remains clear.

Dignifiedquire:
This will screw up the javascript logic, as it relies on these links being achor links

eksperimental:
well, you only access the one that are local, so you never set the page.html.
can we have something like we have in the search results?

Issue #428, opened 2015-09-22
Close Search button not working


eksperimental:
now i get a javascript error about MetaKey

josevalim:
How come? Which browser are you using? It is supposed to be supported on all of them: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/metaKey

eksperimental:
sorry, it is not when i use the metakey, but when i click on the button. 
the error is about the metakey

> Uncaught TypeError: Cannot read property 'metaKey' of undefined

josevalim:
I got that. The question is the same: How come? Which browser are you using? `metaKey` exists on all mouse events on all browsers: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/metaKey

eksperimental:
Chromium 43 / Ubuntu

eksperimental:
firefox gives me this:
> TypeError: event is undefined

josevalim:
I have no idea, the event is received explicitly:

https://github.com/elixir-lang/ex_doc/blob/93ee626aef65b56557914c05630f738dc6410ba0/assets/js/search.js#L117-L118

Maybe change both to `e` and see if it works?

PR #429, opened 2015-09-22
Fix error when closing search

eksperimental:
Closes https://github.com/elixir-lang/ex_doc/issues/428

Can you test it on mac?

the issue was called by calling closeResults with `()`

I had to put it in two separate calls, because preventDefault() would not work. and the page would be reloaded.


eksperimental:
second commit push.
it fixes https://github.com/elixir-lang/ex_doc/issues/430
I have added the `altKey` to the exception just in case it has a special meaning in certain custom configurations.
it deals with right and middle button click.

/cc @Dignifiedquire

eksperimental:
@josevalim can you test this on a mac? so I can merge

assets/js/search.js
```diff
@@ -114,9 +114,13 @@ function search (nodes, value) {
   $oldContent.hide()
   $content.append($results)
 
-  function closeResults (event) {
-    if (event.metaKey) {
-      return
+  function closeResults (e) {
+    var event = e || window.event
+    if (typeof event !== 'undefined' && typeof event === 'object' && event !== null) {
```
josevalim: Honestly, this looks like a smell. Don't  we only need `typeof event === 'object'`? Can it have a type of an object and still be null?

eksperimental: well, we I guess we can get rid of `typeof event !== 'undefined'`, 
but we need the rest because.

```js
> typeof null
"object"
```

josevalim:
Meregd manually

Issue #423, opened 2015-09-20
Javascript Error: Uncaught TypeError: Cannot read property 'top' of undefined


eksperimental:
originally posted: https://github.com/elixir-lang/ex_doc/commit/2d3c9fbadaeea9d532cb614726ef8d1cbd82ee44#commitcomment-13331577

@Dignifiedquire [this line](https://github.com/elixir-lang/ex_doc/commit/2d3c9fbadaeea9d532cb614726ef8d1cbd82ee44#diff-e9cc8298549e98be96351e6176bc85d5R21
) is giving me a JS error,
```
Uncaught TypeError: Cannot read property 'top' of undefined
scrollTo
@ helpers.js?415c:21identifyCurrentHash
@ events.js?0bbb:151initialize
@ events.js?0bbb:166(anonymous function)
@ app.js?113f:20jQuery.Callbacks.fire
@ jquery.js?27d9:3099jQuery.Callbacks.self.fireWith
@ jquery.js?27d9:3211jQuery.extend.ready
@ jquery.js?27d9:3417completed
@ jquery.js?27d9:3433
```

Dignifiedquire:
On which page/action?

eksperimental:
It happens when the element in the #hash in the url doesn't exist.
for example http://elixir-lang.org/docs/master/elixir/Kernel.SpecialForms.html#%/2

eksperimental:
I will try to fix this one myself. if no one else wants to jump in

josevalim:
Please go ahead, ping me on IRC if you want to chat.

PR #432, opened 2015-09-24
fix JS error when element doesn't exit on helper.scrollTo

eksperimental:
Closes https://github.com/elixir-lang/ex_doc/issues/423


josevalim:
do you know why and which element does not exist?

eksperimental:
nope, now it works on all elements, because all the links in the sidebar have been fixed.

eksperimental:
this one was one
http://elixir-lang.org/docs/master/elixir/Kernel.SpecialForms.html#%/2

but just enter an id that wouldn't exist to test it.

PR #439, opened 2015-09-26
remove trailing semicolon from  Synopsis

eksperimental:
this commit improves the removal of trailing punctuation from summary
semicolon has been added as it appeared showing up in a quiet a few
cases in summaries

tests added

ExDoc.Formatter.HTML.Templates.api_reference_page/5 added to ease testing.

and also added specs for  ExDoc.Retriever.docs_from_*

### DIFF
--- a/lib/ex_doc/formatter/html/templates.ex
+++ b/lib/ex_doc/formatter/html/templates.ex
@@ -47,10 +47,23 @@ defmodule ExDoc.Formatter.HTML.Templates do
     end
   end
 
-  # Get the first paragraph of the documentation of a node, if any.
-  defp synopsis(nil), do: nil
-  defp synopsis(doc) do
-    String.split(doc, ~r/\n\s*\n/) |> hd |> String.strip() |> String.rstrip(?.)
+  @doc """
+  Gets the first paragraph of the documentation of a node. It strips
+  surrounding spaces and strips traling `:` and `.`.
+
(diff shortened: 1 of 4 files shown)


lib/ex_doc/formatter/html/templates.ex
```diff
@@ -14,6 +14,15 @@ defmodule ExDoc.Formatter.HTML.Templates do
                     modules, exceptions, protocols)
   end
 
+  @doc """
+  Generate content for the api-reference page, provided the `modules`,
+  `exceptions` and `protocols`.
+  """
+  def api_reference_page(title, modules, exceptions, protocols, config) do
```
josevalim: It seems we are using this only on tests?!

eksperimental: yes. same for the function above: ExDoc.Formatter.HTML.Templates.module_page/5

eksperimental: Can we find a better place for helpers like these ones?

josevalim: module_page is called by Formatter.HTML. We should not add a function if its purpose is to be called by tests. Let's make this easier: make synopsis public and just call it directly with strings. No need to add fixtures. We will need to make synopsis public anyway when we support EPUB.

eksperimental:
we are good to go

josevalim:

PR #445, opened 2015-10-04
Code clean up

milmazz:
This commit include the following fixes:

 * Promote that each function chain begins with a bare value
 * Put a final newline in files where is missing
 * Remove the usage of the match operator in conditional constructs
 * Avoid the usage of tautological predicate names in functions
 * Replace with ~S or ~s those strings that contains double quote chars
 * Remove trailing whitespace at the end of a line


milmazz:
@eksperimental Could you please review this?

lib/ex_doc.ex
```diff
@@ -56,11 +56,12 @@ defmodule ExDoc do
   defp find_formatter(modname) when is_atom(modname), do: modname
 
   defp find_formatter("ExDoc.Formatter." <> _ = name) do
-    Module.concat([name]) |> check_formatter_module(name)
+    [name] |> Module.concat |> check_formatter_module(name)
```
eksperimental: I think the rule we have is one |> operator, we put it in one line, and more we break them into several lines.

lib/ex_doc/formatter/html.ex
```diff
@@ -44,7 +44,7 @@ defmodule ExDoc.Formatter.HTML do
   end
 
   defp normalize_config(%{main: "index"}) do
-    raise ArgumentError, message: "\"main\" cannot be set to \"index\", otherwise it will recursively link to itself"
+    raise ArgumentError, message: ~S("main" cannot be set to "index", otherwise it will recursively link to itself)
```
eksperimental: I'm thiking if we should use a less common delimiter than "()" that can be easily found in message (regardless of whether this message has any parenthesis.
maybe ~S{...}  is better

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -59,7 +59,8 @@ defmodule ExDoc.Formatter.HTML.Templates do
   def synopsis(nil), do: nil
   def synopsis(""),  do: ""
   def synopsis(doc) when is_bitstring(doc) do
-    String.split(doc, ~r/\n\s*\n/)
+    doc
+    |> String.split(~r/\n\s*\n/)
     |> hd
```
eksperimental: I'm working seein the line bellow, if we should do `|> hd()`, i know hd has only one function with arity one, but in the unlinkely event of a having hd/2 in the future, this code can break, i guess. just a convention maybe that can be good to follow. what do you guys think?

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -79,7 +80,7 @@ defmodule ExDoc.Formatter.HTML.Templates do
 
   @doc false
   def enc_h(binary) do
-    URI.encode(binary) |> h
+    binary |> URI.encode |> h
```
eksperimental: I will break this into three lines

eksperimental: actually add parenthesis.
```elixir
binary
|> URI.encode()
|> h()
```

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -119,7 +120,8 @@ defmodule ExDoc.Formatter.HTML.Templates do
       ~s/{"id":"#{node.id}","title":"#{node.id}"}/
     else
       types =
-        group_types(node)
+        node
+        |> group_types
```
eksperimental: |> group_types()

lib/ex_doc/markdown/hoedown.ex
```diff
@@ -12,10 +12,11 @@ defmodule ExDoc.Markdown.Hoedown do
...
   def to_html(text, opts \\ []) do
-    Markdown.to_html(text,
-      autolink: Keyword.get(opts, :autolink, true),
-      fenced_code: Keyword.get(opts, :fenced_code, true),
-      tables: Keyword.get(opts, :tables, true))
+    text
+    |> Markdown.to_html(
+         autolink: Keyword.get(opts, :autolink, true),
+         fenced_code: Keyword.get(opts, :fenced_code, true),
+         tables: Keyword.get(opts, :tables, true))
```
eksperimental: I will ident it only with two spaces.
you can read an interesting discussion about this https://github.com/elixir-lang/elixir/pull/3640#discussion-diff-41086777
read @whatyouhide's comment, i think it is a really neat way to address this indentention issue

eksperimental: particularly in this case, i would be more inclined to leave it as it was before, because it's only two blocks and the first one has so many options, I think it's more readable  as it was. but it doens't really matter to be honest.

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -22,20 +22,20 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   defp all_docs(module, modules) do
     locals = Enum.map module.docs, &(doc_prefix(&1) <> &1.id)
 
-    if moduledoc = module.moduledoc do
```
josevalim: This syntax is fine. :) I know some languages warn, but we are fine, really.

test/ex_doc/cli_test.exs
```diff
@@ -32,7 +32,7 @@ defmodule ExDoc.CLITest do
     File.write!("test.config", ~s(%{"extras" => "README.md"}))
 
     assert_raise RuntimeError,
-                 "expected a keyword list from config file: \"test.config\"",
+                 ~S(expected a keyword list from config file: "test.config"),
                  fn -> run(["ExDoc", "1.2.3", "...", "-c", "test.config"]) end
```
eksperimental: Just a style comment, i would format it like this these assert_raise functions:
```elixir
assert_raise RuntimeError,
  ~S(expected a keyword list from config file: "test.config"), fn ->
    run(["ExDoc", "1.2.3", "...", "-c", "test.config"])
  end
```

test/ex_doc/retriever_test.exs
```diff
@@ -147,7 +147,8 @@ defmodule ExDoc.RetrieverTest do
   end
 
   test "undocumented callback implementations get default doc" do
-    [node] = docs_from_files(["CustomBehaviourOne", "CustomBehaviourTwo", "CustomBehaviourImpl"])
+    [node] = ["CustomBehaviourOne", "CustomBehaviourTwo", "CustomBehaviourImpl"]
+             |> docs_from_files
              |> Enum.filter(&match?(%ExDoc.ModuleNode{id: "CustomBehaviourImpl"}, &1))
```
eksperimental:
```elixir
    [node] =
      ["CustomBehaviourOne", "CustomBehaviourTwo", "CustomBehaviourImpl"]
      |> docs_from_files()
      |> Enum.filter(&match?(%ExDoc.ModuleNode{id: "CustomBehaviourImpl"}, &1))
```

eksperimental:
I have sent a few comments.
:+1: with the use of sigils,
it's pretty much the indentation (which is a style thing only) and the user of `()` when using the pipe operator.
other than that it looks good to me, but i wasn't able to read every line, as I'm kind of leaving...

lib/ex_doc/markdown/hoedown.ex
josevalim: Calculate the options before and just pass `options` as argument.

lib/ex_doc/formatter/html.ex
milmazz: Until now I didn't scape any `(` symbol inside of a `~S` or `~s`. I think that the delimiter usage depends of the context, for example, in some tests you can find `~s[..]`

lib/ex_doc.ex
milmazz: :+1:

lib/ex_doc/formatter/html/templates.ex
milmazz: I agree to change `hd` to `hd()` to be consistent. But, just to be clear, the *pattern matching* will avoid the issue that you mention:

> but in the unlinkely (sic) event of a having hd/2 in the future, this code can break

lib/ex_doc/formatter/html/templates.ex
milmazz: :+1:

lib/ex_doc/formatter/html/templates.ex
milmazz: :+1:

lib/ex_doc/formatter/html/autolink.ex
milmazz: Yes, I know that the syntax is valid :+1:

I just feel (please, correct me if I'm wrong) that this kind of tricks at some point could lead a bad practice, I don't expect to do a *match* inside of a *conditional statement* (the original goal), it's like I'm trying to accomplish more than one thing in just one statement. Also, it could be possible that somebody assume something like this: *Ah, this guy actually wants to do a `==` and after that change :boom:*, or what happens in case of a bad match? I know this not the case, but just thinking about another possible scenarios.

test/ex_doc/cli_test.exs
milmazz: :+1:

test/ex_doc/retriever_test.exs
milmazz: :+1:

milmazz:
@eksperimental Thanks for your suggestions, I already updated this PR. /cc @josevalim

milmazz:
@eksperimental @josevalim Is this ready to merge? wdyt?

josevalim:
:+1: from here

PR #448, opened 2015-10-06
improved instrucitons for getting help

ghitchens:
Getting help for ex_doc wasn't obvious.   First you have to find the `mix help docs` reference which was in a non-obvious place, and then you had to figure out why it didn't work (because deps hadn't been compiled yet).    

I moved the line describing this under using with mix and added guidance about building deps.


josevalim:

PR #451, opened 2015-10-15
Give example of -e flag.

Havvy:


eksperimental:
Use /extra-page.md instead of /docs/extra-age.md since this docs is the def. Dir for generating docs

PR #452, opened 2015-10-15
Remove `/doc` from `-e` example

Havvy:
As per eksperimental's note in #451.


josevalim:
No, it is alright. We generate stuff to "doc" singular and we don't want people to dump everything into root. :)

eksperimental:
We are deleting the output dir before generating any docs.
So that file will be deleted.
https://github.com/elixir-lang/ex_doc/blob/da430a9ad3a4c2d7021498dad78d7bea055d461d/lib/ex_doc/formatter/html.ex#L18

josevalim:
The example use /docs and we output to /doc

-- 

PR #454, opened 2015-10-26
Fix autolink anchors for non URL safe function names

padde:
The function names need to be URL escaped in the auto-generated links as well. Other spots I found like the links in the sidebar were already corrected.

For example [this section in the elixir docs](http://elixir-lang.org/docs/stable/elixir/Kernel.SpecialForms.html#::/2) links to `<<>>/1`, which generates the following markup:

```html
<a href="#&lt;&lt;&gt;&gt;/1">...</a>
```

However, it should be like this

```html
<a href="#%3C%3C%3E%3E/1">...</a>
```

Note that I also changed the specs, which now assert that surrounding spaces are removed. This is now in line with the way `Autolink.project_functions` works, which also strips spaces.


josevalim:

PR #455, opened 2015-10-28
Add a max-width property to img tags

whatyouhide:
This way `` tags don't go far over the right margin of the page, like here:

screen shot 2015-10-28 at 17 12 24

What does the ex_doc team thinks of this? Btw as far as I can tell (I'm not the best you'll find at CSS) this is what GitHub does in their READMEs.


Dignifiedquire:
Good idea, do you have a link to a documentation where this is an issue currently?

whatyouhide:
@Dignifiedquire nope, it was causing them for a documentation I was generating but that I still don't publish on hexdocs. But to reproduce you can just take any big-ish image and shove it into a markdown file, that's pretty much what I did :)

milmazz:
@whatyouhide Could you please also provide the bundled CSS file?. You can find more info about this process [here](https://github.com/elixir-lang/ex_doc/tree/master/assets)

whatyouhide:
@milmazz oops, sorry :) `push --force`'d the CSS file, let me know how it looks :).

eksperimental:
@whatyouhide I think you will have to rebase to the latest master, since there's a merge conflict

whatyouhide:
@eksperimental whoops, forgot to do that :) Sorry today my brain is being stupid. `push --force`'d again!

milmazz:
@whatyouhide Thanks!, your change now is in `master`

PR #453, opened 2015-10-28
** (ExDoc.Retriever.Error) module MyListLexer was not compiled with flag --docs

sunaku:
Hello,

I'm using ex_doc 0.10.0 with Elixir 1.1.1 and Erlang 18.1 under 64-bit Linux. :smiley:

[Following this guide](http://andrealeopardi.com/posts/tokenizing-and-parsing-in-elixir-using-leex-and-yecc/), I created a `src/Elixir.My.List.Lexer.xrl` file containing:

```erlang
Definitions.

INT        = [0-9]+
ATOM       = :[a-z_]+
WHITESPACE = [\s\t\n\r]

Rules.

{INT}         : {token, {int,  TokenLine, list_to_integer(TokenChars)}}.
{ATOM}        : {token, {atom, TokenLine, to_atom(TokenChars)}}.
\[            : {token, {'[',  TokenLine}}.
\]            : {token, {']',  TokenLine}}.
,             : {token, {',',  TokenLine}}.
{WHITESPACE}+ : skip_token.

Erlang code.

to_atom([$:|Chars]) ->
    list_to_atom(Chars).
```

Afterwards, I get the following error when running `mix docs`: :scream:

```sh
$ mix docs
Compiled src/Elixir.My.List.Lexer.xrl
Compiled src/Elixir.My.List.Lexer.erl
...
Generated my app
** (ExDoc.Retriever.Error) module My.List.Lexer was not compiled with flag --docs
    lib/ex_doc/retriever.ex:82: ExDoc.Retriever.verify_module/1
    lib/ex_doc/retriever.ex:71: ExDoc.Retriever.get_module/2
    (elixir) lib/enum.ex:1043: anonymous fn/3 in Enum.map/2
    (elixir) lib/enum.ex:1387: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir) lib/enum.ex:1043: Enum.map/2
    lib/ex_doc/retriever.ex:50: ExDoc.Retriever.docs_from_modules/2
    lib/ex_doc.ex:37: ExDoc.generate_docs/3
    lib/mix/tasks/docs.ex:88: Mix.Tasks.Docs.run/3

exit 1
```

Unfortunately, `:leex.file/2` (which compiles the `*.xrl` file into a `*.erl` file) does not support the `--docs` compiler flag that ex_doc is telling me to use. :cold_sweat:  How would you suggest fixing this error? :bug:

Thanks for your consideration.

### DIFF
--- a/test/ex_doc/retriever_test.exs
+++ b/test/ex_doc/retriever_test.exs
@@ -121,6 +121,16 @@ defmodule ExDoc.RetrieverTest do
     end
   end
 
+  test "docs_from_modules warns when module was not compiled with flag --docs" do
+    module = ExDoc.RetrieverTest.ModuleNotCompiledWithFlagDocs
+    {:ok, ^module} = :compile.file('#{__DIR__}/retriever_test/#{module}.erl')
+    "Elixir." <> module_name_suffix = module |> to_string
+
+    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
+      Retriever.docs_from_modules([module], %ExDoc.Config{})
+    end) == "module #{module_name_suffix} was not compiled with flag --docs\n"
+  end
+
   ## EXCEPTIONS
 
   test "docs_from_files properly tags exceptions" do
--- a/test/ex_doc/retriever_test.exs
+++ b/test/ex_doc/retriever_test.exs
@@ -121,14 +121,14 @@ defmodule ExDoc.RetrieverTest do
     end
   end
 
-  test "docs_from_modules warns when module was not compiled with flag --docs" do
+  test "docs_from_modules warns when module does not export __info__/1" do
     module = ExDoc.RetrieverTest.ModuleNotCompiledWithFlagDocs
     {:ok, ^module} = :compile.file('#{__DIR__}/retriever_test/#{module}.erl')
     "Elixir." <> module_name_suffix = module |> to_string
 
     assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
       Retriever.docs_from_modules([module], %ExDoc.Config{})
-    end) == "module #{module_name_suffix} was not compiled with flag --docs\n"
+    end) == "module #{module_name_suffix} does not export __info__/1\n"
   end
 
   ## EXCEPTIONS
--- a/test/ex_doc/retriever_test/Elixir.ExDoc.RetrieverTest.ModuleNotCompiledWithFlagDocs.erl
+++ b/test/ex_doc/retriever_test/Elixir.ExDoc.RetrieverTest.ModuleNotCompiledWithFlagDocs.erl
@@ -1,0 +1,4 @@
+% This Erlang module is deliberately named like an Elixir module in order to
+% trigger a warning.  See https://github.com/elixir-lang/ex_doc/issues/453
+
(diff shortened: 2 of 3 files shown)


sunaku:
I've converted this issue into a PR by attaching my first attempt to change the error into a warning. :gift: 

Any tips on how to capture `:stderr` (maybe in a StringIO) to assert the warning in the test case? :confused:

sunaku:
Okay, I figured out how to capture IO in the test case.  Please review my PR and merge. :game_die:

sunaku:
I've updated my patch to use a plain old Erlang module instead of going through leex. :bulb:

test/ex_doc/retriever_test.exs
```diff
@@ -121,6 +121,16 @@ defmodule ExDoc.RetrieverTest do
     end
   end
 
+  test "docs_from_modules warns when module was not compiled with flag --docs" do
+    module = ExDoc.RetrieverTest.ModuleNotCompiledWithFlagDocs
+    {:ok, ^module} = :compile.file('#{__DIR__}/retriever_test/#{module}.erl')
+    "Elixir." <> module_name_suffix = module |> to_string
+
+    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
```
milmazz: Why not add this as an import at the beginning of the file and then just call `capture_io` here?

```elixir
import ExUnit.CaptureIO
```

test/ex_doc/retriever_test/Elixir.ExDoc.RetrieverTest.ModuleNotCompiledWithFlagDocs.erl
```diff
@@ -0,0 +1,4 @@
+% This Erlang module is deliberately named like an Elixir module in order to
+% trigger a warning.  See https://github.com/elixir-lang/ex_doc/issues/453
+
+-module('Elixir.ExDoc.RetrieverTest.ModuleNotCompiledWithFlagDocs').
```
milmazz: This looks like a fixture, right? Is there any issue to put this file into `test/fixtures`?

milmazz:
@sunaku Thanks for your PR, I made a few suggestions, please review it.

Also, please, do not forget to update this comment:

> Get all the information from the module and compile
>  it. If there is an error while retrieving the information (like
>  the module is not available or it was not compiled
>  with --docs flag), we raise an exception.

See `get_module` [function](https://github.com/elixir-lang/ex_doc/pull/453/files#diff-10b022e5b0ae070e758670dc4876efeeL79).

@josevalim @ericmj The approach that @sunaku is following is the right one? wdyt?

josevalim:
I don't think it is correct because we have two situations:

1. An Erlang module
2. An Elixir module compiled with --docs

We should still raise for the second, but not for the first. I think the best is to check if the module exports the __info__ function and, if not, simply ignore it. No need to warn.

sunaku:
:+1: PR updated according to Jose's suggestion.  Sorry for the delay. :sweat_smile:

sunaku:
:tada: Hurray! :+1: Thanks for merging! :bow:

josevalim:

PR #458, opened 2015-11-02
Add project's name in the content on small screens

### LINKED ISSUE #435: [Mobile] No sign of what project's documentation we are reading
If you follow the link for the main page for the documentation:
http://elixir-lang.org/docs/master/elixir/

there is no sign of what is the documentattion project we are reading about.

img_20150920_141614 small

the same for the default page (API reference) if no landing page is specified.
img_20150925_151456 small

milmazz:
Closes #435 

capture


josevalim:
Let's wait @Dignifiedquire to sign this one off. :)

lib/ex_doc/formatter/html/templates/dist/app.css
```diff
@@ -1 +1,1137 @@
...
\ No newline at end of file
+@import url('https://fonts.googleapis.com/css?family=Lato:400,300,700,900|Merriweather:300italic,300,700,700italic|Inconsolata:400,700');
```
Dignifiedquire: You forgot to run the minified build

Dignifiedquire:
Just need to rebuild with minified then all good :+1:

milmazz:
@josevalim @Dignifiedquire Done!

eksperimental:
How abouy including the version number?

milmazz:
@eksperimental I wanted to keep it simple, if the user wants to know the version they can perfectly view it in the menu. Right?

eksperimental:
The user landing from the elixir website or anywhere else will only see the project name and may not that he can find the version.

milmazz:
Done!

screen shot 2015-11-03 at 8 00 00 am

Dignifiedquire:
:+1:

eksperimental:
I am wondering what is the best way to tackle this.

Semantically speaking, the project name is part of the title (h1).
I am wondering if that's ok, and if it makes sense to put a small inside a h1 (I'm just wondering, I just don't have a taken position on this).

Aesthetically speaking, I am thinking is that the project name should be at least in bold.
I have played with making the version number 80% of the project name.
here's a screenshot.
ex_doc_project_name_h1

here's is another version with the project name with the same size as h1. and line-height smaller.
ex_doc_project_name_title_big

I think in the mobile version give to project name the same  relevance as the title of the page is the right way to go.

what do you guys think about this?

eksperimental:
this the edited HTML code for the last screenshot.

```html
<h1 style="line-height: 1.2em; ">
  <small class="visible-xs" style="font-weight: 800; font-size: 1em; ">Mix <small style="font-weight: 800; ">v1.2.0-dev</small></small>
  API Reference
</h1>
```

which makes me wonder if it makes any sense to still use small at all.

PR #459, opened 2015-11-02
Assets revision

### LINKED ISSUE #408: [assets] Generate hashed assets to ensure cache busting on cdns
When running `gulp build --type production` it should generate 

```
app-<hash>.js
app-<hash>.css
```
and the elixir process needs to pic up those names to insert them correctly into the templates

cc @josevalim

milmazz:
Closes #408 

/cc @Dignifiedquire @josevalim


milmazz:
@Dignifiedquire @josevalim How we handle the case that after execute `gulp build --type=production` now we get new files for `app-<hash>.{css,js}` from the Git point of view? Just delete the old files and add the new ones? wdyt?

Dignifiedquire:
@milmazz I guess that's the only solution at the moment that I know :/

lib/ex_doc/formatter/html/templates/footer_template.eex
```diff
@@ -13,6 +13,6 @@
   </div>
 </section>
 </div>
-    <script src="dist/app.js"></script>
+  <script src="<%= asset_rev Path.expand(config.output), "dist/app-*.js" %>"></script>
```
josevalim: Maybe we should call `Path.expand` inside `asset_rev`? :)

josevalim:
:shipit:

lib/ex_doc/formatter/html/templates/footer_template.eex
milmazz: Thanks for the feedback, fixed on master 62857db

PR #460, opened 2015-11-03
Identify external links in the content area

### LINKED ISSUE #374: Issues and Suggestions
## List of issues:

* __Footer__
  - [x] All issues Moved to #383
* __Sidebar__
  - [x] 1) Reduce size in version number: Moved to #366
  - [x] 2) If not logo provided, reduce size: Moved to #366
  - [x] 3) Overview white color/bolder: Obsolete b/c of changes in #395  ~~[Proposal](https://github.com/elixir-lang/ex_doc/issues/374#issuecomment-140719583)~~
* __Content Area__
  - [x] 1) Lineup title the Project name in sidebar. (cc0b0b2f63170a484c04e3c899cb62f139e58ef0)
  - [ ] 2) Linked Code
  - [x] 3) Underline color of Code inside link
* __General__
  - [x] 1) Rename html elements with underscore (assigned to: @eksperimental)
  - [ ] 2) External link ICON missing
  - [ ] 3) Vertical spacing Summary without description. [Proposal](https://github.com/elixir-lang/ex_doc/issues/374#issuecomment-140043635)

======

_Disclaimer: I apologize in advance for any duplicate reporting, specially on issues that have been discussed and agreed before, what I ´m looking as a reference are the current open issues, so if anything 's been discussed I appreciate a link to the discussion._

### FOOTER
**THE WHOLE FOOTER DISCUSSION HAS BEEN MOVED TO #383**
**FIXED in #387**

1) I think it should have the font a bit smaller (right now its 16px as the rest of the page)
**FIXED in #387**

2) Optionally a tad bit greyer, not full black.
**FIXED in #387**

3) Can we do a preferred breaking point in between the two sentences. Right now, if i used a zoomed view (125%) in my PC broswer , or in landscape view in my phone,  I see it like:
> Designed by Friedel Ziegelmayer and the ExDoc team. Built using ExDoc
> (v0.9.1-dev).

and I think it's better if it will be 
> Designed by Friedel Ziegelmayer and the ExDoc team.
> Built using ExDoc (v0.9.1-dev).

The only concern is that this won't add a fourth line to the portrait view in mobiles, but, if we the font sized is decided to be reduced a bit, it's quiet possible it will fit into three lines (that's the number it takes now).
there are tricks for this http://stackoverflow.com/a/24357132 but there could be one for CSS3, since that post is a bit old.
**FIXED in #387**

4) I will change the order of the sentences. I will go with:
> Built using ExDoc (v0.9.1-dev). Designed by Friedel Ziegelmayer and the ExDoc team.

instead of 
> Designed by Friedel Ziegelmayer and the ExDoc team. Built using ExDoc (v0.9.1-dev). 

the reason is that when I'm reading the docs, usually i see the copyright notice at the bottom of the page, and what for latter gives me the impression is that the current project that i'm reading its documentation has been designed by "Friedel Ziegelmayer & friends". Placing ExDoc before removes the ambiguity.
**FIXED in #387**

### SIDEBAR

1) I think the space in the sidebar is a [constant constraint](#366). One thing that we can definitly reduce its size is the version number. maybe may it bold? I know it will still be 64px high, because of the size of the logo, This will reduce the size with projects with long names: https://github.com/elixir-lang/ex_doc/issues/361#issuecomment-139825040
**DISCUSSION MOVED TO #366**

2) If no logo is provided, i think we can reduce the size of the header (most projects will not provide a logo i guess) (see point 1 for more info)
**DISCUSSION MOVED TO #366**

3) OVERVIEW should be WHITE when view the overview page. right now there is not distinction from when viewing any other page. and it's confusing, because what looks like it's being selected is MODULES. 
** OBSOLTE DUE TU CHANGES IN #395 **
~~**PROPOSAL: [HERE](https://github.com/elixir-lang/ex_doc/issues/374#issuecomment-140719583)**~~

### CONTENT AREA

1) (suggestion) Wouldn't it be neater to have the title of the page lined up at the same height of the project name in the side bar?
ex_doc_title_linedup

2) Link containing `code` doesn't display underline under the code area. Making it impossible to determine whether its a link to Module/function or not.
ex_doc_code_link_underlined

This behaviour is not consistent. When zoommed in (110%) it will show it.
See the difference between 100% and 110% zoom
ex_doc_code_link
ex_doc_code_link_zoomed

3) Change color of the underline for code inside links, I think pure black creates too much distraction. i find it too strong compared to how light code is styled. the challenged would be to make it work where code and regular text are part of the link.

### GENERAL

1) Get rid of the elements with underscore and use only hyphen. Example: `<h1 id="modules_summary" class="section-heading">Modules summary</h1>` 
I think this should be fixed ASAP, since we are going to start having people linking to from the sidebar to particular places like: Access.html#callbacks_details

2) External link ICON. We lost this feature in the new design. It's not possible to know know whether the link will still take you inside the docs or outside.

3) Vertical spacing Summary without description: It looks a bit odd when most of the elements in the summary have no description, but one does. http://elixir-lang.org/docs/master/elixir/overview.html#exceptions_summary
ex_doc_exceptions_odd
**PROPOSAL: [HERE](https://github.com/elixir-lang/ex_doc/issues/374#issuecomment-140043635)**

_That's it for now. It looks really great. I'm impressed by the quality of everything. I will post more of the usual stuff as I get to use it. Thank you._

milmazz:
Closes #374 

@Dignifiedquire, please review this.


assets/less/icons.less
```diff
@@ -29,22 +29,25 @@
 }
 
 .icon-link:before {
-	content: "\e005";
```
Dignifiedquire: Please remove the indentation changes

Dignifiedquire:
Can't build right now, could you add a screenshot how it looks now please?

assets/less/icons.less
milmazz: Why? Actually this is to normalize the indentation. Other `.less` files have 2 spaces for indentation.

Dignifiedquire: sorry, was confusing red and green :D

milmazz: :+1:

milmazz:
@Dignifiedquire Sure!

capture

eksperimental:
This looks great @milmazz and not intrusive..which is very important.. i also cannot build now. Just one question, when the link has not been yet visited, do the link and the icon look the same color?

josevalim:
I am, honestly, not a fan. It pollutes the links and I don't think it adds useful information. If I care about this information, my browser shows it to me on mouse over. I think there are even extensions that does this? I would prefer to keep the data clean as it is today.

However, given my argument is based on aesthetics, I will leave my judgement to @Dignifiedquire. If he thinks this is an improvement, then please go ahead.

milmazz:
@josevalim Sad to read that, but that's ok :(

Let's wait then for @Dignifiedquire's opinion

The following screenshot shows a different behavior for visited external links as @eksperimental suggested:

screen shot 2015-11-03 at 9 15 48 pm

eksperimental:
thanks for the screenshots @milmazz . it looks great.

one thing to consider is that in mobile devices there's no way to know the link unless you long press the link.

Dignifiedquire:
I'm afraid I have to agree with @josevalim :/ I don't think the added value justifies the added visual clutter, sorry.

milmazz:
@Dignifiedquire @josevalim No problem at all, I'll close this and #374

PR #468, opened 2015-11-07
Get rid of the 'extras-' prefix

milmazz:
This PR removes the `extra-` prefix from the additional documentation (specified with the option `extras`), this was suggested by @wsmoak, you can see more details about the discussion [in the mailing list](https://groups.google.com/d/msg/elixir-lang-core/UzhyIhvQsC8/S_W5sk3yDAAJ) and #462

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -6,7 +6,7 @@ defmodule ExDoc.Formatter.HTML do
   alias ExDoc.Formatter.HTML.Templates
   alias ExDoc.Formatter.HTML.Autolink
 
-  @main "extra-api-reference"
+  @main "api-reference"
 
   @doc """
   Generate HTML documentation for the given modules
@@ -56,7 +56,7 @@ defmodule ExDoc.Formatter.HTML do
 
   defp generate_api_reference(modules, exceptions, protocols, output, config) do
     content = Templates.api_reference_template(config, modules, exceptions, protocols)
-    File.write!("#{output}/extra-api-reference.html", content)
+    File.write!("#{output}/api-reference.html", content)
   end
 
   defp generate_not_found(modules, exceptions, protocols, output, config) do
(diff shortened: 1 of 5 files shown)


wsmoak:
Happy to see the "extras-" prefix removed!  Thank you again for working on this.

It looks like this introduced another feature, (on line 106) and you can now configure: `extras: [ {"path/to/file.md", "Custom Menu Title"}, "path/to/otherfile.md", ]`

At least that's what I thought it was. :)  Looking closer, the second element of the tuple is named `output_file_name` and it does get used as both the menu title *and* the name of the output file.  

So I get a nice menu title, but the url ends in Custom%20Menu%20Title.html . 

I don't think the menu item needs to be tied to the filename... what do you think of this change? 
https://github.com/elixir-lang/ex_doc/compare/master...wsmoak:preserve-filename

With the config above, you would get a menu title of "Custom Menu Title" and the output in "file.html".

lib/ex_doc/formatter/html.ex
```diff
@@ -100,35 +100,74 @@ defmodule ExDoc.Formatter.HTML do
           generate_extra(&1, output, module_nodes, modules, exceptions, protocols, config)
          end))
       |> Enum.map(&Task.await/1)
-    [{"extra-api-reference", "API Reference", []}|extras]
+    [{"api-reference", "API Reference", []}|extras]
+  end
+
+  defp generate_extra({input_file, output_file_name}, output, module_nodes, modules, exceptions, protocols, config) do
+    title = input_to_title(output_file_name)
```
josevalim: It should be `input_to_title(input)` (always based on input file).

lib/ex_doc/formatter/html.ex
```diff
@@ -100,35 +100,74 @@ defmodule ExDoc.Formatter.HTML do
...
   end
 
-  @h2_regex ~r/^##([^#].*)$/m
+  defp valid_extension_name?(input) do
+    file_ext =
+      input
+      |> Path.extname()
+      |> String.downcase()
+
+      if file_ext in [".md"] do
```
josevalim: Wrong indentation. :)

milmazz: Ups, fixed in ac6a9a6

josevalim: :heart:

Issue #450, opened 2015-10-09
RFC: More options for extra pages


bitwalker:
So my only issue with ExDoc (and it's a small one), is that I can't seem to control the way extra pages are named and ordered in the sidebar. I would really like the ability to override those. Thoughts?

josevalim:
Thanks @bitwalker. They should be ordered by the order they are defined in your mix.exs file with the same name as in the filesystem. So ordering should not be an issue... and can't you name the files as you want them to show up?

bitwalker:
@josevalim The seem to alphabetically sorted, because they show up in a different order than I defined them. As far as naming goes, let's say I have `getting_started.md`, this is translated to `GETTING_STARTED.html` and displays as `GETTING_STARTED` in the sidebar. Perhaps I'm doing something wrong, but I'd like to be able to name it more naturally, i.e. "Getting Started".

josevalim:
I will take a look at the ordering thing. About the naming, just call the file "Getting Started.md" and you will be fine. :)

bitwalker:
@josevalim What about something like "Upgrades/Downgrades" or anything which uses special characters? I mean, there are workarounds obviously (like choosing a different name), but having the ability to explicitly define the name in `mix.exs` decouples the naming from how the files themselves are organized on disk.

bitwalker:
I am of course coming from the perspective of not knowing how complicated the ask is here, so maybe this is a lot of added complexity for little benefit?

josevalim:
I just checked and in 0.10.0 I used this:

    extras: ["guides/README.md", "guides/Getting Started.md"]

The order kept was the one I defined. Regarding the naming, pls just pick a different character for now. I know it is not ideal but there are other things we would rather focus (like supporting custom assets and so on). :)

bitwalker:
@josevalim Ah damn, didn't check my mix.lock, I was on 0.9.0. Sorry about that!

wsmoak:
I ran into this as well.  Perhaps ExDoc could use the text of the first `<h1>` as the menu item?  I see that it already makes sub-menu items out of the `<h2>`'s. (Lines beginning with `#` and `##`, respectively.)

wsmoak:
Can this be reopened please?

I've got my idea of taking the text of the first `<h1>` as the menu item working in this branch:
https://github.com/wsmoak/ex_doc/tree/h1-as-title

[...]

PR #469, opened 2015-11-08
Preserve the original filename for extras

wsmoak:
https://github.com/elixir-lang/ex_doc/pull/468 introduced the ability to specify the menu item text for an extra file, but the text is also used as the output filename, leading to things like Custom%20Menu%20Title.html.

This preserves the original filename, so my-extra-file.md simply becomes my-extra-file.html, and the second element of the tuple is only used as the menu item text.

(Note that José had a different comment on line 107.)

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -103,12 +103,11 @@ defmodule ExDoc.Formatter.HTML do
     [{"api-reference", "API Reference", []}|extras]
   end
 
-  defp generate_extra({input_file, output_file_name}, output, module_nodes, modules, exceptions, protocols, config) do
-    title = input_to_title(output_file_name)
+  defp generate_extra({input_file, menu_title}, output, module_nodes, modules, exceptions, protocols, config) do
 
     options = %{
-      title: title,
-      output_file_name: title,
+      title: menu_title,
+      output_file_name: Path.basename(input_file, ".md"),
       input: to_string(input_file),
       output: output
     }


josevalim:
The discussion in the other thread is that the value in keyword lists would be the filename. However, we could make the value be the title, as in this PR. Another alternative is to make the title be the first header (`#`) as today it is also inflected from the filename.

wsmoak:
Yes, when "extras-" was still there, I wanted the ability to stop the prefix from getting added, and it seemed like being able to configure the filename would be good.  But without that... why change the filename at all?

I would also like to make it use the first `<h1>` as the menu title by default, (and I had that working before the change in PR 468.)  But I like this ability to specify a custom menu title as well, if the file doesn't have a `<h1`> for some reason.

If we go back before the "extras-" prefix was ever added... what's the simplest behavior that makes sense?

josevalim:
I am honestly not sure. Today, the filename is used for three things:

1. It represents how you want those stored in disk
2. inflecting the url. For example, "Intro to Phoenix.md" becomes "intro-to-phoenix.html";
3. inflecting the title. Which is copied from the filename as is;

For example, I could have this: the "1 Intro to Phoenix.md" filename must become "intro-to-phoenix.html" in the URL and have title "Intro to Phoenix". In this case, we want to customize both url and title. That said what I had in mind is that we could use `:extras` to configure the url (solving 1 and 2) because we can always inflect the title from the first header. However, I guess someone could always say the title in the menu does not have to be the one in the page.

Maybe the best option after all is to allow the following:

    "foo.md": [path: "foo-bar", title: "Foo bar"]

This way, Phoenix can even organize everything with numbers, let's say:

* 1 Intro.md
* 2 Models.md
* 3 Controllers.md

And you will be able to generate a menu without numbers in both title and URL as follows:

```elixir
File.ls!("guides") |> Enum.sort |> Enum.map(fn name ->
  {_int, " " <> title} = Integer.parse(name)
  {Path.join("guides", name),
   title: Path.rootname(name),
   url: title |> String.downcase |> String.replace(" ", "-")}
end)
```

Thoughts?

wsmoak:
(I am not sure where the numbers in the filename come from?  The order is already defined in the configuration.  Currently Phoenix uses A_filename.md, B_other_filename.md but I think that is just a ReadMe.io requirement.)

Yes, over on https://github.com/elixir-lang/ex_doc/issues/450 I think I arrived at the same idea... let them configure BOTH the title and the output filename, since no two of the input filename, output filename, 
menu item and first h1 really need to be tied together.

Closing this one and I will try again (unless @milmazz beats me to it!) :)

milmazz:
@wsmoak @josevalim So, just to be clear, now we want to include in the `extras` option the following:

* Path of the additional files as the key (as always).
* Per each listed file the user can set the path of the output (for now we just take the `Path.basename |> Path.rootname`), as a fallback we select the same name as the input (in this case we apply `String.downcase |> String.replace(" ", "-")` to that name).
* Per each listed file the user can set the title, as a fallback we select the header.

Is this the new requirement? @wsmoak Are you working on this?

wsmoak:
Not working on it at the moment @milmazz!  I think José wanted the output filename to "look like" the menu title, but I may have gotten lost along the way.  

I'm actually fine with a modified output filename as long as it's sane (no %20's) (and doesn't get an unwanted prefix of course. :D)  Making it all configurable and grabbing the h1 by default would be great.

For the h1 I mostly had a copy of the h2 regex, except I used Regex.run because it will stop at the first match.  Looking forward to seeing the solution, thank you again!

PR #471, opened 2015-11-11
Use first <h1> as menu title if title is not configured

wsmoak:
Take the first line beginning with `# ` (the first `<h1>`) as the menu title, if a title wasn't specified in the configuration.

ExDoc already uses the `<h2>`'s as sub-menu items, so this seems natural to do.

Without this, I have to duplicate the header text in both the markdown file and the configuration.

Advice on implementation and testing is appreciated!  @bbense suggested using multiple defined functions instead of the conditional, see [1], but this seems to fit better with the other code in the module.

[1] http://wiki.wsmoak.net/cgi-bin/wiki.pl?ElixirSlack20151110

### DIFF
--- a/test/ex_doc/formatter/html_test.exs
+++ b/test/ex_doc/formatter/html_test.exs
@@ -191,6 +191,14 @@ defmodule ExDoc.Formatter.HTMLTest do
     assert content =~ ~r{"id":"readme","title":"Getting Started"}
    end
 
+  test "run uses first <h1> as menu title" do
+    generate_docs(doc_config(extras: ["test/fixtures/ExtraPage.md"]))
+    content = File.read!("#{output_dir}/extrapage.html")
+    assert content =~ ~r{<title>Extra Page Title – Elixir v1.0.1</title>}
+    content = File.read!("#{output_dir}/dist/sidebar_items.js")
+    assert content =~ ~r{"id":"extrapage","title":"Extra Page Title"}
+  end
+
   test "run normalizes options" do
     # 1. Check for output dir having trailing "/" stripped
     # 2. Check for default [main: "api-reference"]
--- a/test/fixtures/ExtraPage.md
+++ b/test/fixtures/ExtraPage.md
@@ -1,0 +1,15 @@
+# Extra Page Title
+
+some text
+
+## Section One
+
+more text
+
+# Another H1
+
+even more text
+
+## Section Two
+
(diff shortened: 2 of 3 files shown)


bbense:
defp extract_title(content) do
+    title = Regex.run(@h1_regex, content, capture: :all_but_first)
+
+    if title do
+      title |> List.first |> String.strip
+    end
+  end

You can also write this as 

defp extract_title(content) do
    case Regex.run(@h1_regex, content, capture: :all_but_first) do
       nil -> nil 
       title_list -> String.strip(List.first( title_list))
    end
end

I think the multiple function version is nice if you can use the function in more than one place. If it's just a one off then either the if version or the case version seems fine to me.

test/fixtures/ExtraPage.md
```diff
@@ -0,0 +1,15 @@
+# Extra Page Title
```
josevalim: Maybe we should just add a `#` to the README instead of adding a new fixture?

test/ex_doc/formatter/html_test.exs
```diff
@@ -191,6 +191,15 @@ defmodule ExDoc.Formatter.HTMLTest do
     assert content =~ ~r{"id":"readme","title":"Getting Started"}
    end
 
+   test "run uses first <h1> as menu title" do
+    generate_docs(doc_config(extras: ["test/fixtures/ExtraPage.md"]))
+    content = File.read!("#{output_dir}/extrapage.html")
+    assert content =~ ~r{<title>Extra Page Title – Elixir v1.0.1</title>}
+    content = File.read!("#{output_dir}/dist/sidebar_items.js")
+    assert content =~ ~r{"id":"extrapage","title":"Extra Page Title"}
+
```
josevalim: Extra space (sorry for nitpicking!)

josevalim:
Thank you @wsmoak! I have added two comments.

test/fixtures/ExtraPage.md
wsmoak: I tried that, but it broke some of the existing tests.  It seems that if we add a `#` to the README, then we can't test the case where there is no `#` in the file, no `title` in the config, and it falls back to the original behavior of using the filename as the menu title.

josevalim: You are right. :)

josevalim:

PR #477, opened 2015-11-16
fix(ExDoc.Retriever): fix invalid arguments for callbacks

Gazler:
When a callback is defined, an attempt is made to derive the argument
type from the name. If the typespec has more than one type option then
the argument name is given as "|"

An explicit check for this format is now in place, and the "arg#{i}"
case is used.


test/ex_doc/retriever_test.exs
```diff
@@ -135,7 +135,7 @@ defmodule ExDoc.RetrieverTest do
     functions = Enum.map node.docs, fn(doc) -> doc.id end
     assert functions == ["hello/1"]
     assert hd(node.docs).type == :callback
-    assert hd(node.docs).signature == "hello(integer)"
+    assert hd(node.docs).signature == "hello(arg0)"
```
josevalim: Let's add a new argument so we can test both things. :)

josevalim:

PR #480, opened 2015-11-17
Unsafe variables and string() type

milmazz:
* Avoid warnings on variables defined inside a conditional (Elixir v1.0.2-dev)
* Avoid `string()` type, for strings, use `String.t()` instead.


josevalim:

Issue #492, opened 2015-12-01
Opportunistically Expand Sidebar Width


Havvy:
Current state: 

![Sidebar horizontal bar showing with main content whitespace taking up 2/3 of main content](http://i.imgur.com/qnO9vmA.png)

Expected: Given the extreme amount of whitespace around the main content, the sidebar would have a max-width that takes it into account so the horizontal scrollbar isn't necessary.

pdilyard:
I think this is a good idea, considering it could help eliminate horizontal scrolling on a lot of screens. If others agree, I'd love to make a PR for this one.

milmazz:
I agree if the idea is just to remove the horizontal scroll in the sidebar area. Please go ahead :)

PR #494, opened 2015-12-02
Take advantage of extra space on large screens by widening sidebar

pdilyard:
Solves #492 

This solution uses media queries and an extra JS function to cater to different screen widths.

While working on this feature, I encountered several different ways this could be done.

I started by just adding a couple CSS media queries on the sidebar, which works great until you show/hide the sidebar. Because the width of the sidebar is statically set in the JS file that animates it, as soon as you toggle the sidebar, the new CSS gets overridden.

The solution I went with for this PR is to add CSS media queries, and change the width variable in JS to change based on screen width. After the sidebar is opened, the inline styles that get added are removed so the CSS takes over again. This probably isn't the absolute *best* solution, but it works and it's simple.

Here are some other potential ideas if we decide this isn't good enough:

1. **Refactor to use CSS animations.** This solution would probably be the best, but would also take the most amount of work.
2. **Disable the show/hide ability on larger screens.** If your screen is big enough to have this much extra whitespace, you probably don't have much of a need to hide the sidebar. This would be a really quick, simple, and reliably solution.

### DIFF
--- a/assets/js/sidebar.js
+++ b/assets/js/sidebar.js
@@ -11,13 +11,24 @@ const sidebar = $('.sidebar')
 
 const bodyClass = 'sidebar-closed'
 const duration = 300
-const width = '300px'
 const displayProps = [
   '-webkit-flex',
   '-ms-flexbox',
   '-ms-flex',
   'flex'
 ]
+
+function getWidth () {
+  const docWidth = $(document).width()
+  if (docWidth >= 1800) {
+    return '425px'
+  } else if (docWidth >= 1600) {
+    return '375px'
+  }
+  return '300px'
+}
+
+let width = getWidth()
 
 function closeSidebar () {
   sidebar.animate({
@@ -35,6 +46,8 @@ function closeSidebar () {
 }
 
 function openSidebar (immediate) {
+  width = getWidth()
+
   body.removeClass(bodyClass)
   displayProps.forEach(prop => {
     sidebar.css({display: prop})
@@ -49,7 +62,9 @@ function openSidebar (immediate) {
     '-ms-flex-basis': width,
     'flex-basis': width,
     width: width
-  }, duration)
+  }, duration, function () {
+    sidebar.attr('style', '')
+  })
 }
 
 function toggleSidebar () {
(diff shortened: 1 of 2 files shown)


Havvy:
:heart: :blue_heart: :green_heart: :purple_heart: :yellow_heart:

josevalim:
Thank you! Waiting on @Dignifiedquire to merge it. :)

assets/js/sidebar.js
```diff
@@ -19,6 +18,18 @@ const displayProps = [
   'flex'
 ]
 
+var width = getWidth()
```
Dignifiedquire: Please use `let` for mutable state.

assets/js/sidebar.js
```diff
@@ -19,6 +18,18 @@ const displayProps = [
...
+
+function getWidth () {
```
Dignifiedquire: Please define functions before they are used

pdilyard:
Updated as per @Dignifiedquire's feedback

assets/js/sidebar.js
```diff
@@ -49,7 +62,9 @@ function openSidebar (immediate) {
     '-ms-flex-basis': width,
     'flex-basis': width,
     width: width
-  }, duration)
+  }, duration, function () {
+    sidebar.attr('style', '')
```
josevalim: I think this line introduced a bug where the sidebar never really toggles for small screens. Why do we need it?

pdilyard: I believe I put this line here because the styles of the animation get set inline, so they override the CC media queries. When the animation completes, this line removes the inline styles and lets the CSS take over.

If you take this line away, and then adjust the size of your screen, the sidebar width won't adjust dynamically. It's probably easier to explain by just commenting out the line and trying resizing your browser window and playing with the sidebar opening and closing.

josevalim: I have no idea. I would love if you could help me find out. The bug report is here: #505 and you can reproduce by just making the window small.

josevalim:
Sorry, I had to revert this commit on master. I have tried to fix it by about 30 minutes and I couldn't come up with anything and I need to ship Elixir v1.2. :)

pdilyard:
Understood, not a big deal. I played around for a bit and I really can't make sense of why this only effects small screens and not "normal" sized ones.

PR #501, opened 2015-12-20
Add support for cmark.ex markdown processing.

tvon:
Note that I'm not sure if it's appropriate to do this or create something external, but it's simple enough so here we are.

This is basically an edited copy-pasta of the `ExDoc.Markdown.Earmark` module.  I'm new to elixir so please accept my apologies if I did anything senseless here.


josevalim:
Thank you so much, everything is perfect. :)

PR #502, opened 2015-12-22
Outdent function headers

joefreeman:
I find that the function headers look a bit like code blocks (e.g., those in specs or examples) so was finding it difficult to scan through the documentation to find a function. This really small change simply outdents the function headers so that the text itself lines up with the content below it (rather than having the background/purple bar line up).

screen shot 2015-12-22 at 15 27 02


alco:
I also have been confused by this style issue. I have another idea for solving this: make the function head background extended to the whole width of the page.

milmazz:
@joefreeman Thanks for your PR! Please keep in mind that you also need to generate `lib/ex_doc/formatter/html/templates/dist/app-*.css` file with `gulp build --type=production`.

@Dignifiedquire @josevalim This is not the first time that somebody mention this *function header* issue, so, I think that we need to discuss some approach to avoid more confusion.

Dignifiedquire:
Let's indent the blocks inside the functions, rather than outdent the headers, to keep clean gutters.

joefreeman:
@alco Yeah, that might look nice. I do like the purple bit though.

@milmazz I've run that command, thanks. (Should the artefacts really be in the repo? If so, would it make sense to add this to the git-commit hook?)

@Dignifiedquire Agree it'd be nice to keep the gutters clean, so I tried that, but it started to feel a bit hacky, since I think the headings and other sections should also line up with the content.

Thanks all of you for taking a look at this. Good to hear that it's not just me that was having this issue.

josevalim:
Hi everyone, I am closing this one because I think the other related changes we pushed, with reducing function headers and so on have been sufficient to remove a lot of the ambiguity and confusion we had here. Thank you!

PR #503, opened 2015-12-25
support doc of @macrocallback

tony612:
I try to generate docs for ecto master, but it failed. I found it's related to `@macrocallback`, so I fixed it. I didn't test this because ex_doc needs to support Elixir 1.0 for the moment, but this should work.


josevalim:

Issue #523, opened 2016-03-23
space bar key is broken when navigating to module documentation pages


Bogdanp:
I generally find the way documentation is handled in this community very nice, but as an Elixir newbie reading lots of docs this issue in particular has been driving me a little nuts. :D

**Steps to reproduce:**

1. Go to https://hexdocs.pm/ecto/Ecto.html
2. Click on `Ecto.Adapter` in the sidebar
3. Hit space

**Expectation:**

The module documentation should scroll up.

**Actual behavior:**

It does not.

PR incoming.

PR #524, opened 2016-03-23
focus on content element on document ready

Bogdanp:
This fixes an issue (#523) where hitting the space bar after loading a module's documentation doesn't work.

I set the tabindex through JS rather than changing the template because I felt like it would be clearer to a reader since the whole thing is in one place.


josevalim:

PR #525, opened 2016-03-23
Sort specs by line to preserve ordering

### LINKED ISSUE #519: specs should be displayed in the same order as they have been defined
See the example of Elixir.Range.range?/1

https://github.com/elixir-lang/elixir/blob/b6dc389f5e3eb2286885c172009333ef414d7860/lib/elixir/lib/range.ex#L73-L74
```elixir
@spec range?(%Range{}) :: true
@spec range?(term) :: false
```

and how it's showed in the docs
http://elixir-lang.org/docs/master/elixir/Range.html#range?/1
```
Specs:
range?(term) :: false
range?(%Range{first: term, last: term}) :: true
```

the order changes the meaning.

bjfish:
Fixes #519

### DIFF
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -193,6 +193,7 @@ defmodule ExDoc.Retriever do
     specs = all_specs
             |> Map.get(function, [])
             |> Enum.map(&Typespec.spec_to_ast(name, &1))
+            |> Enum.reverse()
 
     %ExDoc.FunctionNode{
       id: "#{name}/#{arity}",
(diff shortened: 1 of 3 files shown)


lib/ex_doc/retriever.ex
```diff
@@ -193,6 +193,7 @@ defmodule ExDoc.Retriever do
     specs = all_specs
             |> Map.get(function, [])
             |> Enum.map(&Typespec.spec_to_ast(name, &1))
+            |> Enum.sort( fn({_,l,_}, {_,l2,_}) -> l[:line] < l2[:line] end )
```
josevalim: I wonder if they are just in reverse order and calling `Enum.reverse` would be enough?

bjfish:
I've updated to use `Enum.reverse`. The test still passes.

josevalim:

eksperimental:
Thank you @bjfish

PR #526, opened 2016-03-23
Remove unneeded enc_h

### LINKED ISSUE #509: Anchor for <<>> is not working in Firefox
In the documentation for [`Kernel.SpecialForms.<<>>/1`](http://elixir-lang.org/docs/stable/elixir/Kernel.SpecialForms.html#%3C%3C%3E%3E/1) the generated anchor is not linking to the correct section in Firefox 43.0.4.

bjfish:
Should fix #509 

It appears that the id's should not be URI encoded and valid id's are any UTF-8 except for space characters:

http://stackoverflow.com/questions/70579/what-are-valid-values-for-the-id-attribute-in-html
> HTML 5 is even more permissive, saying only that an id must contain at least one character and may not contain any space characters.

http://stackoverflow.com/questions/10474972/named-anchor-a-with-name-same-as-a-div-id-conflict
> If this step was not skipped and there is an element in the DOM that has an ID exactly equal to decoded fragid, then the first such element in tree order is the indicated part of the document; stop the algorithm here.

### DIFF
--- a/lib/ex_doc/formatter/html/templates/detail_template.eex
+++ b/lib/ex_doc/formatter/html/templates/detail_template.eex
@@ -1,4 +1,4 @@
-<div class="detail" id="<%=enc_h link_id(node) %>">
+<div class="detail" id="<%= link_id(node) %>">
   <div class="detail-header">
     <a href="#<%=enc_h link_id(node) %>" class="detail-link" title="Link to this <%= pretty_type(node) %>">
       <i class="icon-link"></i>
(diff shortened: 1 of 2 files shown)


lib/ex_doc/formatter/html/templates/detail_template.eex
```diff
@@ -1,4 +1,4 @@
-<div class="detail" id="<%=enc_h link_id(node) %>">
+<div class="detail" id="<%= link_id(node) %>">
```
josevalim: Can we call `h link_id(node)` instead? We don't need to encode, but we still need to escape it, right? :) Thank you! :heart:

bjfish:
I'm going to close this because after testing I think javascript changes would be needed to make this change.

Also, I don't fully understand the escaping yet. For example, I don't understand why `<%= %>` converts `&lt;` back to `<`'
```
 <div id="<%=h "<<>>/1" %>" 
 <div id="<<>>/1"> 
 
 <div id="<%= "&lt;&lt;&gt;&gt;/1" %>" 
 <div id="<<>>/1"> 
```

PR #529, opened 2016-03-30
Fixes #528, autolink regexp now allows digits in function names.

### LINKED ISSUE #528: 'See also' breaks for method names that contain numbers
I really like the feature where you can create a local link to another function in the documentation by typing its signature (such as `Module.function_name/1` ), enclosed in backtics. However, while writing on the [Math](https://github.com/folz/math) package, I found out that names like `Math.log10/1` or `Math.atan2/2`do not automatically generate these references.

Qqwy:
Note that the regexp is updated in a way that theoretically allows digits at the front of a function name. In practice this will of course not happen because Elixir itself will reject variable/function names that start with numbers (which the regexp results are filtered against on line 209 in autolink.ex).

I've done it this way because I think the regexp will stay more readable like this. The same is also true for underscores and `?` and `!` by the way, that already were this way.

:smiley:

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -202,7 +202,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   will get translated to the new href of the function.
   """
   def local_doc(bin, locals) when is_binary(bin) do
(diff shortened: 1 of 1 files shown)


josevalim:

PR #535, opened 2016-05-13
Fix `source_path` return of char list

vadim-moz:
It is normally masked by `Path.relative_to` processing, but if
`source_root` set to `nil` in the configuration, `ex_doc`
fails (`lib/ex_doc/retriever.ex:410:` `** (FunctionClauseError) no
function clause matching in Regex.replace/4`)


josevalim:

vadim-moz:
Thank you!

PR #544, opened 2016-06-02
#540 improve sidebar animation

### LINKED ISSUE #540: Improving sidebar animation.
At the moment the sidebar animation is handled via jQuery updating the width/flex-basis of sidebar.
It causes unnecessary repainting of the page while the animation is occurring. 
Moreover currently implemented animation causes content of the sidebar to stutter while it's opening / closing.

Although trying to change this on the desktop version of the page doesn't give too much performance gain(because `width` transition requires repainting), on mobile the gain is stable 60 fps even on the slower devices.

I have a working implementation of the animations and the statistics show arround 10 fps gain on desktop (from 30 to 40) and super stable 60fps on mobile (Macbook Pro 13 Early 2015).

If there will be any interest in improving the animation I would be pleased to submit a PR changes.

michalwarda:
This pull request closes #540 .

It involves changing the animations from changing width of sidebar to sliding the main onto the sidebar and on mobile sliding the sidebar onto main.

It also changes the animation of the toggle button to nicely flow over other layers.
I'm not sure how I could record the animations to show them so if anyone want's to see them before merge please just checkout to the branch and look at them yourself.

I'm also attaching the results of benchmarks with different animations. The upper graph show's the fps during animations. Red markers show the moments when fps fell below 30. (The fewer red the better).
Also lower you can see the chart of time spent on specific tasks by the browser. (More idle time is better).
**EDIT**: The benchmark was made on the generated documentation for "Ecto" project because it was one of the biggest I could find to display the difference to the fullest.

So to the profiles:
First old desktop animation profile:
old desktop animation profile

After changes:
new desktop animation profile

Old mobile animation profile:
old mobile animation profile

After changes:
new mobile animation profile

If there will be any suggestions I would be glad to change the code accordingly.


josevalim:
@michalwarda hey, how is it going? :D Thanks for the PR!

dignifiedquire:
Thanks this looks pretty sweet, what browsers have you tested this in?

michalwarda:
@dignifiedquire I just tested it on IE and found out it doesn't support transition on `flex-basis` so I switched it back to width. It doesn't change the page structure at all. Nor does the performance.

Besides that I tested it on 46+, Safari 9.1, Edge, Chrome 50+

josevalim:
@michalwarda you need to run a command so the assets files in source are updated, can you please do so?

michalwarda:
@josevalim sure! I'll do that later today. 
Unfortunately I had to change the logic of code to support IE because it didn't support animation with `calc()`.

Thankfully it didn't have an impact on the new benchmarks!

I've manage to test it on all modern browsers including ie10+ so it should be a-ok now!

michalwarda:
@josevalim did it!

josevalim:

michalwarda:
@josevalim  Unfortunately as I now tested the master build I noticed that on some screens the sidebar is not entirely covered by the content background. I prepared a fix for it on a new branch. I'm sorry for the inconvenience but I didn't notice it previously even with extensive testing for some reason :/. I'm not creating a separate issue for it because it's fully connected to this one.

michalwarda:
Also I didn't take into consideration the night 🌙 mode background color.

PR #545, opened 2016-06-04
Fix content background to cover sidebar on all device heights

michalwarda:
#540 This is a fix to not fully covering the sidebar on monitors with big heights and night mode with new sidebar implementation.


josevalim:

PR #548, opened 2016-06-05
Change night search colors.

### LINKED ISSUE #546: Search results color in night mode is hard to read.
Currently in _night mode_ the visited search result links and top right `x` are in black just like on screen:
night_mode_colors
_from https://hexdocs.pm/absinthe/Absinthe.html_

Because of that the contrast is too small to easily read the content.

I think it would be much nicer if we had different colors for the links and `x` in _night mode_. 
For example `#d2d2d2` just like other links:
night_mode_colors_suggestion

This won't show which links have been **visited** but it already doesn't on the _daily mode_.

I can make a PR implementing the new font colors later today or tomorrow if you think it's a good idea.

michalwarda:
This closes #546.

Unfortunately I had to duplicate the css because of selector importance. 

I also sneaked a tiny bit of refactor to #540 because I wanted to adjust code to existing conventions. I hope it's not a problem. If so I can create a new PR for that.


josevalim:

PR #549, opened 2016-06-06
Fix the vertical align of inline code

### LINKED ISSUE #464: Text inside code looks a few pixels off now
This was probably introduced recently with changes trying to fix the style in code.

Before
ex_doc_code_link

Now
code_off

when zoomed in it can be appreciated more clearly.

code_off_zoomed

nscyclone:
This should fix #464. 
/cc @eksperimental.
Here are 2 zoomed images: the first is similar to the one mentioned as `Before` in the appropriate issue, the second is what I'd like to suggest. I suppose that we should align the text to baseline but not its container.
Top
Baseline


dignifiedquire:
lgtm. It's not ideal because the fonts are different in size but better than the current solution.

josevalim:

eksperimental:
Thank you for the fix

PR #550, opened 2016-06-06
Add support for optional behaviour callbacks

### LINKED ISSUE #541: Optional callbacks
Optional behaviour callbacks (`@optional_callbacks`) recently landed in Elixir (https://github.com/elixir-lang/elixir/pull/4705), we should mark that a callback is optional somehow in the documentation.

whatyouhide:
This PR is supposed to add support for `@optional_callbacks`, that will be introduced in 1.3.0 (and close #541). I'm not sure what the process of rolling this out would look like: do we wait for when 1.3 will be released? I think this changes are harmless when `@optional_callbacks` is not supported.

Also, I really don't like this solution's implementation because:

1. we get to reimplement something very very similar to `Kernel.Typespec.beam_callbacks/1`: should we introduce `Kernel.Typespec.beam_optional_callbacks/1` in Elixir itself?
2. this solution displays optional callbacks in a separate section ("Optional callbacks") than regular callbacks, but ideally I think we should just mark callbacks as optional (e.g., with `(optional)` near the signature); sadly, I lack the design skills for that 😛 

You may be wondering why I opened this PR if I don't like it :), but I did just because the 1.3 release is very very close and I would love to see some kind of support (even if rudimentary) for optional callbacks in the docs for 1.3. Also, we can use this PR to get the conversation started on an implementation :)

### DIFF
--- a/lib/ex_doc/formatter/html/templates/detail_template.eex
+++ b/lib/ex_doc/formatter/html/templates/detail_template.eex
@@ -9,6 +9,9 @@
        <i class="icon-code"></i>
      </a>
     <% end %>
+    <%= if Map.get(node, :annotations) && Map.get(node.annotations, :optional?) do %>
+      <span class="note">(optional)</span>
+    <% end %>
   </div>
   <%= if specs = get_specs(node) do %>
     <div class="specs">
--- a/lib/ex_doc/formatter/html/templates/summary_item_template.eex
+++ b/lib/ex_doc/formatter/html/templates/summary_item_template.eex
@@ -1,6 +1,9 @@
 <div class="summary-row">
   <div class="summary-signature">
     <a href="#<%=enc_h link_id(node) %>"><%=h node.signature %></a>
+    <%= if Map.get(node, :annotations) && Map.get(node.annotations, :optional?) do %>
+      <span class="note">(optional)</span>
+    <% end %>
   </div>
   <%= if node.doc do %>
     <div class="summary-synopsis"><%= to_html(synopsis(node.doc)) %></div>
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -13,7 +13,8 @@ defmodule ExDoc.FunctionNode do
   """
 
   defstruct id: nil, name: nil, arity: 0, doc: [],
-    source: nil, type: nil, signature: nil, specs: []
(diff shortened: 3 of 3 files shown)


josevalim:
>  ideally I think we should just mark callbacks as optional (e.g., with (optional) near the signature)

That's the way to go definitely!

Just add something like this for now after `<span class="signature">`:

    <span class="note">optional</span>

And we will find a way to make it pretty. :)

whatyouhide:
@josevalim ok, updated to just show `(optional)` near the signature; we definitely need to drop some design skills on this, right now it looks horrible 😃

josevalim:
Implementation wise, it would probably be better to not rename those to
optional_callback but keep it as callback and keep the optional bucket
apart, no?


-- 


whatyouhide:
@josevalim yes, definitely, I just didn't know where to put this "optional bucket". I don't think adding a field to the node struct would be ok, given "optional" concerns only callbacks.

josevalim:
Maybe we can have a tags or notes field where we put extra information.
Right now, we will add only optional but it could be used for more in the
future.


-- 


whatyouhide:
What about `meta` for the field name?

whatyouhide:
Actually, `annotations` or `notes` may be better :)

josevalim:
I am fine with any.

On Monday, June 6, 2016, Andrea Leopardi <notifications@github.com> wrote:

> Actually, annotations or notes may be better :)
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/elixir-lang/ex_doc/pull/550#issuecomment-224058174>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAAlbhBj3QozKVTzW2hboFNtsNX_CsUhks5qJHE6gaJpZM4IvE6G>
> .
>


-- 


whatyouhide:
@josevalim updated and moved to have an `:annotations` key in the `FunctionNode` struct; now, the only code I find nasty is the condition in the EEx templates (we have to check for both `:annotations` as well as `:optional?` inside `:annotations`), but the rest is definitely nicer. Thoughts?

josevalim:
I thought of annotations as a list and we just add them all to the template.  I will fix it now. :)

josevalim:

whatyouhide:
Awesome 👍

PR #552, opened 2016-06-13
Remove an unused class from `functions.less`

nscyclone:
Actually, `.spec-title` should be `.specs-title` but anyway there are no rules specified for this.


josevalim:

PR #553, opened 2016-06-14
Fix a `TypeError` at the first attempt to toggle sidebar

nscyclone:
When you click at the sidebar toggle the first time, `body` has no `class` attribute so `jQuery`'s `attr` returns `undefined`. Then a `TypeError`  (`Uncaught TypeError: Cannot read property 'includes' of undefined`) is raised [here](https://github.com/elixir-lang/ex_doc/blob/master/assets/js/sidebar.js#L54).


josevalim:
Thank you @nscyclone! I have decided to fix this one at the javascript level and I have pushed a couple other fixes. Thanks for all of the improvements so far!

PR #554, opened 2016-06-15
Fix sidebar js lint errors

bjfish:
Fixes js linting errors to fix the build


josevalim:
That was my bad, thank you! :heart:

PR #559, opened 2016-06-23
Generate hover links for subtitles in moduledocs

### LINKED ISSUE #508: Generate anchors for subtitles
Currently anchors are only generated for sections and functions. It would be great if they could be also generated for subtitles. This is especially important for moduledocs, as sometimes they are quite long and it's hard to point someone to a specific point.

joshprice:
Search for secondary headers (`##`) in moduledoc section and link to section with a hover link style.

Move `HTML#header_to_id` to `Templates` module.

Another very minor refactoring.

Closes #508


assets/less/content/general.less
```diff
@@ -32,6 +32,10 @@ h2 {
   font-size: 1.625em;
   margin: 1em 0 .5em;
   font-weight: 400;
+
+  &.section-heading {
+    margin: 1.5em 0 .5em .25em;
```
joshprice: This fixes a small mis-alignment but not sure if there is a better way to do this.

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -174,6 +171,31 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
+    |> String.strip()
+    |> String.replace(~r/\W+/, "-")
+    |> String.downcase()
+    |> h()
+  end
+
+  @h2_regex ~r/##([^#].*)\n/m
+  defp link_moduledoc_headings(content) do
+    Regex.replace(@h2_regex, content, fn _, title -> """
+      <h2 class="section-heading" id="#{header_to_id(title)}">
```
joshprice: This could potentially be a small template. Not sure if it's worth it?

joshprice:
Since these headings are user-generated, it's worth thinking about how to make them unique within the document.

Some initial thoughts:
- prefix `id` with `module-` to ensure that each heading won't conflict with the same anchor outside the moduledoc section
- if there is a repeated heading inside the moduledoc (eg. `Example`) then should we make this unique by adding numbers as a suffix?

josevalim:
Thanks @joshprice! Implementation wise, I think most markdown engines support an option that converts titles to IDs, maybe we should simply rely on it instead? I would like to avoid regular expressions if possible. :)

joshprice:
I agree that regexes aren't ideal, but if you want to avoid them then there is some refactoring to do on existing code. I followed what was going on here https://github.com/elixir-lang/ex_doc/blob/b62d0dcdcd0c5649bb91f9eadbb403623833892a/lib/ex_doc/formatter/html.ex#L180-L203

Will investigate the Markdown engine ID generation as I didn't realise this was possible.

josevalim:
@joshprice yeah, i know we use them in other places but IIRC we don't have alternatives there (since we need to extract them and not only generate IDs). :)

joshprice:
@josevalim I'm not seeing where in Earmark you can generate IDs for headings (https://github.com/pragdave/earmark/blob/master/lib/earmark/html_renderer.ex#L57) unless an entirely new HTML renderer is passed in.

It's not obvious to me how best to change the behaviour of just one function of the renderer (header handling in this case). At least not without creating some tight coupling between `ex_doc` and the Markdown processors.

Also this is just Earmark, I haven't looked into any of the others as yet. Am I missing something here?

joshprice:
Looks like Pandex generates IDs automatically but doesn't seem to be a way to generate the rest of the required HTML markup https://github.com/FilterKaapi/pandex/blob/master/lib/pandex.ex#L49-L50

josevalim:
@joshprice the issue with the regex is that we will for example match code like this:

```markdown
See this example:

~~~
## Here is some code:
...
~~~
```

(replace `~~~` by triple backticks).

However we already have this bug today. So let me fix this bug and we port it back to this discussion, ok? This way we don't need to rely on the markdown engine. :)

josevalim:
@joshprice actually, it is not possible for me to fix this. I think we need to start pushing this all to the markdown parser as soon as possible.

joshprice:
Seems like a major contract change between `ex_doc` and the markdown parsers (x4) but agree that the parser is the right place for such code.

It'd be cool to have some way of customising the generated output of the markdown engine on a template by template basis. Passing in templates/functions is the only way I can think of right now to achieve this feature easily without pre- or post-processing.

Happy to help out to make this happen.

joshprice:
Just thinking about the id uniqueness constraints and the need to avoid potential id collisions, we'd also need to be able to pass in id parser/generator functions. ie. prepending `module-` to a dasherized slug id

josevalim:
@joshprice right, we would need to support those in the markdown in ExDoc and delegate to the proper key inside each markdown parser but it seems like the best way to go. :)

eksperimental:
I disagree here. ID generation is a "standard" MarkDown feature. I encountered this to be a problem while working with different engines with Jekyll,
different engines will have different rules for non-ASCII chars, with one end leaveing it as it is, the other removing it completely and anything in between, such as removing the accents and leaving the bare char.

Since links are a key feature, and they will stay forever and ever, I would like ex_doc to take care of this, so if people decide to change their engine, or if ex_doc decides to change its default MD engine, we will not be breaking all the internal links in thousand of documentations.

What we can do, to avoid what José mentions above is do a post-filtering in the generated html.
I haven't looked into the code, but it sounds like a reasonable approach,
if that doesn't work, we need to look for a better solution, but I wouldn't leave it up to the markdown engine.

my 2¢

joshprice:
Good points. I did consider the post-processing approach initially but went with pre-processing mainly because it seemed bit cleaner but it doesn't work for the reasons @josevalim points out.

Also the question arose about exactly how to post-process the HTML, (dark path of regex? http://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454, HTML parser? which one?)

As mentioned in my previous links there are discrepancies between the markdown engines for generating header links (Earmark doesn't, Pandoc does, etc) and as @Eksperimental points out there is no real standard here.

I'll take a look at a post-processing approach so we can contrast and compare.

eksperimental:
there are 2 html parsers that I know of:
- https://github.com/philss/floki
- https://github.com/zambal/eml

I'm sure it'll be better than crafting our own parser using regex

eksperimental:
This thread may also be of your interest @joshprice 
https://talk.commonmark.org/t/feature-request-automatically-generated-ids-for-headers/115
it's about header id's in commonmark.

joshprice:
I found `floki` but it was focussed on parsing only and we need transformation. `eml` on the other hand looks like it handles transformation (https://github.com/zambal/eml#transforming-eml) so will give that a try.

The commonmark article is great for 2 reasons:
1. a surprising use of the word "lacunae"
2. reasonable approach to handling unicode, etc

If we are to continue handling id generation within ex_doc then it should be robust and follow whatever standard that is already out there. Since I'm reusing the id generation we already had then that probably warrants a separate issue.

joshprice:
Reading further in the commonmark thread (https://talk.commonmark.org/t/feature-request-automatically-generated-ids-for-headers/115) it seems clear that the duplicate header problem needs more discussion as it is non-trivial in the general case.

josevalim:
I think we have four options here:

1. Preprocess markdown (done today)
2. Postprocess the HTML using regexes (a regex here will likely generate less false negatives than approach)
3. Postprocess using a HTML parser (the downside is an extra dependency)
4. Add ids dynamically using javascript (the question is if it won't be late enough for links to be broken)

I have the impression doing this with Javascript may be the easiest approach although it wouldn't fully solve the issue we have with ExDoc pages (where the titles are extracted from the document). I am also fine with 2) because it will very likely be better than whatever we have today.

joshprice:
Latest commit attempts 3) with the Eml lib but agree that the extra dep isn't worth it unless it's used elsewhere.

Happy to try 2) post process with regex if everyone is ok with that approach.

Not sure yet how I feel about the JavaScript approach, mainly because it splits the doc processing unnecessarily.

josevalim:
@joshprice cool, I think we will be fine with regexes, it is very likely all of them will generate the same h2 and there is no nesting. :)

joshprice:
This should now be ready to go. 
- Use regex for h2 post-processing
- Allow unicode letters and numbers, strip other unicode chars
- Don't link unless there is a usable ID

assets/less/content/general.less
joshprice: @dignifiedquire see above comment, is there a nicer way to fix this?

dignifiedquire: I can't think of any, looks fine to me :)

josevalim:
@joshprice thank you!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

Would you be so kind to change the "Pages" generation of titles and headers to also use a `h2` post-processing? I would love to make things consistent. :)

michalmuskala:
This is great! 🎉

joshprice:
Thanks everyone, I've opened #576 to deal with the linking of Pages. It uses the same process but not sure if it makes sense to post-process since it's dealing with the raw markdown (this was mentioned previously in this thread).

PR #564, opened 2016-06-25
Improve visibility of the menu icon

### LINKED ISSUE #562: Hamburger icon is not clearly visible when left menu is hidden
ex_doc_humburger_not_visible

and the wider your screen resolution, the more lost it gets.

I don't know if we should have a different icon with darker lines,
or if we should put it a dark purple circle.

another issue to take into consideration while fixing this is https://github.com/elixir-lang/ex_doc/issues/434

milmazz:
This commit increases the font size of the menu icon taking in consideration the current screen size using media queries. Fixes #562


eksperimental:
Before:
screenshot from 2016-06-26 08-29-35

Now:
screenshot from 2016-06-26 08-29-28


I think the icon is too big, and not enough right margin, but other than that, I love the way it looks
I have changed in my browser `font-size: 0.85pc`, and add `opacity: 0.85` and it looks good.

dignifiedquire:
I agree with @eksperimental the margin should be the same to the top and the right and it's too big now in the screenshot.

milmazz:
@eksperimental @dignifiedquire @josevalim You already notice that my design skills are really bad, but I'll keep trying 😛 

wdyt?

screen shot 2016-06-29 at 8 51 25 pm
screen shot 2016-06-29 at 8 50 58 pm
screen shot 2016-06-29 at 8 49 45 pm

eksperimental:
for me it looks great, but i don't know if it's following the rule of the margin to the top as to the right in all the cases. or if it's even possible when you are using a proportional distance.

josevalim:
To me it looks good. Let's wait for @dignifiedquire's blessing both code and style wise. :)

dignifiedquire:
LGTM thanks for this @milmazz :octocat:

josevalim:

PR #565, opened 2016-06-26
Issue 561

RobertDober:
This is kind of a scetch of a fix. The obviously missing things are:
* Multiple warnings as some docs are converted several times to html.
* Could not come up with sensible tests in the time I had available.

Less obvious things that concern me:
* Not sure I covered all bases, the need to leave `def to_html(text) when is binary(text) do` in line 43 of `markdown.ex` is somehow odd.
* And on the same account, I just do not understand the code base well enough (yet ;) ) to be sure my approach was a sensible one.

Nevertheless, in the hope that this is useful, and maybe I can improve on it if aided.


josevalim:
Thank you @RobertDober.  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

Unfortunately I could not merge this as it was outdated (our bad) but I have pushed code that brings functionality similar to this. I have also used the opportunity to unify a couple things related to our docs generation.

RobertDober:
@josevalim good to know

PR #568, opened 2016-06-28
Fix flash of unstyled content

aphillipo:
When loading the page in night mode there is a small delay that causes the page to flash white while the javascript loads.  This is obviously more problematic the longer page is and as such affects documentation more than most.

### DIFF
--- a/lib/ex_doc/formatter/html/templates/head_template.eex
+++ b/lib/ex_doc/formatter/html/templates/head_template.eex
@@ -13,4 +13,4 @@
     <script src="dist/sidebar_items.js"></script>
   </head>
   <body data-type="<%= sidebar_type(page.type) %>">
-    <script>if(localStorage.getItem('night-mode')) document.body.className += 'night-mode';</script>
+    <script>if(localStorage.getItem('night-mode')) document.body.className += ' night-mode';</script>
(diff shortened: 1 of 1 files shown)


josevalim:
This looks good to me. What about you @dignifiedquire ?

@aphillipo I am assuming that with this change we don't end-up with night-mode added as a class twice thanks to jquery?

aphillipo:
@josevalim I'll test that now and get back to you.

aphillipo:
No duplication that I can see. For me this all works perfectly; only testing in Webkit+Firefox latest though.  Do I need to test it in IE - I've done this before and am 99% certain it works.

I really wonder if there is a need for jquery on a project like this when you could have a couple of npm dependencies for making element lookup easier and the classlist polyfill.

joshprice:
I actually noticed this last week and tried the same thing and noticed that although this massively reduced the length of the flash, it was still there. What is everyone else seeing?

Not sure if there is anyway to make this completely flash free, any thoughts?

josevalim:
> I actually noticed this last week and tried the same thing and noticed that although this massively reduced the length of the flash, it was still there. What is everyone else seeing?

I tried to do the same last week too. :P

Merging this... if anyone has better tips, please let us know!

josevalim:

PR #574, opened 2016-07-06
Update version of earmark to 1.0

pragdave:


josevalim:
Hi @pragdave! It seems earmark requires v1.2 so we will need to plan accordingly before merging this, thank you!

PR #576, opened 2016-07-06
Hover links for secondary headings in Pages

joshprice:
As requested by @josevalim in #559 this refactors the extra page generation to use the same template as the h2's in moduledocs.


lib/ex_doc/formatter/html/templates.ex
```diff
@@ -184,23 +184,30 @@ defmodule ExDoc.Formatter.HTML.Templates do
     |> h()
   end
 
+  @doc """
+  Link secondary headings found with `regex` with in the given `content`.
```
nscyclone: Probably, `within` was meant to be used instead of `with in`.

josevalim:

PR #577, opened 2016-07-07
Fix page shaking when switching modes

### LINKED ISSUE #572: Switching between night mode and day mode changes position in page
You can clearly see this if you go to the bottom of a really long page [Enum](http://elixir-lang.org/docs/stable/elixir/Enum.html) module page. Scroll all the way to the bottom, and click SWITCH TO NIGHT MODE. You will see that the page moved.
You need to scroll again to be able to see the SWITCH TO DAY MODE link.

This affects the usability of the page.
You can barely see if in small pages, but as you scroll down you can see it's a pixel or two off, and at the end of a long page it has accumulated to a considerable amount of pixels.

I'm not sure if this could be the highlightjs theme for night mode used, because from the CSS I haven't seen code that changes the sizes deliberately.. but i haven't had the time to look thoroughly though. 
but we need to guarantee that everything shares the same size in night mode as in day mode. 

/cc @pdilyard

nscyclone:
This should close #572.


josevalim:

PR #583, opened 2016-07-15
Add (callback) to search results

### LINKED ISSUE #560: Search results should differentiate functions from callbacks and types
I 'm not sure if we should do this all the time, but when more than one result with the same are displayed, we are not telling the user which one is which.

search for "handle_body" in the Elixir docs.

ex_doc_funs_callbacks

antipax:
Closes #560.

Types don't appear to show up in search results so only implemented "(callback)" for callbacks.
screen shot 2016-07-15 at 2 50 06 pm


antipax:
Do we want to do this for macros too?

josevalim:
Now, I think callbacks only are enough.

josevalim:

PR #584, opened 2016-07-15
Fix the opened `.sidebar-toggle` position on small screens

nscyclone:
I suppose that the vertical position of the toggle icon should remain the same regardless of the sidebar state: on small screens it was `top: 5px;` when the sidebar is closed and `top: 8px;` otherwise.


josevalim:

PR #585, opened 2016-07-15
Remove extra CSS and rebuild

antipax:
I don't know how it happened, but an extra CSS file got in to the dist directory and it's breaking the build. This should fix the build, but make sure it's passing before this is merged.


antipax:
Looks like it got added without removing the old one in #584. Is there any documentation around on the asset build process/commit hook? I might take a shot at enhancing the README with some info about that.

nscyclone:
Yes, @antipax, it seems like #584 was based on pre-#583's master which was merged just a few moments before #584. I could actually rebase the branch but it got merged too fast so I didn't even notice that. My bad.

antipax:
@nscyclone no prob :)

antipax:
Ah, docs for the asset build are all here https://github.com/elixir-lang/ex_doc/tree/master/assets so no need to write them. :)

josevalim:

PR #586, opened 2016-07-19
Remove unused fonts from font list

joshprice:
- Switch base h2 to use font-weight 700
- Use font-weight 700 instead of 600 and 800
- Remove duplicate font-weight specification

Addresses #579


antipax:
lgtm 👍

josevalim:
Great job @joshprice! Let's wait for @dignifiedquire to confirm and then we can ship it. During the night we will also build a new version from the docs and we will be able to see if it implies any difference in practice.

eksperimental:
nice!

dignifiedquire:
Thanks LGTM

josevalim:

joshprice:
Thanks @antipax @josevalim @eksperimental @dignifiedquire. I've only done limited testing on my own machine across a couple of Elixir core modules and everything seemed to look OK to me, but if everyone could keep their eyes out for any potential issues around those fonts once the docs are rebuilt that'd be awesome.

PR #590, opened 2016-07-24
Autolinking to types

alfert:
Evaluating the `t:` prefix to a type with function syntax. The implementation took a little detour, because it took a while to realize that links to types without arity shall not work. Essentially, this make the code easier. 

I added also a better matching Module regex, since numbers and underscores are valid characters in module names.

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -302,10 +302,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
 
 
   defp split_function(bin) do
-    [modules, arity] = case String.split(bin, "/") do
-      [m, a] -> [m, a]
-      [m]    -> [m, ""]
-    end
+    [modules, arity] = String.split(bin, "/")
     {mod, name} =
       modules
       |> String.replace(~r{([^\.])\.}, "\\1 ") # this handles the case of the ".." function
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -203,22 +203,17 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   will get translated to the new href of the function.
   """
   def local_doc(bin, locals) when is_binary(bin) do
-    fun_re = ~r{((c:)?([a-z\d_!\\?>\\|=&<!~+\\.\\+*^@-]+)/\d+)} |> Regex.source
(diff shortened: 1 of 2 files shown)


lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -202,14 +203,15 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
+    fun_re = ~r{(([ct]:)?([a-z\d_!\\?>\\|=&<!~+\\.\\+*^@-]+)/\d+)} |> Regex.source
+    ~r{(?<!\[)`\s*(#{fun_re})\s*`(?!\])}
     |> Regex.scan(bin)
     |> Enum.uniq()
     |> List.flatten()
     |> Enum.filter(&(&1 in locals))
     |> Enum.reduce(bin, fn (x, acc) ->
          {prefix, _, function_name, arity} = split_function(x)
-         escaped = Regex.escape(x)
+           escaped = Regex.escape(x)
```
josevalim: Wrong indentation? :)

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -220,6 +222,9 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   """
   def project_doc(bin, modules, module_id \\ nil) when is_binary(bin) do
     project_funs = for m <- modules, d <- m.docs, do: doc_prefix(d) <> m.id <> "." <> d.id
+    project_types = for m <- modules, d <- m.typespecs, do: "t:" <> m.id <> "." <> d.id
```
josevalim: Does `typespecs` include only types? Or does it include callbacks and specs too?

josevalim:
Beautiful! I have included one tiny comment and a question. Other than that it looks good to go! :D

lib/ex_doc/formatter/html/autolink.ex
alfert: Oh, yes, will be fixed.

lib/ex_doc/formatter/html/autolink.ex
alfert: Callbacks go into `docs`, otherwise linking to callbacks wouldn't work (as they did earlier). Specs are part of a function node, you can retrieve it via `d.specs` in the above code snippet. So, `typespecs` only contain the types defined in the module.

josevalim:

PR #591, opened 2016-07-28
Stop escaping linked h2s

### LINKED ISSUE #587: HTML in headers when the header contains an atom
Looks like atoms inside of headers may cause some issues.

     Git options (<code class="inline">:git</code>)

http://elixir-lang.org/docs/stable/mix/Mix.Tasks.Deps.html#module-git-options-code-class-inline-git-code

joshprice:
Ensure that moduledocs and extra pages use the markdown HTML entity escaping.

Remove HTML entities from header IDs.

Fixes #587

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -147,7 +147,7 @@ defmodule ExDoc.Formatter.HTML do
       output_file_name = "#{options.output_file_name}.html"
       config = set_canonical_url(config, output_file_name)
       html = Templates.extra_template(config, title, modules,
-                                      exceptions, protocols, link_headers(content))
+                                      exceptions, protocols, content)
 
       output = "#{options.output}/#{output_file_name}"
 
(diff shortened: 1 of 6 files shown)


josevalim:
@joshprice regarding the failing test and escaping characters, if we are processing the HTML, should the markdown engine escape the `>`? Or does it escape only `<`?

josevalim:
If the latter, maybe we should rewrite the test to use `<` instead?

joshprice:
You're right the Markdown engine should escape this:

```elixir
iex(4)> ExDoc.Markdown.to_html("## <>")        
"<h2>&lt;&gt;</h2>\n"
```

I'll fix the test.

joshprice:
After a little investigation, the tests are using Hoedown (which also escapes the headers like Earmark example above), but the entity escaping doesn't seem to be happening in this test and I can't yet see why. Will keep investigating...

milmazz:
@joshprice Hi, are you still working on this? Thanks in advance.

joshprice:
My initial investigation didn't reveal anything obvious, apologies for not posting an update. I'll look into it again and report back my findings. Suspect I might need a hand figuring out what is going on though...

josevalim:

PR #596, opened 2016-09-04
Link h2s in detail templates (functions, macros)

joshprice:
Headings are scoped (prefixed) by the macro or function they are in, avoiding potential ID duplication.

The motivation for this was noticing the extensive use of h2 headers in some Macro and Function sections. ie http://elixir-lang.org/docs/stable/elixir/Kernel.SpecialForms.html#&/1

This change has the (possibly undesirable) side-effect of linking h2s for functions such as `Examples`. ie http://elixir-lang.org/docs/stable/elixir/List.html#delete/2


josevalim:

PR #598, opened 2016-09-09
WIP: Keep track of generated files

### LINKED ISSUE #530: Generator blindly wipes directory
On https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc/formatter/html.ex#L18, the `File.rm_rf! output` call wipes the `doc` directory regardless of contents. This is an annoyance when the directory contains a git repository for publishing docs to Github Pages, because the `.git` directory is removed as well.

It would be nice if the generator only deleted files and directories it might have created.

milmazz:
Fixes #530

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -15,10 +15,11 @@ defmodule ExDoc.Formatter.HTML do
   def run(module_nodes, config) when is_map(config) do
     config = normalize_config(config)
     output = Path.expand(config.output)
-    File.rm_rf! output
-    :ok = File.mkdir_p output
+    build = Path.join(output, ".build")
 
-    assets() |> assets_path() |> generate_assets(output)
+    output_setup(build, output)
+
+    assets = assets() |> assets_path() |> generate_assets(output)
 
     all = Autolink.all(module_nodes)
     modules    = filter_list(:modules, all)
@@ -35,13 +36,15 @@ defmodule ExDoc.Formatter.HTML do
     generate_api_reference(modules, exceptions, protocols, output, config)
     extras = generate_extras(output, module_nodes, modules, exceptions, protocols, config)
 
-    generate_index(output, config)
-    generate_not_found(modules, exceptions, protocols, output, config)
-    generate_sidebar_items(modules, exceptions, protocols, extras, output)
+    generated_files =
+      generate_index(output, config) ++
+      generate_not_found(modules, exceptions, protocols, output, config) ++
+      generate_sidebar_items(modules, exceptions, protocols, extras, output) ++
+      generate_list(modules, modules, exceptions, protocols, output, config) ++
+      generate_list(exceptions, modules, exceptions, protocols, output, config) ++
+      generate_list(protocols, modules, exceptions, protocols, output, config)
 
-    generate_list(modules, modules, exceptions, protocols, output, config)
-    generate_list(exceptions, modules, exceptions, protocols, output, config)
-    generate_list(protocols, modules, exceptions, protocols, output, config)
+    generate_build(extras, generated_files ++ assets, build)
 
     Path.join(config.output, "index.html")
   end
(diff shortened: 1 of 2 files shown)


lib/ex_doc/formatter/html.ex
```diff
@@ -88,13 +118,14 @@ defmodule ExDoc.Formatter.HTML do
   """
   @spec generate_assets(list, String.t) :: :ok
   def generate_assets(source, output) do
-    Enum.each source, fn {pattern, dir} ->
+    Enum.reduce source, [], fn {pattern, dir}, acc ->
```
josevalim: `Enum.flat_map` would be ideal here. :)

lib/ex_doc/formatter/html.ex
```diff
@@ -15,10 +15,21 @@ defmodule ExDoc.Formatter.HTML do
...
     output = Path.expand(config.output)
-    File.rm_rf! output
-    :ok = File.mkdir_p output
+    build = Path.join(output, ".build")
+
+    if File.exists? build do
+      File.stream!(build)
+      |> Enum.map(fn(line) ->
+        line = String.trim(line)
+        File.rm!(Path.join(output, line))
```
josevalim: I think `File.rm` here and on the lines below would be enough. If someone removed the file, we don't want the build process to fail forever until the user deletes the whole directory.

lib/ex_doc/formatter/html.ex
```diff
@@ -15,10 +15,21 @@ defmodule ExDoc.Formatter.HTML do
   def run(module_nodes, config) when is_map(config) do
     config = normalize_config(config)
     output = Path.expand(config.output)
-    File.rm_rf! output
-    :ok = File.mkdir_p output
+    build = Path.join(output, ".build")
+
+    if File.exists? build do
+      File.stream!(build)
+      |> Enum.map(fn(line) ->
```
josevalim: My recommendation is to avoid using pipe if you are doing it only once, so I would rather write this one as `Enum.each(File.stream!(build), fn...)`. I would also suggest moving the whole if check to its own function. :)

lib/ex_doc/formatter/html.ex
```diff
@@ -53,29 +66,46 @@ defmodule ExDoc.Formatter.HTML do
     %{config | main: main || @main}
   end
 
+  defp generate_build(extras, generated_files, output) do
+    extras = for {file, _, _} <- extras, do: "#{file}.html"
+
+    build_files =
+      MapSet.new(generated_files ++ extras)
```
josevalim: Theoretically there is no need for MapSet. I would also call `Enum.flat_map(&[&1, "\n"])` to ensure all entries have a new line (including the last one).

josevalim:
@milmazz beautiful. I have added some comments! 

lib/ex_doc/formatter/html.ex
milmazz: I agree.

lib/ex_doc/formatter/html.ex
milmazz: Yeah, I was thinking about moving this section into a new function. Thanks for reminding me of this :)

lib/ex_doc/formatter/html.ex
milmazz: > Theoretically there is no need for MapSet.

I used `MapSet` to avoid duplicate entries coming from the `generate_api_reference/5` and `generate_extras/6` (example: `api-reference.html` and `#{config.main.html}), those duplicate entries make the build process fail at the beginning when we apply `File.rm!`, but, since we're changing `File.rm!` with `File.rm` is not necessary anymore, also, I'll restore `generate_api_reference/5` to its original position. That way we avoid to produce duplicate entries.

>  I would also call Enum.map(generated_files ++ extras, &[&1, "\n"]) to ensure all entries have a new line (including the last one). Keep in mind you don't need to convert to a string. File.write! accepts a list of binaries too.

@josevalim Awesome! with every review, I learn something new :)

josevalim: So instead of a MapSet, why not Enum.uniq?

Glad to help!

milmazz:
@josevalim I added some tests and applied your suggestions, please let me know what do you think.

josevalim:
@milmazz :shipit: :D

seancribbs:
Thank you so much @milmazz 😺

PR #600, opened 2016-09-18
add support for handling settext headers and stripping links from the primary header for link generation

JakeStaTeresa:


josevalim:
Thank you @JakeStaTeresa. I believe we have decided to not parse the HTML because we can run into further complexities, such as having attributes in the headers, other html attributes and so forth. If you want ExDoc to properly generate the sections, you should stick with the `#` format for now. Maybe we should document it as such as well.

JakeStaTeresa:
@josevalim, thanks for the feedback. I agree with the goal of simplifying the implementation. My PR is trying to solve the following issues with the # style headers. I have a couple of questions as to what is the preferred approach in the following cases. May I solicit your feedback on each of them?

- Settext style headers. If ex_doc only supports # style headers, should it throw an error when a settext style header is encountered to document the intent?

- Its not uncommon nowadays to have badges on the header. It contains build and coverage status and maybe even more. Below is an example. My opinion is ex_doc should strip this links during sidebar generation. If not, maybe it should also throw an exception?

``` Markdown
# Currencies [![Build Status](https://travis-ci.org/JakeStaTeresa/Currencies.svg?branch=master)](https://travis-ci.org/JakeStaTeresa/Currencies) [![Coverage Status](https://coveralls.io/repos/github/JakeStaTeresa/Currencies/badge.svg?branch=master&cache=1)](https://coveralls.io/github/JakeStaTeresa/Currencies) [![hex.pm version](https://img.shields.io/hexpm/v/currencies.svg)](https://hex.pm/packages/currencies) [![hex.pm downloads](https://img.shields.io/hexpm/dt/currencies.svg)](https://hex.pm/packages/currencies)
```

- A readme file containing only level 2 headers and some example elixir code will hit the h1 regex and capture the elixir example as a header. Below is an example. What should ex_doc do in this cases?

## Introduction

more text. even more text

## Examples

some descriptions

```Elixir
# Some comment which will be captures as header.

Currencies.filter(&(String.contains?(&1.name, "Peso"))) |>
  Enum.map(&(&1.name))
# ["Argentina Peso", "Chile Peso", "Colombia Peso", "Cuba Convertible Peso", "Cuba Peso",
# "Dominican Republic Peso", "Mexico Peso", "Philippines Peso", "Uruguay Peso"]

Currencies.filter(&(String.contains?(&1.name, "Peso"))) |>
  Enum.map(&(&1.name)) |>
  Enum.count
# 9

josevalim:
Hrm, maybe moving to `h1` and `h2` then is the best way to go although we would need to make it more resilient to attributes. Thanks for extra points @JakeStaTeresa.

@joshprice, do you remember why we picked markdown over the html one?

JakeStaTeresa:
@josevalim, I am happy to make changes to make it more resilient. I just don't fully understand what you meant when you say "we would need to make it more resilient to attributes". Do you mean we need to be resilient to the presence of html attributes that could be present on the h1/h2 tags? If yes, then I dont think this is an issue. Because the MD specs dont support adding attributes on the header tags. If your concern is on html attributes for tags nested inside the header tag, then this is already addressed by the ```strip_html/1``` function.

josevalim:
@JakeStaTeresa IIRC some of the markdown engines could include attributes in the h1/h2 tags. For example, IDs automatically extracted from the source or some sort of annotation.

JakeStaTeresa:
@josevalim, thanks for the quick response. I now updated the regex to ignore attributes on h1 and h2 headers. I haven't added tests yet that verify the behaviour. I will have a look at the different markdown engines supported by ```ex_doc``` and will verify which one of them adds id or class attribute. Then I will create a test that verifies the behaviour.

JakeStaTeresa:
hi @josevalim, i just added a test to ensure sidebar generation using pandoc doesnt capture the attributes on header tags.

josevalim:
Beautiful! Let's ping @joshprice once again and see if he has anything to add as he was working on this part of the codebase.

joshprice:
Sorry for the delay. I've had a quick look, but wouldn't mind a bit more
time to fully digest what is happening here to ensure we're dealing with
headers/markdown consistently across the codebase.

Should be able to comment further later today once I've had a chance to
investigate.

joshprice:
My understanding is that header attributes can occur when:

1. linking headers (providing `id`s and also links to those anchors)
2. adding custom attributes as described in https://github.com/pragdave/earmark#adding-html-attributes

Not sure if I've missed other ways attributes can occur?

Is the intention that we support custom html attributes for headers/blocks (as per https://github.com/pragdave/earmark#adding-html-attributes)?

If so then there are limitations adding an `id` or `class`es might get overridden by doc processing, depending on which element.

Couple of other points:

- I'm getting some errors running the tests, I suspect because of a pandoc dependency (I don't believe I have it installed). So may need to add `pandoc: true` tags to those tests which need it.

- Having regexes for parsing html is generally a terrible idea, but I went with it on the basis that the output from the markdown engines is pretty predictable so think it's probably ok in this case. My concern is that the regexes https://github.com/elixir-lang/ex_doc/pull/600/files#diff-fc9775865b82e656523d7cbbe16526d2R216 and https://github.com/elixir-lang/ex_doc/pull/600/files#diff-fc9775865b82e656523d7cbbe16526d2R226 need to match this https://github.com/elixir-lang/ex_doc/blob/da63fc03c3daff978fb9b3859a9c1a2c6bed1d4b/lib/ex_doc/formatter/html/templates.ex#L201

josevalim:
> Is the intention that we support custom html attributes for headers/blocks (as per https://github.com/pragdave/earmark#adding-html-attributes)? If so I suspect this PR will strip them out.

Great catch. I am fine with this decision given we are adding the link-headings ourselves and in a way that there are no conflicts.

> I'm getting some errors running the tests, I suspect because of a pandoc dependency (I don't believe I have it installed). So may need to add pandoc: true tags to those tests which need it.

@JakeStaTeresa could you please make sure you have pandoc installed so it runs the pandoc based tests?

> My concern is that the regexes ... and ... need to match this ...

@JakeStaTeresa could you please look into unifying the regexes?

Thank you @joshprice. This was a fantastic feedback! :heart:

JakeStaTeresa:
hi @josevalim, @joshprice thanks for the great feedback. i just pushed the changes as per code review comments. cheers!

josevalim:

PR #603, opened 2016-09-23
Define new types

milmazz:
The idea behind this PR is to use `ExDoc.Config.t` instead of `%ExDoc.Config{}` when we're referring to that struct. The same behavior for `ExDoc.{Module,Function,Type}Node.t`. wdyt?


antipax:
I'm 👍  on this, by removing the explicit reference to the struct it should also break the compile-time dependencies those references impose and speed up partial compiles.

josevalim:
:shipit:

PR #613, opened 2016-10-10
Add support for page groups

josevalim:
The idea behind this feature is to support grouping pages for longer guides. It is [a feature request from the Phoenix team](https://github.com/phoenixframework/phoenix/pull/1960#issuecomment-252551608). 

Support is added by adding a group key to each page entry:

```elixir
"guides/a.md": [group: "Testing", title: "Intro", path: "testing"],
"guides/b.md": [group: "Testing", title: "Testing Models", path: "testing-models"],
"guides/c.md": [group: "Testing", title: "Testing Controllers", path: "testing-controllers"],
"guides/d.md": [group: "Testing", title: "Testing Views", path: "testing-views"],
"guides/e.md": [group: "Testing", title: "Testing Channels", path: "testing-channels"],
```

Support for the `:group` was added [in a previous commit](https://github.com/elixir-lang/ex_doc/commit/bd4aa03b5fabe11008fd605b195d78fcd1d86bbf). This commit only adds support to the html formatter. Here is a screenshot:

screen shot 2016-10-10 at 11 02 06

Feedback:

  * @dignifiedquire - I need your feedback on the styling and JS code
  * @milmazz - Can you please add this feature to the epub formatter (if relevant)?
  * @wsmoak, @Gazler - Please let me know what you think
  * @mobileoverlord, @fhunleth - This may be useful for Nerves too


josevalim:
Here is a screenshot for using the same color as the other links for the group header instead of white:

screen shot 2016-10-10 at 11 15 00

And here is using a shade of grey (#DDD):

screen shot 2016-10-10 at 11 16 23

I have 49 others to try if you want me to.

Gazler:
I like it. Here's a build from ex_doc master on Phoenix (including the guides) https://s3.amazonaws.com/gazler.phoenix.guides/a_overview.html

josevalim:
Btw, I prefer the last screenshot. It is as intrusive as the white one and not too similar as the second one.

milmazz:
@josevalim Sure, I can work on this feature for the EPUB format, I'll create another PR for this.

BTW, I think that the subsection header list looks better with `ul` instead of  `ol`, wdyt? /cc @dignifiedquire 

 `ol`    |      `ul` |
----- | ----- |
capture | capture1

wsmoak:
I definitely prefer it without the numbers.

Working with this branch I noticed that clicking on the title in the menu does _not_ load the page.  It takes two clicks -- on the title and then on one of the sub-items.

Did the invented 'Top' anchor get added in this PR or was that there before?  Is it necessary?

I'd prefer to see 'Top' removed, and clicking the page title work to load the page / jump back to the top.

(Or maybe float a 'back to top' link somewhere in the content, there's a ton of whitespace over there.)

dignifiedquire:
Pretty much agree with was said by @josevalim and @milmazz. `ul` + `#ddd` looks really nice.

josevalim:
@wsmoak The top was added around 0.14 to be in sync with the modules page. The idea is that you can navigate the guides without moving away from the current page you are on.

@milmazz another PR is perfect!

josevalim:
Regarding the Top. I would like to understand if the issue is the alignment or the numbers. Here are a couple other suggestions.

  1. All left aligned:

    screen shot 2016-10-10 at 20 38 34

  2. Aligned with a base arrow:

    screen shot 2016-10-10 at 20 44 33

  3. Aligned with a simple arrow:

    screen shot 2016-10-10 at 20 45 32

josevalim:
I am merging this but let's continue the discussion regarding the top here. We can also create a new issue for that if you prefer.

wsmoak:
I like the arrow, that sets it apart from the lines that came from the document. 

You stated the reason for the 'Top' and the title not opening the document (so you can navigate w/o leaving the page.)  I hadn't updated ExDoc in a while, and it was a surprise.  So, okay. :)

On the numbers, I just prefer it without (ul instead of ol).  What if they already have letters in the text of h2's, or they want Roman numerals instead?  (Hmm... maybe that sort of thing is already configurable by adjusting the CSS? Is that easy to do?)

Having the numbers over there on the menu but not _in_ the document also makes things visually not match up.

Only since you asked!

josevalim:
Alright! Numbers removed on master!

eksperimental:
it would make "top" look different than the rest of the items.
maybe with a border, looking like a key in the keyboard, or something like that

michalmuskala:
I really don't like that I have to click twice to change the page. This completely breaks the expectation I have when using a website, where a link transitions me to another page.

Could we have a separate thing to "unroll" the list and the main link be a regular link?

josevalim:
@michalmuskala we can come up with another mechanism but I think it is important to align your expectations to see if it is whether "I expected it work like this because that's how it worked before" or "I expected to work like this because those entries are visually telling me they are links". The second means we need to improve the styling so they don't look as links (albeit I personally don't think they do).

michalmuskala:
It's mainly super annoying I have to click twice to go to a page. Usually in tree-view style things it's a separate button to go to an element and to expand it.

nscyclone:
I fully agree with @michalmuskala. Usually, tree nodes with children behave as a regular node but they also have a special button to expand their contents.

josevalim:
As I said, we can come up with a new one or improve the current mechanism. However, someone needs to do so, otherwise the problem won't be solved. :) We won't revert to the previous behaviour because it drastically hurt navigating and exploring the documentation.

josevalim:
Btw, I will be glad to implement the feature if someone comes up with the design. My original implementation also added a + sign or a ▶ before each module name, but that became too distracting. See screenshot:

screen shot 2016-10-11 at 14 36 30

If @dignifiedquire approves the screenshot above (we can use ▶, +, or a mixture) or someone comes up with a design he does, I can do the changes.

eksperimental:
@josevalim regard your last screenshot, we can use something like this
a hollow arrow that gets's filled up when submenu is open.

jump-menu-vista-style

josevalim:
If it helps, my ultimate source of arrows is here: http://xahlee.info/comp/unicode_arrows.html :P

josevalim:
@eksperimental interestingly enough, the first level of icons in your screenshot does not contain the arrow/triangle, similar to Elixir docs. 😄

eksperimental:
@josevalim yeah.. it's interesting. we could try: no icon in the first level, and arrows as in the example in the second level. i kind of like that tilted triangle thing.

to be honest i don't like the bare +/- signs. i think they should be an icon. 

sources for icons could be:
http://fontawesome.io/icons/
https://icomoon.io/#preview-free

search for keywords: arrow, plus

eksperimental:
on the other hand, adding an icon on the first level will free us from having to click twice to get to module documentation. We can "train" readers to link on the icons if they want to navigate the whole menu.
and also get rid of the "Top" menu item.

PR #625, opened 2016-10-15
Generate top-level docs for an umbrella project

### LINKED ISSUE #567: Add ability to generate top-level docs in umbrella project
Currently in an umbrella project, docs will be recursively generated for sub directories. However, sometimes it would be useful to generate top level documentation for the whole umbrella project.

wojtekmach:
Fixes #567

Looking forward to hearing feedback about implementation, I'll add tests afterwards.

Demo using this branch: http://wojtekmach.pl/acme_bank_docs/main.html

### DIFF
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -84,9 +84,9 @@ defmodule ExDoc.Retriever do
   alias Kernel.Typespec
 
   @doc """
-  Extract documentation from all modules in the specified directory or directories
+  Extract documentation from all modules in the specified directory or directories.
   """
-  @spec docs_from_dir(Path.t | list(Path.t), ExDoc.Config.t) :: [ExDoc.ModuleNode.t]
+  @spec docs_from_dir(Path.t | [Path.t], ExDoc.Config.t) :: [ExDoc.ModuleNode.t]
   def docs_from_dir(dir, config) when is_binary(dir) do
     files = Path.wildcard Path.expand("*.beam", dir)
     docs_from_files(files, config)
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -84,12 +84,15 @@ defmodule ExDoc.Retriever do
   alias Kernel.Typespec
 
   @doc """
-  Extract documentation from all modules in the specified directory
+  Extract documentation from all modules in the specified directory or directories
   """
-  @spec docs_from_dir(Path.t, ExDoc.Config.t) :: [ExDoc.ModuleNode.t]
+  @spec docs_from_dir(Path.t | list(Path.t), ExDoc.Config.t) :: [ExDoc.ModuleNode.t]
   def docs_from_dir(dir, config) when is_binary(dir) do
     files = Path.wildcard Path.expand("*.beam", dir)
     docs_from_files(files, config)
+  end
+  def docs_from_dir(dirs, config) when is_list(dirs) do
+    Enum.flat_map(dirs, &docs_from_dir(&1, config))
   end
 
   @doc """
--- a/lib/mix/tasks/docs.ex
+++ b/lib/mix/tasks/docs.ex
@@ -89,6 +89,16 @@ defmodule Mix.Tasks.Docs do
 
     * `:canonical` - String that defines the preferred URL with the rel="canonical"
       element; defaults to no canonical path.
+
+  ## Umbrella project
+
+  ExDoc can be used in an umbrella project and generates a single documentation for all child apps.
+
+  Generating documentation per each child app can be achieved by running:
+
+      mix cmd mix docs
+
+  See `mix help cmd` for more information.
   """
 
   @doc false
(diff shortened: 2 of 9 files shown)


milmazz:
@wojtekmach Great job!

I think we should provide a switch for this feature, I mean, something like: `mix docs --compact` in an umbrella project, and leave the current behavior as is now. wdty?

What happens when you have *extras* in the sub-apps?, I think this are ignored for now.

lib/ex_doc/retriever.ex
```diff
@@ -84,13 +84,16 @@ defmodule ExDoc.Retriever do
   alias Kernel.Typespec
 
   @doc """
-  Extract documentation from all modules in the specified directory
+  Extract documentation from all modules in the specified directory or directories
   """
-  @spec docs_from_dir(Path.t, ExDoc.Config.t) :: [ExDoc.ModuleNode.t]
+  @spec docs_from_dir(Path.t | list(Path.t), ExDoc.Config.t) :: [ExDoc.ModuleNode.t]
```
josevalim: `[Path.t]` is more common. :)

lib/ex_doc/retriever.ex
```diff
@@ -84,13 +84,16 @@ defmodule ExDoc.Retriever do
   alias Kernel.Typespec
 
   @doc """
-  Extract documentation from all modules in the specified directory
+  Extract documentation from all modules in the specified directory or directories
```
josevalim: Let's add a `.` to the end of the sentence.

lib/mix/tasks/docs.ex
```diff
@@ -141,8 +140,31 @@ defmodule Mix.Tasks.Docs do
...
+      Mix.Project.in_project(dep.app, dest, fn module ->
+        module.project
+        |> Keyword.put_new(:build_per_environment, true)
+        |> Mix.Project.compile_path()
+      end)
+    end)
+  end
+
+  defp umbrella_apps do
+    Mix.Dep.loaded([])
```
josevalim: `Mix.Dep` is not public API so I would prefer to use `Mix.Dep.Umbrella.unloaded` which also not public API but would at least give correct results. Once you do that, you can look at `Path.join dep.opts[:build], "ebin"` for the compile path.

wojtekmach:
Hey, thanks for feedback everyone! I pushed some changes and I also added a test showcasing the current behavior which probably needs adjusting.

> I think we should provide a switch for this feature, I mean, something like: mix docs --compact in an umbrella project, and leave the current behavior as is now. wdty?

Personally, I think I'd want to keep the docs centralized. With the old behavior, the user would have to manually publish docs from each app and probably create a separate index page (similar to http://elixir-lang.org/docs.html) at which point I think he/she might as well just run `mix docs` manually for each app. But yeah, it's perhaps a weak argument, I think more flexibility is good too.

I'd need help from implementation side to make it happen, is it possible to make a task recursive/non-recursive based on options?

> What happens when you have extras in the sub-apps?, I think the option extras (in the sub-apps) is ignored for now.

Yeah, `extras` option in sub-app is ignored for now, however e.g. this would work in the umbrella: `extras: ["apps/foo/foo.md"]`. I guess it goes back to whether or not we want to support "compact" and "old" ways of generating docs - if only "compact", I think it would reasonable to expect users setting all `docs: []` option in just the umbrella root.

One area where "old" way of generating docs feels better is working with multiple READMEs - right now, if I do in the root: `docs: [extras: ["apps/foo/README.md", "apps/bar/README.md"]]`, only one `readme.html` would be created. I think for now we should at least warn/raise error, in the future perhaps: `extras: ["README.md", {"apps/foo/README.md", "foo-readme"}]` would make sense.

sorry for long post :)

josevalim:
>  I think he/she might as well just run mix docs manually for each app.

This is easy to done recursive with: `mix cmd mix docs`. So :+1: for leaving it as manual. We should probably improve the docs for the `mix docs` task though to mention how it works under umbrellas and the `mix cmd mix docs` tip.

lib/mix/tasks/docs.ex
```diff
@@ -141,8 +140,26 @@ defmodule Mix.Tasks.Docs do
...
+      if Mix.Project.umbrella?(config) do
+        umbrella_compile_paths()
+      else
+        Mix.Project.compile_path
+      end
+
+    Keyword.put_new(options, :source_beam, compile_path)
+  end
+
+  defp umbrella_compile_paths do
```
josevalim: I think this could should be like this:

```
build = Mix.Project.build_path()
for %{app: app} <- Mix.Dep.Umbrella.unloaded do
  Path.join([build, "lib", Atom.to_string(app), "ebin"]
end
```

Also please add a TODO that says: "TODO: Use Mix.Project.apps_path when we require Elixir v1.4+".

josevalim:
One last comment and we should be good to go!

lib/mix/tasks/docs.ex
wojtekmach: thanks, this is way better! Fixed in e3debf00352fb7e9b4c326e75298c4ad81015c86 where I also added some docs for running within an umbrella.

josevalim:

PR #626, opened 2016-10-22
Do not highlight function or type name left of ::

### LINKED ISSUE #607: Some functions are highlighted as types in typespecs
If the function name is the same as a type name it will be highlighted and linked to the type http://elixir-lang.org/docs/master/logger/Logger.html#level/0

screenshot 2016-09-28 19 18 57

wojtekmach:
Fixes #607

Before:
master1
master2

After:
fix1
fix2

Not sure if the change to types (dropping highlight on type name on left) is desired, but a) the link jumps to the same exact place, b) it seems more consistent as this is how functions are highlighted and c) implementation-wise it was easier to do. It's def. possible to scope this change to just functions though.

Btw, here're the changes this PR would have made on Elixir 1.3.4 docs: https://gist.github.com/wojtekmach/601771cbedac5a47d63c1c47d21d47e5

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -120,16 +120,15 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     typespec_to_string(other, typespecs, aliases, lib_dirs)
   end
 
-  defp normalize_left({:::, _, [{name, _, args}, right]}, typespecs, aliases, lib_dirs) do
-    placeholder = {:::, [], [{:_, [], args}, right]}
-    "_" <> rest = typespec_to_string(placeholder, typespecs, aliases, lib_dirs)
-    Atom.to_string(name) <> rest
+  defp normalize_left({:::, _, [{name, meta, args}, right]}, typespecs, aliases, lib_dirs) do
+    new_args = Enum.map(args, &[self(), typespec_to_string(&1, typespecs, aliases, lib_dirs)])
+    new_left = Macro.to_string {name, meta, new_args}, fn [pid, string], _ when pid == self() -> string; _, string -> string end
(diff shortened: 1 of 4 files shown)


lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -120,6 +120,22 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     typespec_to_string(other, typespecs, aliases, lib_dirs)
   end
 
+  defp normalize_left({:::, _, [{name, _, args}, right]}, typespecs, aliases, lib_dirs) do
+    placeholder = {:::, [], [{:_, [], args}, right]}
+    "_" <> rest = typespec_to_string(placeholder, typespecs, aliases, lib_dirs)
+    Atom.to_string(name) <> rest
```
josevalim: I am afraid though this breaks for operators. For example: `@type integer + integer :: integer` :(

wojtekmach: good catch, indeed it's broken: https://gist.github.com/wojtekmach/601771cbedac5a47d63c1c47d21d47e5#file-diff-diff-L710

josevalim: Maybe what we could do is to first convert the args and then call Macro.to_string with the converted args as strings:

```iex
iex(2)> {name, meta, args} = {:foo, [], [1, 2, 3]}
{:foo, [], [1, 2, 3]}
iex(3)> new_args = Enum.map(args, &[self(), typespec_to_string(&1, typespecs, aliases, lib_dirs)])
iex(4)> Macro.to_string {name, meta, new_args}, fn [pid, string], _ when pid == self() -> string; _, string -> string end
"foo(1, 2, 3)"
```

josevalim:

josevalim:
I have just pushed a change that makes the left side consistently add parens. So it will be `level() :: :error | :info | ...` which is more consistent with the signature we are just seeing.

PR #627, opened 2016-10-26
Fix autolinking to opaque type

wojtekmach:
Noticed that opaque types aren't correctly auto-linked. See e.g.: `t` here: http://elixir-lang.org/docs/master/elixir/MapSet.html#delete/2


josevalim:
What about `@typep`?

josevalim:

wojtekmach:
> What about @typep?

I noticed we're not currently rendering it in summary so I left if alone

josevalim:
Ah yes, good point. They are private after all.

PR #634, opened 2016-11-07
Fix missing assets in the header section

### LINKED ISSUE #619: Fix bugs in EPUB format
- [x] Mimetype file entry is missing or is not the first file in the archive.
- [x] Delete `OEBPS/assets/logo.png` at the end of the process, is still remaining.
- [ ] Fix missing assets in the header section: `<link type="text/css" rel="stylesheet" href="" />` or `<script src=""></script>`.
- [ ] Declare all the assets (images, JS, CSS) in the OPF manifest.
- [ ] Support the `:group` option from extras

milmazz:
* Reduce the number of args required in some functions.
* Fix references to assets in the OPF manifest.
* Compress JS files when creating ZIP

Fixes #619

### DIFF
--- a/lib/ex_doc/formatter/epub.ex
+++ b/lib/ex_doc/formatter/epub.ex
@@ -12,18 +12,14 @@ defmodule ExDoc.Formatter.EPUB do
   """
   @spec run(list, ExDoc.Config.t) :: String.t
   def run(module_nodes, config) when is_map(config) do
-    output =
-      config.output
-      |> Path.expand()
-      |> Path.join("#{config.project}-v#{config.version}")
-
-    File.rm_rf!(output)
-    File.mkdir_p!(Path.join(output, "OEBPS"))
-
     config = normalize_config(config)
-    HTML.generate_assets(output, assets(config))
+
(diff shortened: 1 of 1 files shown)


lib/ex_doc/formatter/epub.ex
```diff
@@ -34,20 +30,25 @@ defmodule ExDoc.Formatter.EPUB do
...
+    File.rm_rf!(config.output)
     epub_file
   end
 
   defp normalize_config(config) do
+    output =
+      config.output
+      |> Path.expand()
+      |> Path.join("#{config.project}-v#{config.version}")
+    config = Map.put(config, :output, output)
```
josevalim: `%{config | output: output}` would be semantically better (as the key `:output` is guaranteed to exist and is being updated).

josevalim:
@milmazz just one tiny comment then go ahead and :shipit: :D

lib/ex_doc/formatter/epub.ex
milmazz: Sweet!

PR #635, opened 2016-11-07
Filter modules by their prefix (fixes: #628)

milmazz:


lib/mix/tasks/docs.ex
```diff
@@ -90,6 +90,8 @@ defmodule Mix.Tasks.Docs do
     * `:canonical` - String that defines the preferred URL with the rel="canonical"
       element; defaults to no canonical path.
 
+    * `:filter_prefix` - Filter the module list by their prefix, Example: "ExDoc.Formatter.EPUB"
```
josevalim: Maybe a dot before "Example"? Also, maybe we could be a bit less concise? What about:

    Include only modules that match the given prefix in the generated documentation. Example: "MyApp.Core"

josevalim:
Another tiny comment and :shipit: :D

PR #639, opened 2016-11-11
Fix issues reported by `credo`

milmazz:
Fix all the checks reported by [credo][] (categories: warning, refactor, consistency, design, and readability), except those related to the line length.

Also, this commit also includes a `.credo.exs`, which allow us to change `credo` default configuration.

Finally, what do you think about to include [ebert][] in our contribution workflow as an automatic code reviewer. 

[credo]: https://github.com/rrrene/credo
[ebert]: https://ebertapp.io/


josevalim:
@milmazz can we please use this credo.exs file? https://github.com/plataformatec/linters/blob/master/.credo.exs

Thats the one ebert uses by default.

I will drop some comments.

lib/ex_doc/formatter/html.ex
```diff
@@ -143,9 +144,9 @@ defmodule ExDoc.Formatter.HTML do
       |> Enum.map(&Task.await(&1, :infinity))
 
     api_reference_headers =
-      if(Enum.empty?(modules),    do: [], else: [{"Modules", "modules"}]) ++
-      if(Enum.empty?(exceptions), do: [], else: [{"Exceptions", "exceptions"}]) ++
-      if(Enum.empty?(protocols),  do: [], else: [{"Protocols", "protocols"}])
+      if Enum.empty?(modules),    do: [], else: [{"Modules", "modules"}] ++
```
josevalim: This is definitely wrong. It changes the precedence of the code. If credo is asking us to remove the parenthesis on `if`, then it is definitely wrong. Maybe if we are using do/end blocks, but not for one-line ifs. /cc @rrrene

lib/ex_doc/retriever.ex
```diff
@@ -263,10 +263,11 @@ defmodule ExDoc.Retriever do
   defp get_defaults(signature, name, arity) do
     case Enum.count(signature, &match?({:\\, _, [_, _]}, &1)) do
       0 -> []
-      defaults -> for default <- (arity-defaults..arity-1), do: "#{name}/#{default}"
+      defaults -> for default <- (arity - defaults..arity - 1), do: "#{name}/#{default}"
```
josevalim: What do you think of `(arity - defaults)..(arity - 1)`. The current way reads weird. :)

lib/ex_doc/retriever.ex
```diff
@@ -263,10 +263,11 @@ defmodule ExDoc.Retriever do
...
     end
   end
 
+  @lint {Credo.Check.Refactor.PipeChainStart, false}
```
josevalim: Let's remove this. The credo.exs I suggested won't warn on those. :)

lib/ex_doc/retriever.ex
```diff
@@ -438,6 +439,7 @@ defmodule ExDoc.Retriever do
         do: behaviour
   end
 
+  @lint {Credo.Check.Refactor.PipeChainStart, false}
```
josevalim: Same here.

lib/mix/tasks/docs.ex
```diff
@@ -164,6 +165,7 @@ defmodule Mix.Tasks.Docs do
     Keyword.put_new(options, :source_beam, compile_path)
   end
 
+  @lint {Credo.Check.Design.TagTODO, false}
```
josevalim: Same here. TODO is disabled in the credo.exs (there are other things that checks TODOs in ebert).

mix.lock
```diff
@@ -1,4 +1,6 @@
-%{"cmark": {:hex, :cmark, "0.5.7", "7a3304e058dcb62dc73089eb2778aea0c7284ffcf67191cc40910bb36bcf5935", [:mix, :make], []},
+%{"bunt": {:hex, :bunt, "0.1.6", "5d95a6882f73f3b9969fdfd1953798046664e6f77ec4e486e6fafc7caad97c6f", [:mix], []},
+  "cmark": {:hex, :cmark, "0.5.7", "7a3304e058dcb62dc73089eb2778aea0c7284ffcf67191cc40910bb36bcf5935", [:mix, :make], []},
+  "credo": {:hex, :credo, "0.5.2", "92e8c9f86e0ffbf9f688595e9f4e936bc96a52e5606d2c19713e9e4d191d5c74", [:mix], [{:bunt, "~> 0.1.6", [hex: :bunt, optional: false]}]},
```
josevalim: We don't need to add credo as a dependency if we are going to use ebert. :)

lib/ex_doc/formatter/html.ex
josevalim: If we want to fix this while complying with Credo, we can write this line and the ones below as:

    (if Enum.empty?(modules),    do: [], else: [{"Modules", "modules"}]) ++

It will maintain the proper precedence and Credo shouldn't warn.

lib/ex_doc/retriever.ex
milmazz: Looks better the way you propose, I'll fix this!

lib/ex_doc/retriever.ex
milmazz: Ok, no problem.

lib/ex_doc/formatter/html.ex
milmazz: Oops, fix it!

milmazz:
@josevalim This [.credo.exs][] file includes the following check: `{Credo.Check.Refactor.FunctionArity}` and in our code base we can find functions with up to 7 arguments (arity). So, what do you suggest to do in this case?

We can change, at least for now, that check to: `{Credo.Check.Refactor.FunctionArity, max_arity: 7}`, wdyt?

[.credo.exs]: https://github.com/plataformatec/linters/blob/master/.credo.exs

josevalim:
> We can change, at least for now, that check to: {Credo.Check.Refactor.FunctionArity, max_arity: 7}, wdyt?

Sounds good.

josevalim:
@milmazz I think this is ready to go! 🎉

josevalim:
@milmazz here is the full report: https://ebertapp.io/github/elixir-lang/ex_doc/reviews/14

I have enabled ESLint on Ebert too and fixed most of the high arity checks in ExDoc by using maps. There are only three entries left and one of them is a TODO.

The TODO should be reported IMO but we can't act on it right now. :)

milmazz:
@josevalim Wow, great work! :heart:

milmazz:
@josevalim I'll work on the other issues, I'll keep you updated.

PR #640, opened 2016-11-12
Reduce function complexity

milmazz:


lib/ex_doc/retriever.ex
```diff
@@ -158,16 +158,18 @@ defmodule ExDoc.Retriever do
   defp generate_node(module, type, config) do
     source_url  = config.source_url_pattern
     source_path = source_path(module, config)
+    config = %{source_url: source_url, source_path: source_path}
 
     specs = get_specs(module)
     impls = get_impls(module)
     abst_code = get_abstract_code(module)
+    module_parts = %{specs: specs, impls: impls, abst_code: abst_code}
```
milmazz: @josevalim The name `module_parts` make sense for you in this context? or is it better `module_section`? any suggestion is more than welcome :)

lib/ex_doc/retriever.ex
```diff
@@ -267,14 +269,14 @@ defmodule ExDoc.Retriever do
     end
   end
 
-  defp get_callbacks(:behaviour, module, source_path, source_url, abst_code) do
+  defp get_callbacks(:behaviour, module, config, abst_code) do
     optional_callbacks = get_optional_callbacks(abst_code)
     (Code.get_docs(module, :callback_docs) || [])
     |> Enum.sort_by(&elem(&1, 0))
-    |> Enum.map(&get_callback(&1, source_path, source_url, optional_callbacks, abst_code))
+    |> Enum.map(&get_callback(&1, config.source_path, config.source_url, optional_callbacks, abst_code))
```
josevalim: Maybe we should pass the whole config down there too?

lib/ex_doc/retriever.ex
josevalim: Let's call it `module_info` and include the module name (`module`) and module type (`type`) on it? Then we can move its creation into something called `get_module_info(module)`.

lib/ex_doc/retriever.ex
```diff
@@ -158,16 +158,18 @@ defmodule ExDoc.Retriever do
   defp generate_node(module, type, config) do
     source_url  = config.source_url_pattern
     source_path = source_path(module, config)
+    config = %{source_url: source_url, source_path: source_path}
```
josevalim: Since we already have a variable named `config`, this can be confusing. What about calling this `source` with fields `url` and `path`?

lib/ex_doc/retriever.ex
josevalim: If we store the module name and type, we will also be able to pass the `module_info` to other functions such as `get_types`. :)

josevalim:
Beautiful!

josevalim:

PR #644, opened 2016-11-18
Declare assets in the OPF manifest

milmazz:


lib/ex_doc/formatter/epub/templates.ex
```diff
@@ -68,4 +70,16 @@ defmodule ExDoc.Formatter.EPUB.Templates do
   EEx.function_from_file(:defp, :head_template,
                          Path.expand("templates/head_template.eex", __DIR__),
                          [:config, :page])
+
+  @media_types
```
josevalim: I think we could do the FIle.read! directly here as we don't use the module attribute anywhere.

lib/ex_doc/formatter/epub/templates.ex
```diff
@@ -68,4 +70,16 @@ defmodule ExDoc.Formatter.EPUB.Templates do
...
+  |> String.split("\n", trim: true)
+  |> Enum.each(fn(line) ->
+    [extension, media] = line |> String.strip() |> String.split(",")
```
josevalim: The strip should not be necessary.

lib/ex_doc/formatter/epub/templates.ex
```diff
@@ -68,4 +70,16 @@ defmodule ExDoc.Formatter.EPUB.Templates do
...
+
+    def media_type("." <> unquote(extension)) do
+      unquote(media)
+    end
+  end)
+
+  def media_type(_), do: nil
```
josevalim: Should we default to "application/octet-stream" or something of sorts? Otherwise the media-type field will be empty, or is that indeed the preferred behaviour?

lib/ex_doc/formatter/epub/templates.ex
milmazz: I thought about it, but, in this case, the EPUB documents only support a mime-types subset, more information about this in [EPUB 3 Core Media Types](https://idpf.github.io/epub-cmt/v3/)

lib/ex_doc/formatter/epub/templates.ex
milmazz: Agree

lib/ex_doc/formatter/epub/templates.ex
milmazz: Actually, we should raise an exception instead of returning `nil`, that's because EPUB documents only accept the mime-type listed in [EPUB 3 Core Media Types](https://idpf.github.io/epub-cmt/v3/) . wdyt?

lib/ex_doc/formatter/epub/templates.ex
```diff
@@ -68,4 +68,18 @@ defmodule ExDoc.Formatter.EPUB.Templates do
...
+  |> String.split("\n", trim: true)
+  |> Enum.each(fn(line) ->
+    [extension, media] = String.split(line, ",")
+
+    def media_type("." <> unquote(extension)) do
+      unquote(media)
+    end
+  end)
+
+  def media_type(_), do: raise "publication resource not supported in EPUB format"
```
josevalim: Let's include the argument in the error message and try to make it clearer. Something like:

    raise "asset with extension #{inspect arg} is not supported by EPUB format"

What do you think?

milmazz:
@josevalim I think this is ready to :ship:

lib/ex_doc/formatter/epub/templates.ex
```diff
@@ -68,4 +68,18 @@ defmodule ExDoc.Formatter.EPUB.Templates do
...
                          [:config, :page])
+
+  "templates/media-types.txt"
+  |> Path.expand(__DIR__)
+  |> File.read!()
+  |> String.split("\n", trim: true)
+  |> Enum.each(fn(line) ->
+    [extension, media] = String.split(line, ",")
+
+    def media_type("." <> unquote(extension)) do
```
josevalim: Btw, this should be `defp`, no?

josevalim:
@milmazz one extra tiny comment and feel free to ship it. 💯

PR #649, opened 2016-11-28
Visual Modifications

tzveeJ:
Ex_doc has a handful of visual design choices that have caused me trouble in understanding where I am on the page and what section I'm looking at. A few months ago I posted an issue which was closed pending an actual PR with the modifications. 

The commits here include some relatively simple visual changes for section headers, function headers, code blocks, as well as converting the summary section into a table. As a whole, I believe these modifications make Ex_doc significantly easier to read.

Please review.


michalmuskala:
It would be wonderful to get some comparative screenshots showing the differences.

josevalim:
@tzveeJ can you please not include changes to the compiled files? Those will very likely cause merge conflicts down the road (it already has), so the sooner we get rid of them, the happier we will all be. :)

josevalim:
Here are some screenshots:

screen shot 2016-11-28 at 19 30 26



screen shot 2016-11-28 at 19 30 03



screen shot 2016-11-28 at 19 30 08

josevalim:
FWIW, I like the increased font-size for function names and I like the grey background for code snippets.

josevalim:
I have incorporated some of the changes above in master so we have a build up soon. I have not incorporated the summary changes because I believe those require further discussion.

dignifiedquire:
I don't think the code background changes are good, they make it much harder to read for me and call unneeded attention to inline code snippets.

Also the large gray ares in the summary are really unfitting to the rest of the design and don't help readability in any way for me.

tzveeJ:
@josevalim Do you still need me to do anything re: removing changes from compiled files? Also, I'm happy to participate in further discussion of changes to the summary section. My reasoning for creating a table was in order to create a clear visual difference between the summary section and the full function descriptions. I've also seen a similar table format in other documentation.

@dignifiedquire In regards to the background behind code, I removed the border in order to actually call less attention to code, and I made the background only slightly darker to compensate. I didn't think it was very significant? Also, by "the large gray areas in the summary" are you referring to empty space in the table or something else you're seeing? I see Jose is holding off on the summary.

josevalim:
@tzveeJ I would love if you could rebase if we are going to merge the summary changes in. But there is no need to rebase before we are sure we will merge this in.

@tzveeJ @dignifiedquire I have moved the background color back to the original and removed the border so we can see how it looks like. I have found it the best of the ones we have seen so far. The border was indeed the one given unnecessary "weight".

screen shot 2016-11-28 at 21 46 00


screen shot 2016-11-28 at 21 46 14

michalmuskala:
I wonder if it would make sense to give negative margin to the code blocks, so that the code in the block is aligned with the regular text, and the background "sticks out" a little.

josevalim:
@michalmuskala don't wonder, use the inspector to try it out and take a screenshot. :P

tzveeJ:
@josevalim I hope my other suggestions for the function names don't get lost though! Those changes were the reason I started doing this. Currently, the gray section around function names creates the impression that the function name is some sort of code block or special section. But the function name is just a regular title, so there's no reason to give it any additional background. There does need to be additional space after each function to make the separation clearer, which is why I added some.

josevalim:
@tzveeJ I have a branch with all of your changes as is and I will be using them locally for a while. But originally I was not convinced of those particular changes and I am waiting for more feedback before porting them to master (the same as summary).

whatyouhide:
I agree with @dignifiedquire that the changes to the summary are quite unfitting with the current design 😕

josevalim:
The master version of the docs is live: https://hexdocs.pm/elixir/master/Kernel.html

It includes borderless code snippets, increase font-size for function headers and slightly less line-height for content.

dignifiedquire:
@josevalim looking at the live version, I'm not convinced any of these changes improve things compared to the current state. I can understand that for some the font-size is too small but there is always browser zoom to adjust for that on a personal basis. I took great care in matching the different font sizes and line heights on the site together and seeing these changes now disrupts that flow in my opinion.
Regarding the border around the code, yes it does remove the importance but it makes it also less separate from the main text, so not convinced here either, sorry :/

josevalim:
Alright. I will revert those changes then, thank you.

josevalim:
> Regarding the border around the code, yes it does remove the importance but it makes it also less separate from the main text, so not convinced here either, sorry :/

Do you think we can skip it at least from the inline code samples? After reverting the changes for a couple hours I do find the borders very harmful for reading.

josevalim:
The reasoning is that we already change the font and add a background, so those two changes should be enough to show it is a `code` mark?

milmazz:
@tzveeJ Can we close this? or how do want to proceed? /cc @josevalim @dignifiedquire

tzveeJ:
I've been waiting on @dignifiedquire to work with me over email. Here's a list of the concerns I was trying to solve that I sent to him over email.

**1. Removing border around code blocks.**

My issue was that the border lines make code blocks appear to be too separated, even though they are generally examples within sections (thus not separated). But without the border, the background was not noticeably different enough, so I made the background slightly darker. Apparently that came off as too dark. I propose to remove the border and make the background somewhere in between.

**2. Adding border line below major section headers.**

The driver for many of the changes I made is my issue with knowing where I am on the page. A simple line below major sections really helps break up the page with some consistency.

**3. Changing details list into a table called summary table.**

Yeah, this was a riskier move. It was inspired by the Android documentation style, e.g. [ArraySet](https://developer.android.com/reference/android/util/ArraySet.html). I wanted to make clearer separation between the description that is a summary and the full documentation, by putting the former in a table. As a new Elixir developer, I occasionally wound up reading the summary documentation and didn't realize there was anything more to find. This clearer separation, combined with blue (clickable) links, would have stopped my confusion.

**4. Convert function titles to normal header-like text style, add padding below functions for separation.**

The current function title style looks to me like a code block. It's not a standard style for a title. The idea of wrapping a title in a dark background with a purple line next to it makes it appear like a separate section, unrelated to the content above or below. In my opinion, a traditional bold header combined with a little extra space below each function description creates some clear spacing and readability.

**5. Slight text spacing tightening.**

It's a matter of opinion, I like text a bit tighter. But that change can be nixed, I don't feel strongly.

dignifiedquire:
Sorry for not responding sooner :/ things got a bit hectic over the holidays. 

- Re 1. Maybe a reduction of the border in strength might help, I'm still not convinced that dropping it + changing the background improves things.
- Re 2. This is a matter of preference i guess, both space and horizontal lines have been historically used to differentiate sections. The reason this is using space rather than lines, is that a lot of the design was oriented on book design/layout which in most cases uses simple spacing to differentiate sections and paragraphs. 
- Re 3. Table style: I am not generally opposed to using a table instead of a list, it does make sense to a degree, but it needs to be carefully designed to fit inline with the rest of the style.
- Re 4. function headers: It was on purpose that they look similar to code blocks, I disagree with that they look like a separate section, I think the distinction helps making it clear where on part starts and one ends. 
- Re 5. Lets keep it like it is currently, I'm not looking forward to redoing all sizing calculations to be honest.

tzveeJ:
Re 1. I can agree on that. I do think making the border lighter comes with a slight need to make the code background a drop darker (just a drop).
Re 2. You've helped me clarify what I'm thinking. I believe what throws me off is that h1 and h2 are similar in size, and the spacing after h1s is also similar to the spacing after h2s, ultimately making it unclear whether you're looking at an h1 or an h2. This could be because it's a computer screen and not a book, I'm not sure - but since I don't think I could use spacing or font size to improve what I'm concerned about visually (maybe we can?), the horizontal bar seems like the answer to me.
Re 3. What were some aspects of the table design in the images above that you disliked?
Re 4. I don't think making a function title look like a code block is a good decision. It throws me off and probably others - it's the wrong visual identifier since a function title is not a code block. What book are we talking about, by the way?
Re 5. Ok.

milmazz:
@tzveeJ Are you working on this? This PR seems to include too many changes. May I suggest split those changes in different PRs? Hopefully, that way the discussion will be more fluent between you and @dignifiedquire, WDYT?

tzveeJ:
@milmazz That sounds like a great idea. You can close this one. Unfortunately, though, I don't have the capacity to take it on right now. Feel free to break anything out if you want.

PR #652, opened 2016-12-12
Automatically detect GitLab source URL patterns

jparise:
GitLab has become quite popular, and its source URL scheme is identical
to GitHub's. That makes it useful and easy to support by default.


milmazz:
Thanks, LGTM!

josevalim:

Issue #646, opened 2016-11-21
Remove the distinction between functions and macros from the sidebar


josevalim:
Today we split entries between functions and macros which is not useful for discoverability. If I am learning a new API, I typically do not care if something is a macro or a function.

I propose to remove the distinction between functions and macros from the sidebar and use the tag to mark macros as macros. The tag works similar to here:

http://elixir-lang.org/docs/master/elixir/GenServer.html#c:format_status/2

Notice the "(optional)" tag on the right-hand side.

jparise:
I took at shot at this in #653, and I'm happy with the results. Discover definitely feels improved. The only negative aspect of the change that I've felt so far is that, for some complex modules (e.g. Kernel), the function list is now quite long. I suspect that's atypical, however, and says more about how the module is organized and less about the ratio of exported macros vs functions.

josevalim:
We are keeping this one open because we removed the distinction between Functions / Macros from the sidebar but we should find a way to make it slightly more obvious that something is a macro in the full listing.

josevalim:
This is in master!

PR #653, opened 2016-12-21
List both functions and macros under "Functions"

jparise:
Per #646, this removes the top-level distinction between functions and
macros in the sidebar. Macros are listed alongside functions and include
an additional "(macro)" annotation tag:

screen shot 2016-12-21 at 9 49 31 am

### DIFF
--- a/lib/ex_doc/formatter/epub/templates.ex
+++ b/lib/ex_doc/formatter/epub/templates.ex
@@ -30,8 +30,7 @@ defmodule ExDoc.Formatter.EPUB.Templates do
   @doc """
   Creates a chapter which contains all the details about an individual module.
 
-  This chapter can include the following sections: *functions*, *macros*,
-  *types*, *callbacks*.
+  This chapter can include the following sections: *functions*, *types*, *callbacks*.
   """
   EEx.function_from_file(:def, :module_template,
                          Path.expand("templates/module_template.eex", __DIR__),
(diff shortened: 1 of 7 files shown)


josevalim:
Thank you! :+1:

I would love some discussion on this before merging too. /cc @whatyouhide @lexmag @ericmj @eksperimental @milmazz

dignifiedquire:
I like it, but don't use the documentation enough currently to know how much sense this makes in the overall picture.

jparise:
My experience so far (repeated from the issue so we can have a single discussion here):

> Discovery definitely feels improved. The only negative aspect of the change that I've experienced so far is that, for some complex modules (e.g. Kernel), the function list is now quite long. I suspect that's atypical, however, and says more about how the module is organized and less about the ratio of exported macros vs functions.

whatyouhide:
I like the idea of having functions and macros in the same place but maybe (maybe) the current design "hides" the fact that some macro is a macro a bit too much. I am not sure how to solve this though, so sorry for the flaky feedback :(

eksperimental:
I haven't installed this version yet, but I agree with what @whatyouhide said.
maybe not putting it in parenthsis, and not in light grey, but in black it could say macro/function/type/callback, etc.

josevalim:
Shall we move this forward and then figure out a way to make annotations clearer?

lexmag:
I also like idea of "merging" functions and macros, and I think it looks good at its current state to move forward and figure out better annotations on the next step. 👍

whatyouhide:
Yeah moving forward makes sense, as I said I don't have suggestions on how to improve so let's go with this for now :) Great work @jparise!

eksperimental:
also. having them separate doesn't make much sense, haven't installed this version but feels like the right direction

michalmuskala:
How about using a different colour for functions and macros for the vertical bar on the left? It doesn't have to be a huge difference, but could give a very easy way to distinguish them.

josevalim:

PR #660, opened 2017-01-09
EPUB format: Split TOC in groups

milmazz:
screen shot 2017-01-08 at 10 00 35 pm


lib/ex_doc/formatter/epub.ex
```diff
@@ -85,6 +88,26 @@ defmodule ExDoc.Formatter.EPUB do
...
+           group = if x.group != "", do: x.group, else: "Extras"
+           extras_by_group = Map.update(extras_by_group, group, [x], &([x | &1]))
+           groups = Map.put_new(groups, group, index)
+           {extras_by_group, groups}
+         end)
+
+    groups
+    |> Map.to_list()
+    |> List.keysort(1)
+    |> Enum.reduce([], fn({k, _}, acc) ->
```
josevalim: You probably want to use a `map` instead of `reduce+Keyword.put+reverse`.

milmazz: Nice catch!

PR #665, opened 2017-01-19
Warn when redirecting to a non-existent file

jparise:
This makes it a little more obvious when `main:` references an invalid
document or node.


jparise:
I originally planned to specifically validate `main:` by comparing it against the full set of possible redirect targets: primarily module pages and extras. That would let us catch misconfigurations earlier (before generation), but the implementation of that check is a bit more involved than this simple file-existence test.

I'm offering this initial attempt for feedback on the idea in general. I'm happy to revisit with an alternate implementation if the idea has merit.

jparise:
The other thing I don't love about my current approach is that it assumes we've written the redirect target file before we write the redirecting file. That's not too fragile of an assumption, but it's a new precondition nonetheless.

josevalim:
Let's give it a try for now. We are only emitting a warning, so if they are false negatives, we can act on them.

josevalim:

Issue #667, opened 2017-01-30
Prettify struct documentation


whatyouhide:
Related to https://github.com/elixir-lang/elixir/pull/5718. I noticed that the `Task` module in the Elixir stdlib was documenting the `%Task{}` struct by using `@doc` above `defstruct`. This caused the [`Task.__struct__/0`](https://hexdocs.pm/elixir/Task.html#__struct__/0) function to be documented; however, I don't think this is a nice way to document structs as `__struct__/0` is more of an internal tool used for working with structs. As the first step to fix this, we could treat `__struct__` specially and show it as documentation for `%MyStruct{}` instead.

@josevalim would it look the same as `__struct__/0` visually for things like positioning in the functions list and so on? Or should we treat this differently, for example by giving structs their own section?

josevalim:
Due to #646, I would keep it all together. You may change the annotation to `(struct)` though. We should probably have it always come on the top as well.

whatyouhide:
I'll look into implementing this.

whatyouhide:
Closed by #670.

PR #670, opened 2017-02-02
Handle __struct__/0 by displaying the struct instead

whatyouhide:
First shot at #667.

With this commit, we 

* annotate `__struct__/0` functions (like we do with macros and optional callbacks)
* display them as `%Task{}` instead of `__struct__/0` (for the `Task` example of course)

I am still showing them as `__struct__/0` in their anchor and in the sidebar, I am not sure how to proceed: for example, the anchor is URL-encoded as hell if we show `%Task{}`. Thoughts?

### DIFF
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -243,12 +243,23 @@ defmodule ExDoc.Retriever do
             |> Enum.map(&Typespec.spec_to_ast(name, &1))
             |> Enum.reverse()
 
-    annotations = case type do
-      :defmacro ->
-        ["macro"]
-      _ ->
-        []
-    end
+    annotations =
+      case {type, name, arity} do
+        {:defmacro, _, _} ->
+          ["macro"]
+        {_, :__struct__, 0} ->
+          ["struct"]
+        _ ->
+          []
+      end
+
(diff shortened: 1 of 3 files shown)


josevalim:
> I am still showing them as __struct__/0 in their anchor and in the sidebar, I am not sure how to proceed: for example, the anchor is URL-encoded as hell if we show %Task{}. Thoughts?

You would have to escape `%` but showing it as __struct__/0 is completely fine IMO.

josevalim:
And we need a test. :)

whatyouhide:
> And we need a test. :)

Yes of course :), I just wanted to gather some feedback on the approach.

> You would have to escape % but showing it as struct/0 is completely fine IMO.

Did you mean `__struct__/0` and the `_`s got formatted by GH? Anyhow we can show `%Task{}` as well, up to the ExDoc team :). Let me know which one you wanna go with. :)

josevalim:
Oops, sorry. Let's continue showing it as `__struct__/0`. It is important so we don't break existing links as well. :)

whatyouhide:
@josevalim updated with tests. Let me know how it looks.

josevalim:

PR #672, opened 2017-02-13
Fix some issues found by ebertapp.io

milmazz:


josevalim:
@milmazz I would prefer to not have a `.credo.exs` file. Why do we need one?

Everything else looks beautiful.

milmazz:
@josevalim In some cases, we want to ignore some issues, for example:

* In the directory `test/fixtures` (e.g. @moduledoc not present for tests)
* Set 6 as maximum arity instead of 5

wdyt?

josevalim:
We should probably fix the arity ones although I am not sure what to do about the test/fixtures. :S Maybe we should use a `@lint` annotation on those?

milmazz:
> We should probably fix the arity ones

@josevalim I agree, I can do that in another PR/commit.

> I am not sure what to do about the test/fixtures. :S Maybe we should use a `@lint` annotation on those?

I think we cannot use the inline configuration via `@lint` attributes at the top level `defmodule`, but, it should work on the nested ones. @rrrene Can you confirm this?

rrrene:
@milmazz You are right. `@lint` does not work on the module level. 

I just decided to deprecate the use of `@lint` in favor of a comment-based format. There is a discussion in rrrene/credo#291 where @josevalim rightly pointed out that maybe `@lint` attributes should not be module attributes since they are not executable code. After thinking about it, I tend to agree.

Now we have to figure out a better syntax for the "magic control comments", but I am optimistic that we can find an elegant solution. Everybody's input on this would be very welcome :+1:

josevalim:
Thanks @rrrene.

@milmazz while I would personally prefer to not have a .credo.exs since when a new credo version is released I will need to update many repos instead of a single source, feel free to proceed as you prefer.

milmazz:
@rrrene Thanks for the info.

@josevalim I already removed the `.credo.exs` file. That way we can see that we have some pending issues to fix and alleviate the maintenance burden a bit.

PR #680, opened 2017-03-03
Fix ExDoc.CLI.main/2 to keep --source-ref 

eksperimental:
--source-ref passed to the CLI was being lost,
meaning all links to commits/tags were being lost and the docs were always
linked to master


lib/ex_doc/cli.ex
```diff
@@ -8,12 +8,30 @@ defmodule ExDoc.CLI do
   transform the Markdown documents into a specified format (default is HTML).
   """
   def main(args, generator \\ &ExDoc.generate_docs/3) do
-    {opts, args, _} = OptionParser.parse(args,
-               aliases: [o: :output, f: :formatter, c: :config, r: :source_root,
-                         u: :source_url, m: :main, p: :homepage_url, l: :logo,
-                         e: :extra, v: :version, n: :canonical, s: :extra_section,
-                         a: :assets, i: :filter_prefix],
-               switches: [extra: :keep])
+    {opts, args, _invalid} = OptionParser.parse(args,
```
milmazz: Why not something like this?

```elixir
{opts, args, _invalid} =
  OptionParser.parse(args, ...
```

lib/ex_doc/cli.ex
```diff
@@ -8,12 +8,30 @@ defmodule ExDoc.CLI do
...
+        r: :source_root,
+        u: :source_url,
+        m: :main,
+        p: :homepage_url,
+        e: :extra,
+        s: :extra_section,
+        a: :assets,
+        l: :logo,
+        n: :canonical,
+        v: :version,
```
milmazz: WDYT about sorting this aliases?

lib/ex_doc/issue446.ex
```diff
@@ -0,0 +1,17 @@
+# file: lib/ex_doc/issue446.ex
```
milmazz: Do you really want to include this file?

test/ex_doc/cli_test.exs
```diff
@@ -65,4 +65,56 @@ defmodule ExDoc.CLITest do
...
+
+        # invalid options
+        "--nonexistant",
+        "--non_existant",
+        "--idontexist", "baz",
+        "--i_dont_exist", "baz",
+      ]
+
+    {project, version, opts} =
+      run(args)
```
milmazz: Why not something like this:

```elixir
{project, version, opts} = run(args)
```

lib/ex_doc/issue446.ex
eksperimental: nope. it was a mistake. it's from a really old PR

lib/ex_doc/cli.ex
eksperimental: nope. the idea (as the title of the commit suggests) is to keep the same order as the `exdoc --help` command lists them

test/ex_doc/cli_test.exs
eksperimental: oops. that was not intented

eksperimental:
@milmazz thank your for your input. please have a look again.

lib/ex_doc/cli.ex
```diff
@@ -8,12 +8,28 @@ defmodule ExDoc.CLI do
...
+      ],
+
+      switches: [
+        extra: :keep,
+      ]
```
milmazz: Missing comma after `]`

lib/ex_doc/cli.ex
milmazz: Ok, I see, then, It would be nice to see an inline note about your suggestion because it's easy to see the pattern when those aliases are ordered alphabetically, but no when they are listed in same order as the docs, wdyt?

lib/ex_doc.ex
```diff
@@ -25,6 +25,7 @@ defmodule ExDoc do
       project: nil,
       retriever: ExDoc.Retriever,
       source_beam: nil,
```
milmazz: WDYT about removing those `nil` from `defstruct/1`?

eksperimental: actually, I was thinking of that. should we use an empty string?

lib/ex_doc/cli.ex
eksperimental: it will be weird to have a `)` after the comma, ex: `switches: [...], )`

lib/ex_doc.ex
milmazz: The default value is `nil`, so, you can have something like this:

```elixir
defstruct [
  :main,
  :project,
  output: "doc",
  retriever: ExDoc.Retriever
]
```

eksperimental: but what's the point since those are not required.

eksperimental: oh. I see now what you mean.
I was thinking why not have a default empty string.
I will be submitting another PR, and because we are using `nil` I need to check agains `nil`, `false` or `""`  to see if an option is unset. If by default we set to to empty, we just need to check against `false` or `""` (and false is because people may have just used `--foo` without value, and it because a boolean argv.

josevalim: > I will be submitting another PR, and because we are using nil I need to check agains nil, false or "" to see if an option is unset.

We should continue only checking for `nil`. In Elixir, `nil` is used for absence of information (except if we have a field that may be boolean).

If the field can be an empty string and an empty string also mean "emptiness", we need to convert the empty string to nil before putting it in the struct.

lib/ex_doc.ex
```diff
@@ -25,6 +25,7 @@ defmodule ExDoc do
...
+      source_ref: "master",
```
josevalim: I don't understand why this is needed. `:source_ref` is not needed in the struct inside ExDoc. We only use it initially when building the source_url.

lib/ex_doc/cli.ex
```diff
@@ -8,12 +8,31 @@ defmodule ExDoc.CLI do
...
+          s: :extra_section,
+          a: :assets,
+          l: :logo,
+          n: :canonical,
+          v: :version,
+        ],
+
+        switches: [
+          extra: :keep,
+          source_ref: :string,
```
josevalim: Same here. This is not needed.

eksperimental: this is the line that was missing and cause source_ref to be ditched. So I would say it is a total must to be here.

lib/ex_doc.ex
eksperimental: shouldn't the struct reflect what can be passed via config, such as CLI or the `docs` task?
https://github.com/elixir-lang/ex_doc/blob/05eb89e93559e129744cf021d580f14cf6fff313/lib/mix/tasks/docs.ex#L62

lib/ex_doc.ex
eksperimental: I have another PR that is waiting for this one to be merged, and it deals with normalizing options individually, I guess I can include this one rule

lib/ex_doc.ex
```diff
@@ -37,16 +38,17 @@ defmodule ExDoc do
        deps: [{ebin_path :: String.t, doc_url :: String.t}],
```
eksperimental: @josevalim according to what you said, I am uncertain whether these typespecs should be `String.t` or `nil | String.t`

lib/ex_doc.ex
josevalim: There is no need to keep the source_ref around if we don't use it internally. The goal of ExDoc.build_config is build a set of options and build a struct that is relevant internally, if we don't use source_ref past building the config, there is no point in keeping it in the struct.

eksperimental: Ok, I guess if we ever need to find out what was passed by the user, we can add it later to it

lib/ex_doc.ex
milmazz: > If the field can be an empty string and an empty string also mean "emptiness", we need to convert the empty string to nil before putting it in the struct.

Yes, we should preprocess those empty strings and convert the data to `nil`. Also, it looks better doing this:

```elixir
if options.string_field do
  # ...
end
```

than this:

```elixir
if  options.string_field != "" do
  # ...
end
```

:)

lib/ex_doc/cli.ex
josevalim: Ah, it is missing it because of the changes in Elixir v1.4 regarding atoms.

josevalim: It would be best for us to use `:strict` then and define all inputs. We can even move to use `parse!` so we get proper argument handling.

josevalim:
I have merged this right now so we get proper docs but further improvements based on the comments are still welcome. :)

PR #694, opened 2017-03-25
657 properly set link in types with parameter

### LINKED ISSUE #657: Parameterized type in function spec generates an extra anchor tag
I have something similar to the following defined:
```Elixir
@type t :: term
@opaque finger_tree(t) :: ...

@spec append(finger_tree(t), t) :: finger_tree(t)
def append(arg, y) do
 ...
end
```
When docs are generated for this, the entry for `append` looks like

image

It appears that it is trying to add another link to the entry for type `t`

ymtszw:
Should fix #657

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -148,7 +148,8 @@ defmodule ExDoc.Formatter.HTML.Autolink do
         arity = length(args)
         if {name, arity} in typespecs do
           n = enc_h("#{name}")
-          ~s[<a href="#t:#{n}/#{arity}">#{h(string)}</a>]
+          {string_to_link, string_with_parens} = split_string_to_link(string)
+          ~s[<a href="#t:#{n}/#{arity}">#{h(string_to_link)}</a>#{string_with_parens}]
         else
           string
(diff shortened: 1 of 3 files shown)


lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -179,6 +181,13 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   end
   defp strip_parens(string, _), do: string
 
+  defp split_string_to_link(string) do
+    case :binary.match(string, "(") do
```
josevalim: Use :binary.split and then you don't need to use match+String.split_at. :)

josevalim:
I have added one tiny comment and this is good to go! Thank you!

lib/ex_doc/formatter/html/autolink.ex
aYuMatsuzawa: 0c851c6 here it is.
I thought it should be OK here to exact-match against 2- or 1-element lists, correct?

ymtszw: ^ sorry, alt account

josevalim:

PR #697, opened 2017-04-08
Fix project name wrapping

### LINKED ISSUE #696: Project name that is too long wraps weirdly
It seems #621 breaks wrapping of a project name that is too long, since it wraps on any character. This is what it looks like on my project:

https://files.nytsoi.net/NFyT.png

This is what it looks like if you remove `word-break: break-all`:

https://files.nytsoi.net/iFNw.png

If I remove the break-all and make my browser window smaller, I cannot see any wrapping issues. But maybe the issue fixed in #621 reproducible only on certain mobiles?

Nicd:
Changes project name wrapping to use normal word wrapping
instead of break-all. This prevents a long project name from
being wrapped mid-word while still allowing long words to
be wrapped properly.

Fixes #696

### DIFF
--- a/assets/less/sidebar.less
+++ b/assets/less/sidebar.less
@@ -62,7 +62,7 @@
     margin: 0;
     padding: 0;
     max-width: 230px;
-    word-break: break-all;
+    word-wrap: break-word;
   }
 
   .sidebar-projectVersion {


josevalim:

PR #706, opened 2017-05-11
Strip nesting html tags in headers. Closes #705

### LINKED ISSUE #705: h2 header with `link` breaks generation navi part without any warning/error?
I didn't notice it immediately, but when i published my package i noticed that navi part of my docs has nothing https://hexdocs.pm/elivndb/EliVndb.html

It seems to happen due to me putting following header in README:
```md
## `EliVndb.Client` types
```

There is no error or warning.
But all other pages are generated without issues. And if i remove quotes then navi part will have content:
```md
## EliVndb.Client types
```

[README.md](https://raw.githubusercontent.com/DoumanAsh/EliVndb/bc007c7c3638d11402494c5414a6ea531446729a/README.md) of commit which got published to hex.pm

Should there be warning\error on invalid header?

DoumanAsh:
I cannot run tests locally as it seems cmark depends on `make` tool and i have no means to build it on windows :(

### DIFF
--- a/lib/ex_doc/formatter/html/templates.ex
+++ b/lib/ex_doc/formatter/html/templates.ex
@@ -144,10 +144,12 @@ defmodule ExDoc.Formatter.HTML.Templates do
   end
 
   @h2_regex  ~r/<h2.*?>(.+)<\/h2>/m
+  @clean_html_regex ~r/<(?:[^>=]|='[^']*'|="[^"]*"|=[^'"][^\s>]*)*>/
   defp extract_headers(content) do
     @h2_regex
     |> Regex.scan(content, capture: :all_but_first)
     |> List.flatten()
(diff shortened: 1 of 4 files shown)


lib/ex_doc/formatter/html/templates.ex
```diff
@@ -209,7 +211,7 @@ defmodule ExDoc.Formatter.HTML.Templates do
   @spec header_to_id(String.t) :: String.t
   def header_to_id(header) do
     header
-    |> String.replace(~r/<.+>/, "")
+    |> String.replace(@clean_html_regex, "")
```
DoumanAsh: This function is used separately in templates to generate ID.
Therefore proper clean-up is needed here too

josevalim: Perfect. :)

PR #708, opened 2017-05-13
Support default configuations

eksperimental:


whatyouhide:
Would it make sense to have these as configurations of the `:ex_doc` application?

lib/ex_doc.ex
```diff
@@ -9,22 +9,41 @@ defmodule ExDoc do
...
+
+    @default %{
+      :formatter => "html",
+      :output => "./doc",
+      :source_ref => "master",
+      :retriever => ExDoc.Retriever,
+    }
+
+    @spec default :: map
+    def default do
```
josevalim: We can remove this function as it does not seem to be used anywhere.

eksperimental:
It would, but I'm afraid we cannot do it, because we need to value in `defstuct` at compile time, and I think we need the application to be compiled to be able to access the values in mix.exs

I'm getting this error, let me know if you think I'm doing something wrong or it could be done some other way.
> ** (ArgumentError) application :ex_doc is not loaded, or the configuration parameter :default_config is not set

lib/ex_doc.ex
eksperimental: @josevalim done!

eksperimental:
:heart:

PR #709, opened 2017-05-15
Introduce ExDoc.normalize_options/3 and new helpers

eksperimental:
Replace ExDoc.normalize_options/1 with a more general function named ExDoc.normalize_options/3

Introduce:
- ExDoc.normalize_source_url_pattern/2
- ExDoc.normalize_directory/2

### DIFF
--- a/lib/ex_doc.ex
+++ b/lib/ex_doc.ex
@@ -93,19 +93,21 @@ defmodule ExDoc do
     find_formatter(config.formatter).run(docs, config)
   end
 
+  @doc false
   # Builds configuration by merging `options`, and normalizing the options.
   @spec build_config(String.t, String.t, Keyword.t) :: ExDoc.Config.t
-  defp build_config(project, vsn, options) do
+  def build_config(project, vsn, options) do
     options = normalize_options(options)
+
     preconfig = %Config{
       project: project,
       version: vsn,
-      main: options[:main],
-      homepage_url: options[:homepage_url],
       source_root: options[:source_root] || File.cwd!,
     }
     struct(preconfig, options)
   end
+
+  # Helpers
 
   # Short path for programmatic interface
   defp find_formatter(modname) when is_atom(modname), do: modname
@@ -129,14 +131,43 @@ defmodule ExDoc do
     modname
   end
 
-  # Helpers
+  @doc false
+  # Removes trailing slash from paths
+  def normalize_paths(options, fields) when is_list(fields) do
+    Enum.reduce(fields, options, fn(field, options_acc) ->
+      if is_binary(options_acc[field]) do
+        Keyword.put(options_acc, field, String.trim_trailing(options_acc[field], "/"))
+      else
+        options_acc
+      end
+    end)
+  end
 
-  defp normalize_options(options) do
-    pattern = options[:source_url_pattern] || guess_url(options[:source_url], options[:source_ref] || ExDoc.Config.default(:source_ref))
-    options = Keyword.put(options, :source_url_pattern, pattern)
+  @doc false
+  def normalize_source_url_pattern(options, default_source_ref \\ ExDoc.Config.default(:source_ref)) do
+    if !options[:source_url_pattern] && options[:source_url] do
+      source_ref = options[:source_ref] || default_source_ref
+      Keyword.put(options, :source_url_pattern, guess_url(options[:source_url], source_ref))
+    else
+      options
+    end
+  end
 
-    if is_bitstring(options[:output]) do
-      Keyword.put(options, :output, String.trim_trailing(options[:output], "/"))
+  @doc false
+  def normalize_options(options) do
+    options = normalize_options(options, :source_ref, ExDoc.Config.default(:source_ref))
+
+    options
+    |> normalize_paths([:assets, :output, :source_root])
+    |> normalize_source_url_pattern(options[:source_ref] || ExDoc.Config.default(:source_ref))
+  end
+
+  @doc false
+  # Updates `field` in `options` with `default` value if current value is `nil` or hasn't been set,
+  # and if `default` is not `nil`.
+  def normalize_options(options, field, default) do
+    if is_nil(options[field]) and not is_nil(default) do
+      Keyword.put(options, field, default)
     else
       options
     end
--- a/lib/mix/tasks/docs.ex
+++ b/lib/mix/tasks/docs.ex
@@ -126,12 +126,15 @@ defmodule Mix.Tasks.Docs do
     options =
       config
       |> get_docs_opts()
+      # CLI options have precedence over config
       |> Keyword.merge(cli_opts)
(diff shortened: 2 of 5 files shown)


josevalim:
Thanks @eksperimental!

> Replace ExDoc.normalize_options/1 with a more general function named ExDoc.normalize_options/3

Why? What are we solving with this? Which issues are we trying to prevent? What do we gain? :)

eksperimental:
I will report back later here, as I have found other issues, and I would like to sort that first and then see what's the best approach here.
Let's hold it off for now until we sort the other issue

eksperimental:
@josevalim, view check the new changes.
after reworking it out for a whileI ended up leaving `normalize_options/1` and adding `normalize_options/3`.

The reason for introducing `normalize_options/3` is because it allows us to modify options directly without the new to create new functions and duplicate code as in https://github.com/elixir-lang/ex_doc/pull/709/files#diff-817174a6c3600e5759b5cf9117b0f39fL164

lib/ex_doc.ex
```diff
@@ -93,20 +93,43 @@ defmodule ExDoc do
...
       source_root: options[:source_root] || File.cwd!,
     }
     struct(preconfig, options)
   end
 
+  @spec normalize_options(keyword) :: keyword
+  def normalize_options(options) do
+    options = normalize_options(options, :source_ref, ExDoc.Config.default(:source_ref))
+
+    options
```
milmazz: What about this option:

```elixir
options
|> normalize_options(:source_ref, ExDoc.Config.default(:source_ref))
|> normalize_directory([:assets, :output, :source_root])
|> normalize_source_url_pattern(options[:source_ref] || ExDoc.Config.default(:source_ref))
```

lib/mix/tasks/docs.ex
```diff
@@ -113,25 +113,35 @@ defmodule Mix.Tasks.Docs do
   def run(args, config \\ Mix.Project.config, generator \\ &ExDoc.generate_docs/3) do
     Mix.Task.run "compile"
 
-    {cli_opts, args, _} = OptionParser.parse(args,
-                            aliases: [n: :canonical, f: :formatter, o: :output],
-                            switches: [canonical: :string, formatter: :keep, language: :string, output: :string])
-
-    if args != [] do
-      Mix.raise "Extraneous arguments on the command line"
+    {cli_options, remaining_args, _invalid} =
```
milmazz: This section does not seems to be related with your original goal, which is add `ExDoc.normalize_options/3`.

lib/mix/tasks/docs.ex
```diff
@@ -113,25 +113,35 @@ defmodule Mix.Tasks.Docs do
...
 
     project = to_string(config[:name] || config[:app])
     version = config[:version] || "dev"
     options =
       config
       |> get_docs_opts()
-      |> Keyword.merge(cli_opts)
-      |> normalize_source_url(config)   # accepted at root level config
-      |> normalize_homepage_url(config) # accepted at root level config
+      # CLI options have presedence over config
```
milmazz: s/presedence/precedence/

lib/mix/tasks/docs.ex
```diff
@@ -113,25 +113,35 @@ defmodule Mix.Tasks.Docs do
...
+      |> Keyword.merge(cli_options)
+      # source_url is accepted at root level config
+      |> (&(if value = config[:source_url],
```
milmazz: At least for me, this *anonymous* function trick brakes the normal *workflow* of the data transformation. I like the way it was before, `normalize_source_url/1` and `normalize_homepage_url/1` are explicit and show their intention without looking the implementation details, if you want to look into the details you can go straight to that function. wdyt?

test/ex_doc/formatter/html_test.exs
```diff
@@ -121,8 +121,9 @@ defmodule ExDoc.Formatter.HTMLTest do
     assert File.regular?("#{output_dir()}/another_dir/CompiledWithDocs.html")
     assert File.regular?("#{output_dir()}/another_dir/RandomError.html")
 
-    assert "#{output_dir()}/another_dir/dist/app-*.css" |> Path.wildcard |> File.regular?
-    assert "#{output_dir()}/another_dir/dist/app-*.js" |> Path.wildcard |> File.regular?
+    for file <- Path.wildcard("#{output_dir()}/another_dir/dist/app-*.{js,css}") do
+      assert File.regular?(file) == true
```
milmazz: You don't need the `== true` at the end, right?

josevalim:
> The reason for introducing normalize_options/3 is because it allows us to modify options directly without the need to create new functions and duplicate code

I don't understand, if this change is about duplicate code, why is it adding more lines of code than removing? Sorry but I really can't see what we are gaining with those changes.

test/ex_doc/formatter/html_test.exs
eksperimental: that's the convention in the Elixir project. I probably got it from there.

lib/mix/tasks/docs.ex
eksperimental: yes. that's correct. I will use those then.

lib/ex_doc.ex
eksperimental: nope. this was done intetionally.
You cannot do that, because in the last line when I access `options[:source_ref]` I need the updated version that `normalize_options($options, :source_ref, ExDoc.Config.default(:source_ref))` updates

eksperimental:
Thank you for the review @milmazz,
I have implement most of your suggestions, please have a look one last time.
I have reworked the helper so it doesn't break the pipe line flow.

@josevalim If you are still not convinced,
please close the PR.

josevalim:
Ok, I will close the PR then. Thank you @eksperimental for the different attempts.

test/ex_doc/formatter/html_test.exs
whatyouhide: I don't think is a convention in Elixir. For example, the tests for `File.regular?/1` (or for `File` in general for that matter) *never* use it.

eksperimental: you are right.
It is used only with similar lines, and the value will be `true` or `false`.

as in
```elixir
    assert ("abcd" =~ ~R/c(d)/) == true
    assert ("abcd" =~ ~R/e/) == false
```

whatyouhide: Even that should be rewritten to 

```elixir
assert "abcd" =~ ~R/.../
refute ...
```

so that ExUnit can potentially produce a diff in case of failure. The only reason to use `== true` (or `false`) as far as I know is when the function can return something else than a boolean and you want to assert that the specific case you're testing returns a boolean (for example `assert foo` would pass if `foo` was anything non-`nil` and non-`false`).

josevalim:
We may check for explicit bools in Elixir because we may be checking that
=~ returns a Boolean. Other than that, there is no reason for a explicit
book check.
-- 

PR #713, opened 2017-05-16
Separate tasks from regular modules

### LINKED ISSUE #712: Generate a separate section for mix task documentation
Today, mix tasks are documented as regular modules. This is fine, but in some cases creates a lot of noise in the documentation. For example this is what we see when we open phoenix docs
screenshot 2017-05-16 17 47 58
Only the tasks!

I propose we document tasks separately in a new "tab" - this way we'd have "Guides", "Modules", "Tasks", "Exceptions" and "Protocols". We could also name the task modules after the task name instead of the module name. This should make the docs much clearer and less cluttered.

michalmuskala:
- [x] separate tasks in html output
- [x] separate tasks in epub output
- [x] html tests
- [x] js tests
- [x] epub tests
- [x] change page title for tasks from `Mix.Tasks.Foo.Bar` to `foo.bar`

Closes #712


michalmuskala:
Here's how it looks in phoenix right now:
screenshot 2017-05-17 07 40 38

test/ex_doc/formatter/html/templates_test.exs
```diff
@@ -130,38 +130,38 @@ defmodule ExDoc.Formatter.HTML.TemplatesTest do
...
-      "Summaries should not be displayed with trailing punctuation"
+  #   assert Templates.synopsis(doc1) ==
+  #     "Summaries should not be displayed with trailing punctuation"
 
-    assert Templates.synopsis(doc2) ==
-      "Example function: Summary should not display trailing puntuation"
-  end
+  #   assert Templates.synopsis(doc2) ==
+  #     "Example function: Summary should not display trailing puntuation"
+  # end
```
michalmuskala: Moving the synopsis generation to the retriever means we can't test it as easily now. 

I'm not 100% sure about this move. Maybe it's fine to use the first line of the moduledoc for tasks as well instead of the shortdoc. It's only used on the "API Reference" page, which is, in my experience, used rather rarely.

whatyouhide:
I would use "Mix task" as the label on the right of the name instead of "task" to be clearer.

josevalim:
@whatyouhide tasks is fine as you can get the context as soon as you click on the first one of them. My only suggestion is to change the title on the right side to be `mix phx.gen.html (task)` instead of `phx.gen.html (task)`. Then I think we are covered from adding mix to the sidebar. What do you think?

michalmuskala:
Should we also change the name in the summary on the left?

test/ex_doc/formatter/html/templates_test.exs
josevalim: Yes, let's revert the synopsis change for now. We can always add the feature if necessary later.

michalmuskala:
I updated it (not pushed yet) to list `mix foo` instead of `foo task` as the title:

screenshot 2017-05-18 11 46 25

josevalim:
@michalmuskala let's keep mix in bold as well, basically as part of the title. Is that possible?

michalmuskala:
Anything's possible if you try hard enough 😏

I'll change it.

michalmuskala:
@josevalim It would look like this:

screenshot 2017-05-18 11 57 18

Or do we want to keep the small "task" there as well?

michalmuskala:
Another question is the page title - should it also include `mix`? And the search page. Currently it looks like this:
screenshot 2017-05-18 12 05 42

josevalim:
> Or do we want to keep the small "task" there as well?

Up to you. It doesn't feel it is necessary though.

> Another question is the page title - should it also include mix? 

Up to you. It doesn't feel it is necessary though.

josevalim:
Let us know when this is ready to ship so we can release a new version. :)

michalmuskala:
This is ready for review

josevalim:

whatyouhide:
\cc @Kapeli, heyo, I think this will affect how documentation is structured in Dash: there's now a new "Mix Tasks" section alongside Protocols, Exceptions, and so on. May be worth mirroring that in Dash.app :)

Kapeli:
Thanks @whatyouhide, I'll make sure Dash works well with this change.

PR #715, opened 2017-05-18
Improve web accessibility support

milmazz:
This commit includes the following:

* [x] Set the primary language of the documents (default: `en`), which allows screen readers to read the content in the appropriate language.
* [x] Fix empty buttons, add a descriptive text for screen readers.
* [x] Fix empty links (e.g "View Source Code"), add a descriptive text for screen readers.
* [x] Add a label in the search input element.
* [x] Introduce a CSS class `.sr-only` to visually hide some text while keeping it accessible to screen readers.
* [x] Fix a markup error, both the starting and ending `span` tags are mandatory.


lib/ex_doc.ex
```diff
@@ -15,6 +15,7 @@ defmodule ExDoc do
       :output => "./doc",
       :source_ref => "master",
       :retriever => ExDoc.Retriever,
+      :language => "en",
```
eksperimental: could we sort this one alphabetically?

lib/mix/tasks/docs.ex
```diff
@@ -112,7 +115,7 @@ defmodule Mix.Tasks.Docs do
 
     {cli_opts, args, _} = OptionParser.parse(args,
                             aliases: [n: :canonical, f: :formatter, o: :output],
-                            switches: [canonical: :string, formatter: :keep, output: :string])
+                            switches: [canonical: :string, formatter: :keep, output: :string, language: :string])
```
eksperimental: could we sort this one alphabetically as well?

josevalim:

PR #718, opened 2017-05-18
Setting homepage_url at root level in the config, as shown in the Con…

eksperimental:
…figuration

example en Mix.Tasks.Docs module didn't take effect.

This can be seen can see that for example in Phoenix
https://github.com/phoenixframework/phoenix/blob/v1.2.4/mix.exs#L22
but the logo in the docs, link to the main page: https://hexdocs.pm/phoenix/Phoenix.html

A test is included that checks for all allowed configs at root level:
app, name, source_url, homepage_url, version


josevalim:

PR #719, opened 2017-05-20
Fix test in Formatter.HTMLTest

eksperimental:


test/ex_doc/formatter/html_test.exs
```diff
@@ -121,8 +121,9 @@ defmodule ExDoc.Formatter.HTMLTest do
     assert File.regular?("#{output_dir()}/another_dir/CompiledWithDocs.html")
     assert File.regular?("#{output_dir()}/another_dir/RandomError.html")
 
-    assert "#{output_dir()}/another_dir/dist/app-*.css" |> Path.wildcard |> File.regular?
-    assert "#{output_dir()}/another_dir/dist/app-*.js" |> Path.wildcard |> File.regular?
+    for file <- Path.wildcard("#{output_dir()}/another_dir/dist/app-*.{js,css}") do
+      assert File.regular?(file)
```
josevalim: Those have different meanings. The first one is going to assert we have only a single version, which is what we have. Your implementation will work for multiple versions, and we don't want that.

eksperimental: I see. the problem was then that there was a left-over JS file.
I have fixed it.

josevalim:
Please rebuild the assets in master and remove all old ones. I think we can
close this PR. :)
-- 

Issue #695, opened 2017-03-29
Merriweather font not loading in Firefox


sheharyarn:
[Opening any HexDocs page](https://hexdocs.pm/elixir/Kernel.html) in Firefox doesn't load the Merriweather font, which causes it to look like this:

<img width="855" alt="screen shot 2017-03-29 at 6 59 36 pm" src="https://cloud.githubusercontent.com/assets/2355108/24480367/f048ee4c-14b1-11e7-98aa-1dd8412e06bb.png">

---

The developer console shows these errors:

<img width="1039" alt="screen shot 2017-03-29 at 7 02 12 pm" src="https://cloud.githubusercontent.com/assets/2355108/24480432/4f526206-14b2-11e7-926d-030ed815d15d.png">

```
downloadable font: Layout: DFLT script doesn't satisfy the spec. DefaultLangSys is NULL (font-family: "Merriweather" style:normal weight:300 stretch:normal src index:2) source: https://fonts.gstatic.com/s/merriweather/v15/ZvcMqxEwPfh2qDWBPxn6nkZRWJQ0UjzR2Uv6RollX_g.woff2  css:106:12
downloadable font: Layout: Failed to parse script table 0 (font-family: "Merriweather" style:normal weight:300 stretch:normal src index:2) source: https://fonts.gstatic.com/s/merriweather/v15/ZvcMqxEwPfh2qDWBPxn6nkZRWJQ0UjzR2Uv6RollX_g.woff2  css:106:12
downloadable font: GSUB: Failed to parse script list table (font-family: "Merriweather" style:normal weight:300 stretch:normal src index:2) source: https://fonts.gstatic.com/s/merriweather/v15/ZvcMqxEwPfh2qDWBPxn6nkZRWJQ0UjzR2Uv6RollX_g.woff2  css:106:12
downloadable font: rejected by sanitizer (font-family: "Merriweather" style:normal weight:300 stretch:normal src index:2) source: https://fonts.gstatic.com/s/merriweather/v15/ZvcMqxEwPfh2qDWBPxn6nkZRWJQ0UjzR2Uv6RollX_g.woff2  css:106:12
[...]

PR #722, opened 2017-05-20
Change font stack, based on availability on different platforms

aphillipo:
The font stack has been chosen carefully to be supported (99%+) on Windows, Mac and Linux assuming Merriweather fails to load.

To some extent (as much as I think is reasonable) closes: #695


josevalim:

PR #723, opened 2017-05-20
Remove left-over JS files (and build new ones)

eksperimental:
Solves the issue mentioned in https://github.com/elixir-lang/ex_doc/pull/719

PR #724, opened 2017-05-20
Add types to the search results

whatyouhide:
I'm not sure this is the correct way but in the HTML It Works™. Tests do not pass but I suspect that's related to #723? Any pointers are appreciated :).


josevalim:

PR #725, opened 2017-05-20
Fix callback signature when arguments are typed as functions

### LINKED ISSUE #690: Wrong variable name in docs for Access.get_and_update/3
See elixir-lang/elixir#5846:

The short signature for [`c:Access.get_and_update/3`](https://hexdocs.pm/elixir/Access.html#c:get_and_update/3) is called `list` [when it should be](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/access.ex#L202) `fun`.

I'm pretty sure that the issue lies in how [this code](https://github.com/elixir-lang/ex_doc/blob/018cc55a436755cf2e6c727b806e4641f97a3352/lib/ex_doc/retriever.ex#L347-L389) is parsing anonymous functions (it doesn't look like it even tries to).

I'm happy to craft a PR for this this week, just wanted to get an issue opened in the right place if I don't get around to it or someone else finds it interesting first.

whatyouhide:
Closes #690. Not sure I should add a test for this since the whole thing of getting the signature from the typespec is only tested "on the side" (for example [here](https://github.com/elixir-lang/ex_doc/blob/527e0602598f7e283fa15d8aad9876b63334f9dd/test/ex_doc/retriever_test.exs#L181)) and it would mean fiddling with the fixtures behaviours.

### DIFF
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -369,6 +369,8 @@ defmodule ExDoc.Retriever do
 
   defp to_var({name, meta, _}, _) when is_atom(name),
     do: {name, meta, nil}
+  defp to_var([{:->, _, _} | _], _),
+    do: {:function, [], nil}
   defp to_var({:<<>>, _, _}, _),
     do: {:binary, [], nil}
   defp to_var({:%{}, _, _}, _),


josevalim:

PR #726, opened 2017-05-20
Use String.trim_trailing/1 over the deprecated String.rstrip/1

whatyouhide:
ExDoc already depends on Elixir ~> 1.3 where `String.trim_trailing/1` was introduced, so there's no backwards compatibility issue.


milmazz:
@whatyouhide Thanks! <3

PR #727, opened 2017-05-21
Camelize config[:app] when used in project name

eksperimental:
So if no :name is provided in config, it set the project name
based on the camelized version of :app

ie. `app: :ex_doc`
will generate a project named "ExDoc"


josevalim:
I would rather have developers explicitly configure it, especially because Hex defaults to use the app name without being camelized, so we keep it consistent with Hex unless user want to change it. Thank you!

eksperimental:
No worries then.

PR #733, opened 2017-05-27
[Windows] Fix unit tests

milmazz:


test/ex_doc/formatter/epub_test.exs
```diff
@@ -130,13 +134,13 @@ defmodule ExDoc.Formatter.EPUBTest do
     config = doc_config([main: "README"])
     generate_docs_and_unzip(config)
 
-    content = File.read!("#{output_dir()}/OEBPS/readme.xhtml")
+    content = "#{output_dir()}/OEBPS/readme.xhtml" |> File.read!() |> normalize_eol()
```
josevalim: Do we need this? None of those seem to be asserting on new lines?

test/ex_doc/formatter/html/templates_test.exs
```diff
@@ -166,14 +170,14 @@ defmodule ExDoc.Formatter.HTML.TemplatesTest do
   ## LISTING
 
   test "site title text links to homepage_url when set" do
-    content = Templates.sidebar_template(doc_config(), @empty_nodes_map)
+    content = doc_config() |> Templates.sidebar_template(@empty_nodes_map) |> normalize_eol()
```
josevalim: We don't need this. We should just remove `\n` from the regexes below because `\s` matches any space and we don't actually care in HTML if it is a new line or a regular space:

    "a\r\nb" =~ ~r/a\s*b/
    true

josevalim:
Hi @milmazz! I think instead of doing the eol conversion, we can adjust the regexes instead. What do you think?

milmazz:
@josevalim Definitely better! Thanks for your suggestions.

Issue #738, opened 2017-06-06
Test helpers in ex_doc output


d4rk5eed:
Hi guys,
I have some documentation in `test/support/graphql_case.ex`.
But can't make `mix docs` to compile documentation for this module.
How to resolve issue?
thanx in advance

josevalim:
We only generate documentation from compiled files in the current environment. Given test/support/graphql_case.ex is compiled only in test, you probably want to do `MIX_ENV=test mix docs`.

Btw, please use the mailing list, stack overflow or elixir forum for questions. More people are available for help there. :)

PR #740, opened 2017-06-07
Updated README.md with docs from test folder

d4rk5eed:
Relates to #738 

Since @josevalim answered 
> We only generate documentation from compiled files in the current environment. Given test/support/graphql_case.ex is compiled only in test, you probably want to do MIX_ENV=test mix docs

It's not obivious to fetch deps for test environment (as guys like me just copy-paste samples of code to project), so some minor README update

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ def deps do
 end
 ```
 
+Consider using `[:dev, :test]` if you wish to fetch docs from modules under `test` folder.
+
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you @d4rk5eed! However, your case is very uncommon, so I would prefer to leave it out of the README as to not confuse those who use it rather the regular way. :heart:

eksperimental:
why don't we create a FAQ file or something of the like?

PR #748, opened 2017-07-02
Fix search history

### LINKED ISSUE #341: [Search] Add history support to the search page
The idea here is to:
1. Change the URL once you search something
2. Make sure the back button works

We don't need to worry about this working with the file protocol.

Ref #332

aphillipo:
Well this was a battle! Do we support browsers that don't have the History API?

https://caniuse.com/#feat=history

Please try to break this obviously, back buttons can be rather awkward to get right!

Fixes #341


aphillipo:
Okay lot's of little things to fix here...

aphillipo:
Okay Ebert should be happy now!

josevalim:
I am not the person to review this one. :) So let's ping @nscyclone, @dignifiedquire, @milmazz and @eksperimental.

PR #754, opened 2017-07-14
Add unencoded link target for functions with special chars

### LINKED ISSUE #509: Anchor for <<>> is not working in Firefox

cybrox:
When merged, this will fix #509

Firefox (and probably other browsers) expect the URI of an anchor to be urlencoded
but the actual target to be NOT encoded. Due to this behaviour, links to functions
with special characters won't work.

While RFC3986 is very clear on how URIs have to be formatted, there is no clear
specification on how link targets actually should (or shouldn't) be encoded.
There are lose guidelines for HTML4/HTML5 but they rarely cover special characters.

Since most browsers seem to have adapted to also match encoded anchor targets,
this commit does not change but only extend the current behaviour by adding a
second, unencoded target to the template, similar to what is already done when
linking to a function with optional parameters, which results in target links
with different arity.

### DIFF
--- a/lib/ex_doc/formatter/html/templates.ex
+++ b/lib/ex_doc/formatter/html/templates.ex
@@ -120,6 +120,10 @@ defmodule ExDoc.Formatter.HTML.Templates do
     |> h()
   end
 
+  defp is_enc?(binary) do
+    h(binary) != enc_h(binary)
+  end
+
   @doc """
   Create a JS object which holds all the items displayed in the sidebar area
   """
--- a/lib/ex_doc/formatter/html/templates/detail_template.eex
+++ b/lib/ex_doc/formatter/html/templates/detail_template.eex
@@ -1,4 +1,7 @@
 <div class="detail" id="<%=enc_h link_id(module_node.id, module_node.type) %>">
+  <%= if is_enc? link_id(module_node.id, module_node.type) do %>
+    <span id="<%=h link_id(module_node.id, module_node.type) %>"></span>
+  <% end %>
   <%= for default <- get_defaults(module_node) do %>
     <span id="<%=enc_h link_id(default, module_node.type) %>"></span>
   <% end %>


josevalim:
Won't this generate duplicate IDs for most cases though? And I am assuming that if we drop the `enc_` in the line above is going to break the behaviour in some browsers, right? Maybe we should test that though?

cybrox:
@josevalim You are right, no idea how I didn't realise that. While it would technically work, it's definitely not a clean solution according to the specs. 

I have not yet found a browser that caused problems with the non-encoded id's using only `h` but there definitely is a chance, that we just invert the problem. So far, I successfully tested it on:
* **Win7** - Internet Explorer 11, Firefox 54.0, Chrome 61.0
* **OSX10.12.5** - Chrome 59.0, Firefox 54.0

The more pragmatic - and probably safer - solution in terms of browser support would probably be to just wrap the unencoded anchor in an `if` with a helper that determines if the encoded string is different from the raw, escaped string. Reasonable?

josevalim:
@cybrox the if approach sounds good to me.

cybrox:
Re-pushed the feature branch with the respective helper :)

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -120,6 +120,11 @@ defmodule ExDoc.Formatter.HTML.Templates do
     |> h()
   end
 
+  @doc false
```
josevalim: I think this function can be private because the template is executed in this module? Or have you ran into issues?

cybrox: It works fine when private and I don't think a similar issue will arise with `epub`s, I'll make it private.

josevalim:

PR #763, opened 2017-08-02
Markdown modules are now 100% responsible for code

tmbb:
Code blocks are now handled by the markdown modules,
and ExDoc doesn't do any post-processing.

The modules bundled with ExDoc were updated to reflect this.

The goal of this change is to make it easier to use alternative Markdown
implementations.

This might break compatibility for people that are already usin Markdown
extensions.

The code for the Hoedown module wasn't changed because it already handled the postprocessing of code blocks itself.

All unit tests pass, and my visual integration test with Earmark on a real project showed no differences.

### DIFF
--- a/lib/ex_doc/markdown/cmark.ex
+++ b/lib/ex_doc/markdown/cmark.ex
@@ -15,6 +15,6 @@ defmodule ExDoc.Markdown.Cmark do
   Generate HTML output. Cmark takes no options.
   """
   def to_html(text, _opts) do
-    Cmark.to_html(text)
+    Cmark.to_html(text)  |> ExDoc.Markdown.pretty_codeblocks
   end
 end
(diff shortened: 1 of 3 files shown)


josevalim:

lib/ex_doc/markdown/cmark.ex
```diff
@@ -15,6 +15,6 @@ defmodule ExDoc.Markdown.Cmark do
   Generate HTML output. Cmark takes no options.
   """
   def to_html(text, _opts) do
-    Cmark.to_html(text)
+    Cmark.to_html(text)  |> ExDoc.Markdown.pretty_codeblocks
```
milmazz: @tmbb This is a nitpick, but it should be: 

```elixir
text |> Cmark.to_html() |> ExDoc.Markdown.pretty_codeblocks()
```

lib/ex_doc/markdown/earmark.ex
```diff
@@ -32,6 +32,6 @@ defmodule ExDoc.Markdown.Earmark do
              file: Keyword.get(opts, :file),
              breaks: Keyword.get(opts, :breaks, false),
              smartypants: Keyword.get(opts, :smartypants, true))
-    Earmark.as_html!(text, options)
+    Earmark.as_html!(text, options) |> ExDoc.Markdown.pretty_codeblocks
```
milmazz: @tmbb Another nitpick:

```elixir
text |> Earmark.as_html!(options) |> ExDoc.Markdown.pretty_codeblocks()
```

PR #766, opened 2017-08-02
Added :markdown_processor option to the mix config

tmbb:
The changes are very simple and they work well both in practice and in ExUnit tests.

The problem is that the only way I can think of testing this is somewhat contrived. The idea is that I created a mock Markdown processor that raises an exception when called, and use the exception as a a way to decide if the right markdown processor was called.

If you think there is a better way of testing this, I can refactor it.

### DIFF
--- a/lib/ex_doc.ex
+++ b/lib/ex_doc.ex
@@ -3,6 +3,8 @@ defmodule ExDoc do
   Elixir Documentation System. ExDoc produces documentation for Elixir projects
   """
 
+  alias ExDoc.Markdown
+
   defmodule Config do
     @moduledoc """
     Configuration structure that holds all the available options for ExDoc
@@ -39,6 +41,7 @@ defmodule ExDoc do
       language: @default.language,
       logo: nil,
       main: nil,
+      markdown_processor: nil,
       output: @default.output,
       project: nil,
       retriever: @default.retriever,
@@ -67,6 +70,7 @@ defmodule ExDoc do
        language: String.t,
        logo: nil | Path.t,
        main: nil | String.t,
+       markdown_processor: nil | atom,
        output: nil | Path.t,
        project: nil | String.t,
        retriever: :atom,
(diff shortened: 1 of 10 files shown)


josevalim:
Thanks @tmbb! I think this change is counter-intuitive because we are using a local option to set a global value - therefore I don't think we should merge this as is. However, the other option is to pass the markdown processor every time we need to convert to markdown in the formatters and that may be a change that is too large and maybe even not worth it.

Thoughts?

tmbb:
> I think this change is counter-intuitive because we are using a local option to set a global value

You're the one who said the app environment was "more local" than the mix, config right? I agree it's a little strange, but it's probably not worth it to pass the markdown processor explicitly.

I thought of that when I started working on this, but ultimately decided against.

tmbb:
On one hand, setting the option from the mix config is just a minor cosmetic change and it's probably not worth it to desctructure perfectly good code because of that.

So we could simply not merge this (any other approach will probably do the same thing as my code).

If we really want this feature, we can pass the paameter around or just decide that one little global variable can't hur that much. In situations like these, the best answer might be to defer to this xkcd strip: https://xkcd.com/292/ :)

Now, seriously, your call. What the code is doing now doesn't shock me. But this feature is just a "nice shiny thing", and nothing fundamental.

josevalim:
Thanks @tmbb, I won't merge it for the reasons above.

Btw, I probably expressed myself incorrectly previously, but the project configuration is more local than the application config. Because the project configuration is about your project and the app config applies to `:ex_doc` application.

tmbb:
> Btw, I probably expressed myself incorrectly previously, but the project configuration is more local than the application config

Ok, then I might have gotten that wrong then. Should I submit a PR that does it the other way around (having the project's config override the app's env)? That is probably even easier and less contrived than what we have now :) Or you don't want this feature at all?

josevalim:
I think changing the order wouldn't remove the fact we would need to rewrite all calls to markdown... so better to leave things as they are IMO. :)

tmbb:
No, I would still set the apps env, just in the "right order". But this is really not a priority, so let's leave it as it is.

tmbb:
Just commenting here in case you want to reopen the issue.

OvermindDL1:
/me is curious...

Why not set it in the mix.exs file.  It's not like it is a runtime configuration after all (well, I guess someone could include ex_doc at runtime, but even then they could set it via a command in that case)...

josevalim:
@OvermindDL1 yes, now i agree. most of the use cases are via the command line, so we shouldn't sacrifice the convenience of doing that via mix.exs because of a very uncommon use case.

tmbb:
> Why not set it in the mix.exs file via a new option there that ex_doc reads

That is exactly what this PR does.

josevalim:
Since this branch is no longer available, I pushed a similar patch to master: 69dd28eb28ce3ff31d1775cd0c0170a575940943

Thank you!

tmbb:
Sorry for that, at the time I deleted it to get a cleaner history for some reason... I didn't think we'd be using it.

josevalim:
@tmbb no problem at all!

PR #768, opened 2017-08-06
before_closing_*_tag works on EPUB too

tmbb:
All tests pass (which means the extra content is actually being included), and he output is good on visual inspection on my reader. If this is merged, we should make a release so that it can be used for hex packages.


josevalim:

Issue #773, opened 2017-09-10
Allow configurable filename


Gazler:
Currently the filename is defined in https://github.com/elixir-lang/ex_doc/blob/e47a52572883ded94af3af979178ba506dcdcb14/lib/ex_doc/formatter/epub.ex#L54

To prevent having to keep this in sync with any documentation that links to the file, it is desirable to not have the version in the filename.

Please see https://github.com/phoenixframework/phoenix/pull/2481

If we want this to be configurable, what do you suggest as a config option? `filename`, `epub_filename`, `output_filename`? If we do `output_filename` then it could work for other formats in the future e.g .`output_filename: "phoenix"` could generate:

 * phoenix.epub
 * phoenix.mobi
 * phoenix.pdf
 * phoenix.zip

josevalim:
`epub_filename` sounds great to me!

PR #775, opened 2017-09-10
Remove version name from epub formatter

Gazler:
Previously, the epub formatter defaulted to a filename in the following format:

> Phoenix-v1.3.0.epub

This commit changes it to:

> Phoenix.epub

When published to hexdocs, the version will already be specified in the url.
This means if the epub file is linked to in the docs, it will not be updated
each time there is a new version.


Gazler:
This closes https://github.com/elixir-lang/ex_doc/issues/773

josevalim:

PR #776, opened 2017-09-12
Update dependencies (earmark, excoveralls)

dkarter:
Cleans up deprecation warnings during dependency compilation.


dkarter:
Not sure why this is failing on firefox tests.

When I run this on my machine everything is green and gulp only runs Chrome.

image

Either way I don't think my changes should have affected it so there must be something wrong with the build on TravisCI.

mix.exs
```diff
@@ -21,7 +21,7 @@ defmodule ExDoc.Mixfile do
   end
 
   defp deps do
-    [{:earmark, "~> 1.1"},
+    [{:earmark, "~> 1.2"},
```
josevalim: Let's not bump the requirement unless we have to because people may still rely on earlier earmark versions if using earlier elixir versions. What do you think?

dkarter: Couldn't they use the previous version of ex_doc in this case?

josevalim:
They could but unless we have a reason to bump this, we should not. We can
avoid the warnings by updating the lock file, so this change seems
unnecessary.
-- 


josevalim:
Why did you close it? :D Sorry if i was not clear but the lock file updates are welcome, we just don't need the mix.exs ones. :)

josevalim:
@dkarter I pushed an update to your branch, let me know what you think.

dkarter:
@josevalim OK that works. Thanks!

josevalim:
Thank *you*! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

Issue #779, opened 2017-09-26
Group option for modules


LostKobrakai:
I've just tried the group option for markdown docs, like it's used for the new phoenix docs. I'd find it useful to have such a grouping possibility for modules as well. With phoenix contexts the number of modules does quickly become quite big and I'd like to document not only the public interface, but also how the internals work. Grouping the modules by the context name would solve that issue quite simply and it seems the necessary frontend stuff could be reused.

josevalim:
I agree. Can you please submit a PR?
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

LostKobrakai:
I could, but I'm not sure what would be the best way to determine the context group. I've thought of the following options:

1. Custom module attribute  
Would be tedious work to add to an already big codebase. Also I'm not sure if custom module attributes are really desirable in general.
2. Just take the second module namespace level  
Like `MyApp.Namespace.Module` => `Namespace`. But this would require `MyApp.Namespace` to  be handled differently if there are deeper nested modules or not. Also this is quite the preconception about the users codebase.
3. Some kind of mapping in the mix.exs  
I'm not sure what would be the most effortless option to assign groups to modules, which does not break as soon as a developer is adding a new module to the project. Maybe something like this `[MyApp.Namespace, MyApp.AnotherNamespace]`, which would group all the modules of that name and nested ones. Maybe with the additional boolean option of "include top level module in groups".

After writing this out I feel like 3. would probably be the best way.

josevalim:
Yup, definitely 3. For pages those are specified alongside each page:

```
      "introduction/learning.md": [group: "Introduction"],
      "introduction/community.md": [group: "Introduction"],

      "up_and_running.md": [group: "Guides"],
      "adding_pages.md": [group: "Guides"],
```

But I am not sure this approach would work for modules exactly because of what you said. We can add a `groups: ` option that accepts a list of regexes and module names:

  groups: ["Super important": ~r/^MyApp.Foo.?/, "Another group": [MyApp.Bar, MyApp.Baz]]

I am not sure I like the inconsistency with how pages work though but given their different contexts, I guess it is a reasonable choice to make?

/cc @Gazler @jeregrine @eksperimental @milmazz

LostKobrakai:
[...]

PR #780, opened 2017-09-27
Add grouping of modules

LostKobrakai:
Regarding #779.

By now I've just added grouping for modules. I can add the changes for the grouping of pages at a later time. Also I'm not sure where/how to test the changes to `ExDoc.Formatter.HTML.Templates`.

Also: 
Damn is it nice to use it even just for the following groups in phoenix projects.
```
[
          "Web - Controllers": ~R/^MyAppWeb\..*?Controller$/,
          "Web - Views": ~R/^MyAppWeb\..*?View$/,
          "Web - Plugs": ~R/^MyAppWeb\..*?Plug$/,
          "Web - Phoenix": ~R/^MyAppWeb.?/
]
```


lib/ex_doc/retriever.ex
```diff
@@ -115,7 +117,7 @@ defmodule ExDoc.Retriever do
     modules
     |> Enum.map(&get_module(&1, config))
     |> Enum.filter(&(&1))
-    |> Enum.sort(&(&1.id <= &2.id))
+    |> Enum.sort_by(&({&1.group, &1.id}))
```
josevalim: I think we should order them by the group index in `module_groups`. So the first group will be first, etc. Thoughts?

lib/ex_doc/retriever.ex
```diff
@@ -458,6 +464,25 @@ defmodule ExDoc.Retriever do
...
 
+  defp module_group(module, group_patterns) do
+    Enum.find_value(group_patterns, fn {group, patterns} ->
+      patterns = List.wrap patterns
+      module_in_patterns(module, patterns) && Atom.to_string(group)
+    end)
+  end
+
+  defp module_in_patterns(module, patterns) do
+    "Elixir." <> module_string = Atom.to_string module
```
josevalim: Do not assume the module name is "Elixir." because of Erlang modules. You should probably use the module id or title.

lib/ex_doc/retriever.ex
```diff
@@ -458,6 +464,25 @@ defmodule ExDoc.Retriever do
...
+
+    Enum.any?(patterns, fn pattern ->
+      case pattern do
+        %Regex{} = regex -> Regex.match?(regex, module_string)
+        string when is_binary(string) -> module_string === string
+        atom -> atom === module
```
josevalim: Please use `==` and above. :)

josevalim:
@LostKobrakai this looks really nice, I have added some comments. Regarding the templates tests, a unit test will suffice for now, just set the group manually.

lib/ex_doc/retriever.ex
LostKobrakai: By now order was alphabetically by module name, so I feel like having groups sorted alphabetically isn't to bad. Also if you look at the example above in the PR description I've used a less strict regex as last group to catch all phoenix related modules, which weren't caught by the more precise earlier ones. So the order of groups might be useful for "catch-all" cases more so than for sorting order of the groups.

lib/ex_doc/retriever.ex
LostKobrakai: I can certainly do this. I'm just curious, how would/should one decide between `==` and `===`?

lib/ex_doc/retriever.ex
LostKobrakai: But looking at the way groups worked with pages before, they are ordered by the order of first occurrence of a group, which might be a needed feature. Especially if we make the setup of both the same and deprecate the old way of grouping pages we should look out for that.

josevalim: > But looking at the way groups worked with pages before, they are ordered by the order of first occurrence of a group, which might be a needed feature.

Yup, precisely. 

Plus if we give them control, they can keep it alphabetically. :)

LostKobrakai: I'd really like to retain the user-defined "order of matching" as well, because otherwise the regex patterns might have to be more complex in places. Anything like "group all the remaining modules starting with MyAppWeb" won't be easy if the order in the list does determine the frontend order instead of the matching order. Having a regex exclude the matches of possibly multiple regexes can be quite a pain.

But maybe that's a topic for a later enhancement to the implementation.

josevalim:

PR #789, opened 2017-10-12
Make the build with yarn work

josevalim:


josevalim:
I give up. It was working, then it suddenly doesn't. Maybe it suddenly will again. 😣

package.json
```diff
@@ -7,9 +7,9 @@
     "test": "test"
   },
   "scripts": {
-    "lint": "node_modules/.bin/gulp lint",
-    "test": "node_modules/.bin/gulp test",
-    "build": "node_modules/.bin/gulp build --type production"
+    "lint": "gulp lint",
```
milmazz: @josevalim I think that the problem is that `gulp` is not installed globally. Please notice that is listed as one of the `devDependencies`.

Actually you don't need this section at all, you can do the following:

```console
$ yarn run gulp lint # this will call node_modules/.bin/gulp lint
$ yarn run gulp test # this will call node_modules/.bin/gulp test
```

If the `yarn` version is pre-1.0 you can do the following:

```console
$ yarn run gulp -- lint
$ yarn run gulp -- test
```

HTH

josevalim: I tried that as well and it didn't work. :(

milmazz: Ouch!

At least in my local copy is working...

```console
ex_doc [master ≡ +0 ~1 -0 !]> yarn run gulp lint
yarn run v1.2.1
$ ex_doc\node_modules\.bin\gulp lint
[11:05:04] Using gulpfile ~\ex_doc\gulpfile.js
[11:05:04] Starting 'lint'...
[11:05:09] 'lint' errored after 4.69 s
[11:05:09] ESLintError in plugin 'gulp-eslint'
Message:
    Extra semicolon.
Details:
    fileName: ex_doc\assets\js\events.js
    lineNumber: 21
[11:05:09]
ex_doc\assets\js\events.js
  21:2  error  Extra semicolon  semi

✖ 1 problem (1 error, 0 warnings)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

And this is Windows btw, which is not too friendly with the command line :S

Gazler:
Since this isn't a NodeJS project, don't we need to run `yarn` somewhere in the script to fetch the dependencies? Adding `yarn: true` to the cache only matters for the caching part.

Gazler:
The answer is yes. Yes you do. https://github.com/elixir-lang/ex_doc/pull/790

PR #794, opened 2017-10-28
Fix .main height to 100%

fklement:
When the page content does not fill the full height the sidebar closing/opening animation looks odd.  

**Before:**  
bevor

**After:**
after


josevalim:

Issue #796, opened 2017-11-09
Link underline invisible when using zoom < 100%


LostKobrakai:
I'm running my browser at 90% zoom because retina does allow for that bit of more screen space, but this does prevent the link underline of ex_doc to show, because scaling border lines doesn't work very well. I'm used to some miss-alignments or missing borders, but not showing an underline for a link is not really nice. Maybe this could be switched to using `text-decorcation: underline` and using the new `text-decoration-skip: ink` property to get the effect currently faked with shadows.

<img width="618" alt="screenshot 2017-11-09 10 23 27" src="https://user-images.githubusercontent.com/2489835/32597941-9647d222-c538-11e7-9015-186fa14c6416.png">
<img width="592" alt="screenshot 2017-11-09 10 23 31" src="https://user-images.githubusercontent.com/2489835/32597942-96662a6a-c538-11e7-9baa-a22094074926.png">

PR #797, opened 2017-11-09
Replace css shadow border link underline hack with text-decoration-skip: ink

LostKobrakai:
Regarding https://github.com/elixir-lang/ex_doc/issues/796


josevalim:
Thanks @LostKobrakai . There are two new files in this commit that i think should not be here? epub.css and app.css? Did you build assets with the --type production flag?

LostKobrakai:
I did a `gulp --type production` by accident, maybe it's those.

LostKobrakai:
I've updated the branch without those two files

josevalim:

PR #798, opened 2017-11-21
Make night mode switch link more obvious

### LINKED ISSUE #558: Make night mode switch link more obvious
Night mode is awesome. As discussed in #491 the night mode switch ended up right at the bottom. 

However I had no idea it was there though, as I'd never noticed the link. The positioning is usually not obvious because most module docs are quite long. This means a lot of scrolling to get to the bottom of the docs to find the toggle eg. http://elixir-lang.org/docs/stable/elixir/String.html

It'd be nice to move this to the top to make this more obvious, and show that it's available (I only found out about it when reading through ex_doc PRs).

nscyclone:
This PR introduces another icon just below the menu hamburger to switch between the `day` and `night` modes. Closes #558. 

<details>
<summary>
Example:
</summary>

night-mode

</details>


josevalim:
Thank you! Is it possible to have the icon in the same color as the hamburger icon?

/cc @dignifiedquire

nscyclone:
Thanks, @josevalim! I'm afraid we can't switch the color of the emoji versions, but we can use the text ones instead. There is an [example pen](https://codepen.io/anon/pen/PORGRb), we could make use of the two glyphs on the left.

Let's wait for a review from @dignifiedquire.

aphillipo:
I think the icons would be better as:

light -> 🌚 dark -> 🌝 

I hate the sun emoji and crescent personally and find them a bit garish and out of place.  You might find the same about these of course!

josevalim:
Should we use something from the icon font? Depending on emoji means we have no control over its exhibition.

OvermindDL1:
> Should we use something from the icon font?

If an icon font is available then that should be used, otherwise I'd just use inline svg (could even have some css animate the transition between both forms).  Simple CSS needed, no javascript at all needed, and only static svg description in the html (the 'shape' is defined in CSS).

josevalim:
AFAIK, we already have an icon font.

nscyclone:
Yes, we [have](https://github.com/elixir-lang/ex_doc/tree/master/formatters/html/fonts) an icon font. It can easily be imported back into [icomoon app](https://icomoon.io/app) to append a couple more icons from free packs.

<details>
  <summary>It could look something like this:</summary>

updated-font

</details>



The [original issue](https://github.com/elixir-lang/ex_doc/issues/558) used a Twitter poll to get some feedback. Maybe we should choose a couple of options and make a new poll to let community decide?

josevalim:
The icon ones look a thousand times better to me so I would go ahead with them unless someone strongly opposes. :)

dignifiedquire:
agree with @josevalim thank you :)

nscyclone:
Huge thanks to everyone for all of your input. I've just updated the branch with icons from a [recent comment](https://github.com/elixir-lang/ex_doc/pull/798#issuecomment-346262622).

<details>
<summary>
The updated version now looks like this:
</summary>

night-mode-with-icons

</details>
Looking forward to receive some feedback on this.

josevalim:
@nscyclone beautiful. One final question: should we remove the link at the bottom of the pages now?

nscyclone:
@josevalim I've already [removed](https://github.com/elixir-lang/ex_doc/pull/798/commits/c9a336edde469a42591b3f79b3cba3d8a28709fc#diff-53e876d0bcd6afcd43e5b975a2e9c97d) it.

josevalim:

Issue #682, opened 2017-03-03
Autolink should deal with built-in types differently


eksperimental:
Let's say I want to link to `t:char_list/0` type, ExDoc should remove the "t:".

Currently, since  `t:char_list/0` cannot be linked to anyhting, then it doesn't do anything to what's in backticks.

This issue came up in: https://github.com/elixir-lang/elixir/pull/5808#discussion_r103186125

PR #802, opened 2017-12-04
Autolink basic and built-in types in typespecs

wojtekmach:
Ref #682 - however here we only handle typespecs in this PR
Ref [elixir-lang-core: Anchor built-in types](https://groups.google.com/d/msg/elixir-lang-core/HgOkmWK--OY/Eb2rrvQ1BQAJ) - however we just link to a section of that page and not to particular item

Example:

Before:

zrzut ekranu 2017-12-04 o 01 17 55

After:

zrzut ekranu 2017-12-04 o 01 00 55

Note: Pretty sure the different indentation is because the "After" version has more `_pX_` placeholders and so it's shorter when formatting.

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -10,6 +10,58 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   @basic_types_page "typespecs.html#basic-types"
   @built_in_types_page "typespecs.html#built-in-types"
 
+  @basic_types [
+    any: 0,
+    none: 0,
+    atom: 0,
+    map: 0,
+    pid: 0,
+    port: 0,
+    reference: 0,
+    struct: 0,
+    tuple: 0,
+    integer: 0,
+    float: 0,
+    neg_integer: 0,
+    non_neg_integer: 0,
+    pos_integer: 0,
+    list: 1,
+    nonempty_list: 1,
+    improper_list: 2,
+    maybe_improper_list: 2,
+  ]
+
+  @built_in_types [
+    term: 0,
+    arity: 0,
+    as_boolean: 1,
+    binary: 0,
+    bitstring: 0,
+    boolean: 0,
+    byte: 0,
+    char: 0,
+    charlist: 0,
+    nonempty_charlist: 0,
+    fun: 0,
+    function: 0,
+    identifier: 0,
+    iodata: 0,
+    iolist: 0,
+    keyword: 0,
+    keyword: 1,
+    list: 0,
+    nonempty_list: 0,
+    maybe_improper_list: 0,
+    nonempty_maybe_improper_list: 0,
+    mfa: 0,
+    module: 0,
+    no_return: 0,
+    node: 0,
+    number: 0,
+    struct: 0,
+    timeout: 0
+  ]
+
   @doc """
   Receives a list of module nodes and autolink all docs and typespecs.
   """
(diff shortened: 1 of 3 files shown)


lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -166,26 +224,34 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
-            {form, placeholders}
+          cond do
+            {name, arity} in typespecs ->
+              n = enc_h("#{name}")
+              url = "#t:#{n}/#{arity}"
+              string = format_typespec_form(form, url)
+              put_placeholder(form, string, placeholders)
+
+            {name, arity} in @basic_types ->
+              url = @elixir_docs <> "elixir/" <> @basic_types_page
```
wojtekmach: TODO: this is not gonna work for hexdocs.pm/elixir/master, v1.5.0, need to find better way of handling this

wojtekmach: FWIW e.g. EEx _master_ [1] links to Elixir _stable_, so it seems solution above is consistent with that. Then, we'd only need to detect whether we build docs for elixir app itself in which case this would be a relative: `url = @basic_types_page`

WDYT?

https://hexdocs.pm/eex/master/EEx.html#compile_file/2

josevalim: Yeah, I think we need to pipe it though the same logic we link functions today. Then when we change one, everything else works.

josevalim:
> Note: Pretty sure the different indentation is because the "After" version has more _pX_ placeholders and so it's shorter when formatting.

Should we pad the placeholder by the amount of letters in which word? So "foobar" should become "_ppp1_"?

Something like:

    type_size = type |> Atom.to_string |> byte_size()
    int_size = counter |> Integer.to_string |> byte_size()
    String.duplicate("p", max(type_size - int_size - 2, 1))

?

wojtekmach:
ppp1 sounds good!

lib/ex_doc/formatter/html/autolink.ex
wojtekmach: @josevalim here's another idea: what if we "define" basic & built-in types in the Kernel module?

zrzut ekranu 2017-12-04 o 11 30 30

Another advantage is it magically works from IEx:

zrzut ekranu 2017-12-04 o 11 20 33

To make it work, I had to comment out some code in `Kernel.Typespec`:

```elixir
# if builtin_type?(name, arity) do
#   compile_error(caller, "type #{name}/#{arity} is a builtin type and it cannot be redefined")
# end
```

There's similar code in Erlang too, and so I can't yet do:

```elixir
@type term() :: any()
```
One way to solve it is to write it like this instead:

```elixir
@builtin_type type ::
  any()
  | none()
  | ...
  | builtin_type()

@builtin_type builtin_type ::
  term()
  | arity()
  | ...

@basic_type any()

@builtin_type term() :: any()
```

The idea would be that `@basic_type` doesn't actually define the type but makes it accessible by tooling.

Not sure if we could represent https://hexdocs.pm/elixir/typespecs.html#literals section here though.

wojtekmach: Oh, one other caveat with this: not sure if it'd be technically correct to have types on the Kernel because perhaps they're considered part of the runtime rather.

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -166,26 +222,36 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
+
+            {name, arity} in @basic_types ->
+              source = get_source(Kernel, [], lib_dirs)
+              url = "#{source}#{enc_h(inspect Kernel)}.html#t:basic/0"
+              string = format_typespec_form(form, url)
+              put_placeholder(form, string, placeholders)
+
+            {name, arity} in @builtin_types ->
+              source = get_source(Kernel, [], lib_dirs)
+              url = "#{source}#{enc_h(inspect Kernel)}.html#t:builtin/0"
```
josevalim: Why `enc_h(inspect Kernel)`? Don't we know before hand what is the output of this operation?

Also wouldn't be better to consistently link to the typespecs page?

wojtekmach:
@josevalim a follow up on the experiment that defines types in the Kernel:

if you pull this ex_doc branch, and apply https://github.com/wojtekmach/elixir/commit/394259aa8784fe8aeb8ebc68fd59b737057658de you'll see that way we could drop typespec page, and document types in the Kernel.

On that branch I also included `doc/elixir` directory if you'd like to see this locally.

There are a few caveats, like the literal type needs to be defined in typedoc, and Elixir throws some warnings about useless types. But yeah, if this is the direction woth pursuing let me know and I keep at it, otherwise I can go back to what I had initially in this PR to get it working with typespecs page, and consider moving this to Kernel in a PR to Elixir.

wojtekmach:
@josevalim updated!

Regarding moving basic types to Kernel, while there might be some merit to what I proposed, there's still too much other considerations (e.g. we don't want to expose literal type for others) so I'm parking it for now.

josevalim:

PR #807, opened 2017-12-16
Autolink basic/built-in types and special forms in docs

### LINKED ISSUE #679: Automatically link to Special Forms
Linking to a special form function such as `alias/2`, does not autolink.
I guess from a user perspective, it is not necessary to know whether it's a Kernel function, macro, or a special form. ExDoc should automatically determine this and link to it.

WDYGT?

### LINKED ISSUE #682: Autolink should deal with built-in types differently

wojtekmach:
Fixes #679
Fixes #682

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -62,6 +62,10 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     timeout: 0
   ]
 
+  @basic_type_strings (for {f, a} <- @basic_types, do: "t:#{f}/#{a}")
+  @built_in_type_strings (for {f, a} <- @built_in_types, do: "t:#{f}/#{a}")
+  @special_form_strings (for {f, a} <- Kernel.SpecialForms.__info__(:macros), do: "#{f}/#{a}")
+
   @doc """
   Receives a list of module nodes and autolink all docs and typespecs.
   """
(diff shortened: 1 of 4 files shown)


lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -339,15 +343,29 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   or trailing `]`, e.g. `[my link link/1 is here](url)`, the fun/arity
   will get translated to the new href of the function.
   """
-  def local_doc(bin, locals) when is_binary(bin) do
+  def local_doc(bin, locals, extension \\ ".html", lib_dirs \\ elixir_lib_dirs()) when is_binary(bin) do
```
wojtekmach: Kernel.SpecialForms calls are neither local as in `local_doc` nor remote as in `elixir_functions` calls, so having this logic doesn't quite fit either.

Another approach is to combine `local_doc` and `elixir_functions` into one function, see: https://github.com/elixir-lang/ex_doc/blob/wm-autolink-bak/lib/ex_doc/formatter/html/autolink.ex#L347

josevalim:

eksperimental:
Thank you @wojtekmach for taking care of many of the issues I have opened up. I wish i would have been able to test them, review them and give some feedback. It looks pretty awesome

PR #811, opened 2017-12-17
Annotate special forms

wojtekmach:
Ref: https://elixirforum.com/t/case-do-is-not-a-usual-macro/10894/4


josevalim:
I have answered there. But in a nutshell, special forms are macros. :)

wojtekmach:
That's what I thought, but I conveniently forgot to leave a question on this PR :D. Thanks!

PR #813, opened 2017-12-18
Allow custom links in Autolink.elixir_functions

### LINKED ISSUE #622: Link to functions with custom text
I'm going thoroughly through the Elixir documentation, and I have come across quiet a common pattern, which is wanting to link to a function/callback without showing the full Module.function/arity thing. (most of the time trying to hide the module)... in this case because it's either Kernel or Kernel.SpecialForms and we don't need to distract the user with this information.

A good example is the Guards page.

> the following “type-check” functions (all documented in the Kernel module):
> - is_atom/1
> - is_binary/1
> - is_bitstring/1
> - is_boolean/1

you want to link to `Kernel.is_atom/1`, but you don't want to show `Kernel.`

My proposal is to support:

```
[custom](`Module.functions_name/arity`)
```

and 

```
[`function_name/arity`](`Module.functions_name/arity`)
```

and/or (not so sure about this last one)

```
`[function_name/arity](Module.functions_name/arity)`
```

what do you guys think?

wojtekmach:
Fixes #622

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -421,11 +421,38 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   will get translated to the new href of the function.
   """
   def elixir_functions(bin, project_funs, extension \\ ".html", lib_dirs \\ elixir_lib_dirs()) when is_binary(bin) do
+    bin
+    |> replace_custom_links(project_funs, extension, lib_dirs)
+    |> replace_normal_links(project_funs, extension, lib_dirs)
+  end
+
+  def replace_custom_links(bin, project_funs, extension, lib_dirs) when is_binary(bin) do
     module_re = Regex.source(~r{(([A-Z][A-Za-z_\d]+)\.)+})
     fun_re = Regex.source(~r{([ct]:)?(#{module_re}([a-z_]+[A-Za-z_\d]*[\\?\\!]?|[\{\}=&\\|\\.<>~*^@\\+\\%\\!-]+)/\d+)})
-    regex = ~r{(?<!\[)`\s*(#{fun_re})\s*`(?!\])}
+    custom_re = ~r{\[(.*)\]\(`(#{fun_re})`\)}
 
-    Regex.replace(regex, bin, fn all, match ->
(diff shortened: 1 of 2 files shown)


lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -421,19 +421,28 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
-    fun_re = Regex.source(~r{([ct]:)?(#{module_re}([a-z_]+[A-Za-z_\d]*[\\?\\!]?|[\{\}=&\\|\\.<>~*^@\\+\\%\\!-]+)/\d+)})
-    regex = ~r{(?<!\[)`\s*(#{fun_re})\s*`(?!\])}
-
-    Regex.replace(regex, bin, fn all, match ->
-      {prefix, module, function, arity} = split_function(match)
+    module_re = ~r{(?|(?|[A-Z][A-Za-z_\d]+)\.)*}.source
+    fun_re = ~r{([ct]:)?(#{module_re}([a-z_]+[A-Za-z_\d]*[\\?\\!]?|[\{\}=&\\|\\.<>~*^@\\+\\%\\!-]+)/\d+)}.source
+    normal_link_re = ~r{(?<!\[)`\s*(#{fun_re})\s*`(?!\])}.source
+    custom_link_re = ~r{(?|\[(.*)\])\(`(#{fun_re})`\)}.source
+    re = ~r{#{custom_link_re}|#{normal_link_re}}
```
wojtekmach: @josevalim instead of making this regex more complicated, we could process custom links in a separate pass, WDYT?

I put together a fresh project to play with this stuff, see how that 2 pass thing could look like: https://github.com/wojtekmach/autolink/blob/master/lib/autolink.ex#L91:L95 In that snippet I'm also unifying `local_doc` and `elixir_functions` and I'm still exploring whether that's a good idea, so let's ignore that for now :)

josevalim: :+1: for two passes!

wojtekmach:
@josevalim ready! we could allow custom links in other places too, e.g. for local calls, modules, erlang functions and whatnot, but maybe let's wait for use cases :)

PR #815, opened 2017-12-21
Copy callback typespecs

### LINKED ISSUE #806: Improve documentation of Callbacks
> Right now when I generate the documentation of my project I noticed that there is no typespec on my functions in the generated documentation.
I am not sure if this belongs to here, which probably means we have to add the typespec inside the macro?! or should belong to ex_doc where they should detect is a callback and add the typespec to the documentation, either way I am expecting to have the typespec on my repo module.

Related to https://github.com/elixir-ecto/ecto/issues/2349

wojtekmach:
Fixes #806

### DIFF
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -239,18 +239,9 @@ defmodule ExDoc.Retriever do
     function = actual_def(name, arity, type)
     line = find_actual_line(module_info.abst_code, function, :function) || doc_line
 
-    doc = docstring(doc, name, arity, Map.fetch(module_info.impls, {name, arity}))
-
-    specs = module_info.specs
-            |> Map.get(function, [])
-            |> Enum.map(&Typespec.spec_to_ast(name, &1))
-
-    specs =
-      if type == :defmacro do
-        Enum.map(specs, &remove_first_macro_arg/1)
-      else
-        specs
-      end
+    impl = Map.fetch(module_info.impls, {name, arity})
+    doc = docstring(doc, name, arity, impl)
+    specs = get_specs(module_info, type, function, name, arity, impl)
 
     annotations =
       case {type, name, arity} do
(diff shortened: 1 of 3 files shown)


lib/ex_doc/retriever.ex
```diff
@@ -297,6 +288,31 @@ defmodule ExDoc.Retriever do
...
+        Enum.map(specs, &remove_first_macro_arg/1)
+      else
+        specs
+      end
+
+    copy_callback_specs(specs, name, arity, impl)
+  end
+
+  defp copy_callback_specs([], name, arity, {:ok, behaviour}) do
+    callbacks = Kernel.Typespec.beam_callbacks(behaviour)
```
wojtekmach: this is private API :( any other options?

wojtekmach:
there's one thing left which is a bit subtle: If the callback spec uses _local_ type defined in behaviour module and we copy the spec over verbatim then that local type won't autolink - we'd need to update AST from local type to remote. ugh 😅

josevalim:
@wojtekmach I have just realized that one of the issues with this approach is that local types defined in the behaviour won't be available in the implementation. In the best scenario they will be out of context (non-existing) but in the worst scenario we will have conflicting versions.

I am starting to think that maybe our best alternative here is to make sure we link for the callback and that then they resort to the callback for more information.

wojtekmach:
@josevalim yes, exactly, see [message above](https://github.com/elixir-lang/ex_doc/pull/815#issuecomment-353457226) :D

Again, I think one way would be to manually rewrite AST of the typespec, so:

```elixir
defmodule Behaviour do
  @type t() :: term()
  @callback foo(t()) :: t()
end
```

would be "copied" as:

```elixir
foo(Behaviour.t()) :: Behaviour.t()
```

---

However, I just realised that private types are problematic:

```elixir
defmodule Behaviour do
  @typep p() :: term()
  @callback foo(p()) :: p()
end
```

I don't think we want to "copy" it as:

```elixir
foo(Behaviour.p()) :: Behaviour.p()
```

as we don't  expose private types in neither ExDoc nor IEx (https://github.com/elixir-lang/elixir/issues/2612)

wojtekmach:
> I am starting to think that maybe our best alternative here is to make sure we link for the callback and that then they resort to the callback for more information.

Yeah, with the caveats above I think it's ok to drop this, and instead relying on linking to the callback module. Which we're already doing with https://github.com/elixir-lang/ex_doc/pull/641 - do you think it's enough as is, or do you have some ideas to extend it?

josevalim:
#641 seems enough to me. :) So I think we can close this?

PR #821, opened 2018-01-11
Fix incorrect moduledoc for :extras

unnawut:
Sorry for the consecutive PR. Just found another spot where the example in the `@moduledoc` is not working. The `CONTRIBUTING.md` part below is not a valid keyword.

```
["README.md", "CONTRIBUTING.md": [filename: "contributing", title: "Contributing"]]
```

This PR changes it to:

```
["README.md", {"CONTRIBUTING.md", [filename: "contributing", title: "Contributing"]}]
```

I made sure to finish my work with ex_doc before submitting this PR. So this should be the last one for now.

### DIFF
--- a/lib/mix/tasks/docs.ex
+++ b/lib/mix/tasks/docs.ex
@@ -79,7 +79,7 @@ defmodule Mix.Tasks.Docs do
     * `:extras` - List of keywords, each key must indicate the path to additional
       Markdown pages, the value for each keyword (optional) gives you more control
       about the PATH and the title of the output files; default: `[]`. Example:
-      `["README.md", "CONTRIBUTING.md": [filename: "contributing", title: "Contributing"]]`
+      `["README.md", {"CONTRIBUTING.md", [filename: "contributing", title: "Contributing"]}]`
 
     * `:filter_prefix` - Include only modules that match the given prefix in
       the generated documentation. Example: "MyApp.Core"


wojtekmach:
`["a.md": []]` is valid syntax, it evaluates to  `[{:"a.md", []}]`. I tested above configuration on local project using ex_doc and it worked as expected - did you run into any issues?
```

**Comment by unnawut** [comment]
According to the existing doc, my config would look like this:

```ex
extras: [
  "README.md": [filename: "introduction", title: "Introduction"],
  "docs/document.md"
],
```

And yes it is returning a syntax error:

```ex
** (SyntaxError) mix.exs:15: syntax error before: "docs/document.md"
   (elixir) lib/code.ex:334: Code.load_file/2
   (mix) lib/mix/cli.ex:26: Mix.CLI.proceed/1
   (elixir) lib/code.ex:376: Code.require_file/2
```

Writing this way works fine:

```ex
extras: [
  {"README.md", [filename: "introduction", title: "Introduction"]},
  "docs/document.md"
],
```

Actually... I tried swapping the two extras I have, now this works:

```
extras: [
  "docs/document.md",
  "README.md": [filename: "introduction", title: "Introduction"]
],
```

Maybe I'm not familiar with the keyword syntax enough, but the [Keyword](https://hexdocs.pm/elixir/Keyword.html) docs state that "A keyword is a list of two-element tuples where the first element of the tuple is an atom and the second element can be any value". This config allows mainly a list of strings and/or tuples with `{"string", []}`, so I'm not sure if the keyword syntax can strictly apply here?

josevalim:
Similar to function calls, you can use the keyword syntax only at the end of the data structure. :) So the example is correct. Thanks for the PR in any case!

unnawut:
This syntax caveat isn't documented in [Keyword](https://hexdocs.pm/elixir/Keyword.html). Are you keen to have a PR to the documentation there?

josevalim:
Good catch. I will change it accordingly! thanks!

PR #827, opened 2018-02-05
Update jQuery

lexmag:
Does not seem like we use any features that were removed or deprecated (https://jquery.com/upgrade-guide/3.0/).
Also I've build Elixir applications and manually clicked on pages—works as before.


josevalim:

PR #838, opened 2018-03-18
Add plugins support to Earmark processor

thecodeboss:
Wanted to add a [task list](https://help.github.com/articles/about-task-lists/) to the docs in a personal project, but found that Earmark does not support them. I want to write a custom plugin to do it, though it doesn't seem to be possible to enable the plugin through ex_doc. This PR should fix that.


josevalim:

PR #839, opened 2018-03-20
Use [rel="noopener"] on external links

nscyclone:
`[rel="noopener"]` is added to external links having `[target="_blank"]`  to deny access to `window.opener`. [Here](https://developers.google.com/web/tools/lighthouse/audits/noopener) is a quick rundown on why it is important.

### DIFF
--- a/lib/ex_doc/formatter/html/templates/footer_template.eex
+++ b/lib/ex_doc/formatter/html/templates/footer_template.eex
@@ -2,11 +2,11 @@
         <p>
           <span class="line">
             Built using
-            <a href="https://github.com/elixir-lang/ex_doc" title="ExDoc" rel="help" target="_blank">ExDoc</a> (v<%= ExDoc.version %>),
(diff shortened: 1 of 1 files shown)


OvermindDL1:
> [target="_blank"] is added to links to source code in order to open them in a new tab;

Please no, like *really* **no**.  If I want to open a new tab I will middle-click.  Left-click should always *only* open in the same tab and it is really **really** irritating when a site overrides that behaviour so I have no option to open in the same tab then...

nscyclone:
Thanks, @OvermindDL1! You're right, there was no sense in changing current behaviour on links to source code. I've rebased the branch and updated the PR description accordingly.

whatyouhide:
Thanks @nscyclone! 💟

whatyouhide:
@nscyclone missed you're a committer on ex_doc, sorry for the merge + usual thanks 😄

PR #841, opened 2018-03-29
Fix autolinking when running as escript

### LINKED ISSUE #840: Impossible to build docs for Earmark
I was not succesfull in finding any trace of this on the net.

As indicated by @pragdave in this issue https://github.com/pragdave/earmark/issues/185
the ex_doc escript fails with the following stacktrace.
```
** (EXIT from #PID<0.68.0>) an exception was raised:
    ** (ArgumentError) argument error
        (ex_doc) lib/ex_doc/formatter/html/autolink.ex:247: anonymous fn/7 in ExDoc.Formatter.HTML.Autolink.format_and_extract_typespec_placeholders/4
        (elixir) lib/macro.ex:257: Macro.do_traverse/4
        (elixir) lib/enum.ex:1397: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
        (elixir) lib/macro.ex:243: Macro.do_traverse/4
        (elixir) lib/enum.ex:1397: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
        (elixir) lib/enum.ex:1397: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
        (elixir) lib/macro.ex:243: Macro.do_traverse/4
        (elixir) lib/enum.ex:1397: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
```
It was launched from Earmark's commit https://github.com/pragdave/earmark/commit/41c84f5f441d2b6201da593447a5790b05551ddd with the following args

`Earmark 1.2.5 _build/dev/lib/earmark/ebin -m Earmark -o doc -p https://github/pragdave/earmark.html -f html -n -n https://hexdocs.pm/earmark/master` 

The env was 1.6.4/20.3

All I could find out was that `elixir_source` is `nil` in line 244, but that code is way over my head.

wojtekmach:
Closes #840

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -561,15 +561,16 @@ defmodule ExDoc.Formatter.HTML.Autolink do
         path -> List.to_string(path)
       end
 
-    if Application.get_env(:ex_doc, :running_as_escript) do
-      Path.dirname(path)
-    else
+    if File.exists?(path) do
       path =
         path
         |> Path.dirname()
         |> Path.dirname()
         |> Path.dirname()
       path <> "/" <> app <> "/ebin"
+    else
+      # if beam file doesn't exists it's likely an escript
+      Path.dirname(path)
     end
   end
 
--- a/lib/ex_doc/cli.ex
+++ b/lib/ex_doc/cli.ex
@@ -5,8 +5,6 @@ defmodule ExDoc.CLI do
   Handles the command line parsing for the escript.
   """
   def main(args, generator \\ &ExDoc.generate_docs/3) do
-    Application.put_env(:ex_doc, :running_as_escript, true)
-
     {opts, args, _invalid} =
       OptionParser.parse(args,
         aliases: [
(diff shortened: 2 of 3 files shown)


lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -542,27 +542,37 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
           end
+
         Application.put_env(:ex_doc, :elixir_lib_dirs, lib_dirs)
         lib_dirs
     end
   end
 
+  defp elixir_lib_dir(app) do
+    path =
+      case :code.where_is_file('Elixir.Kernel.beam') do
```
wojtekmach: unfortunately we won't be able to autolink things from Mix, IEx etc because they're not loaded.

RobertDober:
Great job, thank you so much.

Would be great if we can install ex_doc inside Earmark as a dev dependency but I guess this is just impossible (unless we render with a different markdown converter but that would be kind of not what we want ;)).

Any thoughts on that?

Thx again and keep your excellent work up.

Issue #842, opened 2018-04-10
" ReferenceError: sidebarNodes is not defined"  when module name contains double quotes


arsssen:
Try to generate docs for this code:

```elixir
defmodule Test do
  for {module_name, _} <- %{ "module.x": nil } do
    defmodule module_name do
    end
  end
end
```

and you'll end up with code like this in dist/sidebar_items-xxx.js:
```javascript
 ... "modules":[{"id":""module.x"","title":""module.x"","group":""} ...
```
which is apparently invalid and doesn't allow sidebar to be rendered

arsssen:
I've created a pull request https://github.com/elixir-lang/ex_doc/pull/843  to fix this issue

PR #843, opened 2018-04-13
Fix generating invalid JS when module name contains double quotes (#842)

arsssen:


lib/ex_doc/formatter/html/templates.ex
```diff
@@ -169,6 +169,13 @@ defmodule ExDoc.Formatter.HTML.Templates do
   end
 
   defp sidebar_items_node(module_node) do
+    escape_quotes = &String.replace(&1, "\"", "\\\"")
```
OvermindDL1: Might be good to escape `\` too in case it appears at the end of a module name (unlikely though not impossible)

josevalim: Instead of `escape`, I think we should call `#{inspect(id)}` instead of `#{id}` when generating the sidebar data in the first place. :) That should take care of everything.

josevalim: Another option is to call `html_escape` or similar.

PR #844, opened 2018-04-17
Fix generating invalid JS when module name contains double quotes (#842)

arsssen:


josevalim:

PR #848, opened 2018-04-27
Add contribution instructions

### LINKED ISSUE #845: Contribution documentation
Hi 👋

I'm coming from hexpm/hexpm#659 and was thinking that I'd make an attempt at fixing the referenced bug. 

However – I'm stumbling a bit in getting `ex_doc` running locally so that I can debug the less and js code. Are there any contribution docs somewhere to get a newcomer up and running?

The `README.md` seems to only contain instructions on how to run ex_doc in the context of a project but I can't find any information on how to run it in the context of contributing.

The instructions in `./assets/README.md` are nice and clear but they fail to mention how I can test my changes in a browser locally.

wojtekmach:
Closes #845


mix.exs
```diff
@@ -35,7 +35,9 @@ defmodule ExDoc.Mixfile do
 
   defp aliases do
     [
-      clean: [&clean_test_fixtures/1, "clean"]
+      clean: [&clean_test_fixtures/1, "clean"],
+      setup: ["deps.get", &setup_assets/1],
+      docs: ["compile --force", &build_assets/1, "docs"]
```
wojtekmach: I think we should be able to get rid of `compile --force` if we add `@external_resource ...` to proper modules, I can look into that later.

README.md
```diff
@@ -108,7 +108,16 @@ Update your project configuration to use Cmark:
 docs: [markdown_processor: ExDoc.Markdown.Cmark]
 ```
 
-# License
+## Contributing
+
+The easiest way to test changes to ExDoc is to locally re-generate it's own docs:
+
+1. Run `mix setup` to install all dependencies
```
wojtekmach: `mix setup` uses just npm, however in https://github.com/elixir-lang/ex_doc/blob/master/assets/README.md#dependencies we mention yarn. I think we can assume folks installed yarn, so mix setup calls `yarn install` etc; wdyt?

josevalim: Agreed. We should use yarn. But I think after the first install they would behave the same.

mix.exs
josevalim: :+1:

josevalim:
:shipit:

PR #853, opened 2018-05-19
better syntax highlighting for elixir code

tmbb:
Use makeup for syntax highlighting of Elixir code.
Other languages will use Highlightjs as a fallback.

As soon as new lexers are written and properly tested, they can be imported by ExDocMakeup and will be made available. Adding more lexers willincrease compile times, so they should be limited to what makes sense, like Erlang, HTML, Javascript and maybe a few others.


josevalim:
@tmbb you updated the Elixir requirement but you also need to update the travis.yml file. Also please update the docs and error messages in the markdown module.

Question: do you need to use earmark because you integrate with Earmark when traversing code blocks or do you simply do a later pass on the html? Does the prompt feature we talked about work on makeup?

michalmuskala:
How does the compile time for makeup look like compared to "raw" ex_doc?

tmbb:
> Also please update the docs and error messages in the markdown module.

Sure, I'll see what needs to be changed.

> do you need to use earmark because you integrate with Earmark when traversing code blocks ... ?

I integrate with Earmark when traversing code blocks.

> or do you simply do a later pass on the html?

I think I could do this too. This way syntax highlighting would be independent of the markdown implementation. That said, any markdown implementarion that provides a configurable way of rendering code is fine.

josevalim:
@tmbb that would be the best as we could handle it completely decoupled from the markdown engine.

josevalim:
> I think I could do this too. This way syntax highlighting would be independent of the markdown implementation.

That would be great! Could you please give it a try? Also, if we are going in this direction, I don't think we need exdoc_makeup as a separate project, we could add support for it directly in ExDoc.

tmbb:
> Could you please give it a try?

Yes, it shouldn't be very hard...

> Also, if we are going in this direction, I don't think we need exdoc_makeup as a separate project, we could add support for it directly in ExDoc.

You don't *need* ex_doc_makeup as a separate project right now. ex_doc_makeup contains very little actual code, and it could be incorporated into ExDoc itself like the other markdown behaviours are. "Getting rid of ex_dox_makeup" is **not** a valid reason to work on the HTML directly.

A valid reason would be to decouple it from earmark, I guess, or to allow the user to use something other than markdown for the docs.

But deciding to work on the HTML directly  requires:

1. Knowing the css classes that the markdown implementation decides to add to the fenced code blocks

2. Parse the language from the css classes

3. Decode the *already escaped* contents of the code tag

4. Highlight the code and escape them again

5. Insert the highlighted code into the HTML

Steps 1 and 2 are a little problematic because they couple us to a concrete markdown implementation.

This whole setup also breaks all extensibility options we've been working these last few months. Instead of having syntax highlighting coupled to highlightjs you have sybtax highlighting coupled to makeup.

tmbb:
Makeup requires some extra assets, of course: a css stylesheet and a small js file so that matching parenthesis and the like are highlighted on mouseover (the JS is 100% optional of course, the syntax highlighting works with JS disabled!).

Those assets would have to be included in ExDoc and declared by the builtin Markdown modules (this require changing only 2 lines of code per module, nothing too big)

tmbb:
> How does the compile time for makeup look like compared to "raw" ex_doc?

Makeup with the elixir lexer takes about 4-5 seconds to compile. The elixir lexer is an abnormally complex lexer because Elixir's "syntax" (I'm using the term very loosely) is very complex and nimble_parsec must generateba large number of function heads (which the BEAM must compile and optimize). Other lexers are much more simple and compile in under 0.5s.

My remark about compilation time is that we shouldn't add the 300 lexers in pygments to ExDoc (I wish I had 300 lexers already...), because even at 0.5 seconds per lexer that would be a lot. Python can get away with it because it generates object code very quickly (and on-demand) and Python's documentation generators are not compiled per project

josevalim:
> Makeup requires some extra assets, of course: a css stylesheet and a small js file so that matching parenthesis and the like are highlighted on mouseover (the JS is 100% optional of course, the syntax highlighting works with JS disabled!).

Alternatively we can include them in the general bundle by making them part of the gulp file. Or we generate a bundle with both highlight+makeup assets.

tmbb:
I've removed support for Elixir 1.3 (because nimble_parsec seems to require 1.4).
I'm aware some tests are failing and I'll solve the issue. I think the problem is that we're trying to match a regex that no longer matches because syntax highlighting happens with HTML.

Makeup is now called on the generated HTML and not on the markdown. You'll notice a new ExDoc.Highlighter module that exports a function that highlights code blocks in an HTML file. ExDocMakeup is no longer required and ExDoc now depends directly on makeup and makeup_elixir.

You can see the Phoenix docs highlighted with the new ExDoc package:

* Normal docs with elixir code: [https://tmbb.github.io/phoenix/endpoint.html]
* A file that shows the highlighting of iex sessions *including errors* (I've added a new dummy section to this pase just to showcase error highlighting: [https://tmbb.github.io/phoenix/ecto.html]

tmbb:
> Also please update the docs and error messages in the markdown module.

@josevalim now that Makeup is included and used in all default makeup implementations in such a way that the user doesn't even notice it's there, I don't think there's anything I need to change in the docs or in the error messages.

mix.exs
```diff
@@ -28,6 +28,9 @@ defmodule ExDoc.Mixfile do
   defp deps do
     [
       {:earmark, "~> 1.1"},
+      # Syntax highlighting (independent of the markdown implementation)
+      {:makeup, "~> 0.5.1"},
+      {:makeup_elixir, "~> 0.6.0"},
```
josevalim: Shouldn't depending on makeup_elixir automatically bring makeup?

lib/ex_doc/highlighter.ex
```diff
@@ -0,0 +1,37 @@
...
+
+  defp render_code(lang, lexer, code) do
+    highlighted =
+      code
+      |> unescape_html_entities()
+      |> Makeup.highlight_inner_html(lexer: lexer)
+    ~s(<pre><code class="nohighlight makeup #{lang}">#{highlighted}</code></pre>)
+  end
+
+  # TODO: this implementation is probably not very efficient.
```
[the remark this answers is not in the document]
josevalim: I would just reverse the implementation in Plug.HTML, which has already been benchmarked. :)

lib/ex_doc/markdown/earmark.ex
```diff
@@ -3,13 +3,29 @@ defmodule ExDoc.Markdown.Earmark do
...
+  # Callback implementations
+
+  @assets [
+    # Read the CSS from the included file.
+    # This allows us to have a custom CSS theme not included in Makeup
+    # that supports both "day mode" and "night mode".
+    {"dist/ex_doc_makeup-css.css", File.read!("assets/dist/ex_doc_makeup.css")},
+    # Get the Javascript snippet directly from Makeup.
+    # If there is any need to customize it further, we can add a "ex_doc_makeup.js" file.
+    {"dist/ex_doc_makeup-js.js", HTMLFormatter.group_highlighter_javascript()}
```
josevalim: I think the distribution files should be called `ex_doc_highlighter` because if in three years we decide to use something other than makeup, we don't want to break existing adapters. Similarly, the javascript thingy should be returned by `ExDoc.Highlighter`. The HTMLFormatter is not really public API and therefore engines should not access it.

lib/ex_doc/markdown/cmark.ex
```diff
@@ -24,6 +40,9 @@ defmodule ExDoc.Markdown.Cmark do
   Generate HTML output. Cmark takes no options.
   """
   def to_html(text, _opts) do
-    text |> Cmark.to_html()  |> ExDoc.Markdown.pretty_codeblocks()
+    text
+    |> Cmark.to_html()
+    |> ExDoc.Markdown.pretty_codeblocks()
+    |> ExDoc.Highlighter.highlight_code_blocks()
```
josevalim: One uses `code_blocks` and the other uses `codeblocks`.

josevalim: Although I don't think we need `pretty_codeblocks` anymore. The functionality in there should be handled by the highlighter.

mix.exs
tmbb: Yeah, it should. We can depend only on makeup and makeup_elixir will be brought as a transitive dependency.

josevalim: You probably mean the other way around. :)

lib/ex_doc/markdown/cmark.ex
tmbb: > Although I don't think we need pretty_codeblocks anymore. The functionality in there should be handled by the highlighter.

We don't and it should. It will probably be even more performant if I handle it in the same regex match.

lib/ex_doc/markdown/earmark.ex
tmbb: I agree

lib/ex_doc/highlighter.ex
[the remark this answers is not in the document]
tmbb: Yes, that seems simple.

lib/ex_doc/markdown/earmark.ex
tmbb: And more than that, the assets (including the filenames) should be defined in the ExDoc.Highlighter module, and there should be an ExDoc.Highlighter.assets/1 function that is simply called by the behaviours. That would centralize the assets in the right place.

tmbb:
A little offtopic: I have an Erlang lexer almost  ready. It needs more unit tests and some integration testing on real code bases. Do you think I should add ir here once I test it a little better?

lib/ex_doc/markdown/cmark.ex
tmbb: And regarding codeblocks vs code_blocks I prefer code_blocks. Is that ok with you?

josevalim:
> A little offtopic: I have an Erlang lexer almost ready. It needs more unit tests and some integration testing on real code bases. Do you think I should add ir here once I test it a little better?

Maybe we could have a discovery system? Every time the `makeup_elixir`/`makeup_erlang`/etc application starts, it registers itself with `makeup`? So adding new engines is a matter of adding new dependencies?

I am thinking something like this:

```elixir
defmodule Makeup.Elixir.Application do
  def start(_, _) do
    Makeup.register(:elixir, Makeup.Elixir)
    Makeup.register(:iex, Makeup.IEx)
    Supervisor.start_link([], strategy: :one_for_one)
  end
end
```

This way we can ship only with Elixir and libraries can add more engines as necessary.

lib/ex_doc/markdown/cmark.ex
josevalim: I would actually say none? `ExDoc.Highlighter.highlight_html`? or `ExDoc.Highlighter.highlight_html_code`?

tmbb:
> Maybe we could have a discovery system? Every time the makeup_elixir/makeup_erlang/etc application starts, it registers itself with makeup? So adding new engines is a matter of adding new dependencies?

I like that. Lexers would register themselves together with a language name (a binary instead of an atom, probably). Makeup could then pick the right lexer from the language name using the registry.

Where would you store the key/value pairs? Genserver? Application config? ETS table?

tmbb:
> I would use the application environment.

That would be my first choice too. Great minds think alike :) This OTP thing with applications and application environments is very useful at times. Those Erlang guys had some cool ideas sometimes

OvermindDL1:
> Makeup requires some extra assets, of course: a css stylesheet and a small js file so that matching parenthesis and the like are highlighted on mouseover (the JS is 100% optional of course, the syntax highlighting works with JS disabled!).

What about instead of javascript have the html be decorated with the appropriate onhover events that just add/remove classes to itself and sibling parenthesis/bracket/etc?  No additional javascript would be needed then and it self-contains it all well?

Or does the javascript do something else as well?

tmbb:
@OvermindDL1, this is not the place to discuss an internal Makeup implementation detail. Let's bring the discussion here instead: https://github.com/tmbb/makeup/issues/12

tmbb:
I still need to rename the assets

tmbb:
@josevalim I'm still not 100% sure of the api for the registry. Could we merge these changes as they are (when all tests pass, etc) and give me some more time to think about the makeup registry?

formatters/epub/dist/app-12c41c88e5.js
```diff
@@ -0,0 +1,2 @@
...nute timezone_region to to_base64 to_date to_days to_seconds todatetimeoffset trace tracking transaction transactional translate translation treat trigger trigger_nestlevel triggers trim truncate try_cast try_convert try_parse type ub1 ub2 ub4 ucase unarchived unbounded uncompress under undo unhex unicode uniform uninstall union unique unix_timestamp unknown unlimited unlock unpivot unrecoverable unsafe unsigned until untrusted unusable unused update updated upgrade upped upper upsert url urowid usable usage use use_stored_outlines user user_data user_resources users using utc_date utc_timestamp uuid uuid_short validate validate_password_strength validation valist value values var var_samp varcharc vari varia variab variabl variable variables variance varp varraw varrawc varray verify version versions view virtual visible void wait wallet warning warnings week weekday weekofyear wellformed when whene whenev wheneve whenever where while whitespace with within without work wrapped xdb xml xmlagg xmlattributes xmlcast xmlcolattval xmlelement xmlexists xmlforest xmlindex xmlnamespaces xmlpi xmlquery xmlroot xmlschema xmlserialize xmltable xmltype xor year year_to_month years yearweek",
```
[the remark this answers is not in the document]
tmbb: I will obviously ignore any issues this ebert bot detects in the artifacts generated by gulp...

tmbb:
## Assets

Assets are now included in the bundles. This makes sense because syntax highlighting is now a feature of ExDoc and not a feature of the markdown implementation.

## Supported Lexers

Currently ExDoc only supports the elixir lexer. This is hardcoded in the source and offers no possibilities for customization. The way forward is to have some kind of registry in which lexers can register themselves on application start. Until I decide which is going to be the final API for such registry, this solution is enough.

In any case, unsupported languages will be highlighted by highlightjs as they are now. So this version of makeup is strictly better than the previous one (it uses a better lexer for Elixir and the same inferior lexers for other languages).

## Decoding HTML

It's not obvious how to invert the Plug.HTML implementation (which only encodes and doesn't decode). Even If I inverted it, it is not obvious it would be the fastest possible implementation. In order to have the fastest implementation possible, I'd need to benchmark several possibilities anyway. That will take a couple days because I'm quite busy, and creating a good microbenchmark takes some work.

The current version is quite fast, and probably not a bottleneck for realistic inputs. The implementation can be trivially switched at any time (it's a pure function after all). For these reasons, I think this could be merged and possibly improved in the near future.

## CSS Themes

Someone with some experience in design and CSS should take a look at the themes, especially the night mode. In particular, the night mode color for erlang tracings is quite unreadable: (for example [here](https://tmbb.github.io/phoenix/ecto.html); you have to switch to night-mode)

I can change that color to something else with better contrast, of course, but the contribution of someone with actual design skills would be appreciated. I can explain how the css classes are used (and should probably document it).

## Merging

I think that *except for the problem with the night-mode theme*, this branch is ready to be merged into master and to be released as ex_doc 19.0. Despite the possible improvements I suggest above, I think this version is strictly better than the previous one and perfectionism isn't our friend here.

lib/ex_doc/highlighter/html_decoder.ex
```diff
@@ -0,0 +1,24 @@
...
+  defp to_iodata(<<>>) do
+    []
+  end
+
+  def unescape_html_entities(html) do
+    html
+    |> to_iodata()
+    |> IO.iodata_to_binary
+  end
+end
```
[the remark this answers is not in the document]
tmbb: @josevalim, is it against the Elixir code of conduct to insult our dear friend Ebert? :)

assets/less/makeup.less
```diff
@@ -0,0 +1,178 @@
...
+.night-mode .makeup .sh {color: #48b685; } /* :string_heredoc */
+.night-mode .makeup .si {color: #f99b15; } /* :string_interpol */
+.night-mode .makeup .sr {color: #48b685; } /* :string_regex */
+.night-mode .makeup .ss {color: #48b685; } /* :string_symbol */
+.night-mode .makeup .sx {color: #48b685; } /* :string_other */
+.night-mode .makeup .sx {color: #48b685; } /* :string_sigil */
+.night-mode .makeup .vc {color: #ef6155; } /* :name_variable_class */
+.night-mode .makeup .vg {color: #ef6155; } /* :name_variable_global */
+.night-mode .makeup .vi {color: #ef6155; } /* :name_variable_instance */
+.night-mode .makeup .vm {color: #ef6155; } /* :name_variable_magic */
```
eksperimental: I think if these classes are grouped they could be simplified

mix.exs
```diff
@@ -28,6 +28,8 @@ defmodule ExDoc.Mixfile do
   defp deps do
     [
       {:earmark, "~> 1.1"},
```
eksperimental: Markdown should be in title case

tmbb: Fixed

assets/less/makeup.less
tmbb: What do you mean by "grouped"? Something like:

```css
.night-mode .makeup .vc, .vg, .vi, .vm {color: #ef6155; }
```

or simply replacing them by a single class? If so, then that's a definite No, because even if this particular theme doesn't show a distinction between these classes, other themes might.

eksperimental: I mean nesting classes

tmbb:
I think this is ready to be merged now :)

milmazz:
@eksperimental ping? Have you had any chance to check the latest changes from @tmbb?

eksperimental:
@milmazz sorry, I haven't . I will try to but please go ahead with it if needs merging

eksperimental:
If it is regarding the CSS changes, YES, that's what I meant, but I haven't had a chance to run this PR locally

tmbb:
If any of my changes is not clear, I'm can help. Just ask :)

josevalim:
Hi @tmbb! Sorry for the delay on this. My plan is to pick this up this week and align it with Elixir v1.7-rc.

josevalim:
Merged, thank you! I have tried this on the elixir repo and generating docs went from 11.5s to 12.4s! So this is performing amazingly well and the result looks better than the highlighter, IMO! Great job! 

tmbb:
> I have tried this on the elixir repo and generating docs went from 11.5s to 12.4s!

The performance impact of using Makeup is only about 8%. This is actually very good, given what it does. Because the impact is so low, improving Makeup's performance won't have a big user-visible impact.

PR #855, opened 2018-05-30
Link directly to page if sidebar item has no subitems

nbw:
# Summary

For sidebar items with no subitems, link directly to the page. 

## Issue
Hex docs show a "Top" sub header even if there are no other contents.  As an example, go to the [Phoenix Hex documentation](https://hexdocs.pm/phoenix/Phoenix.html), click **Guides** then **Installation** and you'll see something like:

screenshot 2018-05-30 13 49 05

In order to actually view the page, you have to click Top. This feels unnecessary.

## Solution

Check if there are any contents in each sidebar item, specifically for:
+ types
+ functions
+ macros
+ callbacks
+ headers

If there are no items link directly to the page.

# Screenshots
## Currently
screenshot 2018-05-30 13 42 13

current_sidebar

## With Top removed

screenshot 2018-05-30 13 44 00

new_sidebar

# Testing

Verified by generating docs for the ex_docs repo.

### DIFF
--- a/assets/js/template-helpers/isEmptyNode.js
+++ b/assets/js/template-helpers/isEmptyNode.js
@@ -1,0 +1,19 @@
+export default function (node, options) {
+  var nodeItems = [
+    node.types,
+    node.functions,
+    node.macros,
+    node.callbacks,
+    node.headers
+  ].filter(Array.isArray);
+
+  if (flatten(nodeItems).length === 0){
+    return options.fn(this);
+  } else {
+    return options.inverse(this);
+  }
+}
+
+function flatten(array) {
+  return array.reduce(function(a, b){ return a.concat(b);}, [])
+}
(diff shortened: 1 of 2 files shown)


assets/js/template-helpers/isEmptyNode.js
```diff
@@ -0,0 +1,19 @@
+export default function (node, options) {
+  var nodeItems = [
+    node.types,
+    node.functions,
+    node.macros,
+    node.callbacks,
+    node.headers
+  ].filter(Array.isArray);
```
josevalim: Wouldn't a simpler implementation instead traverse this list of nodeItems and then check that they are null or an empty list? This way we don't need to flatten nor build a very large list only to check its length later.

nbw: Fair point @josevalim. Refactored to iterate through nodeItems and return true as soon as it detects content. Otherwise, false.

PR #856, opened 2018-06-10
Fix Elixir module matcher RegEx

CharlotteDunois:
This PR adds a negative lookbehind for multiline codeblocks, to fix an issue where a module name in an inline codeblock after a multiline codeblock didn't get correctly matched, due to the regex matching from the multiline codeblock to the inline codeblock.

Before: https://regex101.com/r/Du8vUp/1
After: https://regex101.com/r/Zit4pj/1


josevalim:
Thank you @CharlotteDunois! Could you please also add unit test to [autolink_test.exs](https://github.com/elixir-lang/ex_doc/blob/master/test/ex_doc/formatter/html/autolink_test.exs) so we don't have any regressions in the future?

CharlotteDunois:
@josevalim Thanks for the heads up! The unit test is added.

josevalim:

PR #861, opened 2018-06-23
Autolinl Kernel functions and special forms with shorthands

milmazz:
With this change, now we can create the following links:

<pre>
* [`for`](`for/1`)
* [`is_boolean`](`is_boolean/1`)
</pre>

Refs: #679


josevalim:
@milmazz I am not sure if we should go ahead with this. The arity is very important and I would like to enforce it. Maybe we are interpreting #679 differently? :)

milmazz:
@josevalim But the arity is there: [some custom text](`for/1`) :)

josevalim:
Oh, I am so sorry for the confusion. Perfect!

wojtekmach:
👏

eksperimental:
Thank you Milton for taking care of this. It helps a lot a write clearer documentation

Issue #867, opened 2018-07-08
Allow app to be excluded from umbrella app docs


aaronrenner:
I've been generating docs for my phoenix-based umbrella apps, but I'd really like my_app_web to be excluded from the docs since it's not part of my app's public API. Ideally, I'd like to be able to whitelist or blacklist applications in an umbrella from the documentation build. This would allow me to only build docs for the apps that I wish for other developers to know about. I know I can generate docs for each project individually, but I do like how all of the docs for an umbrella are compiled into one. Is this possible?

josevalim:
@aaronrenner we can support a configuration on your Mix.Project that allows that, as I believe we don't support this right now. A pull request is welcome!

aaronrenner:
Thanks for being open to adding a config like this! I'll try to find some time to work on a PR.

leonardobork:
I'm actually working on a PR for this.

Thought something like 

In `/lib/ex_doc/retriever.ex`.

```elixir
def docs_from_files(files, config) when is_list(files) do
    files
    |> Enum.filter(&valid_files(&1, config))
    |> Enum.map(&filename_to_module(&1))
    |> docs_from_modules(config)
end

defp valid_files(name, config) do
    name = Path.basename(name, ".beam")
    [project | _] = String.split(name, ".")
    !Enum.member?([config.ignored_apps], project)
end
```
Might do the trick... But I don't know if it is the best way to do it.

Any suggestions?

josevalim:
@leonardobork for umbrellas it should be a matter of filtering the application names here and not adding them to the compilation list in the first place: https://github.com/elixir-lang/ex_doc/blob/master/lib/mix/tasks/docs.ex#L263-L265 :)

lackac:
@josevalim, I think this can be closed now as #871 has been merged. Or is there anything else to be done for this?

aaronrenner:
@lackac Thanks for building this. I'm looking forward to using it!

lackac:
@aaronrenner the work was done by @leonardobork in #871, he deserves the praise. I was just helping with keeping the issue tracker tidy. :)

PR #871, opened 2018-07-13
Allow apps to be excluded from umbrella app docs

leonardobork:
Allowing to exclude docs from umbrella app docs utilizing `ignored_apps: [your selected apps]`.

I also have created tests for umbrella projects and tests for this specific case.

This PR is related to this [issue](https://github.com/elixir-lang/ex_doc/issues/867).

### DIFF
--- a/lib/mix/tasks/docs.ex
+++ b/lib/mix/tasks/docs.ex
@@ -249,7 +249,7 @@ defmodule Mix.Tasks.Docs do
   defp normalize_source_beam(options, config) do
     compile_path =
       if Mix.Project.umbrella?(config) do
-        umbrella_compile_paths()
+        umbrella_compile_paths(get_ignored_apps(options))
       else
         Mix.Project.compile_path()
       end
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @leonardobork, a couple suggestions:

1. Let's call the option `:ignore_apps`

2. The value of the option should be atoms and not strings

3. Please do not format the test file as it makes it hard to review this feature

4. We don't need to store `ignored_apps` in ExUnit.Config as it should only be used in the Mix task

Thanks!

lib/mix/tasks/docs.ex
```diff
@@ -249,19 +249,23 @@ defmodule Mix.Tasks.Docs do
   defp normalize_source_beam(options, config) do
     compile_path =
       if Mix.Project.umbrella?(config) do
-        umbrella_compile_paths()
+        umbrella_compile_paths(get_ignored_apps(options))
```
josevalim: `get_ignored_apps(options)`  could be written as `Keyword.get(options, :ignore_apps, [])`.

josevalim:
Also, please document the new configuration under `mix/tasks/docs.ex`.

leonardobork:
@josevalim Thanks, I'll fix this when I get home!

josevalim:

aaronrenner:
@leonardobork Thanks for this! I'm looking forward to using it! ❤️

Issue #851, opened 2018-05-14
Slow scrolling on Safari for macOS 10.13 due to additional style rule


evadne:
The page https://hexdocs.pm/mix/Mix.Tasks.Profile.Fprof.html scrolls very slowly (20fps – 30fps) on rMBP 2016 when integrated graphics is used on Safari 11.1 (13605.1.33.1.4) for macOS 10.13 due to this rule:

    .content-inner .specs pre, .content-inner code {
      font-family: Inconsolata,Menlo,Courier,monospace;
    }

When scrolling up/down the frame rate drops to 30fps. Unfortunately when recording screen content with QuickTime Player the frame rate is capped at 30fps anyway. But disabling this particular rule in Inspector restores performance (55fps – 60fps).

josevalim:
Thanks @evadne! I could not reproduce on Safari 11.1.1. Can you?

josevalim:
Ping. :)

net:
Same behavior for me on both 11.1.1 and 12.0. I agree it is likely a Safari bug; however, it may still be worth considering a interim patch.

josevalim:
Thanks @net! Since I cannot reproduce it locally, could you take a look at it and see if you can come up with a fix?

michalmuskala:
I can reproduce this, but disabling the style rule that @evadne pointed out does not fix it for me.

smfr:
Slow scrolling triggered by certain fonts was fixed in macOS Mojave.

josevalim:
Glad to hear!

josevalim:
Thanks for the updates!

evadne:
Hi, on 10.13 this still happens somehow if the window is larger, I can get it to lag on a Vega 64 when the window is 1920 by 2160 pixels. I have done some poking around today, and found that this is likely because the sidebar and content elements are overlapping. The content element sits under the sidebar and has `width: 100%; padding-left: 300px`. I have tried to change this to `width: calc(100% - 300px)` and it seems much better now.

kzvi:
Just for the record I still have this issue on Mojave.

archdragon:
I have a theory about where the problem might be. I've tried to reproduce and fix it locally.

@evadne @kzvi If you have a moment can you take a look at:
1) [My demo](http://krowinski.com/ex_doc/safari_scroll/Kernel.html)
2) [Hexdocs](https://hexdocs.pm/elixir/Kernel.html)

[...]

PR #878, opened 2018-07-26
Fix Safari scrolling lag

net:
Removing the JavaScript focus dispatch fixes the Safari scrolling issue (#851). I also removed the `overflow-y: auto` on `.content` to keep the current behavior where main content is immediately keyboard-scrollable.

The downside to this approach is that scrolling up on the sidebar when the sidebar is fully scrolled to the top, or down when it is scrolled to the bottom, will cause the main content to scroll. This can be fixed, but I wanted to see if we were happy with this simpler option.

Tested on the latest Safari and Chrome.


josevalim:
Thanks @net! Given the purpose of this code was precisely to avoid the sidebar scrolling, I wonder what other alternatives we have to solve this issue. I am quite sure that if we remove this behaviour, we will have bug reports asking to add it back. :)

OvermindDL1:
Why aren't both the sidebar and main page their own scrolling contexts?  If the main body doesn't scroll (just 100% all around) then then should be easily optimized by the browsers?

mnussbaumer:
Not sure if this is solved, if not, just chipping in because I ran into this issue on one of my projects, running macOSX 10.13.6 safari 12.0, this is probably a bug in Safari since neither chrome nor ff do this sluggish scroll. In my own case, specifying `will-change: scroll-position;` for my div removed the sluggishness but it doesn't work on the hex pages (and it shouldn't since there's no transparencies nor anything that could be optimised). What does work is removing the `height: 100%` from the `.content` container, if it's a `block` it will always extend to 100% height automatically, and I think work correctly in any situation.

net:
@mnussbaumer Excellent. Would you like to submit a PR?

mnussbaumer:
> @mnussbaumer Excellent. Would you like to submit a PR?

@net since you already worked on the PR and have it ready I'm totally fine if you do that?

RudManusachi:
In my Safari 12.0.1 it scrolls just fine while safari's "scroll view" is in a background of a system and focus is on a different app, such as finder, or even safari's webinspector in separate window
 
<details>
  <summary>I tried to capture it here</summary>

safariscroll2

</details>

josevalim:
Many are reporting the bug no longer exists, so hopefully it is permanently fixed with the Safari update. Let us know if it isn't. Thanks!

Issue #876, opened 2018-07-24
Arbitrary grouping of functions with `@doc group: :some_group`


lackac:
The new documentation metadata and specifically the `@doc keyword()` construct opens up new possibilities for enriching the ExDoc UI and providing better ways to navigate and explore.

ExDoc is already making use of the `guard: true` metadata to group guards together in `Kernel` and other modules. It could similarly allow arbitrary grouping of functions:

``` elixir
defmodule Flow do
  @doc group: :mappers
  def each(flow, each)

  @doc group: :mappers
  def filter(flow, filter)

  @doc group: :reducers
  def reduce(flow, acc_fun, reducer_fun)

  @doc group: :reducers
  def group_by(flow, key_fun, value_fun)
end
```

The code above would instruct ExDoc to create two new groups in the sidebar: "Mappers" and "Reducers". The same grouping would also be used on the main content side. The behaviour would be exactly the same as the existing "Callbacks" and "Types" grouping. Optionally we could also introduce a way to toggle grouping.

This is mostly useful for modules that define a high number of functions where separating them conceptually would provide a better experience for the reader. (e.g. documentation of Flow, or some ExAws packages with larger footprints)

Some details of how this would behave still needs to be worked out and can be discussed if this finds support.

josevalim:
I like this, although some people had reservations around groups. /cc @ericmj @benwilson512 @michalmuskala

josevalim:
@lackac btw, it may be better to have a bit of indirection here and allow groups to be created based on the metadata. This means that we would do:

    @doc mapper: true
    @doc reducer: true

Or

    @doc flow_type: :mapper

And then we tell ExDoc how to build this:

```elixir
groups_for_functions: %{
  Mappers: & &1[:flow_type] == :mapper
}
```

Similar approach could be done for `groups_for_modules`. But in the `groups_for_modules` case I like that all information is in a single place (the mix.exs). This is much harder to do for functions.

benwilson512:
I like this idea. A good example would be ExAws.S3 which has a reasonably large surface area, and the main AWS docs categorize the actions into "Actions on Buckets", "Actions on Objects" etc. Being able to reproduce that structure in ExDoc would definitely improve the doc experience.

lackac:
I wouldn't mind working on this, but I might not be able to get to it in the next couple weeks. In other words, don't wait on me if you prefer to have this sooner. :)

josevalim:
No worries, take your time! There is no rush.

[...]

PR #879, opened 2018-07-30
Group functions based on metadata

lackac:
Related to #876.

- [x] initial documentation for gathering feedback
- [x] add `:groups_for_functions` to configuration
- [x] write tests
- [x] collect functions into groups in Retriever
- [x] update templates and JS
- [x] implement Guards group as a default grouping


josevalim:
@lackac beautiful. :+1:

lackac:
I was thinking about dogfooding with ExDoc, but since no modules have an abundance of functions it doesn't make much sense to start breaking them into groups.

Instead I would love to get feedback from library authors that might use this feature. I'm going to look into grouping functions in Flow. In Elixir I think `Kernel`, `Enum`, `Stream`, `String`, `File`, and `IO.ANSI` would benefit most from grouping. Any takers? :)

@benwilson512 could you try this on `ExAws.S3`?

josevalim:
@lackac for Elixir, I would go easy and just start with Kernel so far. People have complained about grouping in the past, that makes it harder to find functions since we lose the alphabetical ordering. I agree on `Flow` though! :+1:

lackac:
There's a catch with the current implementation that I haven't made clear. It allows a function to be put into multiple groups. The rules are currently the following:

1. if a function is in any group other than the default "Functions" group, then it won't be put into "Functions"
2. guards will always be in the "Guards" group and optionally in other groups as well, but never in "Functions"
3. all functions without any specific grouping will be grouped into "Functions"
4. since a function can appear in multiple groups it's possible that we violate the uniqueness of html ids in the document. I'm not too worried about this since the browsers handle it well, but would be nice to find a solution to this. In any case there will be an extra id generated for custom groups in order to make the sidebar linking work.

josevalim:
@lackac to avoid the chance of conflicts, I would prefer if we uniquely identify them for now. Otherwise it feels like a tag system, which is something we could add, but maybe out of scope for what we are doing. Thoughts?

lackac:
@josevalim you mean a function could only belong to a single group? That would simplify some decisions. If that's what you want I can make that clear in the docs and make the necessary changes today.

josevalim:
@lackac yes, single group for now. We will say the first matching group is the one that sticks (so they are order dependent). If none is found, it defaults to functions.

lackac:
@josevalim last question: would you put "Guards" in front of custom groups or just before "Functions"? This matters if we only allow a function to appear in a single group.

I'm leaning towards giving custom groups higher priority. ie. if a guard functions is put into a custom group it should only appear there and not in "Guards". This behaviour is then similar to the default "Functions" group.

josevalim:
@lackac The configuration is a keyword list, right? We can keep the same order as in the keyword list. Functions always come last. What do you think?

lackac:
@josevalim that's right, but ExUnit adds an extra default item to that keyword list: `[Guards: & &1[:guard] == true]`. Should this be prepended or appended to the list that comes from configuration?

josevalim:
We won't have a "Guards" default. Elixir will have to add it and other packages will have to add that too if they want it.

lackac:
@josevalim I apologise if I didn't make it clear, but I consider this complete and it can now be reviewed/merged.

josevalim:

PR #880, opened 2018-08-01
Format codebase and check formatting on CI

lackac:


test/ex_doc/formatter/html/autolink_test.exs
```diff
@@ -393,7 +398,11 @@ defmodule ExDoc.Formatter.HTML.AutolinkTest do
     end
 
     test "autolinks inside parameterized types" do
-      assert Autolink.typespec(quote(do: parameterized_t(foo())), [parameterized_t: 1, foo: 0], []) ==
+      assert Autolink.typespec(
+               quote(do: parameterized_t(foo())),
+               [parameterized_t: 1, foo: 0],
+               []
```
josevalim: Can we move at least one of those to a var so the call fits a single line?

test/ex_doc/formatter/html/autolink_test.exs
```diff
@@ -95,21 +92,20 @@ defmodule ExDoc.Formatter.HTML.AutolinkTest do
...
                "[`MyModule.spaces/0`](MyModule.html#spaces/0)"
 
-      assert Autolink.elixir_functions(
-               "`ModA.example/1` and `ModB.example/2`",
-               ["ModA.example/1", "ModB.example/2"]
-             ) ==
+      assert Autolink.elixir_functions("`ModA.example/1` and `ModB.example/2`", [
+               "ModA.example/1",
+               "ModB.example/2"
+             ]) ==
```
josevalim: Can we please keep this example and previous ones in the previous syntax? I.e. one argument per line with the whole list on a single line? Thanks!

test/ex_doc/formatter/html/autolink_test.exs
lackac: Would shortening the type name be a good compromise? For example `p_t(foo())`? The test description should be enough to learn that these are parameterized.

josevalim: sounds good to me.

josevalim:

PR #881, opened 2018-08-06
add link to changelog

jlgeering:
direct link from https://hex.pm/packages/ex_doc to changelog

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -55,6 +55,7 @@ defmodule ExDoc.Mixfile do
       files: ["formatters", "lib", "mix.exs", "LICENSE", "CHANGELOG.md", "README.md"],
       links: %{
         "GitHub" => "https://github.com/elixir-lang/ex_doc",
+        "Changelog" => "https://github.com/elixir-lang/ex_doc/blob/master/CHANGELOG.md",
         "Writing documentation" => "https://hexdocs.pm/elixir/writing-documentation.html"
       }
     ]


josevalim:
There is nothing wrong with including a link to the CHANGELOG but we don't have this habit in our projects and, since we include the link to the source, the information can easily found. Thanks @jlgeering!

jlgeering:
@josevalim you are correct, it is just 1 additional click, but it is an annoying one. Damn, why are habits so hard to change ;-)

But maybe there is a better / easier way that I don't know of, to got from the output of mix hex.outdated to the changelog of a package, without having to search for the package on hex.pm? Or would that be a feature request for hex?

jlgeering:
@josevalim and would you also reject a PR to add the changelog as a page to the documentation (extra), like for example https://hexdocs.pm/quantum/changelog.html ?

josevalim:
@jlgeering also note that linking to master will definitely be wrong for most releases. Since pages are not per release, I don't think including the changelog there is a good idea. Including it as part of the docs is definitely a better option. We can try that! :+1:

jlgeering:
well I would tend to disagree on linking to master not being a good thing, as for example https://github.com/elixir-lang/ex_doc/blob/master/CHANGELOG.md actually contains all versions and so is appropriate for all releases ;-)

anyways, created a PR for plan B

PR #884, opened 2018-08-12
Pushing Earmark dependency to 1.2.5

RobertDober:
Not sure this is useful, but I saw some warning and given that 1.7 is out it might be a good idea to force a modern version of `Earmark`.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -27,7 +27,7 @@ defmodule ExDoc.Mixfile do
 
   defp deps do
     [
-      {:earmark, "~> 1.1"},
+      {:earmark, "~> 1.2.5"},
       {:makeup_elixir, "~> 0.7"},
       {:cmark, "~> 0.5", only: :test},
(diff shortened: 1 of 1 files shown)


mix.exs
```diff
@@ -27,7 +27,7 @@ defmodule ExDoc.Mixfile do
 
   defp deps do
     [
-      {:earmark, "~> 1.1"},
+      {:earmark, "~> 1.2.5"},
```
josevalim: Shouldn't we do `~> 1.2` instead so we automatically pick 1.3 when it is out?

RobertDober:
You are right, but I was surprised that 1.2.3 instead of 1.2.4 was used, probably I overlooked a thingy in `mix.lock` 
I'll fix it soon.

RobertDober:
Just in case if you want to merge this ;). I have corrected the version to 1.2.

josevalim:

PR #896, opened 2018-09-12
[WIP] Autolink revamp - Round 2

eksperimental:
Round 1 is here: [WIP] Autolink revamp #893

This PR doesn't change the API as requested by @josevalim in https://github.com/elixir-lang/ex_doc/pull/893#issuecomment-419695702

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -2,9 +2,13 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   @moduledoc """
   Conveniences for autolinking.
   """
-
   import ExDoc.Formatter.HTML.Templates, only: [h: 1, enc_h: 1]
 
+  @type language :: :elixir | :erlang | :markdown
+  @type kind :: :function | :module
+  @type link_type :: :normal | :custom
+
+  @backtick_token "<B706848484895T>"
   @elixir_docs "https://hexdocs.pm/"
   @erlang_docs "http://www.erlang.org/doc/man/"
   @basic_types_page "typespecs.html#basic-types"
@@ -71,7 +75,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   @special_form_strings for {f, a} <- special_form_exports, do: "#{f}/#{a}"
 
   @doc """
-  Compiles information used during autolinks.
+  Compiles information used during autolinking.
   """
   def compile(modules, extension, extra_lib_dirs) do
     aliases = Enum.map(modules, & &1.module)
@@ -112,25 +116,21 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   end
 
   defp project_doc(bin, module_id, locals, compiled) when is_binary(bin) do
-    %{
-      aliases: aliases,
-      docs_refs: docs_refs,
-      extension: extension,
-      lib_dirs: lib_dirs,
-      modules_refs: modules_refs
-    } = compiled
+    options =
+      compiled
+      |> Map.merge(%{
+        module_id: module_id,
+        locals: locals
+      })
+      |> Map.to_list()
 
-    bin
-    |> locals(locals, aliases, extension, lib_dirs)
-    |> elixir_functions(docs_refs, extension, lib_dirs)
-    |> elixir_modules(modules_refs, module_id, extension, lib_dirs)
-    |> erlang_functions()
+    link_everything(bin, options)
   end
 
   @doc """
-  Autolinks all modules nodes.
+  Autolinks all module nodes.
 
-  This is the main API to autolink any modules nodes.
+  This is the main API to autolink any module nodes.
   """
   def all(modules, compiled) do
     opts = [timeout: :infinity]
@@ -311,6 +311,20 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     |> IO.iodata_to_binary()
   end
 
+  @kinds [:module, :function]
+  @languages [:elixir, :erlang]
+  @link_types [:custom, :normal]
+
+  @regexes (for link_type <- @link_types,
+                language <- @languages,
+                kind <- @kinds do
+              %{
+                kind: kind,
+                language: language,
+                link_type: link_type
+              }
+            end)
+
   @doc """
   Helper function for autolinking locals.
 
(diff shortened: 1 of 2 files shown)


lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -370,112 +378,308 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
+  # - `:locals` - A list of local functions
+  # - `:module_id` - Module of the current doc. Default value is `nil`
+  # - `:modules_refs` - List of modules available
+  #
+  # Internal options:
+  # - `:preprocess?` - `true` or `false`. Do preprocessing and postprocessing, such as replacing backticks
+  #                     with a token
+  @doc false
+  @spec link(String.t(), language, kind, keyword) :: String.t()
+  def link(string, language, kind, options) do
```
josevalim: Why is this function public?

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -370,112 +378,308 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
+    options = Keyword.put_new(options, :preprocess?, true)
+
+    string
+    |> link_process(:preprocess, options[:preprocess?])
+    |> link(language, kind, :custom, options)
+    |> link(language, kind, :normal, options)
+    |> link_process(:postprocess, options[:preprocess?])
```
josevalim: Wrong option here. It seems we do not have tests for this then?

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -370,112 +378,308 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
+
+  # used temporarily for debugging
+  defp output(term, module_id, print_only_module_list) do
+    if module_id in print_only_module_list do
+      IO.puts(inspect(term))
+      IO.puts("****************")
+    end
+
+    term
+  end
```
josevalim: Please remove debugging code.

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -370,112 +378,308 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
+
+  @doc false
+  # Replaces all backticks inside the text of custom links with @backtick_token.
```
josevalim: Did we do pre-processing before? Why do we need it now?

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -370,112 +378,308 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
+  @doc false
+  # Reverts the changes done by `preprocess/1`.
+  def postprocess(string) do
+    String.replace(string, :binary.compile_pattern(@backtick_token), "`")
+  end
+
+  @doc false
+  # The heart of the autolinking logic
+  @spec replacement(String.t(), language, kind, String.t(), keyword) :: String.t()
+  def replacement(string, language, kind, match, text \\ nil, options) do
```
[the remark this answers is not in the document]
josevalim: Why is this function public too?

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -112,25 +116,21 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
-      lib_dirs: lib_dirs,
-      modules_refs: modules_refs
-    } = compiled
+    options =
+      compiled
+      |> Map.merge(%{
+        module_id: module_id,
+        locals: locals
+      })
+      |> Map.to_list()
```
josevalim: I don't like that we are converting structured list into unstructured one. My suggestion is to keep using a map internally everywhere.

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -370,112 +378,308 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
+  """
+  def erlang_functions(string) when is_binary(string) do
+    link(string, :erlang, :function, [])
+  end
+
+  defp replace_fun(language, kind, link_type, options) do
+    case link_type do
+      :normal ->
+        fn all, match ->
+          replacement(all, language, kind, match, options)
```
josevalim: I think we should merge `replace_fun` and `replacement` such that `replace_fun` already returns a function considering the `link_type`, `language` and `kind`. :)

josevalim:
Thanks @eksperimental! I have reviewed the code and now I can see with more clarity the improvements it brings. However, I still believe we have some extra work to do. In particular, I think the `replacement` function should be split and be concerned with at least `erlang` or `elixir`, instead of a large function that worries about all scenarios.

lib/ex_doc/formatter/html/autolink.ex
eksperimental: We didn't. But it fixes the bug of linking inside the custom text. Before it was only working if either the opening backtick is touching the opening bracket, or the closing backtick is touching the closing bracket. 

This wouldn't pass before IIRC.
```elixir
      assert Autolink.erlang_functions("[the `:erlang.apply/2` function](other.html)") ==
               "[the `:erlang.apply/2` function](other.html)"
```

So this behaviour was inconsistent. I think it worked for local and Elixir functions and it didn't work for the rest

lib/ex_doc/formatter/html/autolink.ex
eksperimental: Sorry, what is wrong here?

lib/ex_doc/formatter/html/autolink.ex
eksperimental: Sorry, but after splitting `replacement` into two functions clauses I cannot see how to merge it with `replace_fun`

I ended up having something like this

```elixir
  defp replace_fun(language, kind, :normal, options) do
    fn all, match ->
      replacement(all, language, kind, match, options)
    end
  end

  defp replace_fun(language, kind, :custom, options) do
    fn all, text, match ->
      replacement(all, language, kind, match, text, options)
    end
  end

  # The heart of the autolinking logic
  defp replacement(string, :erlang, kind, match, text \\ nil, options) do
    lib_dirs = Keyword.get(options, :lib_dirs, default_lib_dirs(language))
    
    pmfa = {prefix, module, function, arity} = split_function(match)
    text = text || default_text(language, kind, match, pmfa)

    if doc = module_docs(:erlang, module, lib_dirs) do
      case kind do
        :module ->
          "[#{text}](#{doc}#{module}.html)"

        :function ->
          "[#{text}](#{doc}#{module}.html##{function}-#{arity})"
      end
    else
      string
    end
  end

  defp replacement(string, :elixir, kind, match, text \\ nil, options) do
    aliases = Map.get(options, :aliases, [])
    docs_refs = Map.get(options, :docs_refs, [])
    extension = Map.get(options, :extension, ".html")
    lib_dirs = Map.get(options, :lib_dirs, default_lib_dirs(language))
    locals = Map.get(options, :locals, [])
    module_id = Map.get(options, :module_id, nil)
    modules_refs = Map.get(options, :modules_refs, [])

    pmfa = {prefix, module, function, arity} = split_function(match)
    text = text || default_text(language, kind, match, pmfa)

    elixir_docs = get_elixir_docs(aliases, lib_dirs)

    case kind do
      :module ->
        cond do
          match == module_id ->
            "[`#{match}`](#{match}#{extension}#content)"

          match in modules_refs ->
            "[`#{match}`](#{match}#{extension})"

          doc = module_docs(:elixir, match, lib_dirs) ->
            "[`#{match}`](#{doc}#{match}.html)"

          true ->
            string
        end

      :function ->
        cond do
          match in locals ->
            "[#{text}](##{prefix}#{enc_h(function)}/#{arity})"

          match in docs_refs ->
            "[#{text}](#{module}#{extension}##{prefix}#{enc_h(function)}/#{arity})"

          match in @basic_type_strings ->
            "[#{text}](#{elixir_docs}#{@basic_types_page})"

          match in @built_in_type_strings ->
            "[#{text}](#{elixir_docs}#{@built_in_types_page})"

          match in @kernel_function_strings ->
            "[#{text}](#{elixir_docs}Kernel#{extension}##{prefix}#{enc_h(function)}/#{arity})"

          match in @special_form_strings ->
            "[#{text}](#{elixir_docs}Kernel.SpecialForms#{extension}##{prefix}#{
              enc_h(function)
            }/#{arity})"

          doc = module_docs(:elixir, module, lib_dirs) ->
            "[#{text}](#{doc}#{module}.html##{prefix}#{enc_h(function)}/#{arity})"

          true ->
            string
        end
      end
    end
```

lib/ex_doc/formatter/html/autolink.ex
eksperimental: done

lib/ex_doc/formatter/html/autolink.ex
eksperimental: forgot to make it private

lib/ex_doc/formatter/html/autolink.ex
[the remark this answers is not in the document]
eksperimental: sorry I forgot to make it private

lib/ex_doc/formatter/html/autolink.ex
eksperimental: fixed

eksperimental:
@josevalim thank you for reviewing my code. I have implemented all the changes you suggested, except [here](https://github.com/elixir-lang/ex_doc/pull/896#discussion_r219462829) and [here](https://github.com/elixir-lang/ex_doc/pull/896#discussion_r219460519)

josevalim:

eksperimental:
The tests for the preprocess/postprocess were there because I found bugs in my previous code. 
Maybe I can conver those tests that were removed to use project_doc/4 to make sure we are dealing with backtick replacement correctly.

I will go through your commit whenever I getg some free time and get back if I have any comments.
thank you

wojtekmach:
I just noticed that besides refactoring, we've introduced here a new feature: auto linking Erlang modules. It can introduce false positives, see: `:file`:

screenshot 2018-10-28 at 00 37 03

probably not a deal breaker but can be a bit jarring.

eksperimental:
@wojtekmach great find!
I think we should revert it, or at least partially revert it, as it is going to be a big annoyance.
By partially I mean, we should remove this features from normal links and allow it only when called explicitly in custom links. WDYGT?

josevalim:
@eksperimental sweet. what do you have in mind? instead of a for comprehensions that builds `@regexes`, should we just list all combinations explicitly? Also, I think we should not support normal nor custom links for erlang modules. That would be a bit inconsistent.

josevalim:
Pushed in 75d32dbc8f17ae71655e12558f939da251c84304.

wojtekmach:
Great, thanks everyone!

PR #900, opened 2018-09-16
updated makeup_elixir version

tmbb:
The new version of makeup_elixir supports correct rendering of PIDs (`#PID<123.456.789>`) and adds correct support for iex prompts inside strings, which can happen in IEx sessions such as this:

```
iex> a = "
...> line1
...> line2
...> "
```


mix.exs
```diff
@@ -28,7 +28,7 @@ defmodule ExDoc.Mixfile do
   defp deps do
     [
       {:earmark, "~> 1.2"},
-      {:makeup_elixir, "~> 0.7"},
+      {:makeup_elixir, "~> 0.10.0"},
```
josevalim: Are you sure you want to depend on "0.10.0" and not on "0.10"?

tmbb: I can depend on "0.10", sure.

josevalim:

tmbb:
I don't know if this is the right place for this request, but I think you should make a patch release and change the script that builds the Elixir docs so that it uses this new release, because the new Elixir lexer actually fixes some highlighting bugs in the Elixir docs.

josevalim:
Elixir master always builds from ex_doc master. Stable updates only in a new release.

Issue #834, opened 2018-02-20
Deeply nested modules do not fit in sidebar


teamon:
Google has recently released multiple libraries for their API.
They contain generated elixir structs for every protobuf model which leads to very deeply nested module names that do not fit in the sidebar. You can see example here [https://hexdocs.pm/google_api_pub_sub/api-reference.html#modules](https://hexdocs.pm/google_api_pub_sub/api-reference.html#modules)

What do you think about (optionally) putting modules list in a tree?

![cloudapp annotation 2018-02-20 at 11 23 00 am png](https://user-images.githubusercontent.com/8083/36418873-800ca714-1630-11e8-87b3-f8a05744d9c5.png)![cloudapp annotation 2018-02-20 at 11 22 28 am png](https://user-images.githubusercontent.com/8083/36418879-827ecd74-1630-11e8-9c6a-8c1a2c633615.png)
> (left - current state, right - my `&nbsp; in inspector` attempt to present the idea)

DarkMarmot:
I could also use this -- and the ability to toggle/hide groups would be an awesome addition.

mhanberg:
I have also run into this using the [Timex](https://hexdocs.pm/timex/getting-started.html) library. 

My proposal would to make the sidebar re-sizable and/or having the sidebar automatically size to the width of the longest module name.

tmbb:
On one hand, this solves the pragmatic issue of not enough space for module names in the sidebar.
On the other hand, this obscures the fact that module names are just atoms. There aren't nested modules in Elixir!

In any case, that's a problem that needs to be solved. I suggest something like this:

```
GoogleApi.PubSub.V1.{
  Api.Projects,
  Connection,
  Deserializer,
  Model.{
    AcknowledgeRequest,
    Binding,
    Empty
  }
}
```

It uses the syntax for aliasing modules, which is familiar, and although it takes up a little more vertical space, I think the resemblance to source code is worth it.

OvermindDL1:
> It uses the syntax for aliasing modules, which is familiar, and although it takes up a little more vertical space, I think the resemblance to source code is worth it.

Ooo, big +1 on this, I like this idea quite a bit!

josevalim:
> It uses the syntax for aliasing modules, which is familiar, and although it takes up a little more vertical space, I think the resemblance to source code is worth it.

It looks great in code but it is hard to do visually. Mockups are appreciated. :)

[...]

PR #905, opened 2018-10-15
Nest module aliases

davidsulc:
(Tests and documentation still need to be written, this is just for preliminary feedback.)

A basic implementation for https://github.com/elixir-lang/ex_doc/issues/834

It leverages `groups_for_modules` to provide the usual UI presentation, and to leverage the current ordering ability. As suggested by José, a `nest_module_aliases` key was added to the docs config.

This new value is an array of alias configs, each of which is one of:
* `ModuleName` (equivalent to `{ModuleName, as: ModuleName}`)
* `{Super.Duper.Really.Long.ModuleName`, as: `Some.ModuleName`}
* `{Super.Duper.Really.Long.ModuleName`, as: "My module"}

The `as:` values are used as group names (as if they had been provided to `groups_for_modules`) and are removed from the module's displayed title.

Hopefully, this will be familiar to language users as it combine the `as:` syntax of module aliases with the multiple value formats as can be used when specifying child configs in supervisors.

The objective for this first implementation is to provide enough flexibility while keeping the code changes simple. It is to be noted that the aliasing is very naive: the first match is kept, without attempting to find the longest prefix possible.

With these modifications, nesting could be specified like this (adapted from ExDoc's `mix.exs`):

```
      groups_for_modules: [
        "Markdown": [],
        "Formatter API": [
          ExDoc.Config,
          ExDoc.FunctionNode,
          ExDoc.ModuleNode,
          ExDoc.TypeNode
        ],
        "ExDoc.Formatter": []
      ],
      nest_module_aliases: [
        {ExDoc.Markdown, as: Markdown},
        #{ExDoc.Markdown, as: "Foo bar"},
        ExDoc.Formatter
      ]
```
Adding empty groups matching `as:` values (or the module name, if absent) in `nest_modules_aliases` allows controlling the ordering of the blocks of nested modules.

Here's the generated result:

nested_ex_doc


davidsulc:
Sorry, meant to open this in my forked repo. I will open a PR here once it is finalized (with tests and docs)

PR #907, opened 2018-10-18
Collapse nested module names

davidsulc:
Here's an attempt at addressing long module names in the sidebar. [Prior](https://github.com/elixir-lang/ex_doc/issues/834) [discussion](https://github.com/elixir-lang/ex_doc/pull/906) has shown that there is no consensus on the best approach to address the issue of long module names getting clipped in the sidebar, but I believe this proposal is the "least bad" solution (and I'll cover why below).

## Context

As you evaluate this, please bear in mind the relevant [issue](https://github.com/elixir-lang/ex_doc/issues/834) was opened in February this year and hasn't seen activity since July, so this is likely an infrequent issue within the community (i.e. if this PR gets merged it hopefully won't besmirch the docs you use daily). In addition, don't forget this option is opt-in: if long module names aren't getting clipped within the sidebar, there's no reason to use it.

Before getting in the thick of things, here's what we're talking about:

timex_before

As you can see, modules with long names have their right hand portion cut off, which is especially unfortunate as that is the most specific portion of the namespace.

This PR proposes to allow users to specify a set of prefixes that would get "collapsed" in order to favor showing the rightmost namespace sections  (to the detriment of the leftmost sections, which are typically of less interest). Here's what the documentation for the same project would look like:

timex_after

This result was obtained by passing in the following option:

```
     collapse_nested_module_names: [
       Timex.Format,
       Timex.Format.Duration.Formatters,
       Timex.Parse,
       Timex.Parse.DateTime,
       Timex.Parse.Duration.Parsers,
       Timex.Parse.Timezones.Posix,
     ],
```

## Why I like it

The difference in font opacity signals that the prefix is "different" than the uncollapsed portion. Even if the user doesn't realize that the prefix is collapsed from the context, the affordance indicated by the font difference will typically lead them to hover the module name, whereupon they will see the full title name displayed.

The darker font of the collapsed prefixes gives a visual indentation communicating the concept of module nesting.

Keeps the straightforward nesting currently used by ExDoc: modules are all on a single level (similar to how they exist within Elixir) with indentation reserved for things that are actually contained within the module (functions, types, etc.).

Of course, this solution isn't perfect either: in the example above, `T.F.DateTime.Formatter` gives the visual impression that it is nested with `Timex.Duration`. This is a trade-off made by the documentation writer, as it enabled the subsequent formatters to be visible within the sidebar. Ultimately, this option puts the onus on the developer to make the trade-offs that best suits their specific use case and makes their documentation most user-friendly. In my opinion, that's as it should be: the developer is precisely who should be empowered to make those trade-offs.

### Grouping

My [initial attempt](https://github.com/elixir-lang/ex_doc/pull/906) sought to leverage the existing `:groups_for_modules` functionality. Having given this more thought, I believe this approach was misguided. There has to be a visual cue that something "unusual" is being done to accommodate excessively long module names so the user can process the information appropriately.

I don't think there's a really good way to do this without needlessly increasing the complexity of front-end documentation code (either by making the current group mechanism more complex, or introducing a new one). Either you have a bold/uppercase/non-clickable title meaning the "root" module has to be repeated below the title, or you need a group title that displays like a normal module, which means that there now needs to be some visual differentiation for the "nested" modules.

In addition, sorting would have to be handled explicitly to ensure "groups by nesting" are displayed in the proper place (i.e. alphabetically, where the user expects them).

### Nesting with indentation

Another idea was nesting modules in the sidebar with (e.g.) nested `ul` lists. I think this idea is also problematic as it just punts the problem down the line: nested modules will eventually indent far enough that the functions they contain will be clipped in the sidebar and we'd be back at square one.

In addition, I think indenting nested modules sends the wrong message, namely that nested modules are "within" parent modules (i.e. there's a module hierarchy). This is already a source of confusion for newcomers to the language, and I don't think the documentation system should add to the confusion.

Finally, using nesting and indentation for modules in the sidebar muddies the waters: currently, the sidebar docs has a nice semantic where modules are all on the same level (i.e. "equal"), and clicking on a module will expand to display an indented list of things that actually exist within the module. This is a really nice analogy, and I think it's worth the effort to keep it that way if we can.

## Conclusion

Now that I've said my piece, please feel free to provide feedback, counter points, etc. As you'll probably have noticed, this PR isn't ready for merging yet: it contains no tests. This is because I'd rather wait until the dust settles on what the community thinks/wants so effort isn't invested on things that aren't a good fit.


axelson:
I'd like to say that I would like to see a few tweaks to this. In particular I do not like the tradeoff that you mentioned "`T.F.DateTime.Formatter` gives the visual impression that it is nested with `Timex.Duration`". My biggest issue is that `Timex.Format` does not display anywhere on the page without hovering with the mouse (how would this work on mobile?)

I do agree that it isn't important to not emphasize nesting too much because strictly speaking the modules are not nested (there is no namespacing, just a common prefix). So I have two potential suggestions.

_Timex.Format_
_T.F_.DateTime.Formatters.Default
_T.F_.DateTime.Formatters.Relative
_T.F_.DateTime.Formatters.Strftime
_Timex.Format.Duration.Formatter_
_T.F.D.F_.Default
_T.F.D.F_.Humanized
Timex.Gettext

Where everything in italics is in the lower contrast color that you're using in the screenshot (which I quite like). I think by keeping each row the same font size and weight that it doesn't appear that we are creating namespaces or a hierarchy. Also by having `Timex.Format` and `T.F` in the same color we indicate a link between them. In addition for desktop users we can still show `Timex.Format` when you hover over `T.F`.

Also, perhaps when you click/tap on the `Timex.Format` text it would use javascript to expand `T.F` and then collapse it when you tap on it again. This would also indicate the difference between an actual `Timex.Format` module and a `Timex.Format` prefix (in addition to the visual color difference)

davidsulc:
@josevalim @teamon @whatyouhide @fertapric @OvermindDL1 @tmbb You've given your opinion on the previous iteration, please feel free to do so again :-)

davidsulc:
@josevalim I think it might be better to keep the concepts orthogonal: groups are for "put this cohesive unit in its own place in the docs to draw attention to it", while collapsing is for "this module name is too long to display properly in the sidebar, and we're cutting off the part with the more specific info".

As ExDoc beautifully mentions (https://hexdocs.pm/elixir/writing-documentation.html), documentation in Elixir is a first-class citizen (and thanks for that!). I feel this has led the community to put more effort in usable/user-friendly documentation which is a great dynamic to have. A key component to this, I think, is making writing readable docs as frictionless for developers as possible (i.e. prevent them from having to fight the tool).

## Grouping
Just because you want to shorten certain module names in the sidebar (so they remain legible) doesn't necessarily imply you want to extract them into a group. After all, the way groups are displayed on their own implicitly suggests they are somehow special and/or form a cohesive unit. I think this should be driven by an explicit request from the documentation writer, not just automatically based on "these are long and have a common prefix". In other words, I'm worried of frustrating documentation writers who just want to shorten the display of their module names without necessarily having a group forced on them (which they might not want or agree to).

Reusing the group mechanism wholesale would also mean that if a developer wants to have a section with a name matching a module prefix (e.g. [Plug](https://hexdocs.pm/plug/readme.html)), they might have modules appear in that group (due to collapsing) that weren't part of the explicit list they provided which would go against the principle of least surprise. Granted, I can't think of a concrete example where this would be an issue (i.e. collapse module names with a common prefix, but only include a subset in an explicit group).

Also how should ExDoc behave when the documentation writer defines a group, but some modules in there are too long to display? Extracting a sub-group from that group would also be somewhat counter-intuitive to me (especially because the extracted group might be displayed far away from its parent due to ordering).

Finally, there are cases where module names are too long due to a prefix that is common to ALL modules (e.g. [google_api_storage ](https://hexdocs.pm/google_api_storage/api-reference.html) or [phoenix_gen_socket_client](https://hexdocs.pm/phoenix_gen_socket_client/api-reference.html). Putting these in a group when there are no modules that exist outside of that one single group would be a bit strange. Of course, those are edge cases but this entire PR is intended for edge cases :p

## Inserting a non-interactive node
I think inserting an "empty module node" in the list (as suggested by @axelson) with the full name of the collapsed prefix (in cases where such a module doesn't exist in the project) would adequately address documentation clarity concerns while not introducing other issues. It also avoids surprises for the documentation writer, as although a new non-interactive module name may be added, the order/structure of the documentation remains as explicitly defined. This would also allow collapsing module names within module groups, if desired (while keeping them in their group).

It also helps with "visual flow" as the font would be darker, there would be clear visual markers in the sidebar to quickly determine which modules are "grouped" under which prefix.

In addition, as this approach is minimally intrusive with respect to the intentions of the documentation writer (with respect to order/structure/flow of the documentation), it would be possible down the road to introduce an opt-in flag of "collapse all modules that are too long". Please note that I'm just pointing out that we could, not necessarily that we should ;-)

## Next iteration
Unless people are vehemently against it, I will move forward with adding the "non-interactive node" idea to this PR so we can discuss the pros/cons further.

And I really hope I'm not coming across as "that guy who asks for feedback but just wants to hear his opinion in a different voice" :p

teamon:
Wow, that's a lot of effort put in here @davidsulc <3 <3 <3

Personally, I'm not a fan of `X.Y.Z.Module` shortening, as they give as much information as `... Module` would give. As my original issue was with google libs, for that case the solution could be:

```
GoogleApi.Storage.V1.Api
  BucketAccessControls
  Buckets
  Channels
  DefaultObjectAccessControls
  Notifications
  ObjectAccessControls
  Objects
  Projects
GoogleApi.Storage.V1.Connection
GoogleApi.Storage.V1.Deserializer
GoogleApi.Storage.V1.Model
  Bucket
  BucketAccessControl
  BucketAccessControlProjectTeam
  BucketAccessControls
  BucketBilling
  BucketCors
  BucketEncryption
  BucketLifecycle
  BucketLifecycleAction
  BucketLifecycleCondition
  BucketLifecycleRule
  BucketLogging
  BucketOwner
  BucketRetentionPolicy
  BucketVersioning
  BucketWebsite
  Buckets
  Channel
  ComposeRequest
  ComposeRequestObjectPreconditions
  ComposeRequestSourceObjects
  Notification
  Notifications
  Object
  ObjectAccessControl
  ObjectAccessControls
  ObjectCustomerEncryption
  ObjectOwner
  Objects
  Policy
  PolicyBindings
  RewriteResponse
  ServiceAccount
  TestIamPermissionsResponse
GoogleApi.Storage.V1.RequestBuilder
```

achieved by this config:

```
nest_modules_with_common_prefix: [
  GoogleApi.Storage.V1.Api, 
  GoogleApi.Storage.V1.Model
]
```

This would be easy enough to include in autogenerated projects (like google-* stuff).
So, 👍 from me for "non-interactive node"

davidsulc:
> I'm not a fan of X.Y.Z.Module shortening, as they give as much information as ... Module would give

While that's true in the Google lib case (since the hierarchy is relatively flat, there's no need for collapsing beyond a single nesting level), X.Y.Z. gives your more info than just ... because you also get info about how deep the current nesting level is by looking at the line itself.

As mentioned in the first comment here, I think indenting won't scale to cases where multiple levels must be nested.

OvermindDL1:
> Personally, I'm not a fan of X.Y.Z.Module shortening, as they give as much information as ... Module would give.

Likewise, plus there can be ambiguity.

I entirely agree with @teamon here.  You could even flatten multiple levels:
```text
Module1
Module2
* Sub1
* Sub2
* Sub3
  * SubSub1
  * SubSub2
```

If you really want to go multiple levels you can like this, or you can flatten the levels like:

```text
Module1
Module2
* Sub1
* Sub2
Modul2.Sub3
* SubSub1
* SubSub2
```

Both of the above should be supported based on how they define the grouping should occur.  Regardless a tiny bit of indentation is *hugely* useful to differentiate such groupings, the abbreviations do not do that.

josevalim:
> I think inserting an "empty module node" in the list (as suggested by @axelson) with the full name of the collapsed prefix (in cases where such a module doesn't exist in the project) would adequately address documentation clarity concerns while not introducing other issues.

I disagree. People would think it is clickable and it will be confusing, unless there is a visual cue. And if we are going to support both groups and nesting, we need to be careful as to how they will work together.

> If you really want to go multiple levels you can like this, or you can flatten the levels like:

Yes, please. There is no chance we will support multiple levels. It just makes navigation deeper without larger benefits.

OvermindDL1:
> Yes, please. There is no chance we will support multiple levels. It just makes navigation deeper without larger benefits.

I don't mean expandable levels, they all 'appear' top level, just indented as appropriate by a tiny amount.  Unless someone starts doing it 12 layers deep then it shouldn't have issues going off the edge or so, literally just indent a tiny amount, a single character's width is fine.

fertapric:
What do you guys think if we keep the discussion in one single place? As @davidsulc mentioned, this discussion is scattered into an [issue](https://github.com/elixir-lang/ex_doc/issues/834), a closed [PR](https://github.com/elixir-lang/ex_doc/pull/906) and this PR.

Ironically, I would create an entry in the [Elixir Forum](https://elixirforum.com/) with a summary of what have been discussed, so we can collect more feedback. I think continuing the conversation there will help to keep the discussion organized (since I don't see we will reach an agreement in the next comments) and out of the issue tracker (or pull requests).

Everybody ok? 😬

davidsulc:
@fertapric If you still wish to get larger community feedback, please wait until I've been able to update this PR with a new iteration. I think I've found a nice solution and I'd like to avoid that what started as a way to address a very specific edge case spirals out of control into a "design by committee" situation. If more voices are going to add to the discussion, I would like them to at least do so against my next iteration which I hope (naively, perhaps :sweat_smile: ) will please most folks.

davidsulc:
So, I've come full circle... I initially though the collapsing approach would nicely address the problem once context headers were added to the sidebar, and would be able to handle long prefix names by recursion. Having experimented with it a bit, it turned out unsatisfactory for reasons that probably aren't of interest here. I also explored displaying truncated context nodes, but that didn't work well either (from a usability point of view).

Here's what I've settled on as the best approach: introduce a `:group_modules_by_nesting` as originally discussed. If a module is prefixed by one of those values (longest match wins), it is truncated and displayed "nested" within the prefix:
timex_example

This option can be used in conjuction with `:groups_for_modules`:
timex_group

At the time of this writing, GitHub is still having issues but you should be able to see the resulting docs here:

* [timex](https://davidsulc.github.io/timex_docs/) ([repo](https://github.com/davidsulc/timex_docs))
* [timex showing nesting within a group](https://davidsulc.github.io/timex_docs_with_group/) ([repo](https://github.com/davidsulc/timex_docs_with_group))
* [Google storage](https://davidsulc.github.io/google_api_storage_docs/) ([repo](https://github.com/davidsulc/google_api_storage_docs))

These were generated with the following relevant options:

```
  # Timex
     group_modules_by_nesting: [
       Timex.Format,
       Timex.Format.Duration,
       Timex.Parse.DateTime,
       Timex.Parse.Duration,
       Timex.Parse.Timezones,
       Timex.Parse.ZoneInfo
     ],
    # this was only added for the "nesting within a group" demo
     groups_for_modules: [
       "Default formatters": [
         Timex.Format.DateTime.Formatters.Default,
         Timex.Format.Duration.Formatters.Default,
       ]
     ],

```

```
# Google Storage
        group_modules_by_nesting: [
          GoogleApi.Storage.V1,
          GoogleApi.Storage.V1.Api,
          GoogleApi.Storage.V1.Model,
        ]
```

I will edit this later to tag people for feedback once GitHub has made the "website" versions of the docs available online so it's easier for people to see the end result without have to clone repos, etc.

## Next steps?

Assuming this revision finds some sort of consensus, I still need to add tests and get ebert bot's blessing: it seems the bot didn't run due to GitHub availability issues. Also: I've added the commits to this branch/PR for discussion/history purposes, let me know if you'd rather I create a new branch with a cleaner commit history (i.e. no detour through the collapsing experiment).

OvermindDL1:
@davidsulc Yes!  Precisely!  That's what I envisioned!  I like that!  :-)

teamon:
Niiiiice!

Personally I think we should merge it and see how is it used in other projects. 
One change I’d make is to group modules nested with the same prefix together (the GoogleApi.Storage.V1 case)

davidsulc:
> group modules nested with the same prefix together

I think that would be a step too far. Modules in Elixir are basically "atoms that contain dots" and I think that suddenly changing their order in the sidebar based on common prefixes would be unexpected. I think the semantic for `:group_modules_by_nesting` should remain "shortening the title displayed in the sidebar".

If modules are to be grouped, the developer should explicitly use the `:groups_for_modules` option.

tmbb:
I like this. There should be an option to change the font size, though, so that even when you use all possile shortening tricks you still have some margin to make it fit.

davidsulc:
> option to change the font size

I don't think we should enable variable font sizing: it adds complexity to both the ExDoc code and the config the user must provide for very little benefit. If a smaller (still readable) font would make it fit, it's probably only a few letters getting cut off which isn't a big deal. Variable font sizes would also mean that docs would no longer have a consistent look & feel.

I think the only thing I'd add (in a separate PR of course) would be a boolean flag to collapse the project name. The top-level project name is one you may have little control of: e.g. including the company/service/hex package that you're providing a wrapper for or extending (i.e. have a name that makes sense for discoverability). This name could be quite long if the "base name" already is, and on top of that more often than not, it prefixes every single module in the package. So not only does it take up room in the sidebar, it causes a sort of [semantic saturation](https://en.wikipedia.org/wiki/Semantic_satiation) and provides no actual information (you already know which package you're dealing with then looking through its docs). Basically a simple thing like `collapse_project_name: true`.

lib/ex_doc/formatter/html.ex
```diff
@@ -79,6 +83,113 @@ defmodule ExDoc.Formatter.HTML do
     end
   end
 
+  defp group_by_nesting([], _), do: []
```
josevalim: What if we move this to the `group_matcher`? This way we can unit test it too.

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -201,6 +195,24 @@ defmodule ExDoc.Formatter.HTML.Templates do
     ~s/{"id":"#{id}","anchor":"#{URI.encode(anchor)}"}/
   end
 
+  defp sidebar_items_json_string(module_node, items) do
```
josevalim: Let's work with iodata instead of keyword list, as it is quite more efficient. Comprehensions should help too. Something like this:

```elixir
    json_attrs =
      for key <- [:id, :title, :title_truncated],
          value = Map.get(module_node, key),
          do: [json_kv(key, inspect(value)), ?,]

    json_attrs = [json_attrs | json_kv(:group, Atom.to_string(module_node.group))]

    json_attrs =
      case items do
        "" -> json_attrs
        items -> [json_attrs, ?, | json_kv(:nodeGroups, "[#{items}]")]
      end

    IO.iodata_to_binary [?{, json_attrs, ?}]
```

lib/ex_doc/nodes.ex
```diff
@@ -20,6 +21,7 @@ defmodule ExDoc.ModuleNode do
   @type t :: %__MODULE__{
           id: nil | String.t(),
           title: nil | String.t(),
+          title_truncated: nil | String.t(),
```
josevalim: What about `nested_title`?

mix.exs
```diff
@@ -90,6 +90,9 @@ defmodule ExDoc.Mixfile do
           ExDoc.ModuleNode,
           ExDoc.TypeNode
         ]
+      ],
+      group_modules_by_nesting: [
+        ExDoc.Formatter
```
josevalim: I don't think we need this for ExDoc itself, as our module names are not that long. :)

assets/js/templates/sidebar-items.handlebars
```diff
@@ -1,50 +1,65 @@
...
-    {{else}}
-      <a href="{{node.id}}.html" title="{{node.title}}" class="expand">{{node.title}}</a>
+      {{! Nodes added purely for nesting purposes (i.e. don't exist in the source) have no id }}
+      {{#if node.id}}
+        {{! When visiting a module page, the link to this module page
+        in the menu should not link to a new page, instead should link
+        to the top of the page itself. }}
+        <a href="{{node.id}}.html" title="{{node.title}}" {{#hasContent node}}class="expand"{{/hasContent}}>
+          {{#if node.title_truncated}}
+            <span class="nested"></span>{{node.title_truncated}}
```
josevalim: I wonder if we should do this as a CSS class instead?

josevalim: The reason why I think this would be better as a class is because we need to indent all of the content inside the link. In the image below, notice how the inner contents are not aligned, which is weird when expanding multiple groups: 

screen shot 2018-10-23 at 10 15 07

mix.exs
davidsulc: Oops, definitely not: it was a quick way to test during development and snuck through when committing :(

josevalim:
Thanks @davidsulc ! This looks great, I have added some comments.

The only nitpick I have is to remove the italic. I believe the italic puts the emphasis in the wrong place. I would maybe even consider removing the italic and the indentation mark, to keep the overall sidebar clean:

screen shot 2018-10-23 at 10 14 42

I am not a designer though :) but it feels the less markup aligns better with the current layout.

lib/ex_doc/formatter/html.ex
davidsulc: That makes sense. I put it here as it was mainly and HTML-related concern, but conceptually it fits within the `GroupMatcher`, especially as it enables unit testing.

lib/ex_doc/nodes.ex
davidsulc: I used `title_...` as the value "goes with" `title`, but I see how that might not be helpful. I'll change it 👍

lib/ex_doc/formatter/html/templates.ex
josevalim: Oh, I forgot to include it. It is this tiny function: `defp json_kv(key, val), do: [?", Atom.to_string(key), ?", ?:, value]`

davidsulc:
@josevalim The markup can definitely be removed/reduced. The reason I added the arrow was because in cases where a LOT of modules share the same nesting level (e.g. the [Google case](https://davidsulc.github.io/google_api_storage_docs/api-reference.html)), it's not immediately obvious you're not seeing the full module name (i.e. normal vs nested module):

untitled

Relying on indentation only means that you can only know you're seeing a nested value if you can see the beginning of the nesting (otherwise it's hard to know whether you're in an indented level or not). But maybe I'm over-thinking that 🤔

josevalim:
@davidsulc I think the google case is fine because on the initial scan of the project, you know that everything is nested, especially when you see the short names themselves. I think the issue is that, for a long list, you don't know which contexet you are right now, but none of them solve it.

assets/js/templates/sidebar-items.handlebars
josevalim: Actually, I would suggest to add the `class="nested"` to the .li itself and not the link. This gives more options to style it later.

assets/js/templates/sidebar-items.handlebars
```diff
@@ -1,50 +1,65 @@
...
+                      {{/each}}
+                    </ul>
+                  </li>
+                {{/each}}
+              {{/isArray}}
+            </ul>
+          {{/hasContent}}
+      {{else}}
+        <span title="{{node.title}}" class="context" aria-hidden="true">
+          {{#if node.title_truncated}}
```
josevalim: This title can never be truncated, right? As we don't support nesting.

lib/ex_doc/formatter/html.ex
```diff
@@ -79,6 +83,113 @@ defmodule ExDoc.Formatter.HTML do
...
+
+    Enum.reduce(nodes, acc, &process_unnested_node/2)
+    |> Map.get(:processed_nodes)
+    |> Enum.reverse()
+  end
+
+  defp process_unnested_node(%ModuleNode{} = module_node, acc) do
+    module_node
+    |> maybe_truncate_node_title(acc.prefixes_to_nest)
+    |> maybe_insert_nesting_context_node(acc)
```
josevalim: I wonder if we should rather insert those nodes in the javascript side, like we do for groups. It is relatively easier to do it there. From the Elixir side, we will emit JS nodes with two information: the nesting and the nested_title. In the JS side, we check if the current node has a nesting, if so and it is different than the previous nesting, then we output the nesting header.

assets/js/templates/sidebar-items.handlebars
davidsulc: I believe you're right. I think this is an artifact from the recursive collapse experimentation, but I'll have to check to be sure.

assets/js/templates/sidebar-items.handlebars
davidsulc: Yes, that's what I understood from your original comment as that would be the way to get consistent indentation.

lib/ex_doc/formatter/html.ex
davidsulc: I'd have to check if we can ensure consistent results if we do it on the js side. In particular, the ["exceptions" section](https://davidsulc.github.io/timex_docs/Timex.Format.FormatError.html) gets nesting "for free" by doing the node insertion in the elixir code. I.e. it's completely transparent whether a node is a module, an exception, or whether it belongs to a group: the nesting will behave the same. I need to check whether there's centralized processing in the js code or not (haven't really dived into that yet).

josevalim: It is exactly the sidebar_items.handlebars that you have changed. Here is the code for the group: https://github.com/elixir-lang/ex_doc/pull/907/files#diff-d85f51765b385ca15bb7242d5936ac89R3

I propose something like very similar to that, in pseudo code it would be:

```
    {{#newNesting ../this node.nesting}}
      <li class="nesting">{{node.nesting}}</li>
    {{/newNesting}}
```

However, keep in mind that, if we start a new group and the first item of the group has a nesting, we need to emit the nesting too. So maybe you also need to pass the group as argument. Alternatively, when you check for newGroup and the group change, you just reset the parent "last nesting" value.

davidsulc:
@josevalim I've updated my PR, and I think I've addressed all your comments. I've also updated the resulting docs: [modified Timex](https://davidsulc.github.io/timex_docs_with_group/getting-started.html) and [Google storage](https://davidsulc.github.io/google_api_storage_docs/api-reference.html)

Let me know if there's anything else I can improve.

assets/js/templates/sidebar-items.handlebars
```diff
@@ -1,50 +1,58 @@
...
-    {{!-- When visiting a module page, the link to this module page
-    in the menu should not link to a new page, instead should link
-    to the top of the page itself. --}}
-    {{#isEmptyNode node}}
-      <a href="{{node.id}}.html" title="{{node.title}}">{{node.title}}</a>
-    {{else}}
-      <a href="{{node.id}}.html" title="{{node.title}}" class="expand">{{node.title}}</a>
+    {{#nestingContextChanged ../this node.nested_context}}
+      <li class="nesting-context" title="{{node.nested_context}}" aria-hidden="true">{{node.nested_context}}</li>
+    {{/nestingContextChanged}}
```
josevalim: Sorry for the super minor nitpick but can we please keep the `newGroup` and `nestingContextChange` function names in sync? We can either call both `groupChanged` and `nestingChanged` or `newGroup` and `newNesting`, I would just like to keep them in sync. :)

assets/js/templates/sidebar-items.handlebars
```diff
@@ -1,50 +1,58 @@
...
+  {{#each nodes as |node nodeId|}}
     {{#newGroup ../this node.group}}
       <li class="group">{{node.group}}</li>
     {{/newGroup}}
-    <li class="{{#isLocal node.id}}clicked open{{/isLocal}}">
 
-    {{!-- When visiting a module page, the link to this module page
-    in the menu should not link to a new page, instead should link
-    to the top of the page itself. --}}
-    {{#isEmptyNode node}}
```
josevalim: I believe you have accidentally removed this functionality in your PR. You can see this in Timex docs themselves, for the module `Timex.Parse.ZoneInfo.Parser.Zone`, it should never expand in the menu, it should directly go to the page because it only has a "Top".

josevalim: Nevermind. I see you just implemented it differently. :)

assets/js/templates/sidebar-items.handlebars
davidsulc: 👍  I'll update them to `...Changed`. I didn't name it `newNesting` because it's confusing: the nesting isn't necessarily "new" as there can be duplicates.

tmbb:
I think I like the marker before the nested modules

davidsulc:
@josevalim I've improved my PR with your suggestions

josevalim:

lib/mix/tasks/docs.ex
```diff
@@ -88,6 +88,15 @@ defmodule Mix.Tasks.Docs do
 
     * `:groups_for_extras`, `:groups_for_modules`, `:groups_for_functions` - See next sections
 
+    * `:group_modules_by_nesting` - A list of atoms (e.g. `Foo.Bar`) that will be truncated
```
fertapric: Minor nitpick: should we say here "A list of module names (e.g. `Foo.Bar`)"? The example says `Foo.Bar` instead of `:"Elixir.Foo.Bar"`

josevalim:
Given that `group_modules_by_nesting` no longer creates a group, I figured out it is best to rename the option to `nest_modules_by_prefix`. If anyone has a better name suggestion, I am all ears! :heart:

lib/mix/tasks/docs.ex
josevalim: Fixed.

Issue #909, opened 2018-10-25
Custom links to modules


wojtekmach:
Ref: https://github.com/elixir-lang/ex_doc/issues/679#issuecomment-385320830

```elixir
     assert project_doc("[`Registry docs`](`Registry`)", %{docs_refs: []}) ==
       "[`Registry docs`](https://hexdocs.pm/elixir/Registry.html)"
```

```
     Assertion with == failed
     code:  assert project_doc("[`Registry docs`](`Registry`)", %{docs_refs: []}) == "[`Registry docs`](https://hexdocs.pm/elixir/Registry.html)"
     left:  "[[`Registry`](https://hexdocs.pm/elixir/Registry.html)](https://hexdocs.pm/elixir/Registry.html)"
     right: "[`Registry docs`](https://hexdocs.pm/elixir/Registry.html)"
     stacktrace:
       test/ex_doc/formatter/html/autolink_test.exs:262: (test)
```

fertapric:
This kind of syntax doesn't work well with command line tools like IEx. This is how the given example would appear in IEx:

![screen shot 2018-10-25 at 08 50 13](https://user-images.githubusercontent.com/651203/47481281-5ffabe00-d833-11e8-95ae-fd90c4d23679.png)

And this is how it would appeared if we do not support this syntax:

![screen shot 2018-10-25 at 08 53 44](https://user-images.githubusercontent.com/651203/47481320-7ef95000-d833-11e8-9a16-e1ddf48c8344.png)

I would favor readability of the documentation in all formats 🙂

fertapric:
As a side note, I would have written it as:

```
`Registry` docs
the documentation for `Registry`
```

which is auto-linked by ExDoc and rendered in IEx as:

```
Registry (https://hexdocs.pm/elixir/Registry.html) docs
the documentation for Registry (https://hexdocs.pm/elixir/Registry.html)
```

wojtekmach:
not sure, I thought that IEx doesn't know about hexdocs.pm and can't generate these links.

But I think you're right, we probably don't need custom links for modules we should just write, I thought it could be an intermediate step to https://github.com/elixir-lang/ex_doc/issues/910 which I think would be nice.

fertapric:
@wojtekmach You are right. I edited my previous comment :)

eksperimental:
> This kind of syntax doesn't work well with command line tools like IEx. This is how the given example would appear in IEx:
> 
[...]

PR #911, opened 2018-10-25
Autolink mix tasks

### LINKED ISSUE #910: Autolink mix tasks
When https://github.com/elixir-lang/ex_doc/issues/909 is done we'd do:

```markdown
Run [`mix test`](`Mix.Tasks.Test`).
```

however we could create the same link with just writing:

```markdown
Run `mix test`.
```

Thoughts?

wojtekmach:
Closes #910

screenshot 2018-10-25 at 14 03 53

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -447,7 +447,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     end
   end
 
-  defp replace_fun(:mix_task, :elixir, _link_type, options) do
+  defp replace_fun(:mix_task, :elixir, :normal, options) do
     extension = options[:extension] || ".html"
     lib_dirs = options[:lib_dirs] || default_lib_dirs(:elixir)
     module_id = options[:module_id] || nil
@@ -754,7 +754,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
 
   defp re_kind_language(:mix_task, :elixir) do
     ~r{
-      mix\ ([a-z]+[a-z0-9\._]*)
+      mix\ ([a-z][a-z0-9\._]*)
     }x
   end
 
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -5,7 +5,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   import ExDoc.Formatter.HTML.Templates, only: [h: 1, enc_h: 1]
 
   @type language :: :elixir | :erlang | :markdown
-  @type kind :: :function | :module
+  @type kind :: :function | :module | :mix_task
   @type link_type :: :normal | :custom
 
   @backtick_token "<B706848484895T>"
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -106,7 +106,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
                 language <- [:elixir, :erlang],
                 kind <- [:module, :function] do
               {kind, language, link_type}
-            end)
+            end) ++ [{:mix_task, :elixir, :normal}]
 
   @doc """
   Autolinks any documentation in the project.
@@ -447,6 +447,37 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     end
   end
 
+  defp replace_fun(:mix_task, :elixir, _link_type, options) do
+    extension = options[:extension] || ".html"
+    lib_dirs = options[:lib_dirs] || default_lib_dirs(:elixir)
+    module_id = options[:module_id] || nil
+    modules_refs = options[:modules_refs] || []
+
+    fn all, text, "mix " <> task_name ->
+      task_module =
+        task_name
+        |> String.split(".")
+        |> Enum.map(&Macro.camelize/1)
+        |> Enum.join(".")
+
+      match = "Mix.Tasks." <> task_module
+
+      cond do
+        match == module_id ->
+          "[#{text}](#{match}#{extension}#content)"
+
+        match in modules_refs ->
+          "[#{text}](#{match}#{extension})"
+
+        doc = module_docs(:elixir, match, lib_dirs) ->
+          "[#{text}](#{doc}#{match}.html)"
+
+        true ->
+          all
+      end
+    end
+  end
+
   ## Helpers
 
   defp default_text(_, :custom, _match, _pmfa, text),
@@ -718,6 +749,12 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   defp re_kind_language(:function, :erlang) do
     ~r{
       #{re_source(:mfa, :erlang)}
+    }x
+  end
+
+  defp re_kind_language(:mix_task, :elixir) do
+    ~r{
+      mix\ ([a-z]+[a-z0-9\._]*)
     }x
   end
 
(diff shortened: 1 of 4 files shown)


test/ex_doc/formatter/html/autolink_test.exs
```diff
@@ -327,6 +327,38 @@ defmodule ExDoc.Formatter.HTML.AutolinkTest do
...
+
+      assert project_doc("`mix hex.publish`", %{
+               docs_refs: ["MyModule"],
+               module_id: "MyModule",
+               extension: ".html",
+               lib_dirs: lib_dirs
+             }) == "[`mix hex.publish`](#{@elixir_docs}hex/Mix.Tasks.Hex.Publish.html)"
+    end
+
+    test "does not autolink task with arguments" do
```
wojtekmach: as the 1st step I skipped it but autolinking these too can be nice, see screenshot.

wojtekmach:
I don't think I need to extend that function because I'm just adding `{:mix_task, :elixir, :normal}` and not also `{:mix_task, :erlang, :normal}` pass; as far as I understand, we need Erlang vs Elixir pass because modules look different whereas here it's always the same: `mix TASK`. If it's gonna make code more consistent though I'll handle that; @eksperimental?

I'll update the type spec, good catch!

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -721,6 +752,12 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     }x
   end
 
+  defp re_kind_language(:mix_task, :elixir) do
+    ~r{
+      mix\ ([a-z]+[a-z0-9\._]*)
```
eksperimental: Should we handle more than one white-space?

fertapric:
@wojtekmach I would also update the [code comment](https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc/formatter/html/autolink.ex#L328):

> `kind` is either `:function` or `:module`.

lib/ex_doc/formatter/html/autolink.ex
wojtekmach: I'd keep it as is, if someone accidentally adds more it won't highlight so it's gonna be easier to spot.

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -106,7 +106,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
                 language <- [:elixir, :erlang],
                 kind <- [:module, :function] do
               {kind, language, link_type}
-            end)
+            end) ++ [{:mix_task, :elixir, :normal}]
```
eksperimental: I guess we could also support `:custom` links.
such as we do with the rest of the funtions.

https://github.com/elixir-lang/ex_doc/blob/5e00c0858aecdfa9506774316218162af25cd3f9/test/ex_doc/formatter/html/autolink_test.exs#L433-L434

```elixir
      assert project_doc("[the :erlang.apply/2 function](`Kernel.apply/2`)", %{}) ==
               "[the :erlang.apply/2 function](#{@elixir_docs}elixir/Kernel.html#apply/2)"
```

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -325,7 +325,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
   #
   # `language` can be: `:elixir`, `:erlang` or `:markdown`.
   #
-  # `kind` is either `:function` or `:module`.
+  # `kind` is either `:function`, `:module`, or `:mix_task`.
```
eksperimental: I'm OK with :mix_task, but I'm wondering. Is there any other kind of task?
Should it be called just `task`?
is there a equivalent of a Mix Task in Erlang?

lib/ex_doc/formatter/html/autolink.ex
eksperimental: My only concern is that if a tests editor wraps the line and we end up with
```
`mix
task argument`
```

or

```
`mix task
arguments`
```

eksperimental:
@wojtekmach: It looks awesome at first glance, I would like to give it a try, tinker with it, and get back to you with more feedback if needed, but unfortunately I'll have to do this later.

lib/ex_doc/formatter/html/autolink.ex
wojtekmach: @fertapric [convinced me](https://github.com/elixir-lang/ex_doc/issues/909#issuecomment-432937018) that custom links can sometimes hurt IEx experience so how about we hold off on this change for now?

lib/ex_doc/formatter/html/autolink.ex
eksperimental: beacuse I'm seeing we are calling `{:mix_task, :elixir, :normal}`, it is a bit redundan i guess.
`{:mix_task, :elixir, :normal}`, and in the future we could call `{:task, :erlang, :normal}`

My only concern is that there are other tools other than Mix and Rebar for creating tasks that we may support

wojtekmach:
@eksperimental great, no rush! Also just wanted to say that it was very easy to add this functionality after your refactoring, thanks for that.

lib/ex_doc/formatter/html/autolink.ex
eksperimental: > @fertapric [convinced me](https://github.com/elixir-lang/ex_doc/issues/909#issuecomment-432937018) that custom links can sometimes hurt IEx experience so how about we hold off on this change for now?

Is that something that we could improve in IEX?

eksperimental:
> @eksperimental great, no rush! Also just wanted to say that it was very easy to add this functionality after your refactoring, thanks for that.

the refactoring came as a must do, after I spent so long trying to fix a bug, and I wasn't able to.
I was wondering how you manged to fixed some of those  before and implement new features :-D

wojtekmach:
easy: fix 1 bug, add 2 more, rinse and repeat, wait for a hero to clean up :joy:

lib/ex_doc/formatter/html/autolink.ex
josevalim: @eksperimental i think we can wait until it happens.

lib/ex_doc/formatter/html/autolink.ex
josevalim: :+1: for `mix_task` because that's also what we call it in the sidebar.

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -447,6 +447,37 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     end
   end
 
+  defp replace_fun(:mix_task, :elixir, _link_type, options) do
```
eksperimental: `since we are only supporting :normal links, this line can read:

`  defp replace_fun(:mix_task, :elixir, :normal, options) do`

lib/ex_doc/formatter/html/autolink.ex
eksperimental: the `+` is redundant in the regex

test/ex_doc/formatter/html/autolink_test.exs
```diff
@@ -327,6 +327,45 @@ defmodule ExDoc.Formatter.HTML.AutolinkTest do
     end
   end
 
+  describe "mix tasks" do
```
eksperimental: Mix tasks

lib/ex_doc/formatter/html/autolink.ex
eksperimental: > @eksperimental i think we can wait until it happens.

already happening
https://github.com/elixir-lang/elixir/pull/8372/files#diff-c12d92120ee0fe3d324a448c404f75a6L26

PR #912, opened 2018-10-27
Always expand Elixir lib dirs

### LINKED ISSUE #874: Application docs doesn't autolink to Mix.Tasks.Compile.App
screen shot 2018-07-15 at 12 03 38

https://hexdocs.pm/elixir/master/Application.html#module-the-application-resource-file

I think it's because Elixir app doesn't depend on Mix app so that's why we can't autolink.

wojtekmach:
Closes #874

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -93,11 +93,14 @@ defmodule ExDoc.Formatter.HTML.Autolink do
           entry <- [doc.id | doc.defaults],
           do: prefix <> module.id <> "." <> entry
 
+    lib_dirs = extra_lib_dirs ++ elixir_lib_dirs() ++ erlang_lib_dirs()
+    lib_dirs = Enum.map(lib_dirs, fn {path, url} -> {Path.expand(path), url} end)
+
     %{
       aliases: aliases,
       docs_refs: docs_refs ++ types_refs,
       extension: extension,
-      lib_dirs: extra_lib_dirs ++ elixir_lib_dirs() ++ erlang_lib_dirs(),
+      lib_dirs: lib_dirs,
       modules_refs: modules_refs
     }
   end
@@ -562,7 +565,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
         nil
 
       path ->
-        path = List.to_string(path)
+        path = path |> List.to_string() |> Path.expand()
 
         lib_dirs
         |> Enum.filter(fn {lib_dir, _} -> String.starts_with?(path, lib_dir) end)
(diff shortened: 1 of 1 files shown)


lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -562,7 +565,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
         nil
 
       path ->
-        path = List.to_string(path)
```
wojtekmach: this was the culprit, when inspecting variables we had:

```
lib_dirs: [
  {"/Users/wojtek/src/elixir-lang/elixir/lib/elixir/ebin",
   "https://hexdocs.pm/elixir/"},
  {"/Users/wojtek/src/elixir-lang/elixir/lib/eex/ebin",
   "https://hexdocs.pm/eex/"},
  {"/Users/wojtek/src/elixir-lang/elixir/lib/iex/ebin",
   "https://hexdocs.pm/iex/"},
  {"/Users/wojtek/src/elixir-lang/elixir/lib/logger/ebin",
   "https://hexdocs.pm/logger/"},
  {"/Users/wojtek/src/elixir-lang/elixir/lib/mix/ebin",
   "https://hexdocs.pm/mix/"},
  {"/Users/wojtek/src/elixir-lang/elixir/lib/ex_unit/ebin",
   "https://hexdocs.pm/ex_unit/"},
  {"/usr/local/opt/asdf/installs/erlang/21.1/lib",
   "http://www.erlang.org/doc/man/"}
]
path: "/Users/wojtek/src/elixir-lang/elixir/bin/../lib/mix/ebin/Elixir.Mix.Tasks.Compile.App.beam"
```

which explains the discrepancy.

I couldn't easily write a test for it, verified by regenerating Elixir docs.

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -93,11 +93,14 @@ defmodule ExDoc.Formatter.HTML.Autolink do
           entry <- [doc.id | doc.defaults],
           do: prefix <> module.id <> "." <> entry
 
+    lib_dirs = extra_lib_dirs ++ elixir_lib_dirs() ++ erlang_lib_dirs()
+    lib_dirs = Enum.map(lib_dirs, fn {path, url} -> {Path.expand(path), url} end)
```
josevalim: I would prefer if we did not do this and the paths were already given expanded. Would that be possible?

wojtekmach: @josevalim agreed. I just realised I was expanding elixir paths twice so doing it just in `elixir_lib_dir/1` now.

Issue #730, opened 2017-05-23
Improving search engine results for functions


jamen:
When I Google a specific function for Elixir looking for API docs, I am not given a link directly to that function.  Instead I get a link to the module page, but this is not always obvious or useful especially if you are new to the language trying to figure out what to use.

Take a few searches as an example:
 - elixir string split
 - elixir binary macro
 - elixir defmodule

I think if each function had their own pages, Google would index more useful information.

josevalim:
We probably don't need to have separate pages but there is likely something we can do for Google to index the page better.

milmazz:
We can include support for *Structured Data* in ExDoc with [schema.org](http://schema.org) or even better with [JSON-LD](https://json-ld.org) in the near future. wdyt?

BTW, this [Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/#url=http%3A%2F%2Fhexdocs.pm%2Felixir) from Google it's very handy for developers. The example shows that we don't provide any Structured Data yet :-(

josevalim:
That's awesome @milmazz. If we can embed it directly in the markup, it definitely gets more vote.

axelson:
I spent some time looking into this today and as far as I can tell Google does not specify how to enable a "jump_to" link in the search results. I also looked at Ruby and Go which do have "jump_to" listed (e.g. try searching for "ruby string chomp") but I do not see any structured data on the ruby page. So to me it is unclear how to accomplish this.

davidsulc:
> We generate these deep links completely algorithmically
[source](https://webmasters.googleblog.com/2009/09/using-named-anchors-to-identify.html)

It doesn't seem there's any way to configure "jump to" links for a page, only things that "increase the chances that they might appear on your pages" which are already in place in the generated documentation markup.

cryic:
I compared the pages that come back in an Elixir search versus a couple of other languages:

No 'jump to' links:
https://www.google.com/search?q=elixir+enum+fetch

'Jump to' links:
https://www.google.com/search?q=ruby+hash+fetch
https://www.google.com/search?q=gorm+query+Count

Something we could try is how we use file-level relative links in the sidebar.

When on https://hexdocs.pm/elixir/Enum.html, the sidebar links for local anchors would be:

`<a href="#fetch/2">fetch/2</a>`

vs.

`<a href="Enum.html#fetch/2">fetch/2</a>`

[...]

PR #913, opened 2018-10-27
[Issue 730] - Improving Search Engine Results - Sidebar links

cryic:
As a potential solution for https://github.com/elixir-lang/ex_doc/issues/730, this adds a check to create same-page anchors if the node is local.

For example, when on https://hexdocs.pm/elixir/Enum.html, the sidebar links under MODULES>Enum>Functions would be:

`<a href="#fetch/2">fetch/2</a>`

vs.

`<a href="Enum.html#fetch/2">fetch/2</a>`

I think this could help the visibility of on-page destinations in search engine's results.

Please let me know your thoughts on refactoring the check into a separate helper. I left it as is to be explicit about the check, however, it is a bit ugly.


eksperimental:
Do you have any documentation that backs this?

cryic:
Unfortunately, not much at all. Having tried to research this, most discussions were around absolute vs relative urls, and had no discussion about anchor links.

I had thought of this as an actionable difference after comparing against other pages. I think it definitely isn't a 'full' solution, but thought it may help. I can close this if people don't think this would help or add very much.

Thanks for your suggestions in the issue thread!

josevalim:
@cryic shall we change this according to #922? Thank you!

cryic:
@josevalim That sounds good. Time got away from me, but I've gotten a start on #922. For now, I'll close this PR and open a new one when that work is finished.

Thanks @eksperimental for suggesting and writing up the issue.

PR #914, opened 2018-10-29
don't emit a redundant entry for nesting

davidsulc:
No entry should be emitted for the nesting context if the nesting prefix matches an existing module.

This bug slipped in when moving the nesting entry logic to the UI code.

Before fix (where `ScrapyCloudEx.Endpoints` is a module that exists in the source):
before

After fix:
after


josevalim:
Good catch! @davidsulc what does it look like when "ScrapyCloudEx.Endpoints" is open to show its summary, functions, etc?

davidsulc:
@josevalim This is with both the "context node" and the "nested node" expanded:
expanded

josevalim:

PR #919, opened 2018-11-08
Autolink `mix help TASK`

wojtekmach:
Screenshot:

screenshot 2018-11-08 at 10 21 41

Although maybe it's taking this too far :)


josevalim:
I am fine with it but let's see what others think. /cc @eksperimental

eksperimental:
This is great and it makes a lot of sense, since we are linking to the help documentation

PR #929, opened 2018-11-30
Disable earmark smartypants option by default

### LINKED ISSUE #928: Quotes are sometimes rendered incorrectly
In the Phoenix 1.4.0 documentation we have the following [snippet](https://github.com/phoenixframework/phoenix/blob/v1.4.0/lib/phoenix/socket.ex#L294):

`for example "room:*", "users:*", "system"`

Which is rendered in [html](https://hexdocs.pm/phoenix/Phoenix.Socket.html#channel/3) as `“room:<em>“, “users:</em>”, “system”`

Note that the second smart quote is "incorrect", which is possibly because the `<em>` tag was inserted between them, although I would expect that the quote after `users:` would have the same issue (abut maybe this is just a browser rendered oddity).

After looking at the generated code I'm not sure that this is necessarily an ex_doc issue since the documentation should either be escaping the `\*` or enclosing them in backticks. I will file a PR on Phoenix to fix the generated docs.

Also in my opinion, it would be better for ex_doc not to automatically convert double quote characters to smart quotes (but that's probably a different discussion)

Edit: Phoenix PR filed: https://github.com/phoenixframework/phoenix/pull/3191

axelson:
Fixes #928

### DIFF
--- a/lib/ex_doc/markdown/earmark.ex
+++ b/lib/ex_doc/markdown/earmark.ex
@@ -29,7 +29,7 @@ defmodule ExDoc.Markdown.Earmark do
       breaks significant (so every line in the input is a new line in the output)
 
     * `:smartypants` - boolean. Turns on smartypants processing, so quotes become curly,
-      two or three hyphens become en and em dashes, and so on. True by default
+      two or three hyphens become en and em dashes, and so on. False by default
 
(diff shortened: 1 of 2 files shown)


josevalim:

axelson:
Great, thanks! ❤️

PR #930, opened 2018-12-06
Warn on undefined remote functions from project's docs

wojtekmach:
This only works for project's docs so when e.g. `Foo.bar/0` from `foo` dependency doesn't exist we won't warn yet.

We also don't warn when parsing extras pages, this is useful for Elixir deprecations page when we reference functions that are deprecated.

### DIFF
--- a/lib/ex_doc/config.ex
+++ b/lib/ex_doc/config.ex
@@ -42,7 +42,8 @@ defmodule ExDoc.Config do
             source_url: nil,
             source_url_pattern: nil,
             title: nil,
-            version: nil
+            version: nil,
+            warn_on_undefined_functions: false
 
   @type t :: %__MODULE__{
           api_reference: boolean(),
@@ -72,6 +73,7 @@ defmodule ExDoc.Config do
           source_url: nil | String.t(),
           source_url_pattern: nil | String.t(),
           title: nil | String.t(),
-          version: nil | String.t()
+          version: nil | String.t(),
+          warn_on_undefined_functions: boolean()
         }
 end
--- a/lib/ex_doc/config.ex
+++ b/lib/ex_doc/config.ex
@@ -43,7 +43,7 @@ defmodule ExDoc.Config do
             source_url_pattern: nil,
             title: nil,
             version: nil,
-            warn_on_undefined_functions: true
+            warn_on_undefined_references: true
 
   @type t :: %__MODULE__{
           api_reference: boolean(),
@@ -74,6 +74,6 @@ defmodule ExDoc.Config do
           source_url_pattern: nil | String.t(),
           title: nil | String.t(),
           version: nil | String.t(),
-          warn_on_undefined_functions: boolean()
+          warn_on_undefined_references: boolean()
         }
 end
--- a/lib/ex_doc/config.ex
+++ b/lib/ex_doc/config.ex
@@ -43,7 +43,7 @@ defmodule ExDoc.Config do
             source_url_pattern: nil,
             title: nil,
             version: nil,
-            warn_on_undefined_references: true
+            skip_undefined_reference_warnings_on: []
 
   @type t :: %__MODULE__{
           api_reference: boolean(),
@@ -74,6 +74,6 @@ defmodule ExDoc.Config do
           source_url_pattern: nil | String.t(),
           title: nil | String.t(),
           version: nil | String.t(),
-          warn_on_undefined_references: boolean()
+          skip_undefined_reference_warnings_on: [String.t()]
         }
 end
--- a/lib/ex_doc/config.ex
+++ b/lib/ex_doc/config.ex
@@ -43,7 +43,7 @@ defmodule ExDoc.Config do
             source_url_pattern: nil,
             title: nil,
             version: nil,
-            warn_on_undefined_functions: false
+            warn_on_undefined_functions: true
 
   @type t :: %__MODULE__{
           api_reference: boolean(),
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -101,7 +101,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
       extension: extension,
       lib_dirs: lib_dirs,
       modules_refs: modules_refs,
-      warn_on_undefined_references: config.warn_on_undefined_references
+      skip_undefined_reference_warnings_on: config.skip_undefined_reference_warnings_on
     }
   end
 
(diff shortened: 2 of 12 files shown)


wojtekmach:
There's one thing left, we get false positives for protocols, here's one from tests:

```
warning: c:CustomProtocol.plus_two/1 is not found (parsing CustomProtocol.Number docs)
```

I get similar warnings for Elixir docs.

Any ideas?

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -444,6 +446,10 @@ defmodule ExDoc.Formatter.HTML.Autolink do
           "[#{text}](#{elixir_docs}Kernel.SpecialForms" <>
             "#{extension}##{prefix}#{enc_h(function)}/#{arity})"
 
+        module in modules_refs && module_id ->
+          IO.warn("#{match} is not found (parsing #{module_id} docs)", [])
```
josevalim: Can we also show the function name/arity in case the docs come from those? Or we don't have this info?

wojtekmach: not sure what you mean, could you elaborate?

`match` is e.g. `c:CustomProtocol.plus_two/1`

wojtekmach: ohh, that would be super helpful. I'll look into it.

wojtekmach: @josevalim Good call, it should be much more useful now!

The build should be fine now but we still get the warnings I mentioned in the first comment.

lib/ex_doc/config.ex
```diff
@@ -42,7 +42,8 @@ defmodule ExDoc.Config do
             source_url: nil,
             source_url_pattern: nil,
             title: nil,
-            version: nil
+            version: nil,
+            warn_on_undefined_functions: true
```
wojtekmach: hm, it's actually gonna warn on undefined callbacks and types. Maybe it should be simply `warn_on_undefined`? 

Btw, I only added the option to be able to disable it in integration tests.

wojtekmach:
> `warning: c:CustomProtocol.plus_two/1 is not found (parsing CustomProtocol.Number docs)`

Ok, here's what happening. `defprotocol` defines callbacks, when we parse docs for _impl_ module we get this node:

```elixir
  %ExDoc.ModuleNode{
    module: P.Integer,
    type: :impl,
    # ...
    docs: [
      %ExDoc.FunctionNode{
        id: "p/1",
        doc: "Callback implementation for `c:P.p/1`.",
        type: :function,
        # ...
      }
    ],
  }
```

However, the _protocol_ module node is this:

```elixir
  %ExDoc.ModuleNode{
    module: P,
    doc: "The Foo protocol.\n",
    docs: [
      %ExDoc.FunctionNode{
        id: "p/1",
        type: :function
      }
    ],
    # ...
  }
```

the documentation of `P.p` is stored as a function, not callback, which explains the warning.

Looks like we don't ever show _impl_ module docs so perhaps we should just skip these nodes?

josevalim:
Don’t we show impl if it is explicitly doc’ed or not even then?

I guess the correct answer though is to not convert to markdown modules
that are explicitly doc’ed false.
-- 


wojtekmach:
We don't show impl docs even if they're explicitly doc'ed.

I'm gonna skip impl `ModuleNode`s for now to fix this issue and we can bring them back if we decide to show impl docs

josevalim:
Sounds good to me then. My only question is why we are trying to autolink
for a module that has moduledoc false. Shouldn’t we filter it out before,
based on the moduledoc being false?
-- 


wojtekmach:
You're saying impls are getting `@moduledoc false` by default? I think that's not the case but I can double check.

wojtekmach:
```
iex(2)> Code.fetch_docs(Enumerable.List)
{:docs_v1, 3306, :elixir, "text/markdown", :none, %{},
 [
   {{:function, :__impl__, 1}, 3306, ["__impl__(atom)"], :hidden, %{}},
   {{:function, :count, 1}, 3307, ["count(list)"], :none, %{}},
   {{:function, :member?, 2}, 3308, ["member?(list, value)"], :none, %{}},
   {{:function, :reduce, 3}, 3311, ["reduce(list, arg, fun)"], :none, %{}},
   {{:function, :slice, 1}, 3309, ["slice(list)"], :none, %{}},
   {{:function, :slice, 3}, 3316, ["slice(list, start, count)"], :hidden, %{}}
 ]}
```

AFAIR `:none` means there's no `@moduledoc` and `:hidden` would be for `@moduledoc false`

eksperimental:
I think we should pass a list pages to be ignored and include extra pages. The case of deprecated functions or CHANGELOG are very specific, and I'm sure first time you encounter them, you know they need to be added to the ignore list. It would be more desired than to leave all extra pages uncovered

wojtekmach:
> I think we should pass a list pages to be ignored and check all extra pages by default.

Good call, we can now do e.g.:

```
  skip_undefined_reference_warnings_on: ["compatibility-and-deprecations", "SomeModule"]
```

josevalim:
@wojtekmach oh sorry. I was thinking the skip would receive the reference you don't want to warn on and not on the module that defines the reference. Thoughts? So if `Foo` calls a non existing `Foo.Bar.baz/0` then you would pass "Foo.Bar.baz/0". I think this is the most fine grained control, hence why I propose it. :)

wojtekmach:
@josevalim yeah, we could do that too. I just thought this approach is less hassle for deprecations page (it's off the hook completely). (same goes for `Warnings` test fixture module)

It's easy to switch this to your approach, it's just we'd need to put the whole deprecations list from Elixir to it's ex_doc config but that shouldn't be too bad.

eksperimental:
What is "the current approach"?

wojtekmach:
the current approach is we can skip certain pages from triggering warnings. Note, what we skip are docs that contain references, not the references themselves.

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -161,26 +163,32 @@ defmodule ExDoc.Formatter.HTML.Autolink do
 
     compiled =
       compiled
+      |> Map.put(:id, nil)
```
josevalim: We don't need this call anymore, do we?

wojtekmach:
I've cherry-picked https://github.com/elixir-lang/ex_doc/commit/e67bd99c75bb077af902c72cfb3f44fc85793c3f on master in the meantime.

Issue #922, opened 2018-11-11
Add title attribute to "Link to this..." and other links


eksperimental:
_Taken from https://github.com/elixir-lang/ex_doc/issues/730#issuecomment-433664519 and https://github.com/elixir-lang/ex_doc/issues/730#issuecomment-433664728_
_Related PR: #730_

One thing that if it will not improve search results in search engines it will still be useful is to use the full mfa in the title of every link, for every module, function, type, etc.

So we could have something link:
```html
<a href="#fetch/2" title="Enum.fetch/2">fetch/2</a>
```

or have this:

```html
<a href="#fetch/2" class="detail-link" title="Link to Enum.fetch/2">
```

instead of this:

```html
<a href="#fetch/2" class="detail-link" title="Link to this function">
```

/cc @cryic

cryic:
Thanks for breaking this into its own issue @eksperimental . I've started this and hope to have something up this week.

wojtekmach:
@cryic ping :)

cryic:
@wojtekmach Thanks for the ping! I got through some of this but haven't completed it yet. I'm going to setup time to finish and post a PR soon. Let me know if you have any thoughts/concerns.

wojtekmach:
@cryic no worries and no rush, I was just checking if you're still working on it in case anyone else wants to pick this up in the meantime. Let us know if you run into any issues!

PR #932, opened 2018-12-09
Add title attribute with mfa on links to improve seo

cryic:
Implementation of https://github.com/elixir-lang/ex_doc/issues/922

To improve the docs SEO for function discoverability, this PR adds a title attribute to links for modules, functions, etc.

Please let me know if you have any feedback.

A screenshot of dev tools showing the title:

screen shot 2018-12-08 at 10 46 40 pm


lib/ex_doc/nodes.ex
```diff
@@ -106,6 +109,7 @@ defmodule ExDoc.TypeNode do
           signature: nil | String.t(),
           source_url: nil | String.t(),
           source_path: nil | String.t(),
-          annotations: list()
+          annotations: list(),
+          module: nil | String.t()
```
josevalim: Do we really need this new field? We should always know the module it belongs to based on the context. For example, for summary_entry_template, we can probably pass it when invoking it here: https://github.com/elixir-lang/ex_doc/blob/59905ee0415cf639136001988f502bd71872dd7a/lib/ex_doc/formatter/html/templates/summary_template.eex#L6

cryic: I had thought that each node may use some info about its containing module, however, passing the module through makes sense. I take out the field and change the template signatures. Thanks!

cryic:
@eksperimental That makes sense. I'll remove the 'Link to' text except for the chain link and screenreader text. Thanks for reviewing!

eksperimental:
I will leave a full review when I get to be with my computer. Thank you for taking care of this

cryic:
I forgot to comment on here that the changes were implemented. Thank you all for reviewing!

assets/js/templates/search-results.handlebars
```diff
@@ -1,31 +1,33 @@
...
+{{#each levels}}
+<h2 class="results-title">{{name}}</h2>
+{{#each results}}
+<div class="result">
+  <h3 class="result-id">
+    <a href="{{id}}.html">{{{match}}}</a>
+  </h3>
+  <ul class="functions">
+    {{#each functions}}
+    <li class="result-elem"><a href="{{../id}}.html#{{{anchor}}}" title="{{../id}}.{{anchor}}">{{{match}}}</a></li>
```
milmazz: Is this whole re-indent actually needed for this feature? It makes a bit difficult to see the _real_ differences unless you ignore the whitespace changes. In the future try to accommodate to the current style of send another commit just doing the re-indent.

cryic: Completely unintentional spacing changes. I neglected to undo the changes from my formatter. I'll change these and comment again once it is fixed up.

eksperimental:
I think we can be consistent with the titles in the sidebar and add them to all of the items.
For instance, where it reads "Guards"  title should be "MODULE_NAME guards" (Kernel Guards), 
and so on for the other items "Kernel functions", "Kernel macros", "Kernel summary".
I'm not sure about "Top", since it links to the same spot as Kernel, it we should use the title "Kernel", or "Kernel - top of the module". I am more inclined to the first one, just to be consistent with the title of the Kernel item.
As for the pages, it can have the same title as the page, and for the sections, "Compatibility between non-major Elixir versions — Deprecations" being "—" an em-dash.

While some of this may sound repetitive, I think while scrolling such longish menus, it's refreshing to mention where we are by mentioning the main module or page.

cryic:
Thanks for the thorough feedback. I'll go back through and work on these.

eksperimental:
I'm  back. I will review the changes and get back to you.
Thank you.

cryic:
Great! I haven’t pushed yet, but I’ll comment here again when everything is ready.

eksperimental:
Hi @cryic
the PR got messed up because I think you merged master instead of rebasing.
My suggestion is  to download master locally, rebase on top of it `git rebase -i master`, fix anything that needs to be fixed if needed, and the do a "forced push"  by doing `git push origin HEAD -f` to your repo

cryic:
@eksperimental Thanks for pointing that out! You're completely correct. I had merged instead of rebased master.

I'm going through your last couple of comments/suggestions to make sure I've covered them. I'll comment here with any notes.

cryic:
@eksperimental From your comment https://github.com/elixir-lang/ex_doc/pull/932#pullrequestreview-199401443 I have accomplished number 5, but the rest are based off an earlier comment https://github.com/elixir-lang/ex_doc/pull/932#issuecomment-449217970 .

I'll use a separate branch to get this work up to the most recent [comment](https://github.com/elixir-lang/ex_doc/pull/932#pullrequestreview-199401443). Let me know if you think the current PR is okay as is.

Thanks very much for your reviewing!

cryic:
As per @eksperimental 's comment above, I've gotten items 1-6 and part of 9 done here. The remaining items will need to be their own issues and/or PRs due to research and some need for insight. Just for quick reference, Items 7-8 need a bit more research in how to recognize private/opaque functions. The ':erlang' module part of 9 and 10 are difficult due to `Autolink` currently not linking Erlang modules or 'prelinked' links.

Concerning this PR, there a bug in Earmark that messes up Markdown parsing when two links that have titles are on the same line, reported in https://github.com/pragdave/earmark/issues/220 . I think it has been fixed, however, until the release is out, it may cause problems with this work.

Let me know what you think about this PR. It covers a lot and was much larger than I'd anticipated. Thank you!

cryic:
I've gotten the latest changes from `master` into my branch and conflicts fixed up.

eksperimental:
@cryic I'm downloading the latest changes, I will test it and get back to you in a few days

eksperimental:
@cryic do you think you could squash everything in one or more commits, it's really hard to review your changes as there are other changes interwoven chronologically speaking. it's 57 commits already

eksperimental:
or if there is a way to run a diff and only show the changes of this PR, it will also help

cryic:
@eksperimental Sorry for misunderstanding, but I'm not sure what you mean. I can definitely squash this PR into 1 commit, but I'm not sure what difference that makes to https://github.com/elixir-lang/ex_doc/pull/932/files . 

Is there a different view/method you use to review? Let me know and I'll squash and push.

eksperimental:
Im reviewing from my terminal. Thats why. One commit is fine for me

cryic:
@eksperimental A couple of things got mixed up during the squash. I'll fix those up, push, then `@` you. Sorry for the trouble.

cryic:
@eksperimental Okay, I think I got everything fixed up. However, the TravisCI build says failed, however, it looks like all the tests passed, but there was an error on report upload:

```
** (ExCoveralls.ReportUploadError) Failed to upload the report to 'https://coveralls.io' (reason: timeout).
    (excoveralls) lib/excoveralls/poster.ex:20: ExCoveralls.Poster.execute/2
    (mix) lib/mix/tasks/test.ex:385: Mix.Tasks.Test.run/1
    (mix) lib/mix/task.ex:316: Mix.Task.run_task/3
    (excoveralls) lib/mix/tasks.ex:51: Mix.Tasks.Coveralls.do_run/2
    (mix) lib/mix/task.ex:316: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:79: Mix.CLI.run_task/2
    (elixir) lib/code.ex:767: Code.require_file/2
The command "mix coveralls.travis" exited with 1.
```
Can you rerun the build?

josevalim:
> Can you rerun the build?

Done!

cryic:
@josevalim Thanks José!

@eksperimental Great. I think this is ready for review. Its 4 commits, after squashing the main piece of work, then needing to fix a couple of things. Should be in better shape. Thanks!

lib/ex_doc/formatter/html/search_items.ex
```diff
@@ -27,14 +27,15 @@ defmodule ExDoc.Formatter.HTML.SearchItems do
     encode(
       "#{id}.html##{HTML.text_to_id(header)}",
       "#{title} - #{header}",
+      "#{title} - #{header}",
```
josevalim: The generated search_items.js is already really large, so I don't think we should add a new field. We already have a title field and we should re-use it. Therefore, can you please revert the changes to this file and re-use the existing file on the `.js` side of things? Thank you!

wojtekmach:
I did a little bit of research (links below) and I wasn’t able to find any evidence that adding
the title attribute helps with SEO and in fact some sources discourage that as it unnecessarily
clutters the UI.

Some references claim that the only benefit of the title attribute is it improves
UX. If it boils down to that I’m not sure if it justifies the change.

The problem we were orignally trying to solve was to have Google “jump to” links, but it’s unclear
whether adding title helps here at all. Ruby doc that was mentioned as example of site that has
these jump to links, does not put title into the links.

We're also investigating a somewhat conflicting feature, which is adding popovers on hover: https://github.com/elixir-lang/ex_doc/issues/971

For these reason, I think we shouldn’t move forward with this change. I’m sorry that I haven’t
raised these concerns sooner. Thank you for your work @cryic nonetheless.

To improve SEO perhaps we should go back to looking into
microformats mentioned in the earlier issues, or look into sitemaps (e.g. each package would have
it’s own sitemap)

This was by no means a comprehensive research, just a cursory look at some google hits when
searching for “seo link title”:
- https://moz.com/community/q/does-the-link-title-attribute-benefit-seo-2
- https://www.searchenginepeople.com/blog/title-attribute.html
- http://forums.seochat.com/search-engine-optimization-28/title-link-title-important-search-engine-optimization-486924.html

josevalim:
Those are very good points @wojtekmach, thanks.

Looking at the entries above, they even recommend to *not* add links if the content is the same as the link itself, as it pollutes the UI. With that said, I think the only case for adding links in today's docs would be in the typespecs (which was really well done by @cryic) but, again, as we are already adding popovers, it indeed feels the feature is redundant. :(

In any case, thanks for all the work @cryic, it is really appreciated 
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:. Sorry that it won't be merged.

cryic:
@wojtekmach Thank you for researching this a bit more. I had thought this might have been helpful originally, however, it does appear that probably isn't good. I had also thought that this would help for accessibility, but also that doesn't appear to the case either. Need to update my assumptions :) Thanks for your input. Looking forward to the popovers!

PR #935, opened 2018-12-09
(2) Δ Refactor retriever.ex

English3000:
I'm guessing you'll reject this one for the [same reasons](https://github.com/elixir-lang/ex_doc/pull/933#issuecomment-445517066).

Hence, I haven't gone further than this (although my next step would be to split this out into more files).

That said, I think at least moving the 400+ lines of code to `data.ex` is doable by your criteria, even if you ignore everything else.


josevalim:
Closing for the same reasons as the others. :)

Issue #903, opened 2018-09-26
Help making my first contribution


English3000:
Currently, `:ex_doc` organizes functions using the navigation hierarchy:

> Module > Alphabet

I [proposed](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/elixir-lang-core/TxDvKFwltZU/wXf_T2_PBwAJ) adding functionality for:

> Category > Module > Alphabet

For starters, I looked at a [similar pull request](https://github.com/elixir-lang/ex_doc/pull/879) by @lackac 

Taking a closer look, it appears I could add `@doc category: :data` in a file,

```elixir
# mix.exs, project/0
docs: [
  groups_for_functions: [
    Data: & &1[:category] == :data
  ]
]
```

and this would generate a `+ Data` dropdown tab (once the next release of `:ex_doc` comes out), creating the nav hierarchy:

> Module > Category > Alphabet

In all honesty, there's a lot going on in `retriever.ex`, where I believe the "action" happens--setting up structs which are used to render the UI--and I'm having a hard time understanding.

If there's the bandwidth, it would be helpful to get a high-level overview (or guidance on how I'd add control over the navigation hierarchy to the codebase). In addition, this would reduce the barrier for others in making a contribution.

josevalim:
Hi @English3000!

The ExDoc team is quite time constrained. When we have extra time, our priority is to focus on fixing existing issues. You may have better luck getting help with more specific questions somewhere like the ElixirForum.

Regarding the feature, note that the `groups_for_modules` already allows you to specify groups -> modules -> alphabet, so make sure to expand what is the difference in the feature you are requesting with the one you have right now, with its benefits.

Also, keep in mind there is no guarantee a contribution will be accepted, especially if it will make the retriever much more complex (as you already noted it is complex today).

Thanks!

English3000:
[image: image.png]

English3000.org: *A data-driven approach to language learning.*

English3000:
![image](https://user-images.githubusercontent.com/29601050/46077900-518f8880-c147-11e8-9e25-3e6121f5440c.png)

PR #936, opened 2018-12-09
(last one) + CONTRIBUTING.md

English3000:
Frankly, I'm a little confused by your [reasons](https://github.com/elixir-lang/ex_doc/pull/933#issuecomment-445517066) against refactoring, namely because I just haven't seen them before.

In terms of "messing up history", is there some special view on GitHub for that? Otherwise, I don't get it...

Whatever the case, **I don't expect you to accept this one either _(it's incomplete)_--rather just to produce something in this vein.**

`retriever.ex` in particular is nastily complicated and without a doc like this, I'd be very surprised by open source contributions (albeit, I think the `groups_for_functions` PR was, so...).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,38 @@
+# Guidance
+
+> Contributing to an open-source codebase for the first time can be intimidating --
+huge files with functions that call other functions in other huge files.
+>
+> This doc aims to make contributing a little friendlier.
+
+Given the nature of this project, you'll most likely want to start at [`docs.ex`](https://github.com/elixir-lang/ex_doc/blob/master/lib/mix/tasks/docs.ex#L230)
+
+## docs.ex
+
+> If you're unfamiliar with [`Mix.Task`s](https://hexdocs.pm/mix/Mix.Task.html)...
+
+1. When a user runs `mix docs`, `run/3` is executed. As noted in the `@moduledoc`,
+the `args` are from the command line. As we'll be generating docs from the task,
+the first step is to compile our user's Mix project.
+
+2. We guard against bad command line `args`
+
+3. After setting up `options`, we generate docs in a number of formats and
+log messages with where our user can find them in the project.
+
+## ex_doc.ex
+
+So what's this [`generator/3`](https://github.com/elixir-lang/ex_doc/blob/master/lib/mix/tasks/docs.ex#L264) anonymous function being called? Well, it [defaults](https://github.com/elixir-lang/ex_doc/blob/master/lib/mix/tasks/docs.ex#L240) to `ExDoc.generate_docs/3`, so let's open up [`ex_doc.ex`](https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc.ex#L18) to see what that function does...
+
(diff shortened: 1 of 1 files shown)


English3000:
_Re: `git-blame`,_ I just added it on Atom and it doesn't seem to appear in chronological order. Also, it seems there are duplicates of the exact same commit. Is there a way to fix those? (Don't see anything about chronological order on a quick search.)

---

_Re: the scope of this PR,_ a few months ago, I was trying to add some more customizability to the nav bar _(specifically, the ability to group functions by category, as is already done for modules)_, but as I said, the way the codebase was written and the lack of documentation made it hard to wrap my head around... and making this more readable in order to add that feature (since no one wanted to help me with that feature...) was the reason I made this set of PR's.

While I do see plenty of commits from different people, I'm not sure we mean the same thing by open source contribution. I'm speaking to a feature-level contribution. For example, in `docs.ex`, the `normalize_*_url` helpers are very similar so I turned them into 1 helper, `normalize_urls`, using `Enum.reduce/3`. That would technically be open source (if I had stayed with the concision approach), and could account for a number as high as 25%.

---

_Re: being time consuming,_ I mean it's obvious (and reasonable) that's the true reason you are closing my PR's. I'm just confused by your implication here that I'm the "only one" (or in the true "minority of people") having this experience of your codebase. You yourself said `retriever.ex` was a pretty nasty file:

> [Also, keep in mind there is no guarantee a contribution will be accepted, especially if it will make the retriever much more complex (as you already noted it is complex today).](https://github.com/elixir-lang/ex_doc/issues/903#issuecomment-424604409)

So wouldn't the benefits of a more readable `retriever.ex` speed up the pace of development for the whole team?

To me, a lot of this just sounds like codespeak (no pun intended) regarding our different approaches/code styles. So while it's in your power to reject my PR's outright, I'm a little confused why the response wasn't:

> @English3000 In your PR's, you change code style along with refactoring. This makes your PR extremely time-consuming (and difficult) for our time-constrained team to review.
>
> If you would like to contribute, I suggest you attempt to make the changes in a manner that changes the codebase as little as possible while achieving your desired result.

I mean, that's what you're insinuating (along with that refactors, per se, don't add value).

And my response is that doing that may make the codebase leaner, but it won't make it more readable (for someone who didn't help create the codebase).

---

Re: `CONTRIBUTING.md`, I'm sure you'll understand that I'm a little skeptical about going further with it. You yourself already stated that the team knows the codebase as is better than I do, so wouldn't it make more sense for one of them to do it?

And if I did do it, you could simply change it after all my disproportionately hard work to a format that I didn't even find helpful.

If you like the idea, go ahead, but the tone of your communications has been you don't want my contributions here, and actions speak louder than words.

josevalim:
*> Re: being time consuming,*
 I mean it's obvious (and reasonable) that's the true reason you are
closing my PR's.

No such thing as true reason. Everything I said earlier counts.

> I'm just confused by your implication here that I'm the "only one" (or in
the true "minority of people") having this experience of your codebase

Quite the opposite. I believe many others will find it hard to contribute
to ExDoc in the same way I also believe others besides the ExDoc team will
be able to contribute features. YMMV.
-- 

PR #940, opened 2018-12-14
Fix code color in html

### LINKED ISSUE #927: Non-elixir code blocks grayed out
I tried `master` ex_doc on `phoenix` and it seems that `.hljs` styling looks bad. See screenshot:
image

Firefox 63.0 on Ubuntu Linux

chulkilee:
Fixes #927

As LESS does not import CSS into scope where @import is called, and HTML
ouput loads both default and night CSS, text color is incorrectly shown.

To avoid that, embed the whole tomorrow-night CSS (BSD-3-Clause)

https://raw.githubusercontent.com/highlightjs/highlight.js/master/src/styles/tomorrow-night.css

---

Note that  this is copying BSD-3-Clause code into Apache 2. We anyway bundle it, so it is not adding *new* licensing issue, I believe.


josevalim:
Thanks! Shall we embed the other theme for consistency too?

chulkilee:
Unlike tomorrow-night, tomorrow is imported in two places - `app.less` and `epub.less`. We may have own `less` file, and include them - but not sure that's more consistent.

I think it's better to take more refactoring, such as using top-level class or including style based on the theme selection. However, I haven't digged into epub output, and JS part of it, which is required to make that change.

If it works, then we may merge it and leave more refactoring to others :)

josevalim:

PR #945, opened 2018-12-21
Use same format as in mix.exs generated with Elixir v1.7 for dependen…

eksperimental:
…cies

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,9 @@ If you are using Elixir v1.7 and later:
 
 ```elixir
 def deps do
-  [{:ex_doc, "~> 0.19", only: :dev, runtime: false}]
+  [
+    {:ex_doc, "~> 0.19", only: :dev, runtime: false},
+  ]
 end
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,15 +19,19 @@ If you are using Elixir v1.7 and later:
 
 ```elixir
 def deps do
-  [{:ex_doc, "~> 0.19", only: :dev, runtime: false}]
+  [
+    {:ex_doc, "~> 0.19", only: :dev, runtime: false},
```
fertapric:
```suggestion
    {:ex_doc, "~> 0.19", only: :dev, runtime: false}
```

josevalim:

josevalim:
> It breaks the rules of the formatter, but thia rule is still under discussion.

FWIW, It is not under discussion. We don't do trailing commas. :)

eksperimental:
I read one PR where you mentioned that we were waiting until v1.8 to give more time for people to use  it and get more feedback, and then make a decission about it

PR #956, opened 2019-01-04
Do not auto-link to module within the same module using normal links

eksperimental:
When referring to the same module, do not auto-link when using normal
links, but auto-link when using custom links.


josevalim:
Why? This adds a bunch of complexity, so what are the gains?

eksperimental:
This has annoyed me for ages, something you are longish documentation and you see a link to the module itself, and you kind of forgot what module you are in, so it is linked, and you click on it, and back to the beginning you go where you started. It add an overhead to the reader: REMEMBER WHAT MODULE YOU ARE AND DO NOT CLICK ON IT IF YOU DON'T WANT TO START OVER.
It is a common practice to refer to the same module though out the docs specially if they are kind of longish.

Maybe we should even make it italics the module that is not being linked.

eksperimental:
> > Maybe we should even make it italics the module that is not being linked.
> Please don't. I often use `` `GenServer` `` to refer to the module and just "GenServer" to talk about it in general. If we introduce a third interpretation (italics), it will be confusing and we really shouldn't change the semantics of markdown of the author.
I totally agree with you.

> I wonder if what is in this PR could be achieved with CSS and JavaScript.
Wouldn't other formats other than html be affected by such a decision, such as epub, which it cannot interpret Javascript if i'm not mistaken. 

> In the worst case scenario, can we then just add this behaviour to modules?
I was planning on working on this in a second PR but I haven't even analyzed all the possible situations.

I understand the scenario you describe can also happen to functions but:
> 
> 1. How frequently does functions link to the themselves?
> 2. How frequently you forget you are reading the doc for _that_ function?
> 3. How far will you going to scroll anyway if you click on the link for the current function?
Let me study this for a bit, and I will get back to you with a more informed input
> 
> Applying to modules should provide the majority of the benefits.

milmazz:
> Wouldn't other formats other than html be affected by such a decision, such as epub, which it cannot interpret Javascript if i'm not mistaken

@eksperimental EPUB supports JS, CSS3, etc. Actually we're using JS to highlight the source code, as a reference you can look at [epub.js](https://github.com/elixir-lang/ex_doc/blob/master/assets/js/epub.js)

josevalim:
Ping!

eksperimental:
Moving the logic to JS will make things even more complicated. Because we are not linking normal links, but we are linking custom links, and when MarkDown is rendered as HTML, how are we going to find which one is which.
It feels just wrong leaving up to JS something that we deal with Elixir everywhere else.

I have the feeling that not having this feature in functions, is not as big of a deal as not having it for modules.

If we merge this, extending it to functions wouldn't will be adding much more code complexity, since all the heavy-lifting had been done by this PR.

josevalim:
> Moving the logic to JS will make things even more complicated IMO. Because we are not linking normal links, but we are linking custom links, and when MarkDown is rendered as HTML, how are we going to find which one is which?

You will still expand custom links as usual in autolink. But then in the JS side you can get all links that point to the current page and replace them by `#` or `""`. The autolink is already complex and I really want to avoid adding more complexity to it if we can.

josevalim:
To be clear, ideally, I would avoid adding this altogether. Are there other developers running into this problem?

eksperimental:
> > Moving the logic to JS will make things even more complicated IMO. Because we are not linking normal links, but we are linking custom links, and when MarkDown is rendered as HTML, how are we going to find which one is which?
> 
> You will still expand custom links as usual in autolink. But then in the JS side you can get all links that point to the current page and replace them by `#` or `""`. The autolink is already complex and I really want to avoid adding more complexity to it if we can.

The rationale of this PR is that any mention to the current module (normal link) won't be linked, and any custom link will. If you end up with a generated <a href="my_module.html">MyModule</a> link, how would you know it was generated by `MyModule` or by [`MyModule`](MyModule). I guess it will be a small margin of a small false possitive but it will be impossible to link to the current module if implemented like that.

eksperimental:
> wouldn't you want to disable the link anyway because the same problem exists? (i.e. you are linking to where you are right now?).

Nope. you are linking to the top of the module documentation. Imagine you are dealing with long moduledoc and you want to refer to the beginning of the doc.
I came up with this rationale, because I didn't wanna go "all or nothing". Also, if the user uses a custom link it means she consciously  want to link to it.

josevalim:
I see. So I am not sure how to proceed because I don't think this is common enough to warrant the complexity in such an already complex part of the codebase. :(

eksperimental:
I know. it's already complex. What I can think is we can mark the custom generated links (which are very few in comparison with the normal ones) with a custom HTML attribute or CSS class. And we can leverage this feature from Javascript

josevalim:
Can we instead open up an issue and wait to see if others manifest interest
for this feature? Because I can also see someone submitting a bug report
that links are broken after we merge this. :(
-- 


michallepicki:
Maybe better way to approach this would be to find a minimal (non-meaningful if possible) markdown syntax that paired with ``` `` ``` could mean "Do not try to generate a link for this string, even if you think such module or function exists somewhere". I'm not saying we should start using this style, but at least it would be possible for @eksperimental and others to solve the problem from this PR.

Offtopic: I like the style of Wikipedia, where each reference is linked only once, for the first mention in this article. Such syntax would allow for writing guides in that style.

eksperimental:
@michallepicki It goes against the aim of this PR. I wouldn't want to add any overhead to the developer when writing documentation. I wouldn't want developers to have to remember whether they should add backtick or double backticks depending where they are writing those docs and what is the module they are putting around backticks.
I take it as a workaround hack. Why might be useful depending on the outcome of this feature getting implemented in JS or Elixir.

eksperimental:
@josevalim If this module is getting too complex, is there a way you can see we can split up in two or more modules? it's a sign that this is something we should be doing, instead of limiting functionality, regardless of whether this feature is considered for inclusion.

josevalim:
I agree that we don't want to push this to users but I still question the narrow scope of the proposal to only remove links written as `` `Foo` `` and not all links to the current page. Hand-written links are super uncommon and are prone to the same confusion. If Wikipedia links only the first time around, we could probably do the same?

In any case, regarding the code complexity, you already did a fantastic job in an earlier PR and maybe we could split it into modules. My point though is not that we will never add features, it is just that I personally don't think this feature is worth it, with the current constraints, given the impact it has on the code right now.

eksperimental:
Ok. Let's hold this one off for now then. I will think work on splitting this module and later we can reconsider it.

PR #959, opened 2019-01-04
Unify code styling

### LINKED ISSUE #958: Unify spacing and colors between elixir and other code blocks
Elixir code blocks (` ```elixir `) have `.3em .5em !important` padding while javascript, console, html and others have `.5em 1em` padding. I think bigger padding looks better.

Also, I think elixir "black" should be the same as javascript (and others') dark grey.

Example screenshot (elixir on top, javascript on bottom):
image

michallepicki:
Closes #958

I still feel like the colors used by makeup could be improved but the blacks were bugging me the most :smiley:


assets/less/makeup.less
```diff
@@ -17,6 +9,7 @@ code.makeup {
   }
 }
 
+@hljs-default-text-color: #4d4d4c;
```
josevalim: Can we call this just `@default-text-color`? One day we will remove hljs and this won't make any sense :D Thanks!

michallepicki: @josevalim Done!

> One day we will remove hljs

This will be a great day, I think hljs causes some FOUCs

josevalim:

PR #960, opened 2019-01-07
Reduce hljs FOUC to colors only

michallepicki:
hljs class is being added to code blocks by javascript, after the browser runs hljs code. For slower computers, browsers, or heavier pages, this can be noticeable by a FOUC. Unstyled code blocks have different height than styled hljs code blocks, so the whole page content "jumps". This can be especially annoying when navigating to anchor links. Some browsers would scroll to linked page fragment, and after the quick page re-render, the user would not see the linked page fragment anymore.

We can use the `:not(.inline)` css selector [(supported by all browsers)](https://caniuse.com/#feat=css-sel3) to style code blocks to limit hljs FOUC to just color changes.


michallepicki:
Should have been fine but rebased

josevalim:
I completely forgot about the issues tracker, apologies.

PR #962, opened 2019-01-07
Fix anchor links sometimes not working by moving js to HTML head

### LINKED ISSUE #961: Anchor links sometimes do not show the linked page fragment
In #960 I fixed a hljs FOUC that caused similar behavior:

> Some browsers would scroll to linked page fragment, and after the quick page re-render, the user would not see the linked page fragment anymore.

We still have a problem that randomly causes anchors link not to work. It happens most often when accessing locally-generated documentation. The page would scroll to the top instead the anchor link header.

To reproduce: generate `ex_doc` documentation locally (but remove build badges from the README first, they seem to prevent this bug from happening on this page, I'm not sure yet), open `doc/readme.html#auto-linking` in Chromium, refresh a few times. On my machine, about 1/3 times it would show top of the page.

I think it may be some javascript from HTML's `<head>` that's causing an unfortunately timed re-render.

edit: After some experiments, I found out it's caused by `<script>` near the end of `<body>`. Fixed in #962

michallepicki:
Closes #961

Even an empty script (`<script></script>`) at the end of `<body>` can break anchor links in Chrome for some reason. Fortunately, we can move the script to HTML `<head>` with `async` tag, which should not cause rendering performance to suffer. We can safely do that because our javascript already makes sure to run after the DOM is ready by using jQuery's [ready](https://api.jquery.com/ready/) handler [here](https://github.com/elixir-lang/ex_doc/blob/7ff86d47a7982d72ee16ef26a93cba790a86b355/assets/js/app.js#L34).

### DIFF
--- a/lib/ex_doc/formatter/html/templates/footer_template.eex
+++ b/lib/ex_doc/formatter/html/templates/footer_template.eex
@@ -14,7 +14,6 @@
   </div>
 </section>
 </div>
-  <script src="<%= asset_rev config.output, "dist/html*.js" %>"></script>
   <%# Extra content needed by the current markdown processor (e.g. custom Javascript) %>
   <%= ExDoc.Markdown.get_markdown_processor().before_closing_body_tag(:html) %>
   <%# Extra content specified by the user (e.g. custom Javascript) %>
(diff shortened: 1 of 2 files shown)


OvermindDL1:
Does the javascript in the file wait until the page is fully loaded before attempting to touch the DOM?  If not then moving it to the top could break things randomly on random platforms and browsers dependent on loading speed and so forth.

michallepicki:
> Does the javascript in the file wait until the page is fully loaded before attempting to touch the DOM? If not then moving it to the top could break things randomly on random platforms and browsers dependent on loading speed and so forth.

As mentioned above,

> We can safely do that because our javascript already makes sure to run after the DOM is ready by using jQuery's ready handler

OvermindDL1:
Ah, I entirely missed that when I started looking at the code changes.  ^.^;

josevalim:

PR #970, opened 2019-02-07
Add a versions dropdown to the html sidebar

### LINKED ISSUE #964: A dropdown for selecting package version
Currently if I want to read documentation for some specific version of a package I have to manually edit the url.I think it would be nice if we had a dropdown where it currently shows the version to select the desired one. I can take this one if you guys think that it makes sense

michallepicki:
* add `javascript_config_path` config
* generate versions dropdown using javascript when `versionNodes` is defined

Closes #964 

[demo](https://michallepicki.github.io/ex_doc/doc/)


lib/ex_doc/formatter/html/templates/head_template.eex
```diff
@@ -11,6 +11,9 @@
       <link rel="canonical" href="<%= config.canonical %>" />
     <% end %>
     <script src="<%= asset_rev config.output, "dist/sidebar_items*.js" %>"></script>
+    <%= if config.javascript_script_path do %>
+      <script src="<%= config.javascript_script_path %>"></script>
```
josevalim: What do you think about `javascript_config_path` to indicate this is a path to configure what is happening on the javascript side of things? :D

lib/ex_doc/config.ex
```diff
@@ -30,6 +30,8 @@ defmodule ExDoc.Config do
             groups_for_modules: [],
             groups_for_functions: [],
             homepage_url: nil,
+            # TODO default to a path that can be supported on hexdocs
+            javascript_script_path: nil,
```
josevalim: Would `"config.js"` be a good default?

lib/mix/tasks/docs.ex
```diff
@@ -200,6 +204,37 @@ defmodule Mix.Tasks.Docs do
   the first will take precedence. Functions that don't have a custom group will
   be listed under the default "Functions" group.
 
+  ## Additional javascript script
```
eksperimental:
```suggestion
  ## Additional JavaScript script
```

lib/mix/tasks/docs.ex
```diff
@@ -200,6 +204,37 @@ defmodule Mix.Tasks.Docs do
...
+
+  Since version `0.19.4` ExDoc includes a way to enrich the documentation
```
eksperimental: I think since we are adding a new feature, this shouldn't be a PATCH release, but a MINOR one, so I would change the version to `0.20.0`

lib/mix/tasks/docs.ex
```diff
@@ -200,6 +204,37 @@ defmodule Mix.Tasks.Docs do
...
+  with new information without having to re-generate it, through a javascript
```
eksperimental:
```suggestion
  with new information without having to re-generate it, through a JavaScript
```

lib/mix/tasks/docs.ex
```diff
@@ -200,6 +204,37 @@ defmodule Mix.Tasks.Docs do
...
+  file that can be shared across documentation for multiple versions of the
+  package. If `:javascript_script_path` is configured when building
+  the documentation, this script will be referenced in each page's `<head>`
+  using a `<script>` tag. The script should define data in global javascript
```
eksperimental:
```suggestion
  using a `<script>` tag. The script should define data in global JavaScript
```

lib/mix/tasks/docs.ex
```diff
@@ -200,6 +204,37 @@ defmodule Mix.Tasks.Docs do
...
+  variables that will be interpreted by `ex_doc` when viewing the
+  documentation.
+
+  Currenly supported variables:
+
+  ### `versionNodes`
+
+  This global javascript variable should be providing an array of objects that
```
eksperimental:
```suggestion
  This global JavaScript variable should be providing an array of objects that
```

lib/mix/tasks/docs.ex
```diff
@@ -200,6 +204,37 @@ defmodule Mix.Tasks.Docs do
...
+  define all versions of this mix package which should appear in the package
```
eksperimental:
```suggestion
  define all versions of this Mix package which should appear in the package
```

lib/mix/tasks/docs.ex
```diff
@@ -88,6 +88,10 @@ defmodule Mix.Tasks.Docs do
 
     * `:groups_for_extras`, `:groups_for_modules`, `:groups_for_functions` - See the "Groups" section
 
+    * `:javascript_script_path` - Path of an additional javascript file to be included on all pages
+      to provide up-to-date data for features like the version dropdown - See the "Additional
+      javascript script" section. Example: `"../versions.js"`
```
eksperimental:
```suggestion
      JavaScript script" section. Example: `"../versions.js"`
```

michallepicki:
We will probably need to test if the version dropdown looks good on all browsers (I tested Chrome and Firefox on Linux and while there are slight differences, it looks OK).

Also for generating the dropdown and options I just stitched together some functions and jQuery but I saw that we have some fancy templates - we could use them instead.

josevalim:
> Also for generating the dropdown and options I just stitched together some functions and jQuery but I saw that we have some fancy templates - we could use them instead.

Good call. It looks like the template can cleanup that part considerably. Could you please give it a try? :D

michallepicki:
@josevalim done

assets/js/versions.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Public Methods
+// --------------
+
+export function initialize () {
+  if (typeof versionNodes !== 'undefined') {
+    addCurrentVersionIfNotPresent()
+
+    let versionsDropdown = versionsTemplate({nodes: versionNodes.map(addIsCurrentVersion), currentVersion: currentVersion})
```
michallepicki: `currentVersion: currentVersion` - this is unnecessary, will remove

michallepicki:
I tried out a few browsers that I have access to and overall the feature works everywhere.

One issue that I see with the current styling is that the dropdown arrow is placed slightly differently (the gap between version number and the dropdown arrow is different). The biggest difference is between Firefox and Chromium. I don't think this is a big issue even though I am one of the very few people using Firefox on Android.

Firefox 65 Android:
firefox_android
Firefox 65 Linux:
firefox
Chromium 71 Linux:
chromium
Edge 17 Windows 10:
edge
IE 11 Windows 10:
ie
Safari 11 macOS High Sierra:
safari

Additional issue on IE 11 is not supporting [CSS initial value](https://caniuse.com/#feat=css-initial-value). We could just set the text color to black there but I'm not sure how it would work on systems with dark system theme. Similarily, I don't think this is a big issue and I would leave it as it is.
ie_drop

Gazler:
*Tries demo*
<details><summary>*Clicks asdf*</summary>
:cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: 
</details>

OvermindDL1:
Tried the demo, not seeing a dropdown element at all?  I'm in elinks and I can normally access dropdowns just fine.

michallepicki:
@OvermindDL1 Just to be clear, it doesn't work for you on elinks with javascript compiled in and enabled, right?

OvermindDL1:
@michallepicki I'm using ubuntu's and redhat's stock elinks, which I don't think do:
```zsh
Ubuntu Features:
Standard, IPv6, gzip, bzip2, UTF-8, Periodic Saving, Viewer (Search
History, Timer, Marks), Cascading Style Sheets, Protocol
(Authentication, File, CGI, Finger, FSP, FTP, HTTP, URI rewrite, User
protocols), SSL (GnuTLS), MIME (Option system, Mailcap, Mimetypes
files), LED indicators, Bookmarks, Cookies, Form History, Global
History, Scripting (Lua, Perl), Goto URL History

RedHat Features:
Standard, IPv6, gzip, bzip2, UTF-8, Periodic Saving, Viewer (Search
History, Timer, Marks), Cascading Style Sheets, Protocol
(Authentication, BitTorrent, File, FTP, HTTP, URI rewrite, User
protocols), SSL (OpenSSL), MIME (Option system, Mailcap, Mimetypes
files), LED indicators, Bookmarks, Cookies, ECMAScript (SpiderMonkey),
Form History, Global History, Scripting (Spidermonkey ECMAScript), Goto
URL History
```

Though the 'expandable menu' doesn't work either, just refreshes the page when I try to access one of them.  They really should be embedded lists that are just hidden and shown as desired using pure CSS (I do the same on my sites and they work just fine for 'expandable menus' identically to ex_doc's, except it always shows them all in elinks, meaning they actually work, there's no need for javascript on those...).  But that's a side issue anyway.

Optimally, since I tend to be GUI-less much of my workday, I'd really *really* like programming documentation to actually fully work in a text browser or a simple `curl ... | html_to_text | less` for at least reading or so, which means no required javascript on things that can be done via standard methods, like expandable menu's can be done purely via html+css (even works on IE11 as I have to test for that at work).

michallepicki:
@OvermindDL1 As mentioned in the PR description:
> * add `javascript_config_path` config
> * generate versions dropdown using javascript when `versionNodes` is defined

The main goal here is not having to regenerate old documentation when a new package version is published.

Maybe we could use some JavaScript polyfills for this SpiderMonkey version, or maybe instead we could use an iframe? I'm not sure.

josevalim:

josevalim:
Now we just need to implement this on Hex side of things. :D

Issue #967, opened 2019-01-22
Provide "Search or go to..." functionality


josevalim:
The functionality should mirror the "search or jump to" functionality on GitHub.

Once the user types "foobar" in the search box, the first entry will be "Search foobar..." but below we will show any existing matching entry:

```
 |-----------------------------------|
 | -> | to_charlist                  |
 |-----------------------------------|
 | Search to_charlist...             |
 | Atom.to_charlist/1 (function)     |
 | Kernel.to_charlist/1 (function)   |
 | List.to_charlist/1 (function)     |
 | CharList (module)                 |
 |-----------------------------------|
```

As we press up and down, we can choose the search or to go straight to a given entry. Note this autocomplete functionality is exactly how the search works today, so we can likely just reuse the inner logic.

I am also aware we may need to change how the search is styled today or even move it elsewhere completely. Suggestions on this area are welcome.

Extra: to improve the usability we also should: provide shortcuts to all entries in the top right of the sidebar. For example, pressing s should start "Search". Pressing "c" should "Collapse/expand". And pressing "n" should toggle "Night/day mode".

eksperimental:
I really like the way this issue was solved. Here are my comments.

First, if you move over your mouse horizontally, it looks like it is all the same link, the module title, and the arrow. I think we need to make a distinction, and change the color or do something to the arrow so when you hover it, it is clear they are two different things.

Secondly, I would change to "Go to" title to something more descriptive, like: "Go to module documentation" or something else, no particular preference here.

eksperimental:
I opened it here, since in the new issue, I think it is more oriented to the search feature. Let me know if I should move it there

josevalim:
This issue is about the search functionality. :) The issue about the
feature you are talking about was closed. Perhaps we should open up a new
issue with your comments as feedback? We can call it “feedback about goto
functionality in sidebar”, as I assume others will have more feedback when
we role it out more. Thanks!
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

archdragon:
I've created a quick prototype to get a better sense of how this feature might work - I think it might be a good starting point to get the conversation going :)

![image](http://krowinski.com/ex_doc/autocomplete.gif)

## Demo
➡️ [Example generated with Elixir repo](http://krowinski.com/ex_doc/autocomplete_demo)


## Summary

I think this feature can work well with existing search and page layout.

[...]

PR #974, opened 2019-02-23
full text search

peillis:


assets/js/search.js
```diff
@@ -19,100 +19,82 @@ const $input = $('.sidebar-search input')
...
     })
 
     $input.val(value)
-    $search.html(results)
+    $search.html(resultsHtml)
   }
 }
+
+function getIndex () {
+  if (sessionStorage.getItem('idx') == null) {
```
josevalim: I believe this can be an issue with Hex. For example, if I am on `hexdocs.pm/foo` and go to `hexdocs.pm/bar` in the same tab, they will share the same tab. Maybe, at the same time we define contentsJSON, we could define a variable called contentsScope, such as "elixir-0.1.0", and use that as key.

assets/js/search.js
```diff
@@ -19,100 +19,82 @@ const $input = $('.sidebar-search input')
...
+    createIndex()
+  }
+  return lunr.Index.load(JSON.parse(sessionStorage.getItem('idx')))
```
josevalim: We also need to check what happens if the data is too large to fit the session storage. I assume it will fail to store and getItem will then fail? so it may be better to write this feature in a way that it works even if sessionStorage fails, something like:

```js
var index, stored = sessionStorage.getItem('idx');
if(stored == null) {
  index = createIndex()
  sessionStorage.setItem('idx', JSON.stringify(idx))
} else {
  index = JSON.parse(stored)
}
return lunr.Index.load(index)
```

assets/js/search.js
josevalim: Alternatively we can store the scope in a `<head><meta>` tag, instead of a JS var.

lib/ex_doc/formatter/html.ex
```diff
@@ -119,6 +120,20 @@ defmodule ExDoc.Formatter.HTML do
     [sidebar_items]
   end
 
+  defp generate_contents_json(nodes_map, extras, config) do
```
josevalim: We should probably unify the `sidebar_items` with the `contents_json`. What do you think? But this can also be done in a separate pull request, I thought I would start the discussion though. :)

peillis: Right, that makes sense, since I've had to make many things as in ```sidebar_items```

assets/js/search.js
peillis: Do you mean use the project name as key?

josevalim: name+version, yeah. This key can either be passed as a JS variable OR be extracted from the meta (but we need to add it to the meta in the first place).

peillis:
Ok, I've done the changes you suggested. By the way, I don't know why tests fail on travis, I run them locally without problems.

josevalim:
Beautiful! I am currently on holidays and I will take another look at it
once I am back!
-- 


josevalim:

josevalim:
Amazing work @peillis! Thank you so much!

I have merged this and I have pushed some improvements to master, namely adding a loading state and loading the contentsJSON only in the search page. I think the contentsJSON late loading is important because for Elixir it is more than 1MB. 

Another change I want to do is to remove the dependency on Poison but once that is done, I think we are almost ready to ship a new version. We will just wait for #967 to merged. I am really excited about this release!

peillis:
You are welcome! I'm very proud to help this amazing community.

PR #979, opened 2019-03-05
Added ARIA-label to the search button

kennethlarsen:
This PR adds an aria-label for the search button in the sidebar. This is necessary for people using assistive technologies to understand what the button actually does.


josevalim:
Merged by hand, thank  you for those commits, they are veery appreciated! 

kennethlarsen:
@josevalim Sure thing - I wasn't sure whether or not all of these assets etc. should be included in the PR though.

josevalim:
@kennethlarsen theoretically your compilation should have resulted in the same assets. This means we are either getting different outcomes on every compilation (or on different OSes or JS versions) or that  someone forget to compile them after changing. :) But yeah, they are not necessary in this change  but it is not your fault they were listed as changed. Ideally they wouldn't.

PR #980, opened 2019-03-07
Fix issues with JS events under Firefox

zoldar:
This PR fixes issue with JS events handling under non-webkit browsers. I have noticed this issue under  Firefox 65.0.1 (OSX Mojave). The code in the current shape uses implicit `window.event` which refers to the event in the current scope, which is not available under Firefox (at least in default configuration: https://developer.mozilla.org/en-US/docs/Web/API/Window/event#Browser_Compatibility).


josevalim:

archdragon:
Thanks for catching that!

PR #982, opened 2019-03-09
Keyboard shortcuts - improvements

archdragon:
Updates based on Twitter suggestions - including extra search shortcut and better night mode switching.

image

## Demo

[Demo link](http://krowinski.com/ex_doc/shortcuts_updates_demo/readme.html)

## Changes included

I've noticed some really good suggestions on Twitter, so in this PR I've added:

### 1. `/` shortcut

I've not realized just how common the "vim search" shortcut is. Almost all popular websites support it (ie. Twitter, Youtube), so people expect it to be supported.

image

### 2. Taking into account "conflicting" external shortcuts

When checking keyboard shortcuts I'm now more strict when looking for the match, ignoring the shortcut if any extra/modifier keys are pressed (that `n` != `ctrl + n` != `cmd + n`  != `shift + ctrl + n` != ...). 

image

### 3. Triggering Night Mode only on keyup

Changing between night and day mode only happens after you let go of the `n` key. That way if you keep `n` pressed for a long time, the website won't keep blinking.


josevalim:
> Changing between night and day mode only happens after you let go of the n key. That way if you keep n pressed for a long time, the website won't keep blinking.

This also happens with collapse and the keyboard shortcuts. So I would suggest to probably use keydown for all of and [maybe avoid triggering keydown again until a keyup event](https://stackoverflow.com/questions/17514798/how-to-disable-repetitive-keydown-in-javascript)?

josevalim:

PR #985, opened 2019-03-26
Search autocomplete

archdragon:
Add autocomplete (drop-down with suggestions) to search

image

### **Demo:** [ [Demo link](http://krowinski.com/ex_doc/autocomplete_mod_fun/elixir/Kernel.html) ]

Additional demos:
* [eex](http://krowinski.com/ex_doc/autocomplete_mod_fun/eex)
* [ex_unit](http://krowinski.com/ex_doc/autocomplete_mod_fun/ex_unit)
* [iex](http://krowinski.com/ex_doc/autocomplete_mod_fun/iex)
* [logger](http://krowinski.com/ex_doc/autocomplete_mod_fun/logger)
* [mix](http://krowinski.com/ex_doc/autocomplete_mod_fun/mix)

## Description

* Adds a drop-down with search suggestions
  * Suggestions update as you type
  * If suggestion belongs to a model (i.e. a function) it will display the module name under it.
  * Marks callbacks/types with a special label.
  * Searches through regular modules, behaviours, mix tasks, regular functions, callbacks and types
  * Supports up/down arrow shortcuts and CTRL+ENTER shortcut for opening the result page in new window
  * Handles special characters (ie. searching for `++` or `@`)
  * Displays max 5 results
  * [suggestion spec file](https://github.com/elixir-lang/ex_doc/compare/master...archdragon:autocomplete?expand=1#diff-538692d99bb5f3267fc46cd527d2103e) describes the autocomplete logic in depth 

* Layout updates
  * Night mode button moved to the footer
  * Also in the footer, added a button for displaying keyboard shortcuts help
  * Search box now has a background and is located at the very top of the page

## Full text search compatibility
This feature is fully compatible with the full-text search. I've brought back the `findIn` function  in the `search.js`. While the search page no longer needs it, autocomplete still uses the basic module/function search.

TODO (in scope of this PR):
- [x] Ensure that 'x' always clears the autocomplete.
- [x] Move findIn to the 'suggestions.js'. 
- [x] Ensure that `Module.function` syntax works
- [x] Refactor `findIn` to cleanup and simplify the remaining 'old search' code


assets/js/search.js
```diff
@@ -90,6 +90,71 @@ export function search (value) {
   }
 }
 
+export function findIn (elements, matcher) {
```
josevalim: What if we move those functions to `autocomplete/suggestions.js` since they are not related to the search anymore?

assets/test/search.spec.js
```diff
@@ -1,4 +1,113 @@
 import * as search from '../js/search'
 
 describe('search', () => {
+  describe('findIn', () => {
```
josevalim: If we move findIn to autocomplete, then we need to move these tests too. :)

wojtekmach:
image

How are we sorting results? I think it'd be good if in this case `reduce/2` is towards top of the list. Sorting search results is not an easy problem; I recently saw an article with some heuristics and I'll try to find the link.

image

Seems we're only autocompleting a module **or** a function/type/callback/etc, what do you think about something like above?

assets/js/search.js
archdragon: 👍

josevalim:
Awesome job @archdragon!

I have dropped two comments and I have one last question: should clicking on the "X" clean up the autocomplete/search field? That was my expectation but I am not sure if it is rooted in reality.

josevalim:
> How are we sorting results? 

Once you find the link please open a separate issue. :) I don't want to block this one on that since the behaviour you describe was also available before (but I guess it is somewhat worse here because we limit the results to only 5).


> Seems we're only autocompleting a module or a function/type/callback/etc, what do you think about something like above?

I also spotted this but I opened a separate issue (since the code behaves like this already today): #986.

archdragon:
> How are we sorting results?  

Currently it depends on the suggestion type:

```
  'Module': 3,
  'Child': 2,
  'Exception': 1,
  'Mix Task': 0
```

Higher number == higher on the list. `Child` means anything that belongs to a module (function/callback/type)


> Seems we're only autocompleting a module or a function/type/callback/etc, what do you think about something like above?

That makes sense - the reason it works as it does right now is that I've used existing search functionality for finding matching elements. Now that we have the full-text search (and module/function search functionality is used by autocomplete only) I might be able to add that functionality, but I'll dig a bit and check!

archdragon:
> Awesome job @archdragon!

Thank you!

> ... I have one last question: should clicking on the "X" clean up the autocomplete/search field? That was my expectation but I am not sure if it is rooted in reality.

👍 This is how it's supposed to work - but looks like the input was sometimes triggering the animation/state change before the `click` event could register. Fixed! (and reuploaded the demo)

archdragon:
Demo link now includes the `Mod.fun` search. I left some extra info in the related issue https://github.com/elixir-lang/ex_doc/issues/986

Now cleaning up and refactoring the "old search" code.

assets/less/sidebar.less
```diff
@@ -131,31 +131,75 @@
...
+
+    .search-input {
+      background-color: darken(@darkGray, 8%);
+      border: none;
+      border-radius: 4px;
       color: @lightestGray;
-      margin-left: 5px;
-      height: 20px;
+      margin-left: 12px;
+      margin-right: 10px;
```
josevalim: The most unimportant nitpick on earth: you either don't need the margin (since we have the width) or you should make the margin be 12px so it totals 300px with the width. :D

lib/ex_doc/formatter/html/templates/footer_template.eex
```diff
@@ -7,7 +7,15 @@
           <span class="line">
             designed by
             <a href="https://twitter.com/dignifiedquire" target="_blank" rel="noopener" title="@dignifiedquire">Friedel Ziegelmayer</a>.
-            </span>
+          </span>
+        </p>
+        <p>
+          <button class="line footer-button night-mode-toggle">
```
josevalim: The "toggle night mode" is not working for me on the latest demo (autocomplete_mod_fun).

assets/less/sidebar.less
archdragon: Removed! :D

lib/ex_doc/formatter/html/templates/footer_template.eex
archdragon: Fixed!

archdragon:
I've added the sort we discussed in https://github.com/elixir-lang/ex_doc/issues/987 to the demo. Since I was refactoring some of the "old search" code it was more or less a one-liner. :)

As we discussed, the results are now sorted based on `match chars cout` / `total chars count`. `reduce` (from the @wojtekmach 's example) now looks like this:

image

josevalim:
> As we discussed, the results are now sorted based on match chars cout / total chars count. reduce (from the @wojtekmach 's example) now looks like this:

This is glorious! Please let us know whenever it is ready to merge, as we are pretty much looking forward to it! :D

archdragon:
> This is glorious! Please let us know whenever it is ready to merge, as we are pretty much looking forward to it! :D

Thank you! Just wanted to do some manual cross-browser testing in the morning - I'm now done with that. 

To sum up:
- I've simplified the matches-finding code and added extra comments
- Fixed issues with night mode button not triggering change on really long pages
- Added better sorting
- Added support for "Mad.fun" search

I've updated the demos and uploaded some examples for ex_unit, eex etc. 

From my side the PR is ready for review/merge :)

josevalim:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart: :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart: :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #988, opened 2019-03-30
Push Earmark to v1.3

RobertDober:
Unless there is any reason on your side, I'd like to get Earmark to v1.3.
1.3.2 was released recently and there are no bugs open, so I think it should be a good idea.
Ty in advance.

BTW I can refork if the old commits are an issue.


josevalim:

Issue #978, opened 2019-03-04
Anchor links are broken


eksperimental:
They only work when they are internal links. 
Try the search and click in any result.

ExDoc is unusable with this bug!

Found in Chromium and Firefox

josevalim:
Can you please provide more information? Is it in master? Released code?
How did you try it? Which app? I have no idea how to start looking into
this.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

eksperimental:
It's in master, and with the Elixir documentation

josevalim:
Where did you click? Which anchor links? In the sidebar? Or in the content?

eksperimental:
in the contents of the search.

josevalim:
I still cannot reproduce it.

josevalim:
Can you please describe precisely what you are searching, what you are clicking and what are you expecting to happen? As members of the ExDoc team we should really be giving  better examples of bug reports. :)

eksperimental:
Sorry, I didn't write a deep report, because it seems to obvious. You cannot click anywhere, it will take you to the top of the page. It takes only 1 minute using ExDoc to be annoyed by it.

On the other hand I said "Try the search and click in any result" with that given information, you find the bug. If you cannot reproduce it, that's a different story

josevalim:
It is not obvious. I searched for the letter "a", clicked on the first result, and everything looked OK because I was taken to the page and I could see the function head I clicked on.

Also, because you said ExDoc became unusable, I was expecting nothing to work *at all*, so once the link actually sent me to the page, it seemed ok.

The bug only became apparent once I clicked a definition that would take me to the bottom of the page and noticed that it didn't scroll.

If you spent 30 seconds more on the report it would have saved me 5 minutes or more.

PR #1001, opened 2019-04-14
Speed up Safari scroll

archdragon:
* Reduce safari scroll lag, by removing overflow-y - https://github.com/elixir-lang/ex_doc/issues/851
* Fix https://github.com/elixir-lang/ex_doc/issues/978

Demo: [Demo link](http://krowinski.com/ex_doc/safari_scroll2)

## Description
This PR does two things:
  1. Removes overflow-y from the .content. 
      * overflow-y is not needed, since full-page scroll will work just as well.
      * Safari scroll is much more fluid when we use the full-page scroll instead of the overflow-y.

  2. Fixes an overflow-related issue described in https://github.com/elixir-lang/ex_doc/issues/978
      * We have a helper that focuses `.content` after each navigation. This ensures that pressing space (or an arrow key etc.) scrolls the main content, not just the navigation. 
      * When an element outside the viewport receives focus, browser will scroll to it. This behaviour was the reason why https://github.com/elixir-lang/ex_doc/issues/978 was happening. 
      * Now I'm setting focus on `.content-inner`, which is a DIV that resizes together with the main content (and is always in the viewport). This fixes the scroll/jump issue.

### 1. Safari scroll is more fluid

1. Open Safari
2. In tab A open the [demo page](http://krowinski.com/ex_doc/safari_scroll2), in tab B open the [hexdocs page](https://hexdocs.pm/elixir/Kernel.html)

**EXPECTED:** Demo page scroll is faster and more fluid than Hexdocs page scroll.

### 2. Anchor links still work as expected
1. Go to the [demo page](http://krowinski.com/ex_doc/safari_scroll2)
2. In the sidebar click on the `Float` -> `functions` -> `ratio/1`

**EXPECTED:** You get redirected to the `Float` page - and the page is scrolled to the `ratio` function description.


josevalim:

Issue #998, opened 2019-04-13
Duplicate headings have the same id


rinpatch:
Example:
```
# Changelog
## 1.0
### Changed
- foo
## 0.9
### Changed
- bar
```
In this case both h3s will have `changed` as their id, resulting in inability to link to the second h3.

Git{lab,hub} solves this problem by giving duplicates an id that looks like this `name-<number of name occurences starting from the top>`. So in this case the first h3 wil have `changed` as it's id and the second will have `changed-1`

cybrox:
Looking into this

eksperimental:
Awesome is has been fixed, it has been in my bucket list for ages.
Thank you @cybrox for fixing it and @rinpatch for reporting it

PR #1002, opened 2019-04-24
Ensure that link_headings/3 generates unique anchor id's

cybrox:
This PR ensures that the anchor id's generated by `link_headings/3` are always unique.

This is achieved by replacing the simple use of `Regex.replace/3` with `Regex.scan/2` and an `Enum.reduce/3`, which keeps track of id's that already occurred and adds a `-n` suffix to them.

#998 will be fixed when this is merged.

### DIFF
--- a/test/ex_doc/formatter/html/templates_test.exs
+++ b/test/ex_doc/formatter/html/templates_test.exs
@@ -109,6 +109,20 @@ defmodule ExDoc.Formatter.HTML.TemplatesTest do
              </h3>
              """
     end
+
+    test "generates headers with unique id's" do
+      assert Templates.link_headings("<h3>Foo</h3>\n<h3>Foo</h3>") == """
+             <h3 id="foo" class="section-heading">
+               <a href="#foo" class="hover-link"><span class="icon-link" aria-hidden="true"></span></a>
+               Foo
+             </h3>
+
+             <h3 id="foo-1" class="section-heading">
+               <a href="#foo-1" class="hover-link"><span class="icon-link" aria-hidden="true"></span></a>
+               Foo
+             </h3>
+             """
+    end
   end
 
   describe "synopsis" do
--- a/lib/ex_doc/formatter/html/templates.ex
+++ b/lib/ex_doc/formatter/html/templates.ex
@@ -229,9 +229,19 @@ defmodule ExDoc.Formatter.HTML.Templates do
   def link_headings(nil, _, _), do: nil
 
   def link_headings(content, regex, prefix) do
-    Regex.replace(regex, content, fn match, tag, title ->
-      link_heading(match, tag, title, HTML.text_to_id(title), prefix)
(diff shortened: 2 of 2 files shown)


josevalim:
Hi @cybrox! Can you please remove the changes to the .js/.css files,. since you have not modified them? Thank you.

cybrox:
Hi @josevalim, sure, just force-pushed the branch without those changes.

josevalim:

test/ex_doc/formatter/html/templates_test.exs
```diff
@@ -109,6 +109,20 @@ defmodule ExDoc.Formatter.HTML.TemplatesTest do
...
+
+    test "generates headers with unique id's" do
+      assert Templates.link_headings("<h3>Foo</h3>\n<h3>Foo</h3>") == """
+             <h3 id="foo" class="section-heading">
+               <a href="#foo" class="hover-link"><span class="icon-link" aria-hidden="true"></span></a>
+               Foo
+             </h3>
+
+             <h3 id="foo-1" class="section-heading">
+               <a href="#foo-1" class="hover-link"><span class="icon-link" aria-hidden="true"></span></a>
```
eksperimental: I would love to see the first one as "#foo" and the second one as "foo-2".

cybrox: What would be the benefit? Personally, I think using the same naming scheme as GitHub/GitLab (see #998) makes sense, but I'm open for suggestions :)

eksperimental: The benefit is that we, humans start counting at 1, not 0 as computers do.
I don't know the rationale behind GitHub/GitLab, probably based on the computer language they solved this issue, or because they are just computer geeks.
But if I ever read "#change-2" I don't have to think it will be the third item, and not the second one.
May I ask in reverse, what is the benefit of using a zero-based approach, besides using the same scheme as github/gitlab

cybrox: I agree with the human argument, however, we also like things to be in order, so we'd have to use `#foo-1`, `#foo-2`, etc. which would add a useless suffix to every first (and probably unique) anchor tag.
The way I understood the zero-based approach is that `#foo` is an original heading,  while `#foo-1` is the first reoccurrence of this heading, `#foo-2` is the second [...] and so on.

eksperimental: I don't see a problem using "foo" and "foo-2", most of the time it will only be "foo".
"foo" will always come before any "foo-N" if you sort them.

cybrox: I think both formats have their pros and cons. Since the PR is merged, I don't think we will come to a conclusion here. I guess it's best to leave it at that and open a new PR if a change to "foo/foo2" is desired.

PR #1006, opened 2019-05-16
Detect if browser prefers night mode

btkostner:
Detects if the browser prefers a dark color scheme and switches to dark mode automatically. Needs testing.

Relevant APIs:
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia

### DIFF
--- a/assets/js/night.js
+++ b/assets/js/night.js
@@ -17,12 +17,14 @@ function activateNightMode () {
 
 function deactivateNightMode () {
   body.removeClass(nightMode)
-  try { localStorage.removeItem(nightMode) } catch (e) { }
+  try { localStorage.setItem(nightMode, false) } catch (e) { }
 }
 
 function checkForNightMode () {
   try {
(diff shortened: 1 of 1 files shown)


assets/js/night.js
```diff
@@ -17,12 +17,14 @@ function activateNightMode () {
 
 function deactivateNightMode () {
   body.removeClass(nightMode)
-  try { localStorage.removeItem(nightMode) } catch (e) { }
+  try { localStorage.setItem(nightMode, false) } catch (e) { }
```
josevalim: Is this change necessary or is it a style change?

btkostner: It's necessary to tell if a user intentionally set the theme to day time. That way we can prefer what the user has set, and fall back to what the browser wants.

I just realized this still has a bug, will fix.

josevalim:

PR #1009, opened 2019-05-23
Do not link to modules with `@moduledoc false`

### LINKED ISSUE #951: Linking to non available functions
The Compatibility and Deprecations page of v1.8.0-rc0 mentins that ~`Kernel.ParallelCompiler.files/2`~ `Kernel.ParallelRequire.files/2` is being hard-deprecated, still a link is generated.
the module `Kernel.ParallelCompiler` is marked as `@moduledoc false`,
neither themodule nor the functions are included in the compiled argument that we pass to autolink.
So beats me where this is coming from, posting this before I forget about it.

eksperimental:
Closes #951

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -10,7 +10,7 @@ defmodule ExDoc.Formatter.HTML do
   @doc """
   Generate HTML documentation for the given modules.
   """
-  @spec run(list, ExDoc.Config.t()) :: String.t()
+  @spec run([module], ExDoc.Config.t()) :: String.t()
   def run(project_nodes, config) when is_map(config) do
     config = normalize_config(config)
     config = %{config | output: Path.expand(config.output)}
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -95,6 +95,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
 
     %{
       aliases: aliases,
+      config: config,
       docs_refs: docs_refs ++ types_refs,
       extension: extension,
       lib_dirs: lib_dirs,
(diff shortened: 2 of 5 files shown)


josevalim:
Thanks @eksperimental!

Unfortunately, if we go through the retriever every time we autolink, that's going to be very expensive. I suggest for us to try to read the "Docs" chunk from the .beam file directly without going through the retriever. We should likely also implement a cache, to avoid doing the same lookup over and over again.

eksperimental:
How do we keep state? :ets

eksperimental:
Please have a look again, no need to cache things.

josevalim:
Thanks @eksperimental, this looks much better! I think it can still be simpler though.

Since we only keep `@moduledoc false` for the current project, we actually don't need to know if the module is hidden or not since, if the module does not have the documentation and the module exists in the project, then it is by definition with doc false.

So my suggestion is to break the retriever pipeline, which does: `docs_from_dir |> docs_from_files |> docs_from_modules` into something like this:

    all_modules = Retriever.modules_from_dir(config.source_beam, config)
    docs = Retriever.docs_from_modules(all_modules, config)

I prefer this approach because it is a less invasive change and it means we don't have to change `docs_from_modules` and friends to return a tuple. Can you please investigate? Thank you.

eksperimental:
> Thanks @eksperimental, this looks much better! I think it can still be simpler though.
> 
> Since we only keep `@moduledoc false` for the current project, we actually don't need to know if the module is hidden or not since, if the module does not have the documentation and the module exists in the project, then it is by definition with doc false.
> 
> So my suggestion is to break the retriever pipeline, which does: `docs_from_dir |> docs_from_files |> docs_from_modules` into something like this:
> 
> ```
> all_modules = Retriever.modules_from_dir(config.source_beam, config)
> docs = Retriever.docs_from_modules(all_modules, config)
> ```
> 
> I prefer this approach because it is a less invasive change and it means we don't have to change `docs_from_modules` and friends to return a tuple. Can you please investigate? Thank you.

Ok. no need to create any new function, but if you still want `docs_from_modules` I can provide it in another PR.
yes, this was the way to to go, Apparently I stuck to the tuple return value, probably since when I was trying to support Elixir modules, or even Erlang.

Have a look and tell me what you think.
This is ready to be merged

michalmuskala:
I wonder - should it produce a warning when this situation happens? It seems to me like this can be a "bug" in the documentation when you refer to a private module.

PR #1011, opened 2019-05-24
Fix incorrect id escaping in sidebar

### LINKED ISSUE #1010: Incorrect page section escaping in sidebar
https://hexdocs.pm/phoenix_ecto/main.html#the-phoenix-ecto-integration

image

cybrox:
This will fix #1010 when merged.

Using triple curly braces causes handlebars to render HTML and thus also escaped characters fine. This should not be an issue, because after converting the docs text with earmark and matching for `h2` headings, the id's are run through `HTML.strip_tags/1`, ensuring that there are no actual html tags in them.

The escaping of characters (e.g. `<` to `&lt;`) is done by earmark, an alternative approach would be to re-unencode all escaped characters for the id's, but since `HTML.strip_tags/1` is used anyways, there is no additional security benefit.


josevalim:

PR #1012, opened 2019-05-27
Rename variables

eksperimental:
This is part of #1009, but split in a separate commit to ease the reviewing process
It renames variables since it was a bit of a mess and confusing to work with.


josevalim:
So the reason why it is called `autolink` is because it is about `autolink` information. It makes sense to call it `compiled` in the Autolink module, but calling it `compiled` in that module misses important information once I pass it around, because it removes all context on  what `compiled` is about. A better name would be `compiled_autolink` but that would take too long. So can we please keep the variable name as is? Thank you.

PR #1015, opened 2019-06-09
Do not link to modules with `@moduledoc false`

### LINKED ISSUE #951: Linking to non available functions

eksperimental:
I'm opening a new PR, so this one replaces #1009, and closes #951

While a big amount of code has to be changed (mostly tests), I think it is the way to go.

So before generating the docs, the `ExDoc.Retriever` cache needs to be started by running `ExDoc.Retriever.start_cache/1`, and a tuple with `{response, cache_id}` is returned where `response` is the `Agent.start_link/1` response when the ETS is created, and `cache_id` is an atom that identifies the ETS table.

```elixir
{_response, cache_id} = ExDoc.Retriever.start_cache([])
ExDoc.Autolink.project_doc("`example/2`", nil, %{cache_id: cache_id, locals: ["example/2"]})
```

Then this cache_id needs to be passed to any function in the API.
I decided to take the approach to not work if `cache_id` is not passed, to speed things up.
I made a simpler version that didn't require much refactoring of the tests, and when no cache_id was provided, a new ETS would be created, and this caused a HUGE slow-down in the execution of the tests, from 18s before the PR to 28s after the PR. Now, making sure the Cache is started and passing the cached_id, it's around 21seconds.

Please care only on what's under lib/ and the tests that deals with `@moduledoc false`, the rest is just refactoring to adjust to the changes.

But `ExDoc.generate_docs/3` doesn't need any change and created the cache_id by itself.

### DIFF
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -16,20 +16,12 @@ defmodule ExDoc.Retriever do
   @spec docs_from_dir(Path.t() | [Path.t()], ExDoc.Config.t()) :: [ExDoc.ModuleNode.t()]
   def docs_from_dir(dir, config) when is_binary(dir) do
     dir
-    |> __docs_from_dir__(config)
-    |> reject_docs_with_moduledoc_false()
+    |> files_from_pattern(config)
+    |> docs_from_files(config)
   end
 
   def docs_from_dir(dirs, config) when is_list(dirs) do
-    dirs
-    |> Enum.flat_map(&__docs_from_dir__(&1, config))
-    |> reject_docs_with_moduledoc_false()
-  end
-
-  defp __docs_from_dir__(dir, config) do
-    dir
-    |> files_from_pattern(config)
-    |> docs_from_files(config)
+    Enum.flat_map(dirs, &docs_from_dir(&1, config))
   end
 
   @doc """
@@ -49,20 +41,12 @@ defmodule ExDoc.Retriever do
   def docs_from_modules(modules, config) when is_list(modules) do
     modules
     |> Enum.flat_map(&get_module(&1, config))
-    |> reject_docs_with_moduledoc_false()
     |> Enum.sort_by(fn module ->
       {GroupMatcher.group_index(config.groups_for_modules, module.group), module.id}
     end)
   end
 
   ## Helpers
-
-  defp reject_docs_with_moduledoc_false(docs) do
-    Enum.reject(docs, fn
-      {:module_doc_hidden, _} -> true
-      _ -> false
-    end)
-  end
 
   defp files_from_pattern(dir, config) do
     pattern =
@@ -96,7 +80,7 @@ defmodule ExDoc.Retriever do
         []
 
       :hidden ->
-        %{module_doc_hidden: module}
+        []
 
       docs_chunk ->
         generate_node(module, docs_chunk, config)
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -15,13 +15,21 @@ defmodule ExDoc.Retriever do
   """
   @spec docs_from_dir(Path.t() | [Path.t()], ExDoc.Config.t()) :: [ExDoc.ModuleNode.t()]
   def docs_from_dir(dir, config) when is_binary(dir) do
-    pattern = if config.filter_prefix, do: "Elixir.#{config.filter_prefix}*.beam", else: "*.beam"
-    files = Path.wildcard(Path.expand(pattern, dir))
-    docs_from_files(files, config)
+    dir
+    |> __docs_from_dir__(config)
+    |> reject_docs_with_moduledoc_false()
   end
 
   def docs_from_dir(dirs, config) when is_list(dirs) do
-    Enum.flat_map(dirs, &docs_from_dir(&1, config))
+    dirs
+    |> Enum.flat_map(&__docs_from_dir__(&1, config))
+    |> reject_docs_with_moduledoc_false()
+  end
+
+  defp __docs_from_dir__(dir, config) do
+    dir
+    |> files_from_pattern(config)
+    |> docs_from_files(config)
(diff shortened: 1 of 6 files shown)


josevalim:
Thanks @eksperimental! I have one suggestions for this PR. Passing the cache_id around adds a lot of complexity to the code. If it is a choice between 21s with complex code and 28s with simple code, I prefer the latter. So I would keep the cache global and disable writing to the cache while the test runs.

But even though, In the future we can do neat things, such as warm-up the cache on the test_helper while still keeping it write disable during the suite. Another important factor to consider is the memory usage but there are also options around this. For now the most important is to merge the code and simplifying the cache_id handling is the first step.

eksperimental:
But note that the 28s was with an ETS table per run, so if no cache_id was passed to the public API, as new table would be created and that cache_id was passed around to the private functions. That's probably what we want if we want to simplify things.

But having a global cache would force use to run all the tests serially.

So which way to go?

josevalim:
How long does it take without cache then? Can you also please benchmark the time to generate the docs for Elixir with and without cache?

eksperimental:
> How long does it take without cache then? Can you also please benchmark the time to generate the docs for Elixir with and without cache?

18s

josevalim:
The cost of creating the ets and passing the cache_id is cheap, so the jump from 28 -> 21 is related to the cache exclusively. As I said, for 25% time in tests, I'd prefer to use a global cache. :) But before we take this decision, can you please let us know the time to:

1. build elixir docs in master
2. build elixir docs with the lookup but without cache?
3. build elixir docs with the lookup and with cache?

Thanks!

eksperimental:
> The cost of creating the ets and passing the cache_id is cheap, so the jump from 28 -> 21 is related to the cache exclusively. As I said, for 25% time in tests, I'd prefer to use a global cache. :) But before we take this decision, can you please let us know the time to:
> 
> 1. build elixir docs in master
> 2. build elixir docs with the lookup but without cache?
> 3. build elixir docs with the lookup and with cache?
> 
> Thanks!

sorry, I don't get what is "with the lookup with cache" and "with the lookup with cache"

that 25% is the time that it takes to create a a new ETS table because there was already an ETS created in that test run by `setup` but since I was not passing the cache_id of it, it had to create a new one and keeps track of that one.

eksperimental:
> sorry, I don't get what is "with the lookup with cache" and "with the lookup with cache"

I think I know what you mean, I will tweak the code later and get back to you with the results

eksperimental:
1. build elixir docs in master : 17.2s
2. build elixir docs with the lookup but without cache? 26.4s
3. build elixir docs with the lookup and with cache? 22.4s

josevalim:
Thanks! Since adding the cache doesn't make such a large difference, I would start without the cache for now and we can figure out a way to add a global cache later.

eksperimental:
please have a look

eksperimental:
@josevalim ping

lib/ex_doc/retriever.ex
```diff
@@ -15,13 +15,21 @@ defmodule ExDoc.Retriever do
...
   end
 
   def docs_from_dir(dirs, config) when is_list(dirs) do
-    Enum.flat_map(dirs, &docs_from_dir(&1, config))
+    dirs
+    |> Enum.flat_map(&__docs_from_dir__(&1, config))
+    |> reject_docs_with_moduledoc_false()
+  end
+
+  defp __docs_from_dir__(dir, config) do
```
josevalim: Those changes do not seem to be necessary?

lib/ex_doc/retriever.ex
```diff
@@ -60,10 +91,15 @@ defmodule ExDoc.Retriever do
...
-    if docs_chunk = docs_chunk(module) do
-      generate_node(module, docs_chunk, config)
-    else
-      []
+    case docs_chunk(module) do
+      false ->
+        []
+
+      :hidden ->
+        %{module_doc_hidden: module}
```
josevalim: This also does not seem to be necessary. Seems to be left-over?

test/ex_doc/formatter/html_test.exs
```diff
@@ -537,4 +537,21 @@ defmodule ExDoc.Formatter.HTMLTest do
       end
     end
   end
+
+  describe "does not link to local modules with @moduledoc false" do
```
josevalim: Can we write unit tests instead through autolink_test instead of adding more modules to our integration suite? 🤔

eksperimental: I'd rather not, because we will be modelling `:modules_refs` in config. I don't think a test like that would be of help

lib/ex_doc/retriever.ex
eksperimental: fixed

lib/ex_doc/retriever.ex
eksperimental: fixed

eksperimental:
@josevalim I have removed the left-over code. And left a comment on your suggestion removing the module from the fixtures.

test/ex_doc/formatter/html_test.exs
josevalim: Sorry, I don't understand it. it is probably my bad but can you please explain whta you mean we cannot unit test it? I understand at least ModuleDocFalse is necessary but I don't understanad why we need the other fixture.

eksperimental:
@josevalim Any new on this?

wojtekmach:
Closing in favour of #1132.

PR #1020, opened 2019-06-14
Fix epub toc without groups

msaraiva:
Remove the parent level when there's no group defined.

Before:

image

After:

image


josevalim:

PR #1024, opened 2019-06-17
Fix 'view source' link positioning inside the 'detail-header' element

archdragon:
## Summary

Link to source appears in the wrong place (really visible if a function does not have specs displayed). This is a side-effect of a recent bugfix.

|Before this PR|After this PR|
|---|---|
|image|image|

## Explanation
Bug fixed in https://github.com/elixir-lang/ex_doc/pull/1019 was causing `notes` and `source_url` elements to appear inside the `h1` (since `h1` was not being closed correctly).

|Before #1019|After #1019|
|---|---|
|image|image|

"source_url" is positioned with `float: right` - this worked fine when it was inside the h1, but started causing problems after the fix. When `source_url` is rendered outside the `h1` (a block element), `float: right` causes it to jump to the next line.

Making `h1` an `inline-block` element fixes this problem - `h1` will not take 100% of the width and will not push `source_url` to the next line.


josevalim:

PR #1026, opened 2019-06-18
Add hotkey for version select dropdown

cybrox:
A proposal-PR insipred by [this comment](https://github.com/elixir-lang/ex_doc/issues/1008#issuecomment-501263437) on #1008.

Press <kbd>v</kbd> to select the version dropdown. The select item does not really make it obvious that it has been selected (for the search, pressing <kbd>s</kbd> will show the cursor in the search field). If anyone has a good idea on how to communicate the shortcut action better on a visual level, I'm open for suggestions.

Technical hint: `#version-select` must be evaluated on button-press time, since the handlebars template is dynamically rendered based on the `docs_config.js`.


cybrox:
As a note, the comment originally suggested the shortcut to be <kbd>g</kbd> <kbd>v</kbd> (since <kbd>g</kbd> <kbd>p</kbd> would be used as well). Personally, I think since we have the space, it's probably best to keep shortcuts as easy as possible, hence why it's just <kbd>v</kbd> now.

josevalim:
Beautiful. Can you please rebase? Then I will merge this in immediately. :)

wojtekmach:
> The select item does not really make it obvious that it has been selected (for the search, pressing s will show the cursor in the search field). If anyone has a good idea on how to communicate the shortcut action better on a visual level, I'm open for suggestions.

Yeah, when testing this locally I didn't even realise what's happening, I thought the feature doesn't work but I needed to press e.g. down arrow key for select to show up. I don't have a suggestion how to improve it unfortunately, just that I think we need to do something before landing this.

@archdragon any ideas?

archdragon:
What do you think about a subtle animation to bring attention to the selected element? Something like "pulse" in https://daneden.github.io/animate.css/ (but maybe a bit faster). 

This way when you press `v` you will get immediate feedback and it will be clearer which element got selected.

josevalim:
Another option is to reimplement the select using something simpler since we only use part of the select functionality.

cybrox:
> Yeah, when testing this locally I didn't even realise what's happening, I thought the feature doesn't work but I needed to press e.g. down arrow key for select to show up. I don't have a suggestion how to improve it unfortunately, just that I think we need to do something before landing this.

Yeah, sorry for the confusion but it proves that something needs to be improved

> Another option is to reimplement the select using something simpler since we only use part of the select functionality.

We could implement a custom dropdown but I like `select` for its scrolling behaviour (for many versions) and the fact that it "just works" on a lot of browsers. I'll see if I can do a pulse-like effect with some simple css

cybrox:
I really liked @archdragon's idea with the pulse animation. Changing size is not ideal in the sidebar layout so I tried using color: 

[http://g.recordit.co/gQXMOki00c.gif](http://g.recordit.co/gQXMOki00c.gif)

josevalim:
I think the pulse is an improvement but to be honest I don't think it is enough. Also, given that not all pages with have a version dropdown, it can be even more confusing (you would press `v` and nothing will happen). Maybe if `v` is pressed and there is no dropdown, we should show a message?

Another alternative is to make this a feature of the upcoming package autocomplete. Imagine that the autocomplete starts empty. Once you type "ph", it starts autocompleting only the package names. Then you choose "phoenix". If you then type "/", so you have "phoenix/", it starts autocompleting particular versions for Phoenix. This means that pressing "v" would start the autocomplete for the current package "CURRENT/" as, which means choosing versions.

I guess we could even use delete the version dropdown then but that is a discussion for later. My suggestion for now is to drop this for the time being and focus on the package autocompletion and then revisit this later.

cybrox:
Yeah, ideally the hotkey would actually _open_ the version dropdown but this is not possible with `select`. (At least not reliably, across browsers)

I like the idea of adding it to the auto-complete the way you suggested. I will continue work on that and PR once it's ready.

PR #1031, opened 2019-06-24
Fix: search autocomplete result hovering causes double selection

cybrox:
This is a PR for @archdragon 

Instead of styling the auto-complete element on-hover, it adds/removes the `selected` class via jQuery. This way, it is not possible for the user to have multiple autocomplete results selected.

I noticed this happening when you have your moues idling on the sidebar, hit <kbd>s</kbd>, type something and then use the arrow keys for selecting a result.


archdragon:
Thanks!

I've intentionally made mouse hover independent of the keyboard selection (at first it had a slightly different style, but decided it created too much clutter/confusions :) ) since that felt more natural to me  - but this is just a personal opinion and I don't really have anything agains changing that behaviour! :D

I don't think that an issue that Jose mentioned in https://github.com/elixir-lang/ex_doc/pull/1028 (hover state "sticking around") exists here - at least I can't reproduce it.

cybrox:
I see. Yeah, for me it was a bit confusing at first but I could see why it would be separated.

Thanks for testing. :) I think it's up to the maintainers to decide which behaviour to adopt.

PR #1032, opened 2019-07-04
Fix Night Mode detection

archdragon:
There is a bug with Night Mode switching.

To reproduce:

1. Go to https://hexdocs.pm/elixir/Kernel.html
2. Switch Night Mode `ON`
3. Switch Night Mode `OFF`
4. Refresh the page

**CURRENT:** Night Mode stays ON.
**EXPECTED:** Night Mode stays switched ON.

Local storage value is a truthy string (ie. "false", "true"), not a boolean. `if(localStorage.getItem('night-mode'))` is always `true` if the key is present.


josevalim:

PR #1034, opened 2019-07-06
Run `mix format` with Elixir v1.9.0

eksperimental:


josevalim:
We also need to update travis, done here: #1035

PR #1038, opened 2019-07-07
Rewrite html formatter tests so it avoid future problems with Travis

eksperimental:
Tests will fail in Travis when a PR is on the works, and master is ahead
of the PR and new JS and CSS assets have been introduced to master.
You can see the errors here: https://travis-ci.org/elixir-lang/ex_doc/jobs/555347013#L422-L422
Travis fetches the PR and applies it to master.

So one way to solve this is to not to check for only one JS or one CSS file,
but to check for at least having one.


eksperimental:
Well, now looking at it, We can add the original test and execute it only when not being run on Travis. I will look into the way to do this

eksperimental:
Please do not merge as I want to update a commit to make sure it fails locally but not on Travis. But I cannot work on it at the right now.

josevalim:
I don't think we should change this. Travis-CI is correct in failing, you have to rebase and rebuild the JS, otherwise it means you have outdated JS code.

PR #1042, opened 2019-07-11
Encode URL fragments with a specific function

### LINKED ISSUE #1041: Links to functions containing `&` characters do not fully work
Links to functions containing `&` will direct to the top of the module page (instead of scrolling to the relevant function description).

Took a quick look at it and from what I can see this happens because we're getting `&amp;` in the hash/link and just `&` in the anchor.

To reproduce:

1. Go to https://hexdocs.pm/elixir/Function.html
2. Click on `&/1`

**Current behaviour**: You are being redirected to the top of the `SpecialForms` module page
**Expected behaviour**: You are being redirected to the top of the `&/1` function description on the `SpecialForms` module page

eksperimental:
Closes #1041

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -251,7 +251,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
               put_placeholder(form, url, placeholders)
 
             {name, arity} in typespecs ->
-              n = enc_h("#{name}")
+              n = URI.encode("#{name}")
               url = "#t:#{n}/#{arity}"
               put_placeholder(form, url, placeholders)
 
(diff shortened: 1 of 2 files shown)


cybrox:
One note: The templating adds additional anchors if the `is_enc?` check returns true, which uses `h()` and `enc_h()` for checking the encoded and unencoded names respectively. Could this cause false-positives, because `enc_h()` still encodes values improperly?

I think this issue could also be fixed by just removing the call to `h()` from `enc_h()` [here](https://github.com/elixir-lang/ex_doc/blob/c8f2a7d0ef399dfc0a6fbb20326c908deb3ce346/lib/ex_doc/formatter/html/templates.ex#L90-L94).

josevalim:
@cybrox I have merged and applied your suggestions, thanks!

Issue #1036, opened 2019-07-07
ExDoc should deal with fully namespaced module names


eksperimental:
I think a reference to `Module` and `Elixir.Module` should link to `Module.html`,
same for `Elixir.Module.function/1`

eksperimental:
@josevalim regarding the commit 098b99fcec956a55d3feeae47957ff76e5269d56,
I think if the developer explicitely wrote the fully namespaced module, it should be kept.

It will defeat the purpose of for example this text:

```elixir
"""
The fully namespaced module `Elixir.Enum` and the alias `Enum` resolve to the same module.
"""
```

```
$ mix test test/ex_doc/formatter/html/autolink_test.exs 
.............

  1) test corner-cases removes optional Elixir namespace (ExDoc.Formatter.HTML.AutolinkTest)
     test/ex_doc/formatter/html/autolink_test.exs:639
     Assertion with == failed
     code:  assert project_doc("The fully namespaced module `Elixir.Enum` and the alias `Enum` resolve to the same module.", %{}) == "The fully namespaced module [`Elixir.Enum`](#{@elixir_docs}elixir/Enum.html) and the alias [`Enum`](#{@elixir_docs}elixir/Enum.html) resolve to the same module."
     left:  "The fully namespaced module [`Enum`](https://hexdocs.pm/elixir/Enum.html) and the alias [`Enum`](https://hexdocs.pm/elixir/Enum.html) resolve to the same module."
     right: "The fully namespaced module [`Elixir.Enum`](https://hexdocs.pm/elixir/Enum.html) and the alias [`Enum`](https://hexdocs.pm/elixir/Enum.html) resolve to the same module."
     stacktrace:
       test/ex_doc/formatter/html/autolink_test.exs:648: (test)
```

josevalim:
@eksperimental we can do that as well. A new PR is welcome!

eksperimental:
I will work on it

eksperimental:
Should we close this issue, or there's anything else that needs to be covered?

PR #1044, opened 2019-07-12
Do not strip "Elixir." from text, in fully namespaced calls

eksperimental:
This is a continuation of 098b99fcec956a55d3feeae47957ff76e5269d56

Related issue: #1036

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -399,8 +399,9 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     modules_refs = options[:modules_refs] || []
 
     fn all, text, match ->
-      pmfa = {_, module, _, _} = split_match(:module, match)
+      pmfa = {_, original_module, _, _} = split_match(:module, match)
       text = default_text("", link_type, pmfa, text)
+      module = strip_elixir_namespace(original_module)
 
       cond do
         module == module_id ->
@@ -430,9 +431,11 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     module_id = options[:module_id]
     skip_warnings_on = options[:skip_undefined_reference_warnings_on] || []
 
-    fn all, text, match ->
-      pmfa = {prefix, module, function, arity} = split_match(:function, match)
+    fn all, text, original_match ->
+      pmfa = {prefix, original_module, function, arity} = split_match(:function, original_match)
       text = default_text("", link_type, pmfa, text)
+      match = strip_elixir_namespace(original_match)
+      module = strip_elixir_namespace(original_module)
 
       cond do
         match in locals ->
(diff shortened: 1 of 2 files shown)


test/ex_doc/formatter/html/autolink_test.exs
```diff
@@ -636,14 +636,31 @@ defmodule ExDoc.Formatter.HTML.AutolinkTest do
...
 
       assert project_doc("`Elixir.Enum.concat/1`", %{}) ==
-               "[`Enum.concat/1`](#{@elixir_docs}elixir/Enum.html#concat/1)"
+               "[`Elixir.Enum.concat/1`](#{@elixir_docs}elixir/Enum.html#concat/1)"
 
       assert project_doc("`t:Elixir.Enum.t/0`", %{}) ==
-               "[`Enum.t/0`](#{@elixir_docs}elixir/Enum.html#t:t/0)"
+               "[`Elixir.Enum.t/0`](#{@elixir_docs}elixir/Enum.html#t:t/0)"
+
+      assert project_doc(
```
wojtekmach: what does this assertion and the one below test that we don't already have covered?

josevalim: Yes, those assertions can be removed. :)

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -399,8 +399,9 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     modules_refs = options[:modules_refs] || []
 
     fn all, text, match ->
-      pmfa = {_, module, _, _} = split_match(:module, match)
+      pmfa = {_, original_module, _, _} = split_match(:module, match)
```
josevalim: I don’t understand why we removed the stripping from the split_match functions, since we don’t use original_module.

eksperimental: We use original_module to build the text. That's the whole point of not stripping it in split_match but outside of it.

test/ex_doc/formatter/html/autolink_test.exs
eksperimental: testing this change
https://github.com/elixir-lang/ex_doc/pull/1044/files#diff-4cca810c5bca54191f6de3e9b8edafb0L557

lib/ex_doc/formatter/html/autolink.ex
josevalim: Ah, sorry. I somehow missed it. :D

test/ex_doc/formatter/html/autolink_test.exs
josevalim: This case is tested on the assertion above. Still, if there are new cases necessary, I would add simple assertions like the ones above. :)

josevalim:

PR #1049, opened 2019-07-17
make use of the makeup lexer registry

tmbb:
Because I've been too lazy to write new makeup lexers, I haven't actually tested the possibility of using lexers for languages other than Elixir. All current tests pass, of course.

### DIFF
--- a/lib/ex_doc/highlighter.ex
+++ b/lib/ex_doc/highlighter.ex
@@ -19,9 +19,17 @@ defmodule ExDoc.Highlighter do
   def before_closing_body_tag(_), do: ""
 
   # If new lexers are available, add them here:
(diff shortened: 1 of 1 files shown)


tmbb:
@josevalim I have no idea why the build fails with OTP 22. It has probably something to do with the `makeup` application not being started properly (it creates the lexer registry on startup), but I don't know why it should be started in previous OTP versions and not when using OTP 22...

josevalim:
@tmbb we run some checks only on latest Elixir/OTP, but it is a general issue. I will work on a fix. :)

josevalim:
@tmbb can you please rebase? Then it should work!

josevalim:
Nevermind, merged manually. Thank you!

josevalim:

Issue #1050, opened 2019-07-17
Potentially remove the requestId on iframes for caching


josevalim:
Hi @archdragon!

You use a `requestId=` when loading the tooltip iframe. Unfortunately that makes it impossible for the page to be cached. Can you explain why the requestId is necessary? Do you know of any other options?

Thank you!

archdragon:
Thank you for catching that  - I should have not trouble removing it. I've added it for two reasons:

First - to avoid race conditions. Sometimes I'll see an invalid response when I: 1) hover over a link `A`, 2) hover over a link `B` and 3) `A` sends the answer back just as I start the `B` hover. But that can be fixed just by checking if the requested function/module matches.

Second - to make sure I'm not getting an unwanted response. I'm disregarding anything that does not matches my latest request ID. But I think I can achieve the same just by checking if the  module/function matches and accepting only one matching response.

I think this can be safely removed (with a few extra changes to the code) - I'll prepare a PR.

josevalim:
> But I think I can achieve the same just by checking if the module/function matches and accepting only one matching response.

Yeah, I agree. I actually think we can just match the URL. See if the URL in src is the same as in the frame?

archdragon:
> Yeah, I agree. I actually think we can just match the URL. See if the URL in src is the same as in the frame?

I'm on it - it's a bit more involved because of the postMessage, which does not give me a full url directly.

josevalim:
@archdragon you can pass the URL as parameter if you want. As long as the parameter is static, we are golden. You may need to escape the URI characters on it though.

archdragon:
@josevalim 

👍

I've made it a bit simpler and prepared a PR here https://github.com/elixir-lang/ex_doc/pull/1054 - I'm sending the page URL back (next to the hint info) and check if it matches the one we just hovered over.

The URL stays static and cache works.

princemaple:
Why did it have to be using an iframe? I'm sure there has been some discussion I didn't see, if I could get a pointer :)

Looking at all the things being passed around I just want to change it to use `(await fetch('https://hexdocs.pm/elixir/master/Kernel.html#content')).text()` and call the extraction functions.

princemaple:
That said, I confirm that the cache works :)

archdragon:
@princemaple Thanks for checking! 

Here is the discussion that contains some info and requirements for the popovers https://github.com/elixir-lang/ex_doc/issues/971 :)

[...]

PR #1053, opened 2019-07-17
Avoid re-fetching the same page

princemaple:
I was going to eliminating more requests as I stated in
https://twitter.com/PrMaple/status/1151430729026269184

However, the way of fetching whole pages via iframe has its limits.
While not having to build extra infrastructure server side to support
this functionality, I cannot optimize away the extra requests being
sent out in the iframe.

As a compromise, I think it's the least we can do to not re-request
the already fetched data. (~50KB transferred each hover)

Code is only tested in my head, if someone could test it for me or
point me to how to test it properly I'd appreciate it!

### DIFF
--- a/assets/js/tooltips/tooltips.js
+++ b/assets/js/tooltips/tooltips.js
@@ -223,7 +223,7 @@ function prepareTooltips () {
 
   if (isSelfLink(href)) { return }
 
-  currentCacheKey = href;
+  currentCacheKey = href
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks @princemaple. My suggestion would be to first try #1050 and see if that boils it down to one cached request. If it doesn't, then we can try the approach here.

PR #1054, opened 2019-07-17
Tooltips - Don't use dynamic requestId to avoid re-fetching the page

archdragon:
Take advantage of cache by removing the dynamic requestId param from the tooltip URL

Demo:
http://krowinski.com/ex_doc/no_request_id/elixir/Kernel.html

image


josevalim:

Issue #1058, opened 2019-07-19
Closing "package switcher" with Esc blocks it from opening again


wojtekmach:
Afterwards we can't open "?" etc.

cc @cybrox

archdragon:
I've tried to reproduce it (on https://hexdocs.pm/elixir/master/ ), but can't.

Did you take any extra steps between opening the switcher and closing it?

wojtekmach:
No, all I'm doing is: <kbd>g</kbd> <kbd>Esc</kbd> <kbd>g</kbd>

But thanks for checking in, I just found out the bug is on Safari but not on Firefox!

archdragon:
@cybrox @wojtekmach I think I know what might be the case, since something similar happened to me in autocomplete. Will prepare a quick PR and demo to test this. :)

PR #1060, opened 2019-07-19
Allow "package switcher" re-opening after pressing Esc

archdragon:
Defocus input field when closing the switcher window. 
Otherwise Safari will keep input focused and won't catch new keyboard shortcuts.

Fixes: https://github.com/elixir-lang/ex_doc/issues/1058

@wojtek - can you check with the demo?

Demo link: http://krowinski.com/ex_doc/safari-switch/readme.html


wojtekmach:
Perfect, that fixes it!

PR #1065, opened 2019-07-24
Use EDoc link for erlang type

### LINKED ISSUE #1005: Can't link to remote types in Erlang projects
Start a new project:

```bash
mix new demo && cd demo
```

Replace `mix.exs` with:

```elixir
defmodule Demo.MixProject do
  use Mix.Project

  def project do
    [
      app: :demo,
      version: "0.1.0",
      elixir: "~> 1.8",
      start_permanent: Mix.env() == :prod,
      deps: [
        {:dialyxir, "~> 0.5.1"},
        {:ex_doc, "~> 0.20.2"},
        {:opencensus, "~> 0.9.2"}
      ]
    ]
  end

  def application, do: [extra_applications: [:logger]]
end
```

Replace `lib/demo.ex` with:

```elixir
defmodule Demo do
  @moduledoc """
  Demonstrates a problem with links to remote types.
  """

  @doc """
  Do nothing with a `t::opencensus.span/0`.
  Or, is it a `t:opencensus.span/0`?
  """
  @spec hello(:opencensus.span() | nil) :: :opencensus.span() | nil
  def hello(span), do: span
end
```

Build the documentation and open it:

```bash
mix do deps.get, deps.compile, docs
open doc/Demo.html
``` 

Observe:

* There doesn't appear to be a way to link from the `@doc` string to the remote type.

* The description of `hello` appears to correctly describe the types of the argument and return value:
  > `hello(span)`
  > `hello(:opencensus.span() | nil) :: :opencensus.span() | nil`

* When you follow the link from `:opencensus.span()`, you get a [broken link](https://hexdocs.pm/opencensus/0.9.2/:opencensus.html#t:span/0), not a [working link](https://hexdocs.pm/opencensus/opencensus.html#type-span)

Double-check with Dialyzer:

```bash
mix dialyzer
```

If you see this, you're going to have long enough to grab a coffee, eat dinner, or perhaps even cook dinner:
> ```
> Adding 903 modules to dialyxir_erlang-21.3.6_elixir-1.8.1_deps-dev.plt
> ```

I've got something else to get to, so I'll click Submit and report back later with how it went. From what happened in my main project, however, I expect a complaint about the missing export in `opencensus` (census-instrumentation/opencensus-erlang#152) but not the reference to `:opencensus.span()`.

chulkilee:
Fixes #1005

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -285,7 +285,17 @@ defmodule ExDoc.Formatter.HTML.Autolink do
 
   defp type_remote_url(source, alias, name, args) do
     name = enc("#{name}")
-    "#{source}#{enc(inspect(alias))}.html#t:#{name}/#{length(args)}"
+
+    if erlang_alias?(alias) do
+      "#{source}#{enc(to_string(alias))}.html#type-#{name}"
+    else
+      "#{source}#{enc(inspect(alias))}.html#t:#{name}/#{length(args)}"
+    end
+  end
+
+  defp erlang_alias?(alias) do
+    first_char = alias |> to_string() |> String.at(0)
+    first_char == String.downcase(first_char)
   end
 
   defp typespec_string_to_link(string, url) do
--- a/test/ex_doc/formatter/html/autolink_test.exs
+++ b/test/ex_doc/formatter/html/autolink_test.exs
@@ -542,9 +542,14 @@ defmodule ExDoc.Formatter.HTML.AutolinkTest do
                ~s[<a href=\"#{@erlang_docs}sets.html#type-set\">:sets.set</a>(<a href=\"#t:foo/0\">foo</a>())]
     end
 
-    test "autolinks shared aliases" do
+    test "autolinks shared Elixir aliases for ex_doc" do
       assert Autolink.typespec(quote(do: Foo.t()), [], [Foo]) ==
                ~s[<a href="Foo.html#t:t/0">Foo.t</a>()]
+    end
+
+    test "autolinks shared Erlang aliases for EDoc" do
+      assert Autolink.typespec(quote(do: :foo.t()), [], [:foo]) ==
+               ~s[<a href="foo.html#type-t">:foo.t</a>()]
     end
 
     test "autolinks inside parameterized types" do


lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -285,7 +285,17 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
-    "#{source}#{enc(inspect(alias))}.html#t:#{name}/#{length(args)}"
+
+    if erlang_alias?(alias) do
+      "#{source}#{enc(to_string(alias))}.html#type-#{name}"
+    else
+      "#{source}#{enc(inspect(alias))}.html#t:#{name}/#{length(args)}"
+    end
+  end
+
+  defp erlang_alias?(alias) do
```
chulkilee: I found ex_doc uses compiled regex - so probably we may use regex here? Any feedback?

josevalim: I think this is fine as is.  Can you please add a test?

chulkilee: Added the test

chulkilee:
@wojtekmach  Oh, I didn't know that's possible (`defmodule :foo`). However, how can we know whether the referenced name is erlang or elixir? That seems impossible without this heuristic..

I'm not sure looking into `alias` is a good idea actually. Probably the language should be passed down to `type_remote_url` function - but I'm not sure that's even possible.

josevalim:
In this case wouldn’t it be better to look at the original text and see if
it stars with “:”?
-- 


wojtekmach:
> However, how can we know whether the referenced name is erlang or elixir? That seems impossible without this heuristic..

As @josevalim wrote [here](https://github.com/elixir-lang/ex_doc/issues/1005#issuecomment-493106281)

> > Would checking in deps for mix.exs or rebar.config be out of the question?

> I actually like this idea. the Mix integration already looks for dependencies in there. We can possibly see if the manager is :mix and not include the dependencies otherwise.

so it's not a matter of whether it's Erlang or Elixir, but what tool was used to build the documentation for that particular module.

As José mentioned elsewhere, given different Erlang projects might use different tools, it's a safer bet to only use EDoc style urls for Erlang stdlib and leave others not linked.

chulkilee:
> As José mentioned elsewhere, given different Erlang projects might use different tools, it's a safer bet to only use EDoc style urls for Erlang stdlib and leave others not linked.

@josevalim mentioned that at https://github.com/elixir-lang/ex_doc/issues/1066#issuecomment-514822870

I'm going to close this PR as it's not clear "adding more link to external erlang project" is the "right" direction to go at this moment.

Issue #1070, opened 2019-07-29
Hide packages without docs in quick-switch


cybrox:
<kbd>g</kbd> + `cowboy` will get search results for cowboy (and some `cowboy_` packages) but none of these have their docs on HexDocs.

Since the feature is now named "Go to a HexDocs package", should we just filter out those entries?

If we decide to do so, I tried to filter by the `releases` key returned from the hex API but all packages seem to have releases with the property `has_docs: false`, even the ones that are on HexDocs. Is there any reliable information on wether or not a package is on HexDocs in the payload?

wojtekmach:
> but all packages seem to have releases with the property `has_docs: false`, even the ones that are on HexDocs

that's a Hex bug, I'm looking into it.

wojtekmach:
@cybrox the fix is live on hex.pm.

cybrox:
@wojtekmach Thanks a lot!

Could it be that some search queries are cached? Querying `ecto*` returns has_docs as false for all the releases of `ecto_audit`, however, querying for `ecto_audit*` returns true for the latest 3 releases. 

If it's just cache, I'll PR the change for quick-switch, it will take a while to be included in an ex_doc release anyways :)

wojtekmach:
Yup, forgot to mention that we have a cache so you might see stale data!

cybrox:
Solved by #1071

PR #1071, opened 2019-07-29
Filter out packages without docs on HexDocs in quick-switch

cybrox:
Fixes #1070 by appending an additional filter to the search results.


josevalim:

Issue #1079, opened 2019-08-15
Support function expressions with multiple arities `fun/1,2`


fertapric:
As discussed in [the Elixir mailing list](https://groups.google.com/forum/#!topic/elixir-lang-core/75Rz_Ds-aEk/discussion).

The first arity in the expression should be the one used to create the link (the Erlang documentation seems to do the same). For example `Module.fun/1,2` would link to `Module.html#fun/1`; and `Module.fun/4,3` to `Module.html#fun/4`. This option allows developers to control were the link is going to point to.

cc @eksperimental

josevalim:
@fertapric I am not a big fan of the inversed arity. The lack of standard will be confusing for readers.  My suggestion is to always point to the last one and you can use a custom link if you want to change it.

cybrox:
I could take a look at this. :)

@josevalim Regarding the _not having a standard_ part, I agree that always linking to the first or last makes sense. For consistency, would it make sense to sort the arities in ascending order, if a user chooses to write `Module.fun/4,3` in their docs or would you just leave that as-is?

fertapric:
> For consistency, would it make sense to sort the arities in ascending order, if a user chooses to write Module.fun/4,3 in their docs or would you just leave that as-is?

@cybrox if we want to make the standard stricter and not support inversed arity (or random arity `fun/2,1,3`, then I would only link if the arities are in ascending order.

fertapric:
@josevalim I suggested the first arity because:

* that's [how it works in the Erlang documentation](http://erlang.org/doc/man/io_lib.html#fwrite-2)
* it feels more natural to scroll down to see the rest of arities than scrolling up

cybrox:
@fertapric Agreed. Also, in a lot of cases the lowest arity version of a function is the "common" one. (e.g. `fwrite/2` for the 'generic' behaviour, `fwrite/3` for when you want to 'customize' it more)

eksperimental:
We may not necessarily have only one link.

The way it should work IMO is to have different links but compact the links as much as possible. It should be based on the list of functions with arities that we see in the side bar and compact.

For instance, we get on the side bar for Enum:
* chunk_every/2
* chunk_every/4

So if we write `Enum.chunk_every/2,4`, then `Enum.chunk_every/2` should link to `Enum.html#chunk_every/2` and `,4` should link to `Enum.html#chunk_every/4`.

[...]

PR #1080, opened 2019-08-15
Support autolinking for multiple arities

cybrox:
This addresses #1079.

Since @eksperimental refactored a lot of this code not too long ago, I wanted to PR this WIP to check if I'm following the imagined structure or implemented a completely off-track solution.

Things that are currently missing:
- ~~Supporting multiple arities for built-in types and kernel functions~~
- Any additional work that is needed for making this work on epubs(?)

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -395,7 +395,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
             "[#{text}](#{doc}#{module}.html)"
 
           :function ->
-            "[#{text}](#{doc}#{module}.html##{function}-#{arity})"
+            "[#{text}](#{doc}#{module}.html##{function}-#{first_arity_of(arity)})"
         end
       else
         all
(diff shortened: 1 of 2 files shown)


test/ex_doc/formatter/html/autolink_test.exs
```diff
@@ -90,6 +104,11 @@ defmodule ExDoc.Formatter.HTML.AutolinkTest do
       assert project_doc("`Mod.example/2`", %{docs_refs: ["Mod.example/2"]}) ==
                "[`Mod.example/2`](Mod.html#example/2)"
 
+      multiple_refs = %{docs_refs: ["Mod.example/1", "Mod.example/2"]}
```
cybrox: Need to move this into the `project_doc`, this was for testing.

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -596,20 +608,29 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
+    {"", Enum.join(mod, "."), hd(name), split_arities(arity)}
   end
 
   # handles "/" function
   defp split_function_list([modules, "", arity]) when is_binary(modules) do
     split_function_list([modules <> "/", arity])
   end
 
+  defp split_arities(arities) do
+    if String.contains?(arities, ","), do: String.split(arities, ","), else: arities
```
josevalim: Doesn't this mean someitmes it is a list and sometimes it is a string? If so,  I would prefer to always make it a non-empty list. It would require less conditionals in the code.

cybrox: It does and you're absolutely right. I am working on changing this, since it would also cause a lot more unnecessary code when supporting the same for built-in types/functions/callbacks.

cybrox: Changed `split_arities/1` to always return a list.    
Consuming functions will use helpers like `first_arity_of/1` now.

lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -596,14 +613,22 @@ defmodule ExDoc.Formatter.HTML.Autolink do
...
   end
 
   # handles "/" function
   defp split_function_list([modules, "", arity]) when is_binary(modules) do
     split_function_list([modules <> "/", arity])
   end
 
+  defp split_arities(arities), do: String.split(arities, ",", trim: true)
+
+  defp print_arities(arities), do: arities |> Enum.join(",")
```
wojtekmach:
```suggestion
  defp print_arities(arities), do: Enum.join(arities, ",")
```

test/ex_doc/formatter/html/autolink_test.exs
```diff
@@ -10,6 +10,9 @@ defmodule ExDoc.Formatter.HTML.AutolinkTest do
     test "autolinks fun/arity in docs" do
       assert project_doc("`example/2`", %{locals: ["example/2"]}) == "[`example/2`](#example/2)"
 
+      assert project_doc("`example/1,2`", %{locals: ["example/1", "example/2"]}) ==
+               "[`example/1,2`](#example/1)"
```
wojtekmach: could you add a test when we reference `example/1,2` but only `example/1` function exists? I think the behaviour should be (untested):

```elixir
assert project_doc("`example/1,2`", %{locals: ["example/1"]}) ==
         "`example/1,2`"  
```

What do you think?

In some cases we also trigger warnings if the function we try to link to doesn't exist (not for local functions though) so it would be good to test that with multiple arities too.

cybrox: Definitely! I did put this on my to-do list but somehow overlooked it :disappointed:   
Added a test for missing one of two/three function definitions and a test for the warning emitted when the module exists but [one of the] function[s] does not.

josevalim:

PR #1082, opened 2019-08-20
Readme: Add link to main configuration listing

axelson:
I think it's important to have a link to the page with the full configuration, otherwise it is difficult to find because it isn't explicitly linked anywhere.

A relative link might be better, but since this is a plain markdown file I'm not aware of a method that would work.


josevalim:

axelson:
Thanks! :heart:

PR #1083, opened 2019-08-23
add support for source_repo_type configuration option

aseigo:
this allows marking a custom source_url as being of type
gitlab/github/bitbucket without resorting to defining a pattern

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,19 +14,19 @@ To see all supported options, see the documetation for [mix docs](https://hexdoc
 ExDoc ships with many features:
 
   * Automatically generates HTML and EPUB documents from your API documentation
-  * The generated HTML documentation is accessible online and offline
+  * Support for custom pages and guides (in addition to the API reference)
+  * Support for custom grouping of modules, functions, and pages in the sidebar
+  * Generates HTML documentation is accessible online and offline
   * Responsive design with built-in layout for phones and tablets
   * Customizable logo on the generated documentation
-  * Support for custom pages and guides (in addition to the API reference)
-  * Support for custom grouping of modules and functions in the sidebar
+  * Each documented entry contains a direct link back to the source code
   * Full-text search
   * Keyboard shortcuts (press `?` inside an existing documentation to bring the help dialog)
-  * Quick search with autocompletion support on the API reference (`s` keyboard shortcut)
+  * Quick search with autocompletion support (`s` keyboard shortcut)
   * Go-to shortcut to take to any HexDocs package documentation with autocomplete support (`g` keyboard shortcut)
   * Support for night-mode (automatically detected according to the browser preferences)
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @aseigo, it is not clear to me how this would be used and what are the benefits over the current approach. Can you please clarify?

wojtekmach:
@josevalim it's when e.g. Phoenix code becomes hosted on `git.phoenixframework.com` - based on the source_url we can't tell whether it's using GitHub, GitLab etc, and we need to know to be able to generate source links.

josevalim:
This is just a corner case though... I would rather go ahead and specify the whole pattern.

wojtekmach:
Oh, we already allow to specify the whole pattern. Agreed, I think that's a decent option.

aseigo:
Yes, it is a corner case in that it won't be used by a majority, but then it would not be first such option supported by ExDoc, either. The rational I had for this was:

Products like gitlab are designed to allow self-hosting. This is a common use case, and I have worked on code across a few different instances of it over time with different organizations. The pattern matching for the well-known software is done based on URL, though obviously that same technology lives elsewhere. 

For those in that situation, It is both more reliable and easier to just say "this is github". What, exactly, is the "full pattern", and how should it be constructed? Why should the developer know how gitlab formats its URLs when the information is already right there in ExDoc? Instead of remembering the long pattern (both url and variables for interpolation), it's easier to just remember "this is gitlab" (or whichever solution is in use). This feel smore natural from a developer POV as this is a supported (community and commercially) option for the software, of which ExDoc already knows about.

This is indeed not a required feature. It is an ease of use and minor "quality of living" improvement.

If you feel it is too much of a corner case, that is, of course, no problem :) But the above was my rational for doing this.

Cheers, and have an amazing weekend!

josevalim:
What if we include in the docs explicit examples on how to configure self-hosted repos for each of those choices? Then it is also quite accessible without increasing the API surface. Thoughts?

josevalim:
And have a great weekend too!

aseigo:
There is an explicit example in the docs, already, though only for github. Gitlab's are pretty well the same, though that isn't explicitly noted. Bitbucket's are a bit different (and they also support on-premise / self-hosting). So, I suppose we could add all three .. but then docs need to be updated if/when code is, though I suspect that would be a very unlikely occurance (too many things use those links in the wild for them to change, probably).

Probably works either way and is a matter of maintainer (you :) preference ... Cheers!

josevalim:
Right. But if they change the URL, then we need to update the code too and
between those the docs are easier to handle. So if you could send a PR
clarifying the docs, it would be very welcome, thank you!
-- 


aseigo:
Ok, will do on Monday ... (am away from my coding nest until then :)

PR #1092, opened 2019-09-10
Change autolink placeholder to something less likely

### LINKED ISSUE #1091: ex_doc 0.21.2: "@spec x_p0_x() :: any" crashes mix docs
Erlang/OTP 22 [erts-10.4.2] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1]
Elixir 1.9.1 (compiled with Erlang/OTP 22)
ex_doc 0.21.2

This compiles fine but `mix docs` crashes:

`defmodule A do
  @spec x_p0_x() :: any
  def x_p0_x() do
    42
  end
end`

Seems to be because of the *_p0_* in the function name:

>  ** (KeyError) key :_p0_ not found in: %{_p1_: "<a href=\"https://hexdocs.pm/elixir/typespecs.html#basic-types\">any</a>"}
        (stdlib) :maps.get(:_p0_, %{_p1_: "<a href=\"https://hexdocs.pm/elixir/typespecs.html#basic-types\">any</a>"})
        (ex_doc) lib/ex_doc/formatter/html/autolink.ex:334: anonymous fn/2 in ExDoc.Formatter.HTML.Autolink.replace_placeholders/2
        (elixir) lib/regex.ex:731: Regex.apply_list/5
        (elixir) lib/regex.ex:726: Regex.apply_list/5
        (elixir) lib/regex.ex:668: Regex.do_replace/4
        (elixir) lib/enum.ex:1336: Enum."-map/2-lists^map/1-0-"/2
        (ex_doc) lib/ex_doc/formatter/html/autolink.ex:205: anonymous fn/6 in ExDoc.Formatter.HTML.Autolink.all_typespecs/2
        (elixir) lib/enum.ex:1948: Enum."-reduce/3-lists^foldl/2-0-"/3

wojtekmach:
Closes #1091

### DIFF
--- a/lib/ex_doc/formatter/html/autolink.ex
+++ b/lib/ex_doc/formatter/html/autolink.ex
@@ -250,6 +250,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
           {{name, [], args}, placeholders}
 
         {name, _, args} = form, placeholders when is_atom(name) and is_list(args) ->
+          ensure_no_placeholder!(name)
           arity = length(args)
 
           cond do
@@ -272,6 +273,7 @@ defmodule ExDoc.Formatter.HTML.Autolink do
 
         {{:., _, [alias, name]}, _, args} = form, placeholders
         when is_atom(name) and is_list(args) ->
+          ensure_no_placeholder!(name)
           alias = expand_alias(alias)
 
           if source = get_source(alias, aliases, lib_dirs) do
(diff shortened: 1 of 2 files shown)


lib/ex_doc/formatter/html/autolink.ex
```diff
@@ -321,17 +321,23 @@ defmodule ExDoc.Formatter.HTML.Autolink do
     end
   end
 
+  @placeholder "ó"
```
josevalim: Let’s use $. But then we will need to escape it in the regex.

wojtekmach: Good call. I got some failing tests on CI, which I didn’t get locally, so there might still be something wrong.

josevalim: Never mind. It has to be a valid identifier which is why ó fails on v1.7. We will have to make it something longer then. A mixture of lowercase, uppercase, underscore and digits.

wojtekmach:
@josevalim updated!

PR #1094, opened 2019-09-16
Link to callback docs instead of copying them

### LINKED ISSUE #1093: Replace callback docs copy with a link
In #641 we made it so that callbacks docs are copied to a function (unless it has `@doc false`).

There was a similar proposal #815 about copying typespecs which ultimately has not been accepted given the copy can be wrong, if it referred to private types. The same problem exists for copying docs although it's less severe given it's, well, docs.

I'd like to propose to replace docs copy with docs that just give a link to appropriate callback, e.g.: `@doc "Implementation for callback c:GenServer.handle_call/3."`. One particular advantage of this solution is we can now hover over the link and we'll see the typespec in the popover.

If this is accepted, we should probably make similar change to IEx which I'll be happy to do.

Thoughts?

wojtekmach:
Closes #1093

Additionally we fix a bug where we haven't been handling macrocallback.

### DIFF
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -253,7 +253,8 @@ defmodule ExDoc.Retriever do
     annotations = annotations_from_metadata(metadata)
 
     line = find_function_line(module_data, actual_def) || doc_line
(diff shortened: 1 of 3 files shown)


wojtekmach:
Here's the difference when generating MyApp.Repo docs.

before:

image

after:

image

Issue #1108, opened 2019-11-22
Code fence with more than 3 backticks


chulkilee:
To include code with three backticks, GFM allows wrapping the code with more than three backticks.

From https://github.github.com/gfm/#fenced-code-blocks

> A code fence is a sequence of at least three consecutive backtick characters (`) or tildes (~).

For example:

`````
````elixir
defmodule Foo do
  @moduledoc """
   Foo bbar

   # Example

   ```elixir
      defmodule MyFoo
      end
   ```
  """
end
````
`````

This is rendered in GFM  as follows:

````elixir
defmodule Foo do
  @moduledoc """
   Foo bbar

   # Example

   ```elixir
      defmodule MyFoo
      end
   ```
  """
end
````

For now, I circumvent it with using `~~~` instead of using four backticks.

Not sure this is earmark bug or ex_doc though.

**Comment by OvermindDL1** [comment]
It's not just a GFW thing, but in every markdown engine I routinely use.  You can use 3-or-more backticks to delinate code blocks:

Like this:
`````
````
Like to delinate code fences:
```elixir
Some code!
```
````
`````

Which becomes:
````
Like to delinate code fences:
```elixir
Some code!
```
````

And of course I wrote it with 5 backticks.  ^.^

wojtekmach:
> Not sure this is earmark bug or ex_doc though.

Yeah, that'd be responsibility of the markdown engine (e.g. earmark) to handle.

PR #1109, opened 2019-11-23
Earmark version bump

RobertDober:


mix.exs
```diff
@@ -1,7 +1,7 @@
 defmodule ExDoc.Mixfile do
   use Mix.Project
 
-  @version "0.21.2"
+  @version "0.19.1"
```
RobertDober: I just pulled from upstream, strange????

josevalim:
Hi @RobertDober! Is there any reason for us to enforce the latest? I typically don't like to enforce the latest and give users the control for them to do if they want to. After all, they can simply do `mix deps.update earmark`. What do you think?

eksperimental:
@josevalim it fixes this issue:
* https://github.com/elixir-lang/ex_doc/issues/1108
* https://github.com/pragdave/earmark/issues/309

You can read it in this commit https://github.com/pragdave/earmark/commit/8f533b11dc4a10ba42101fd9514b234e6c64c72a

josevalim:
Yes, but folks can get those changes without us imposing a stricter requirement. I will bump to ~> 1.4 but generally speaking we should not be very strict on the requirements unless we have strong reasons to. Thanks!

josevalim:

eksperimental:
Correct!. `~> 1.4.3` wouldn't have included `1.5.0` when this one is released.

RobertDober:
I completely yield to you, the ex_doc team for that issue.
I will only send PRs for minor version bumps in the future.
Makes sense too :)

PR #1113, opened 2019-12-01
Remove plugins from Earmark config options

maartenvanvliet:
With https://github.com/elixir-lang/ex_doc/pull/1109 the Earmark
dependency was set to `~> 1.4`

Earmark 1.4 no longer supports plugins see (https://github.com/pragdave/earmark/blob/master/RELEASE.md#140-20190905)
so setting this option through `ex_doc` is not possible anymore.


josevalim:

Issue #1115, opened 2019-12-06
Allow custom generators?


ityonemo:
I'm building a library which makes a nice interface for ziglang NIFs in elixir (https://github.com/ityonemo/zigler).  One of the features is that it takes zig code, parses the docstrings in the zig code, and marshals it into elixir documentation.  I can readily accomplish this by passing a custom generator to `Mix.Tasks.Doc.run/3` inside of a custom mix task.  However, this won't be executed as a part of the `mix hex.publish` pipeline.  It would be nice to be able to supply a custom generator as part of the config! thanks!  (Discussion welcome -- is this a bad idea because it could make forward improvements difficult?; PR forthcoming)

wojtekmach:
Note, you can achieve the same result by overriding `docs` task with an alias that calls Mix.Tasks.Doc.run/3 with desired arguments.

But I think your change about customising generator is fine too, not sure how common it would be though, I believe passing custom generator so far was only used for tests.

josevalim:
Oh, you can override the docs task by doing this:

```
def project do
  [aliases: [docs: "your_other_mix_task"]]
end
```

PR #1116, opened 2019-12-06
Allow custom generators (#1115)

ityonemo:


josevalim:
I am afraid this will force ExDoc to expose a bunch of implementation details that we are not really ready to expose. ExDoc.Config for example is private. All of the options are private. By making it part of a generator API, it means we would be unable to change any of the keys or any of the values, because that would break your code depending on it. I am afraid the best option for you right now is to run a fork.

We do plan for a major refactoring when we add Erlang support and that should get us really close to v1.0, which is when we would be comfortable with accepting this, but until then I think exposing our internal generation is going to harm our ability to move forward.

In any case, thanks for the pr!

Issue #1104, opened 2019-10-30
Make Exceptions a sub-grouping inside modules


josevalim:

archdragon:
Took a look at this one today.

One quick question - should we change the grouping rules on the "frontend" (in the JS menu generation) or on the "backend" side (in the Elixir doc generation code)?

### a) changing the groupings just on the "frontend" side


Currenly JS code receives a data structure that looks like:

```
{
  "extras": ...
  "exceptions": ...
  "modules": ...
  "tasks":...
}
```

So we can just ignore "exceptions" when generating  root menu elements and add them in the part that displays modules.

### b) changing the groupings on the "backend" side

We can change the elixir code to generate the data structure that will look more or less like this:

```
{
  "extras": ...
  "modules": [
    ...
    (append exceptions here)
  ]
  "tasks":...
}
```
This seems like a cleaner solution, but will still need some adjustments/cleanup in JS/HTML and EPUB generators.

b) is also a potentially breaking change for other, custom generators that people might be using (ie. if they assume that the `exceptions` key is always present).

I assume b) is what we want, but just want to make sure there is no additional context behind this issue.

archdragon:
This is how I see it (and how I understand the "sub-grouping inside modules"):

![image](https://user-images.githubusercontent.com/1810732/72680548-5eae5480-3abb-11ea-8a08-7977f8adf208.png)

^ I'm assigning Exceptions to an "Exceptions" group, which is always displayed last under the "Modules" top menu item.

josevalim:
Precisely! And i would make "Exceptions" the last group (exactly as you did below deprecations).

PR #1124, opened 2020-01-26
Make Exceptions a sub-grouping inside modules

archdragon:
Issue: https://github.com/elixir-lang/ex_doc/issues/1104

|Before|After|
|---|---|
|image|image|

## Changes
* `Exceptions` sidebar category was removed
* Exceptions are now treated like regular modules and available in a separate group, under `Modules`. This group is always displayed at the last position.
  * As a side-effect `API Reference` looks more cluttered now, since exceptions are treated as regular modules and displayed in the module list. We've discussed potentially introducing module grouping there (as a separate PR), to keep the list clean and readable.
  * Other functionality (like autocomplete search) is not affected by the change.

## Demo
[[Elixir docs with new exception grouping](http://krowinski.com/ex_doc/exceptions_demo/elixir/)]


assets/js/autocomplete/suggestions.js
```diff
@@ -199,7 +199,33 @@ function findMatchingChildren (elements, parentId, term, key) {
...
+ * Chooses the right sorting priority for the given item.
+ *
+ * @param {Object} item Search result item.
+ *
+ * @returns {number} Sorting priority for a given item. Higher priority means higher position in search results.
+ */
+function getSortingPriority (item) {
+  if (isException(item)) {
+    return sortingPriority['Exception']
+  }
```
josevalim: I am thinking we should remove this and not special case Exceptions in any way. Thoughts?

josevalim:

josevalim:
Hi @archdragon! I have dropped a comment that we can tackle later if you think it makes sense. :)

PR #1126, opened 2020-02-03
Simplify exception handling in autocomplete suggestions

archdragon:
Based on the suggestion from @josevalim https://github.com/elixir-lang/ex_doc/pull/1124#discussion_r372889229 , I've cleaned-up the code, removing special autocomplete rules for exceptions.

With recent changes we're treating exceptions like regular modules, so handling them the same way in autocomplete code makes sense. In addition moving exceptions to the `modules` sorting priority group has  basically no effect on the autocomplete search.


josevalim:

Issue #1128, opened 2020-02-12
generate both html and epub by default


supersimple:
per: @josevalim 
We want to generate both versions of the documentation by default.

wojtekmach:
Closing in favour of PR.

PR #1129, opened 2020-02-12
Adds epub to the default formatters

supersimple:
solves #1128


lib/mix/tasks/docs.ex
```diff
@@ -12,7 +12,8 @@ defmodule Mix.Tasks.Docs do
       rel="canonical" link element, defaults to no canonical path
 
     * `--formatter`, `-f` - Which formatters to use, "html" or
-      "epub", default: "html" (may be given more than once)
+      "epub". This option can be given more than once, by default,
```
wojtekmach:
```suggestion
      "epub". This option can be given more than once. By default,
```

wojtekmach:
Thank you!

PR #1134, opened 2020-03-13
Fix hidden text selection, hide tooltips for details link

archdragon:
Two fixes:

1. Copy-paste issue

Fix for an issue described here https://github.com/hexpm/hexdocs/issues/15

We have a hidden, screen-reader-only "tip" element, that was being selected when highlighting the function name. Fixed via `user-select` CSS attribute.

2. When fixing that I've noticed that tooltips were appearing on the permalink icon. They are not needed there.

image


josevalim:

PR #1137, opened 2020-03-21
Recognize : in function/operator names (for ::/2)

glv:
This will remove the need for [this workaround](https://github.com/elixir-lang/elixir/pull/9903#discussion_r395922872).


josevalim:
Thanks @glv! Note we already have a big refactoring of the autolinker in #1132 to work with ASTs instead of regexes, so we will probably merge that instead (where this issue has been fixed). But other than that, this is perfect!

glv:
Oh, nice! Yeah, that's a much better approach. I'll just watch for that to be merged then. Thanks!

PR #1140, opened 2020-03-26
Auto-link to extras

### LINKED ISSUE #1075: EPUB: Broken links on Elixir embedded documentation
After creating an EPUB document from the Elixir source code I noticed that the embedded documentation introduce a couple of _broken links_ (basically the markdown link assumes that the extension is `.html`) in the following modules:

* [`Kernel`](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/kernel.ex#L130-L142)
* [`Application`](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/application.ex#L451-L454)
* [`Config`](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/config.ex#L36-L39)

Also, some _broken links_ can be found on some guides like: `guards`, `library-guidelines`, and `syntax-reference`.

So, my question is how we want to handle those cases from ExDoc?, at least for the EPUB document generation. Should we explicitly transform the extension on those cases in the ExDoc code-base, seems a bit ugly, but I'm open to suggestions. /cc @josevalim

wojtekmach:
closes #1075

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,10 @@ ExDoc will automatically generate links across modules and functions if you encl
 
 ExDoc supports linking to modules (`` `MyModule` ``), functions (`` `MyModule.function/1` ``), types (`` `t:MyModule.type/2` ``) and callbacks (`` `c:MyModule.callback/3` ``). If you want to link a function, type or callback in the current module, you may skip the module name, such as `` `function/1` ``.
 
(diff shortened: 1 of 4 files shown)


josevalim:
Beautiful!

PR #1141, opened 2020-03-26
Modules grouped by prefix

javiergarea:
Module nodes are ordered by `nested_context`. Thus, the sidebar does not show the same nested group multiple times.

closes #1123

Issue #1123, opened 2020-01-23
nest_modules_by_prefix does not group all modules at once


chulkilee:
### Environment

- ex_doc: 0.21.2
- Elixir 1.9.4
- Erlang 22.2.2

`docs` option:

```
nest_modules_by_prefix: [
  B,
  B.B
]
```

Modules:

```elixir
defmodule B.A do
end

defmodule B.B.A do
end

defmodule B.C do
end
```

### Current behavior

Sidebar structure is:

```
B
    A
B.B
    A
B
    C
````

### Expected behavior

Sidebar should group modules like this, so that the nested group does not appear multiple times.

```
B
    A
    C
B.B
    A
````

josevalim:
Thanks! A PR would be welcome!

PR #1146, opened 2020-04-17
sort modules by group, nested_context, nested_title, id

SpaceEEC:
When using both `:nest_modules_by_prefix` and `:groups_for_modules` together, groups might be rendered more than once:
duplicated groups example

This is due to the modules getting sorted first by groups and then later in the html formatter again by nested_context and nested_title. (introduced in #1141)

Including these sort criteria where modules are initially sorted seems to fix this while not breaking #1123 again.

no longer duplicated groups example


josevalim:

PR #1148, opened 2020-04-18
Display moduledoc headings in the sidebar

### LINKED ISSUE #1145: Add Module Doc headings to the side nav
# Feature Idea
The Elixir standard library, as an example, has a lot of great and useful module documentation. Take the [Supervisor](https://hexdocs.pm/elixir/Supervisor.html) module for example. There are several headings that contain useful information that are linkable

* https://hexdocs.pm/elixir/Supervisor.html#module-examples
* https://hexdocs.pm/elixir/Supervisor.html#module-child-specification
* https://hexdocs.pm/elixir/Supervisor.html#module-shutdown-values-shutdown
* https://hexdocs.pm/elixir/Supervisor.html#module-restart-values-restart
* https://hexdocs.pm/elixir/Supervisor.html#module-child_spec-1
* https://hexdocs.pm/elixir/Supervisor.html#module-module-based-supervisors
* https://hexdocs.pm/elixir/Supervisor.html#module-start_link-2-init-2-and-strategies
...etc

Now, when I'm directly linked to one of these sections from somewhere else, great! However, what if I'm just browsing the docs? Navigating the description can be quite tedious looking for a section, or even just discovering what is available.

My suggestion is adding a section in the sidebar, call it `Description`, `Module Docs`, etc. It would be a dropdown similar to `Types`, `Functions`, and `Callbacks` that when clicked, it reveals the names of all these sections, and links to them.

Screen Shot 2020-04-15 at 12 27 57 PM

**I would really like to implement this feature myself**, but I am first looking for guidance.

1. Is there a reason that this doesn't already exist?
2. Are there any technical hurdles I should know about?

## Thank you
I love `ex_doc`. It is best in class, IMO, and I would love to help improve it further.

davydog187:
Closes #1145 

This PR adds moduledoc headings into the sidenav to make them more discoverable and easier to navigate.

- [X] Refactor implementation (it's mostly hacked together at the moment)
- [x] Support duplicated heading anchors
- [x] Add tests

Screen Shot 2020-04-18 at 6 17 26 PM


lib/ex_doc/formatter/epub/templates.ex
```diff
@@ -10,7 +10,7 @@ defmodule ExDoc.Formatter.EPUB.Templates do
   Generate content from the module template for a given `node`
   """
   def module_page(config, module_node) do
-    summary = H.module_summary(module_node)
+    summary = Enum.reject(H.module_summary(module_node), &match?({:Sections, _}, &1))
```
davydog187, line 13: This is a hack, I'm thinking on how to restructure this so its not necessary

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -157,7 +172,19 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
 
   def module_summary(module_node) do
-    [Types: module_node.typespecs] ++
+    sections =
+      case module_node.rendered_doc do
+        nil ->
+          []
+
+        moduledoc ->
+          Enum.map(extract_headers(moduledoc), &Map.put(&1, :type, :heading))
```
davydog187, line 181: This also is problematic. If there are multiple headings with the same text, this will always link to the first. I think I'll have to break out some of the code in `link_headings`

.gitignore
```diff
@@ -33,3 +33,6 @@ formatters/html/dist/html.js
 
 # Ignore escript when built
 /ex_doc
+
+# Elixir LS
+.elixir_ls/
```
josevalim, line 38: This belongs to your global gitignore. :)

lib/ex_doc/formatter/html/templates.ex
josevalim: I believe for pages, if we have duplicate titles, we add a indexed suffix.

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -157,7 +172,19 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
+      end
+
+    [
+      Sections: sections
+    ] ++
```
josevalim, line 186: I don't know how tricky it would be to do this, but can we included the Sections before the Summary? Since they appear before the Summary on the page.

josevalim:
Hi @davydog187, this looks promising! I have dropped some comments. Also, please do not include the compiled artifacts in the PR - as we cannot accept precompiled code from PRs. We will generate them after merging. Thanks!

davydog187:
Sounds good, thanks for the feedback @josevalim

davydog187:
@josevalim this is ready for review. I hope it is ok that I introduced Jason as a test dependency. It was actually already included as a transitive dependency for `excoveralls`.

Separate from this PR, I was chatting on the side with @wojtekmach about helping clean up `ex_doc`. I feel like some refactoring could have make this PR a bit simpler. The way that data is being passed to the frontend sidebar through a global variable seems unnecessary. Unless I am missing something (I probably am), I think this could be refactored to be all server-rendered. In order to properly render the sidebar, the javascript would need to change considerably. 

The net goal would to make it easier to contribute to `ex_doc`, with better test coverage.

lib/ex_doc/formatter/html/templates.ex
davydog187: Done!

assets/js/templates/sidebar-items.handlebars
```diff
@@ -34,6 +34,20 @@
             </li>
           {{/each}}
         {{else}}
+          {{#showSections node}}
+            <li class="docs open">
```
davydog187, line 38: Note that I rendered this open by default. I'm not sure if that's desired though.

davydog187:
Also, it's hard to tell if I hit the right test coverage. If there are other places where I should add tests I am happy to.

josevalim:
Hi @davydog187. So the reason the sidebar is rendered client side is because it can get really large for large projects like Elixir, so embedding the sidebar on every page would take a lot more space. We tried iframes but the usability was poor. And give that we want the docs to work offline, the current approach is the best option we could find.

In any case, we do agree a refactoring would be nice. We have a large refactoring in mind once we start fully supporting Erlang, which would require a lot of internal changes, but it may still take some time until we effectively get to it.

davydog187:
🤔 is it really that much more data? I would think it would be marginally more, considering that the only extra data is the tags surrounding the data.

In any case, thank you for the explanation.

assets/js/templates/sidebar-items.handlebars
josevalim: I think this would be ok only if this is the current page. Otherwise, it will probably make navigation difficult. WDYT?

test/ex_doc/formatter/html/templates_test.exs
```diff
@@ -256,6 +256,34 @@ defmodule ExDoc.Formatter.HTML.TemplatesTest do
                ~s(sidebarNodes={"extras":[{"group":"","headers":[{"anchor":"foo","id":"Foo"}],"id":"","title":""}]})
     end
 
+    test "it pulls sections out of moduledocs" do
```
josevalim, line 259:
```suggestion
    test "builds sections out of moduledocs" do
```

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -146,6 +153,31 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
+            case Map.fetch(acc, header.id) do
+              {:ok, id} ->
+                {%{header | anchor: "module-#{header.anchor}-#{id}"},
+                 Map.put(acc, header.id, id + 1)}
+
+              :error ->
+                {%{header | anchor: "module-#{header.anchor}"}, Map.put(acc, header.id, 1)}
+            end
+          end)
+          |> elem(0)
```
josevalim, line 175: Do we have this logic elsewhere? If so, we can re-use it? Or would it be too complicated?

davydog187: We do

```elixir
  @heading_regex ~r/<(h[23]).*?>(.*?)<\/\1>/m
  @spec link_headings(String.t(), Regex.t(), String.t()) :: String.t()
  def link_headings(content, regex \\ @heading_regex, prefix \\ "")
  def link_headings(nil, _, _), do: nil

  def link_headings(content, regex, prefix) do
    regex
    |> Regex.scan(content)
    |> Enum.reduce({content, %{}}, fn [match, tag, title], {content, occurrences} ->
      possible_id = HTML.text_to_id(title)
      id_occurred = Map.get(occurrences, possible_id, 0)

      anchor_id = if id_occurred >= 1, do: "#{possible_id}-#{id_occurred}", else: possible_id
      replacement = link_heading(match, tag, title, anchor_id, prefix)
      linked_content = String.replace(content, match, replacement, global: false)
      incremented_occs = Map.put(occurrences, possible_id, id_occurred + 1)
      {linked_content, incremented_occs}
    end)
    |> elem(0)
  end
```

The problem I see is that this is working on the derivative rendered data, and not the AST. A lot of this code could be greatly simplified if these actions were taken upstream of the rendering part of the process

davydog187: To answer your question. I suppose I could consolidate this logic, but its a much bigger project. I don't feel great about duplicating it.

If its any reconciliation, I would like to help to refactor this code to make it easier to work with.

assets/js/templates/sidebar-items.handlebars
davydog187: that makes sense to me, i'll make that change.

lib/ex_doc/formatter/html/templates.ex
josevalim: I understand. Yes, we are slowly migrating to the AST but I don’t think we did so for extras pages, so perhaps that’s the first step before we can address the duplication.

davydog187:
@josevalim updated so that Sections expands only if its the current page

Screen Shot 2020-04-25 at 8 10 58 AM

davydog187:
When you're on the page

Screen Shot 2020-04-25 at 8 13 10 AM

josevalim:

PR #1150, opened 2020-04-19
Fix chunk extraction for Elixir module on case-insensitive fs

### LINKED ISSUE #1149: Elixir.module_info/1 is undefined (module Elixir is not available)
While working on #1148, I noticed that the current master of `ex_doc` is broken when running against Elixir (both master and 1.10.2).

```
~/Development/elixir [tags/v1.10.2] λ make docs
==> ex_doc (elixir)
bin/elixir ../ex_doc/bin/ex_doc "Elixir" "1.10.2" "lib/elixir/ebin" --main "Kernel" --source-url "https://github.com/elixir-lang/elixir" --source-ref "v1.10.2"  --output doc/elixir --canonical "https://hexdocs.pm/elixir/v1.10/ " --homepage-url "https://elixir-lang.org/docs.html" --formatter "html" --config "lib/elixir/docs.exs"
** (UndefinedFunctionError) function Elixir.module_info/1 is undefined (module Elixir is not available)
    Elixir.module_info(:exports)
    (ex_doc 0.22.0-dev) lib/ex_doc/refs.ex:92: ExDoc.Refs.from_chunk/2
    (ex_doc 0.22.0-dev) lib/ex_doc/refs.ex:52: ExDoc.Refs.public?/1
    (ex_doc 0.22.0-dev) lib/ex_doc/autolink.ex:340: ExDoc.Autolink.module_url/2
    (ex_doc 0.22.0-dev) lib/ex_doc/autolink.ex:63: ExDoc.Autolink.walk/2
    (elixir 1.10.2) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (elixir 1.10.2) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (ex_doc 0.22.0-dev) lib/ex_doc/autolink.ex:71: ExDoc.Autolink.walk/2
make: *** [docs_elixir] Error 1
```

I got around it by changing this line to an empty list `[]` https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc/refs.ex#L109

I'm happy to fix this bug if someone can point me in the right direction.

wojtekmach:
Closes #1149

On such filesystems we'd get:

    iex> :code.get_object_code(Elixir)
    {Elixir, <<70, 79, 82, , ...>>, '(...)/../lib/elixir/ebin/Elixir.beam'}

And so we're getting the beam file for the private `:elixir` module.

### DIFF
--- a/lib/ex_doc/utils/code.ex
+++ b/lib/ex_doc/utils/code.ex
@@ -4,11 +4,6 @@ defmodule ExDoc.Utils.Code do
   # TODO: this is vendored from Elixir v1.11.0.
   #       Remove and use Code.fetch_docs/1 in the future.
 
-  # special case required for Elixir
-  def fetch_docs(Elixir) do
-    false
-  end
-
   def fetch_docs(module) when is_atom(module) do
     case :code.get_object_code(module) do
       {_module, bin, beam_path} ->
--- a/lib/ex_doc/utils/code.ex
+++ b/lib/ex_doc/utils/code.ex
@@ -3,6 +3,12 @@ defmodule ExDoc.Utils.Code do
 
   # TODO: this is vendored from Elixir v1.11.0.
   #       Remove and use Code.fetch_docs/1 in the future.
+
+  # special case required for Elixir
+  def fetch_docs(Elixir) do
+    false
+  end
+
   def fetch_docs(module) when is_atom(module) do
     case :code.get_object_code(module) do
       {_module, bin, beam_path} ->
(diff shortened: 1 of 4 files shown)


lib/ex_doc/utils/code.ex
```diff
@@ -3,6 +3,12 @@ defmodule ExDoc.Utils.Code do
 
   # TODO: this is vendored from Elixir v1.11.0.
   #       Remove and use Code.fetch_docs/1 in the future.
+
+  # special case required for Elixir
+  def fetch_docs(Elixir) do
+    false
+  end
```
wojtekmach, line 10: we don't need to special case it in two places, the other place is enough, but I'm planning to send this particular change upstream (this fetch_docs/1 is vendored) so we may as well keep them in sync.

PR #1151, opened 2020-04-19
Update undefined reference warning and :skip_undefined_reference_warnings_on option

wojtekmach:
Previously we were warning like this:

    warning: documentation references function ExUnit.Case.__using__/1 but it doesn't exist or isn't public (parsing Kernel.use/2 docs)

Now, to be consistent with other Elixir warnings, we output:

    warning: documentation references function ExUnit.Case.__using__/1 but it is undefined or private
      lib/elixir/lib/kernel.ex: Kernel.use/2

We're also making a breaking change to
`:skip_undefined_reference_warnings_on`, previously we could give it the
generated id of an extra e.g.: `["compatibility-and-deprecations"]`, now
we'd give it the path instead:
`["lib/elixir/pages/compatibility-and-deprecations.md"]`.

### DIFF
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -8,16 +8,22 @@ defmodule ExDoc.Autolink do
   #
   # * `:module_id` - id of the module being documented (e.g.: `"String"`)
   #
-  # * `:id` - id of the thing being documented (e.g.: `"String.upcase/2"`, `"library-guidelines"`, etc)
+  # * `:file` - source file location
+  #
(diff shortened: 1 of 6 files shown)


josevalim:
I believe this text is using the docs definition of private (i.e. `@doc false` is private) but we can probably clarify.

eksperimental:
> I believe this text is using the docs definition of private (i.e. `@doc false` is private) but we can probably clarify.

I can submit a PR

wojtekmach:
> I believe this text is using the docs definition of private 

Yeah, that was the idea.

PR #1152, opened 2020-04-25
Display line numbers for undefined reference warnings

wojtekmach:
Currently we simply display the line number of the beginning of the docs
if present (i.e.: line of `@moduledoc`, `@doc`, etc. We don't do it for
extras)

In the future, when Markdown engine will return line number in AST,
we'll make the warning even more accurate.


eksperimental:
@wojtekmach this is awesome!

PR #1155, opened 2020-04-26
Upgrade dependencies, use latest recommendations for webpack

davydog187:
Freshens up the frontend dependencies, and uses webpack's recommended packages.

- [X] Uglifyjs :point_right: Terser


assets/webpack.js
```diff
@@ -1,4 +1,4 @@
-const UglifyJSPlugin = require('uglifyjs-webpack-plugin')
+const TerserPlugin = require('terser-webpack-plugin');
```
thiamsantos, line 1:
```suggestion
const TerserPlugin = require('terser-webpack-plugin')
```
@davydog187 :nail_care: Seems that we are not using semicolons in the rest of the config file.

josevalim:

PR #1167, opened 2020-05-16
Adding spaces on paragraph endings for search results

### LINKED ISSUE #1164: Replace newlines with spaces in search data/results
Search functionality seems to be joining separate paragraphs / lines of documentation without any separator. This looks bad in search results and can potentially result in unexpected results showing up. Example:
image

I think in previous `ex_doc` versions, spaces were used.

kelvinst:
Fixes #1164

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -348,8 +348,8 @@ defmodule ExDoc.Formatter.HTML do
   @doc """
   Strips html tags from text leaving their text content
   """
-  def strip_tags(text) when is_binary(text) do
-    String.replace(text, @clean_html_regex, "")
+  def strip_tags(text, replace_with \\ "") when is_binary(text) do
+    String.replace(text, @clean_html_regex, replace_with)
   end
 
   @doc """
--- a/lib/ex_doc/formatter/html/search_items.ex
+++ b/lib/ex_doc/formatter/html/search_items.ex
@@ -59,7 +59,7 @@ defmodule ExDoc.Formatter.HTML.SearchItems do
   defp clean_doc(doc) do
     doc
     |> Kernel.||("")
-    |> HTML.strip_tags()
+    |> HTML.strip_tags(" ")
     |> String.replace(~r/\s+/, " ")
     |> String.trim()
   end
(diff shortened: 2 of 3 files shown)


kelvinst:
So this should be enough to fix it, could not find any test for the search templates, can you point me where I can implement the tests for it? Or is the tests I added enough already in your opinion?

josevalim:
I guess before when we generated the html from markdown, we would automatically add a newline between paragraphs. But now we don't. This fix looks good to me but maybe @wojtekmach prefers to fix it on the HTML generation side. Thoughts?

kelvinst:
I supposed the same, thought the change was probably intentional to make htmls smaller, that’s why I went with this solution. IMHO fixing it on HTML generation would make it brittle, as we might be getting the same problem with other tags that add spaces (like divs for example) that can be added by users on @doc manually. I don’t think my solution is the PERFECT one, as it would add spaces even for tags that don’t, I just don’t think it’s worth the hassle of searching for that perfect one.

Best,
Kelvin Stinghen
Kelvin.stinghen@me.com

wojtekmach:
Yeah, I think it would be nicer to fix it on the html generation side, I'll briefly look that but it's likely we'll just end up with this patch. :)

davydog187:
I will take a stab at fixing this on the HTML generation side.

davydog187:
@wojtekmach this can't be fixed on the HTML generation side. 


> "EnhancementsSeparate tasks into &quot;Mix Tasks&quot; in sidebarAdd types to the search resultsImprove accessibilityBug fixesStrip nesting HTML tags in h2 headersRemove the old search results after every new search attempt"

This is the value that the template is working with

davydog187:
Sorry, in my previous comment I misunderstood what was meant by HTML generation side (I think?)

Regardless, this PR still has issues

Screen Shot 2020-05-17 at 10 25 27 AM

kelvinst:
Well, that's odd, I didn't touch the actual tag regex, can you test the same with the old code? Maybe that's an old bug.

davydog187:
@kelvinst i think its due to the way your logic for adding spaces works

michallepicki:
@davydog187 I pulled the branch and I'm not seeing the issue from your screenshot:
image

Or maybe I'm missing it, what is the issue exactly?

davydog187:
@michallepicki its been a while, so I'm not sure.

kelvinst:
Just revisiting this old PR, I have tested it and couldn't repeat @davydog187's problem either. But mine doesn't even show the link for sidebar, maybe that changed on the recent version, IDK 🤷‍♂️

Captura de Tela 2021-03-29 às 07 44 55

josevalim:

PR #1180, opened 2020-05-28
Bring back auto-linking to Kernel and Kernel.SpecialForms

wojtekmach:


lib/ex_doc/autolink.ex
```diff
@@ -393,10 +395,21 @@ defmodule ExDoc.Autolink do
 
     if Refs.public?(ref) do
       fragment(tool(module), kind, name, arity)
+    else
+      if kind == :function do
+        try_autoimported_function(name, arity, config)
+      end
```
josevalim, line 401:
```suggestion
    cond do
      Refs.public?(ref) -> fragment(tool(module), kind, name, arity)
      kind == :function -> try_autoimported_function(name, arity, config)
      true -> nil
    end
```

?

eksperimental:
What was the reason behind for this feature to be initially removed?

wojtekmach:
the thinking was to nudge people to use fully qualified name even for Kernel* functions as it it's more explicit and hopefully more approachable to beginners that may not be familiar with all built-in functions. But sometimes it is really useful to auto-link to these after all, I guess that's why we introduced this capability in the first place!

eksperimental:
thank you @wojtekmach

PR #1182, opened 2020-06-05
Handling snake case terms in search results

### LINKED ISSUE #1178: Searching for partial matches in snake_cased terms doesn't work
Problem:

Case came up on Elixir Slack this morning, someone trying to set up Phoenix PubSub was looking for how to set the pool size.  In PubSub, there's a configuration option named, `pool_size` but searching just for 'pool' did not return a match to that section (it did return a match to the Phoenix.Tracker behavior section, because it had the word 'pools' in it).

Possible Solution?

It seems like adding underscores as "word breaks" for searching would make it easier to discover things in the documentation when you're not certain of a complete snake_cased term.  Not sure how flexible/smart Lunr is with regards to indexing and searching.  This MIGHT lead to 'verbose' search results for some terms, but that seems better than 'false negatives', as the user can either refine their search or just weed through the results to find what they're looking for -- if there are no results, they have no "starting point".

Optimal behavior would be that searching for 'pool' or 'size' would match on 'pool_size' but searching for 'pool_size' would _only_ match on 'pool_size' (and not be treated as "pool OR size"), but that might be asking too much of Lunr.

jwarwick:
Fixes #1178


assets/js/search.js
```diff
@@ -122,21 +122,37 @@ function titleExtractor (document) {
...
     title = title + ' ' + modFun + ' ' + modOrFun + ' ' + parts[parts.length - 1]
   }
 
   return title
 }
 
+function snakeCaseSplitter (builder) {
+  function snakeCaseFunction (token) {
+    var snakeTokens =  token.toString().split("_").map(function (str) {
+      return token.clone().update(function () { return str })
```
josevalim, line 136: My JavaScript-fu is weak, so what does `clone().update(function() { return str })` does? It feels unnatural but I am sure there is a reason. :D

assets/js/search.js
```diff
@@ -122,21 +122,37 @@ function titleExtractor (document) {
 
   if (type === 'function' || type === 'callback' || type === 'type') {
     var modFun = title.replace(/\/\d+/, '')
-    var modOrFun = modFun.replace('.', ' ')
+    var modOrFun = modFun.replace(/\./g, ' ')
```
josevalim, line 125: I assume this change is necessary as otherwise it replaces only the first one?

jwarwick: Correct. This is expanding the name to make it searchable, so `Mod.Mod2.Fun` was only getting split into `Mod Mod2.Fun` so you couldn't search for `Mod2` or `Fun` independently.

assets/js/search.js
jwarwick: My Javascript is non-existent :) It's from the example they provide of splitting up email addresses (https://lunrjs.com/guides/core_concepts.html). The `token` is an object, not just a string, so I think it's capturing the meta-data of the base token but changing the text associated with it.

josevalim:
Excellent PR, excellent replies! Glad to have you back and have a great weekend! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

jwarwick:
@josevalim thanks! Happy to help where I can.

Issue #1186, opened 2020-06-11
Support for markdown with path depedent uri 


calvinsadewa:
Currently there is already support for autolink markdown file (`.md`) to generated html documentation page (`.html`).

However this doesn't extend to path dependent markdown uri, for example `../subscription.md` is still rendered as `../subscription.md` in generated doc (would like to be `subscription.html`).

Would it be okay to extend autolinking for markdown to path dependent uri?

here is example taken from [Absinthe Guide](https://github.com/absinthe-graphql/absinthe/blob/master/guides/client/javascript.md) for `../subscriptions.md`
![image](https://user-images.githubusercontent.com/7587616/84419473-da639600-ac42-11ea-866f-80f14d6dae48.png)

wojtekmach:
A PR would be very welcome!

calvinsadewa:
Here is the PR https://github.com/elixir-lang/ex_doc/pull/1187

PR #1187, opened 2020-06-11
Add support for path dependent markdown autolink

calvin-kargo:
add support for local path markdown to be autolinked as html in documentation generated
(as in https://github.com/elixir-lang/ex_doc/issues/1186)

example `../subscriptions.md` autolinked as `subscriptions.html`:
image


josevalim:

PR #1183, opened 2020-06-07
Earmark version bump

RobertDober:
RELEASE NOTE:
# 1.4.5 2020/06/06                                                                                                                                                      
                                                                                                                                                                        
This is mostly a bugfix release, as there were edge cases that resulted in                                                                                              
Earmark crashing, notably                                                                                                                                               
                                                                                                                                                                        
  - Bare IAL                                                                                                                                                            
  - unquoted attributes in html tags                                                                                                                                    
                                                                                                                                                                        
Also autolinks (GFM extension) delivered incorrect URLS where parenthesis were involved,                                                                                
for better GFM compatibility we therefore                                                                                                                               
                                                                                                                                                                        
  - Fixed broken parenthesis links (99% of all cases)                                                                                                                   
  - introduced the same URL encoding/decoding in links and link names of autolinks as GFM does                                                                          
                                                                                                                                                                        
And last but not least all numeric options in the CLI can now be written with                                                                                           
underlines for readability.                                                                                                                                             
                                                                                                                                                                        
- [343-error-parsing-unquoted-atts](https://github.com/pragdave/earmark/issues/343)                                                                                     
                                                                                                                                                                        
- [342 parens in pure links](https://github.com/pragdave/earmark/issues/342)                                                                                            
                                                                                                                                                                        
- [340 IAL might cause error](https://github.com/pragdave/earmark/issues/340)                                                                                           
                                                                                                                                                                        
- [339 Typos fix](ihttps://github.com/pragdave/earmark/pull/339)                                                                                                        
    Kudos to [Ondrej Pinka](https://github.com/onpikono)                                                                                                                
                                                                                                                                                                        
- [336 Smartypants: Convert three hyphens to em dash](https://github.com/pragdave/earmark/pull/336)                                                                     
    Kudos to [Jony Stoten](https://github.com/jonnystoten)                                                                                                              
                                                                                                                                                                        
- [324 Fix AST for links with nested elements](https://github.com/pragdave/earmark/pull/324)                                                                            
    Kudos to [Wojtek Mach](https://github.com/wojtekmach)                                                                                                               
                                                                                                                                                                        
- [320 Nested Blockquotes](https://github.com/pragdave/earmark/issues/320)                                                                                              
                                                                                                                                                                        
# 1.4.4 2020/05/01                                                                                                                                                      
                                                                                                                                                                        
- [338  Deprecation warnings in mixfile removed](https://github.com/pragdave/earmark/issues/338)

PR #1188, opened 2020-06-12
Update Earmark

wojtekmach:
Note, we already have version requirement set to `{:earmark, "~> 1.4.0"}`.

Earmark 1.4.5 started returning comments as `{:comment, _}` (previously
a 3-element tuple) and per our discussions in the future AST will always
be a 4-element tuple, so for now we're simply loosening up our catch-all
clause.


lib/mix/tasks/docs.ex
```diff
@@ -140,11 +140,15 @@ defmodule Mix.Tasks.Docs do
 
       For GitLab/GitHub:
 
-          https://mydomain.org/user_or_team/repo_name/blob/master/%{path}#L%{line}"
+      ```
+      https://mydomain.org/user_or_team/repo_name/blob/master/%{path}#L%{line}"
+      ```
```
wojtekmach, line 145: Ref: https://github.com/pragdave/earmark/issues/349

test/ex_doc/formatter/html/templates_test.exs
```diff
@@ -148,14 +148,21 @@ defmodule ExDoc.Formatter.HTML.TemplatesTest do
       ## Example:
       """
 
-      assert Templates.synopsis(Earmark.as_html!(doc1)) ==
+      assert Templates.synopsis(to_html(doc1)) ==
```
wojtekmach, line 151: Something else changed in Earmark that caused this regression in our test suite around rendering paragraphs: https://github.com/elixir-lang/ex_doc/pull/1183#issuecomment-640247967

With this change we make the test better, we don't use Earmark to generate html anymore but our own custom code, so this way it's closer to reality too.

Issue #1194, opened 2020-06-24
Clicking same-page anchor in floating sidebar should close sidebar


henrik:
Hi! Whenever I check docs on my phone, I find this a little frustrating:

![issue](https://user-images.githubusercontent.com/216/85635567-100e7300-b676-11ea-902e-2c942e483f0e.gif)

If I click/tap a function in the floating sidebar, I want to read about it, not keep navigating – so I think it'd make sense for the sidebar to close when I click on it.

On a small display the sidebar is effectively modal. So it may also make sense to close it if clicking/tapping outside the sidebar – though perhaps that should be filed as a separate issue.

josevalim:
I understand the need but, I often click on the wrong place on my phone, given the combination of large thumbs and small phone, and I would be extremely upset if I had to go through the whole dance again. :) So I will keep this one as a feature. :D

henrik:
@josevalim Hehe, that's fair. What about if tapping outside the sidebar would close the sidebar, then? That would be intuitive IMO and might address both concerns to some extent.

josevalim:
@henrik I like that :+1: A PR is welcome, but it is probably easier said than done. :)

henrik:
Great. I may look into it sometime then :) Thank you!

henrik:
Another approach I just thought of would be to make it more fat-finger-friendly on small displays - so it does close on click but it’s much harder to misclick.

PR #1195, opened 2020-06-25
Hide sidebar-over-content on click/tap outside it

henrik:
On a small display, clicking anchor links in the sidebar (e.g. a
function while you're on that module page already) would navigate to
the anchor but not close the sidebar, requiring you to first click
the anchor and then the sidebar toggle.

José feels we shouldn't close the sidebar on the anchor click, because
that would be frustrating in case of misclicks, so instead we make the
sidebar easier to dismiss.

This is only in effect when the sidebar shows on top of the content, and
not on wide displays (where you might click the content just to
highlight some text, and wouldn't expect the sidebar to close).

Refs #1194.

Also snuck in a little cleanup in the immediate vicinity :)

x


assets/js/sidebar.js
```diff
@@ -83,14 +83,20 @@ export {breakpoint, closeSidebar, openSidebar, toggleSidebar, focusSearchInput}
...
       lastWindowWidth = window.innerWidth
       setDefaultSidebarState()
     }
   }, 100))
-  $('.sidebar-toggle').click(function () {
-    toggleSidebar()
+
+  $('.sidebar-toggle').click(toggleSidebar)
+
+  body.click((e) => {
```
archdragon, line 97: What do you think about - instead of listening to the `body.click` - using `$('.content').click(...)` ?

This way we won't have to check if sidebar/sidebar-toggle is one of the parents. :)

henrik: Hi! Thank you for the feedback :) Good call – I considered it initially but incorrectly ruled it out thinking it would have issues. Seems to work perfectly. Force-pushed those changes.

assets/js/sidebar.js
```diff
@@ -83,14 +83,19 @@ export {breakpoint, closeSidebar, openSidebar, toggleSidebar, focusSearchInput}
...
     }
   }, 100))
-  $('.sidebar-toggle').click(function () {
-    toggleSidebar()
+
+  $('.sidebar-toggle').click(toggleSidebar)
+
+  $(".content").click((e) => {
+    const sidebarCoversContent = window.innerWidth <= breakpoint
+    if (sidebarCoversContent) { closeSidebar() }
```
henrik, line 99: This could be made a one-liner, of course, but I like how this reads :)

henrik:
Force-pushed again – realised I had a console.log left in the dist JS.

josevalim:

PR #1190, opened 2020-06-17
Fix escaping of special HTML chars

eksperimental:
Closes #1168, #1189

/cc @wojtekmach

Issue #1189, opened 2020-06-15
Rendering regression in 0.22


imanhodjaev:
Hey,

In version `0.21` it was rendering markdown with HTML blocks as expected for example the following block 

```html
<h1 align="center">
Bla bla
</h1>
<p align="center">
  <img width="150" height="150" src="https://image.svg"/>
</p>
```

In version 0.22 `ex_doc` renders tags as plain text.

Also parser is unable to properly parse text blocks for example if you put
```html
<h1 align="center">Title</h1>
```
It will render

```html
Title</h1>
```

Elixir version: 1.10.3
earmark: 1.4.5
makeup: 1.0.3
makeup_elixir: 0.14.1

Thanks.

josevalim:
Can you please provide a sample project that reproduces the error? Thanks.

imanhodjaev:
Hi Jose,

Project documentation is here https://hexdocs.pm/dust/readme.html and GitHub link is https://github.com/imanhodjaev/dust

eksperimental:
The generated HTML is fine and it looks fine in my browser.
Did you try with a different browser or with no extensions installed?

imanhodjaev:
@eksperimental yes I tried on different browsers also hex docs https://hexdocs.pm/dust/readme.html show the same thing
<img width="869" alt="Screenshot 2020-06-16 at 19 36 26" src="https://user-images.githubusercontent.com/354868/84808282-c506b700-b008-11ea-9535-0beece1b85f9.png">

eksperimental:
@imanhodjaev I see what you mean now.
I went head and wrote a test and ran `git bisect` and the commit that introduced the bug is:

```
f553cc0c406ab0b589be652bd4148893c785bd8b is the first bad commit
commit f553cc0c406ab0b589be652bd4148893c785bd8b
Author: Wojtek Mach <wojtek@wojtekmach.pl>
Date:   Thu Mar 5 15:53:42 2020 +0100

    Use new autolinker
    
    * Replace `ExDoc.Markdown.to_html/2` with `to_ast/2`
    * Change the flow from `markdown->autolink->html` to `markdown->ast->autolink->html`
    * Add `:app` to ExDoc configuration (for generating project-local links)
```

/cc @wojtekmach 
I can share the test and the shell script if needed

imanhodjaev:
@eksperimental awesome! Thank you very much for your investigation. ✨🚀✨

eksperimental:
a reduced case is this
[...]

PR #1196, opened 2020-06-26
Introduce  Markdown.AST

### LINKED ISSUE #1168: Breaking change in 0.22: HTML entity support
Not mentioned (explicitly?) in the change log:
```elixir
  @doc "&mdash;"
```
… in 0.21.3 will result in the HTML
```html
&mdash;
```
… but in 0.22.0 will result in the HTML
```html
&amp;mdash;
```
… with similar results for `&hellip;` _etc._ This diverges from 0.21 and [GitHub Flavoured Markdown](https://github.github.com/gfm/#entity-and-numeric-character-references).

eksperimental:
- Introduce ExDoc.Markdown.AST and t:Markdown.AST.html/0.
- Replaces ExDoc.Formatter.HTML.ast_to_html/1 with ExDoc.Markdown.AST.to_html/2 which is a port of Earmark.Transform
- It uses a more recent git version of Earkmark since due to a bug in the library.
- It adds Floki to deal with HTML code in Markdown docs.

Closes #1168, #1189
Supersedes #1190


josevalim:
Thank you @eksperimental, this looks amazing but I am afraid it is nudging us in the wrong direction. Earmark is already capable of parsing HTML, so why do we need to reimplement it here and rely on Floki?

In particular, I would expect `markdown |> Earmark.to_html` and `markdown |> Earmark.as_ast |> simple_ast_to_html` to be enough. If it isn't, I would say `Earmark.as_ast` is not functioning correctly and we should push those fixes upstream.

eksperimental:
The use of Floki was a last resort. Actually this PR was done when I added more tests to realize that we were still not covering it.
Earkmarks has certain limitations,
1. Currently it does not part HTML recursively. https://github.com/pragdave/earmark/issues/358
2. HTML needs to be written in a certain way. https://github.com/pragdave/earmark/issues/357,  https://github.com/pragdave/earmark/pull/360/files

Boths PRs stemmed from reports I have made upstream  by working in this PR.
I have reported many others, but still I don't think all cases will be covered unless they use something like Floki.
Floki is exclusively used in exactly one line, in one case, when  `verbatim: true`, and I'm  in the talks to see if can get a copy of the original HTML code in the Earmark AST https://github.com/pragdave/earmark/issues/356#issuecomment-649906841

josevalim:
Right, so I would rather push those concerns to Earmark rather than in ExDoc. It is not our job to solve them.

Perhaps one solution you can adopt in Earmark is to have a copy of [`floki_mochi_html`](https://github.com/philss/floki/blob/master/src/floki_mochi_html.erl), properly renamed to `earmark_mochi_html`. It doesn't bring a whole dependency and it will import only the parts that you need (which is effectively the parser).

eksperimental:
@RobertDober What do you think can be done about this?

RobertDober:
@eksperimental has had a tough time discovering how basic  and also broken HTML parsing is in Earmark. Now I believe that I can come up with a pretty good HTML parser in 1.5 but I admit that

- Floki already does this and probably better than Earmark will ever
- personally I am totally confused how HTML _should_ be parsed in Markdown and therefor try to orient myself on GFM but that is a high bar
- I really liked @eksperimental's idea to provide the pure HTML string in the AST, (it would be a great safety net, I mean Earmark is not really professionally maintained,  I am very much aware of my shortcomings) **and that I could push into 1.4.6 I guess**
- I'll have a look at https://github.com/philss/floki/blob/master/src/floki_mochi_html.erl and come back to you on that

All that said I am working hard on the 1.4.6 release with the now AST format and most of the gross errors fixed (ETA Sunday :fearful: ) do we agree that this is the first priority? If not I can rearrange my time.

Please note too that  pragdave/earmark#358 is downwards incompatible and therefore scheduled for 1.5 (ETA????)

josevalim:
My understanding is that the issues described here are "old issues", so I would keep your current priority.

@eksperimental if we close this, we should still merge #1190, right?

eksperimental:
> My understanding is that the issues described here are "old issues", so I would keep your current priority.
> 
> @eksperimental if we close this, we should still merge #1190, right?

no, never. That PR will render the Epub unusable whenever a `<` or `>` is not escaped

wojtekmach:
My understanding, in this issue we're discussing and trying to tackle:

1. #1168 (note: it's already fixed)
2. #1189
3. other improvements to html parsing (via floki/floki_mochi/etc)

does that sound about right? @eksperimental could we just tackle 2. for now in this PR and leave 3. for the future?

eksperimental:
We could just leave out the Floki call which is one or two lines, and leave what Earmark gives as is.

josevalim:
Right, it is not our job to parse html and adding markdown specific logic is going to be counter intuitive to direction we are taking ExDoc which is to support multiple input formats. :)

eksperimental:
> Right, it is not our job to parse html and adding markdown specific logic is going to be counter intuitive to direction we are taking ExDoc which is to support multiple input formats. :)

I know, but we have no other others, and we still don't know whether this will be fixed in Earkmark

josevalim:
If we have no other options, then we will stick with no other options. We have lived without those so far and I really don't want to be responsible for handling this complexity. :)

eksperimental:
I have rebased to master, meaning now we work with EarmarkParser. I have added some temporary fixes until EarkmarkParser always works with 4-element tuples in the AST.
Floki has been removed and the comments that are not passing due to the limitation in nested HTML code have been commented out.

eksperimental:
> If we have no other options, then we will stick with no other options. We have lived without those so far and I really don't want to be responsible for handling this complexity. :)

For the record, this limitation recently came in with the introduction of AST if I'm not mistaken.

I think the easiest solution is to store the original HTML in the metadata, as I have suggested to @RobertDober and we can inject it back in on our side.

josevalim:
But why do we need the original HTML? Doesn’t markdown say we should parse any markdown inside the HTML too? So returning the original HTML could override the nested parsing.

josevalim:
Btw, thanks for the clarification! :D

eksperimental:
> But why do we need the original HTML? Doesn’t markdown say we should parse any markdown inside the HTML too? So returning the original HTML could override the nested parsing.

Earmark is returning a crippled version of the original HTML or something like that (I can't recall exactly). As of when I wrote this PR we cannot fix this limitation, as we were not able to reconstruct the HTML that we passed to it. I will look later if something changed in EarmarkParser

eksperimental:
Sorry, the crippled version it was a bug that has been fixed. We are getting the HTML string. The issue is that we cannot escape the text within the HTML elements.  This PR and also the current version of ExDoc escape the special characters of the whole string

RobertDober:
@josevalim Almost no Markdown parser parses content inside HTML https://babelmark.github.io/?text=%3Cdiv%3E+%0A++-+a%0A++-+b%0A%3C%2Fdiv%3E , nor does EaramrkParser

josevalim:
Ah, apologies for the confusion then. So storing the original HTML should work, if that’s what we want.

eksperimental:
Sorry, since Earkmark was parsing the first level of the HTML I thought the intention was to build the whole tree.
What we need is to escape the content of those html tags, but not the tags themselves.

josevalim:
Why do we need to escape them? Maybe they shouldn’t be unescaped in the first place?

eksperimental:
> Why do we need to escape them? Maybe they shouldn’t be unescaped in the first place?

Because we produce invalid HTML and Epub.
Worst case is that the Epub will not be rendered some pages in some ebook readers.

RobertDober:
@eksperimental parsing the whole tree is planned for 1.5, scheduled for 2020-08-30

eksperimental:
Regardless the parsing of the HTML code which is just two lines and I will remove them later today., I would like to know what is your opinion on this PR

josevalim:
I think it is too early for us to formalize the AST. We also don't really care about Markdown AST per se as we will likely introduce our own so it works with both Markdown and the Erlang docs AST. So I would rather focus on fixing bugs instead of trying to formalize something that will definitely change.

PR #1197, opened 2020-06-26
Replace <kbd> with <kbd><kbd> when it represents keys to be hit

eksperimental:
Such is the case in the shortcuts.

Reference: https://www.w3.org/TR/html52/textlevel-semantics.html#elementdef-kbd
> When the kbd element is nested inside another kbd element, it represents an actual key or other single unit of input as appropriate for the input mechanism.


josevalim:

PR #1199, opened 2020-06-26
Make HTML valid

eksperimental:


josevalim:

eksperimental:
@josevalim if we would have wanted to insert this script element with, how would we done it with config.before_closing_body_tag ?

PR #1200, opened 2020-06-26
Allow files with no extension in extra files

eksperimental:
The files with no extension are considered to be plain text.

Additionally it adds LICENSE as an extra file.

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -128,14 +128,7 @@ defmodule ExDoc.Formatter.HTML do
 
   def ast_to_html({tag, attrs, ast}) do
     attrs = Enum.map(attrs, fn {key, val} -> " #{key}=\"#{val}\"" end)
-
-    case tag do
-      tag when tag in [:pre, :code] ->
-        ["<#{tag}", attrs, ">", Enum.intersperse(ast_to_html(ast), "\n"), "</#{tag}>"]
-
-      _ ->
-        ["<#{tag}", attrs, ">", ast_to_html(ast), "</#{tag}>"]
-    end
+    ["<#{tag}", attrs, ">", ast_to_html(ast), "</#{tag}>"]
   end
 
   defp output_setup(build, config) do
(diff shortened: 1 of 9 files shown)


lib/ex_doc/formatter/html.ex
```diff
@@ -128,7 +128,14 @@ defmodule ExDoc.Formatter.HTML do
 
   def ast_to_html({tag, attrs, ast}) do
     attrs = Enum.map(attrs, fn {key, val} -> " #{key}=\"#{val}\"" end)
-    ["<#{tag}", attrs, ">", ast_to_html(ast), "</#{tag}>"]
+
+    case tag do
```
josevalim, line 132: Are those changes related?

josevalim:
Looks good! We just need some tests.

lib/ex_doc/formatter/html.ex
eksperimental: They are. Otherwise the text inside `<pre>` will be printed in one long line

eksperimental:
@josevalim tests have been added. Since we were already supporting plain text, I have added an additional extension .txt

lib/ex_doc/formatter/html.ex
```diff
@@ -128,7 +128,14 @@ defmodule ExDoc.Formatter.HTML do
...
+      tag when tag in [:pre, :code] ->
+        ["<#{tag}", attrs, ">", Enum.intersperse(ast_to_html(ast), "\n"), "</#{tag}>"]
```
josevalim, line 134: I see. My question though is: why are we breaking .txt over multiple lines? Shouldn't we keep it as a single binary? I am afraid we are adding new interpretations to the AST while not necessary. We can also add the newlines after splitting in the splitting code instead of here.

eksperimental: > I see. My question though is: why are we breaking .txt over multiple lines? Shouldn't we keep it as a single binary? I am afraid we are adding new interpretations to the AST while not necessary. We can also add the newlines after splitting in the splitting code instead of here.

yes, that is because of Earmark's and the new ExDoc's AST that I have been working on. That is how it is represented, because there can be tags inside a `<pre>`. I will simplify this for now.

eksperimental: fixed!

josevalim:

PR #1201, opened 2020-06-27
Run: mix build

eksperimental:
We were having two different set of .js and .css files due to the last two commits merged in parallel


josevalim:
I will fix it locally, thanks!

josevalim:
Btw, it is best if the PRs do not build assets, I forgot to pay attention to it. :)

eksperimental:
> Btw, it is best if the PRs do not build assets, I forgot to pay attention to it. :)

so the one who merges them build them?

josevalim:
Yes, we will rebuild them from master time to time. It is not even safe to accept PRs with precompiled code. :)

wojtekmach:
I think it is fine to `mix build` just before the release.

We could prevent folks from changing dist/ on CI (or rebuild assets on CI and see if we get the same output), but honestly I don't think it's worth it, doing it manually seems good enough.

eksperimental:
I thought they were needed in the tests of the UI.
We should update the docs because it says otherwise.

## Contributing

The easiest way to test changes to ExDoc is to locally re-generate its own docs:
```

  1. Run `mix setup` to install all dependencies
  2. Run `mix build` to generate docs. This is a custom alias that will build assets, recompile ExDoc, and output fresh docs into the `doc/` directory
  3. Commit both `assets/*` and `formatters/*` changes (after running `mix build`)
```

I can send a PR.

wojtekmach:
@eksperimental re: contributing instructions, good catch!

eksperimental:
@wojtekmach 
https://github.com/elixir-lang/ex_doc/pull/1207
done

PR #1202, opened 2020-06-27
Fix travis

eksperimental:
It reverts changes introduced in 900a040 (#1193)

The errors in the config can be found here:
https://travis-ci.org/github/elixir-lang/ex_doc/builds/701021629/config

> root: both matrix and jobs given, matrix overwrites jobs

/cc @adrianomitre


eksperimental:
there are some fixes that were not previously caught by the CI due to the faulty config file.

josevalim:

PR #1205, opened 2020-06-27
Improve extra links generation

eksperimental:
1. Improve the warning messages when linking with a custom link, such as in ```the [license](`LICENSE`)```, provided that the file is listed in the :extras option in config,  it warns the users:
> warning: file name surrounded by backticks "`LICENSE`, please remove them to link to the existing file "LICENSE"

****
2. When using backticks around the file name, it assumes that the user could have meant a module, and it suggests:

```the [kernel](`Kernels`) module```

> documentation references module or file "`Kernels`" but it doesn't exist.

******
3. If not backticks is used, it uses the existing message.

```the [kernel](Kernels) module```

> documentation references file "`Kernels`" but it doesn't exist.

/cc @wojtekmach

### CODE REVIEW

File: test/ex_doc/autolink_test.exs [hunk 1]
```diff
@@ -161,11 +161,55 @@ defmodule ExDoc.AutolinkTest do
...
-      #       Though backticks are markdown specific, is that ok?
-      # assert_warn(fn ->
-      assert_unchanged(~m"[custom text](`Unknown`)")
+      captured =
+        assert_warn(fn ->
+          assert_unchanged(~m"[custom text](`Unknown`)")
+        end)
+
+      assert captured =~
+               "documentation references module or file \"`Unknown`\" but it doesn't exist"
```
wojtekmach, line 170: > "documentation references module or file (...)

I'd remove the "or file" piece, since custom links are special I believe we can be more strict:

1. `` [foo](`Foo`) `` - if something is within backticks, it is a module/function/type/etc, and if we cannot find it, we warn
2. `` [foo](Foo) `` - if something is without backticks, then it can only be an extra, and if we don't have it, we warn

IMHO the warning is not going to be as useful but still useful enough, and we'd keep things a bit simpler. WDYT?

thank you for handling this TODO, btw!

eksperimental: I agree with simplifying things. My only concern is that people will get confused when using custom links swap the terms (I did that myself A LOT).

eksperimental: done

test/ex_doc/autolink_test.exs
```diff
@@ -161,11 +161,30 @@ defmodule ExDoc.AutolinkTest do
...
+
+      assert captured =~
+               "documentation references module \"Unknown\" but it doesn't exist"
+
+      captured =
+        assert_warn(fn ->
+          assert_unchanged(~m"[custom text](Unknown)")
+        end)
+
+      # File LICENSE exists, but backticks are meant for modules
```
wojtekmach, line 177: I think this comment is misplaced, I'd either put it on line 183 or simply remove it.

lib/ex_doc/autolink.ex
```diff
@@ -92,17 +89,37 @@ defmodule ExDoc.Autolink do
...
+            build_extra_link(href, config)
+        end
+
+      _ ->
+        nil
+    end
+  end
+
+  defp build_custom_link(link, config) do
+    cond do
```
wojtekmach, line 108: nitpick: change this to an if?

lib/ex_doc/autolink.ex
```diff
@@ -92,17 +89,37 @@ defmodule ExDoc.Autolink do
...
+      url = url(link, :custom_link, config) ->
+        url
+
+      true ->
+        uri = URI.parse(link)
```
wojtekmach, line 113: what's the reason for URI.parse/1 here? By this point, we are supposed to be between `` `(...)` `` here anyway. If it's needed, could you add a test?

lib/ex_doc/autolink.ex
```diff
@@ -92,17 +89,37 @@ defmodule ExDoc.Autolink do
...
+        message = "documentation references module \"#{uri.path}\" but it doesn't exist"
```
wojtekmach, line 114: saying _module_ in the warning is too generic, people could have written:

```
[bad](`Bad.bad/0`)
```

or even worse (please don't!)

```
[bad](`bad bad`)
```

and the warning would be incorrect. I'd just remove the word _module_ but hopefully we could use other parts that are already distinguishing between modules, functions, types, etc.

I also noticed that this test:

```
    test "bad" do
      assert_unchanged(~m"[bad](`String.upcase/9`)")
    end
```

warns twice:

```
warning: documentation references function String.upcase/9 but it is undefined or private
  nofile

warning: documentation references module "String.upcase/9" but it doesn't exist
  nofile
```

notice the first warning is more accurate so that's the one I'd rather keep here!

test/ex_doc/autolink_test.exs
eksperimental: removed.

lib/ex_doc/autolink.ex
eksperimental: what do you suggest for avoiding the double warning?

lib/ex_doc/autolink.ex
eksperimental: done

lib/ex_doc/autolink.ex
wojtekmach: from the two warnings we want to keep the 1st one, so let's not trigger the 2nd one at all. Now, we're not triggering the 1st under certain circumstances (e.g. `` `Unknown` ``, `` `Unknown.unknown/1` ``, etc) so I think we could have an option to _do_ trigger them when we have a custom link. WDYT?

eksperimental: > from the two warnings we want to keep the 1st one, so let's not trigger the 2nd one at all. Now, we're not triggering the 1st under certain circumstances (e.g. `` `Unknown` ``, `` `Unknown.unknown/1` ``, etc) so I think we could have an option to _do_ trigger it when we have a custom link. WDYT?

please have a look

test/ex_doc/autolink_test.exs
```diff
@@ -27,15 +27,28 @@ defmodule ExDoc.AutolinkTest do
...
 
     test "unknown module" do
-      assert_unchanged("Unknown")
+      captured =
+        assert_warn(fn ->
+          assert_unchanged("Unknown")
+        end)
+
+      assert captured =~
+               "documentation references module \"Unknown\" but it is undefined or private"
```
wojtekmach, line 50: I don't think we should warn here, we want to write docs that have e.g. `` `MyApp` ``.

However, if we write `` [foo](`MyApp`) `` then we are referring to a particular module so if it doesn't exist, we should warn.

test/ex_doc/autolink_test.exs
```diff
@@ -183,7 +220,13 @@ defmodule ExDoc.AutolinkTest do
...
       assert_unchanged("mix compile.elixir --verbose")
 
-      assert_unchanged("mix unknown.task")
+      captured =
+        assert_warn(fn ->
+          assert_unchanged("mix unknown.task")
+        end)
+
+      assert captured =~
+               "documentation references module \"unknown.task\" but it is undefined or private"
```
wojtekmach, line 229: I don't think we should warn here, we want to write docs that have e.g. `` `mix myapp.mytask` ``.

However, if we write `` [foo](`mix myapp.mytask`) `` then we are referring to a particular mix task so if it doesn't exist, we should warn. I also think, the warning should be slightly different, I think one of the below:

- `documentation references "mix unknown.task" (...)`
- `documentation references mix task "unknown task" (...)`
- `documentation references module "Mix.Tasks.Unknown.Task" (...)`

but we don't need to worry about the warning message too much right now imho.

test/ex_doc/autolink_test.exs
eksperimental: so `String.Unicode` should warn, or not?

eksperimental: To make it clear: linking to a module that has been defined with `@doc false` should warn or not?

wojtekmach: I think it should warn (and I initially implemented it as such) but I don't have a very strong opinion about it.

eksperimental: done

test/ex_doc/autolink_test.exs
eksperimental: I have taken care of this already.

eksperimental:
@wojtekmach please have a look at my last two commits.
I have revamped `Refs` to pretty make the distinction between undefined modules and hidden/private.

lib/ex_doc/autolink.ex
```diff
@@ -383,21 +392,40 @@ defmodule ExDoc.Autolink do
...
+    case {mode, Refs.get_visibility({:module, module})} do
+      {_, :public} ->
+        module_url_app(tool(module), module, config)
+
+      {:regular_link, :undefined} ->
+        nil
+
+      {_, visibility} ->
+        module_string =
+          if mix_task? do
```
wojtekmach, line 410: since we already have an if statement here, maybe let’s make the warning specific for mix tasks here? “(...) references module mix foo (...)” isn’t totally accurate.

test/ex_doc/autolink_test.exs
```diff
@@ -26,22 +26,35 @@ defmodule ExDoc.AutolinkTest do
...
 
     test "erlang module" do
       assert_unchanged(":array")
     end
 
     test "unknown module" do
-      assert_unchanged("Unknown")
+      captured =
+        assert_warn(fn ->
+          assert_unchanged("Unknown")
```
wojtekmach, line 46: if we expect not to warn, let’s remove assert warn. Otherwise, since we “refute =~”, if the warning message ever changes this test won’t catch that.

test/ex_doc/autolink_test.exs
```diff
@@ -190,7 +236,13 @@ defmodule ExDoc.AutolinkTest do
 
       assert_unchanged("mix compile.elixir --verbose")
 
-      assert_unchanged("mix unknown.task")
+      captured =
+        assert_warn(fn ->
```
wojtekmach, line 240: same as above, if we expect not to warn let’s remove assert_warn.

eksperimental: done

test/ex_doc/autolink_test.exs
eksperimental: done

lib/ex_doc/autolink.ex
eksperimental: how about this message?

test/ex_doc/autolink_test.exs
```diff
@@ -51,16 +57,23 @@ defmodule ExDoc.AutolinkTest do
...
       ])
 
       assert autolink("Foo.foo/1") == ~m"[`Foo.foo/1`](Foo.html#foo/1)"
       assert autolink("Foo../2") == ~m"[`Foo../2`](Foo.html#./2)"
       assert autolink("Foo.../2") == ~m"[`Foo.../2`](Foo.html#../2)"
-      assert_unchanged("Bad.bar/1")
+
+      captured =
+        assert_warn(fn ->
+          assert_unchanged("Bad.bar/1")
```
wojtekmach, line 72: We should not warn here. Taking a step back, warnings are mostly for finding typos and preventing referring to internals. We can’t warn on undefined modules or functions on undefined modules because then we would warn when docs mention user code. However if we have a custom link we always warn because custom link would never refer to user code.

If the rules around warnings are too complicated/arbitrary/etc we should simplify them.

eksperimental: fixed

eksperimental:
I will submit a new commit documenting when ExDoc warns, and another onedocumenting `:skip_undefined_reference_warnings_on` with the new changes.

josevalim:
> warning: documentation references module "Logger.Backends.Console" but it is hidden
>  lib/logger/lib/logger.ex:2: Logger

We should probably fix this in Elixir tbh. We should make it public somehow while hiding its callbacks.

eksperimental:
I

> I used this branch to generate Elixir docs and there is a lot of warnings, could you check? Notably, whenever a spec refers to a private type we warn, but we shouldn't if this happen in the same module.
> 
> I was having 2nd thoughts about warning about private modules (I honestly thought we already did that, so thank you for adding this capability), e.g. this is perhaps too much:
> 
> ```
> warning: documentation references module "Logger.Backends.Console" but it is hidden
>   lib/logger/lib/logger.ex:2: Logger
> ```
> 
> (or maybe we that module shouldn't be private)
> 
> but then the warning also found:
> 
> ```
> warning: documentation references module "ExUnit.TestCase" but it is hidden
>   lib/ex_unit/lib/ex_unit/formatter.ex:2: ExUnit.Formatter
> ```
> 
> which I'm glad, that module is deprecated. So I think the warning is useful after all, thanks again.
> 
> Taking a step back we are doing a lot in this PR so could you update the PR description with a brief bullet list of changes and/or split it up into smaller more focused PRs?

I am splitting this into two PRs or more 
here' s the first one. Please merge so I can continue working on the following changes.
https://github.com/elixir-lang/ex_doc/pull/1217

wojtekmach:
Are you planning to rebase this after all the prerequisites are done, or should we close it and create a new one later? Up to you.

eksperimental:
I will create another one. thank you!

PR #1209, opened 2020-06-30
Preparing ex_doc to switch to EarmarkParser

RobertDober:
Well maybe this is a little bit early, Earmark will remain at your disposal if you prefer

If you want to wait until Earmark uses EarmarkParser for some versions or not switch at all I will not be offended (well not much)

However ultimately I'd like to use `ex_doc` as an indirect dependency in `Earmark` so if you want to switch one day that would be great.

This PR just gives you the option and I'll try to sync it with upstream as much as I can


lib/ex_doc/markdown/earmark.ex
```diff
@@ -1,14 +1,15 @@
 defmodule ExDoc.Markdown.Earmark do
   @moduledoc """
-  ExDoc extension for the Earmark MarkDown parser.
+  ExDoc extension for the EarmarkParseer MarkDown parser.
```
wojtekmach, line 3:
```suggestion
  ExDoc extension for the EarmarkParser Markdown parser.
```

mix.lock
```diff
@@ -1,6 +1,7 @@
 %{
   "certifi": {:hex, :certifi, "2.5.1", "867ce347f7c7d78563450a18a6a28a8090331e77fa02380b4a21962a65d36ee5", [:rebar3], [{:parse_trans, "~>3.3", [hex: :parse_trans, repo: "hexpm", optional: false]}], "hexpm", "805abd97539caf89ec6d4732c91e62ba9da0cda51ac462380bbd28ee697a8c42"},
   "earmark": {:hex, :earmark, "1.4.5", "62ffd3bd7722fb7a7b1ecd2419ea0b458c356e7168c1f5d65caf09b4fbdd13c8", [:mix], [], "hexpm", "b7d0e6263d83dc27141a523467799a685965bf8b13b6743413f19a7079843f4f"},
```
wojtekmach, line 3: oh, could you also unlock this dependency? `mix deps.unlock --unused`.

RobertDober:
The split is risky and I might regret it, but I am a perfectionist and a risk taker.
 I do not consider these traits being a quality but sometimes I just cannot hold back. (as they say above the Oracle of Delphi)

Ok enough psychoanalysis of an egocentric developer 😓 just as you might have noticed when upgrading Earmark beyond 1.4.7 you will use EarmarkParser anyway.

RobertDober:
> LGTM!
> 
> Since we are not using many features from Earmark, depending on just the parser piece is definitely a win. I'm a little bit concerned that the split might generate more work for you (people will report issues in the "wrong" repo, syncing releases, etc) but hopefully these would be just growing pains and things will stabilise over long-term.

Definitely it will remain a problem but I am prepared to deal with it. The release procedure will take more time too as the bulk of changes will be in the parser, I am just a mad man I guess ...

eksperimental:
:clap:

RobertDober:
This is reallt great for my dev circle. Thank you so much.

PR #1217, opened 2020-07-14
Revamp Refs

eksperimental:
Refs no longer store wether the ref is public or not,
but adds the concept of visibility, such as :public, :hidden, :undefined


josevalim:

PR #1218, opened 2020-07-14
Fix basic types

eksperimental:


eksperimental:
these ones came up with the new warnings when run against Elixir core.

eksperimental:
By the way, I think we have avoided this had we have a helper function in Elixir that lists the built-in and basic types, with the exact same lists that we have here.

josevalim:

PR #1220, opened 2020-07-14
Introduce :private visibility in Refs

eksperimental:
Addresses the issue of warning about private types, which is not tackled in this PR per se.

### DIFF
--- a/lib/ex_doc/refs.ex
+++ b/lib/ex_doc/refs.ex
@@ -5,7 +5,7 @@ defmodule ExDoc.Refs do
   #
   # The cache consists of entries:
   #
-  #     entry() :: {ref(), visibility :: :hidden | :public | :undefined}
+  #     entry() :: {ref(), visibility :: :hidden | :private | :public | :undefined}
   #
   #     ref() ::
   #       {:module, module()}
@@ -49,9 +49,13 @@ defmodule ExDoc.Refs do
       [] ->
         case load(ref) do
           # when we only have exports, consider all types and callbacks refs as matching
-          {:exports, entries} when elem(ref, 0) in [:type, :callback] ->
+          {:exports_and_privates, entries} when elem(ref, 0) in [:type, :callback] ->
             :ok = insert([{ref, :public} | entries])
             :public
+
+          {:exports_and_privates, entries} when elem(ref, 0) == :typep ->
+            :ok = insert([{ref, :private} | entries])
+            :private
 
           {_, entries} ->
             :ok = insert(entries)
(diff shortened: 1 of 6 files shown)


eksperimental:
After the other big PR is merged, I would like to introduce the ref of private functions. So we can properly make the distinction and warning between private and undefined.

wojtekmach:
Yeah, I think we can track “private” functions, I think they are rather uncommon so I wouldn’t worry about them too much but I think it makes sense to do it for completeness. Instead of “private” maybe a better name is “hidden”? That’s the term we use in the chunk spec and there is no ambiguity with truly private functions (I.e. defp).

eksperimental:
We are already tracking those hidden functions (declared with doc false). What I mean here is to track truly private functions (`defp`, `defmacrop`), actually this PR does that with `typep`).

wojtekmach:
There is no way to track truly private functions (as they are not visible to the runtime system), is there? :)

eksperimental:
> There is no way to track truly private functions (as they are not visible to the runtime system), is there? :)

I was able to access private function definitions with `:beam_lib.chunks(beam, [:debug_info])`

eksperimental:
I have implemented tracking private macros and functions. Please have a look.
There is one build failing, it looks like OTP 19 was not tracking these functions, unless it is something related to Elixir v1.7.
This would mean that we will have to move to Elixir v1.8 as the minimum Elixir version.

josevalim:
Apparently I commented in the wrong place, so I will do it here again:

This is using many private APIs from Elixir. Code.Typespec, the debug chunk, none of them are public. I also don't believe we should go down this path, it will make things more complex, more expensive, for little to not benefit. Not even Elixir tries to track these when showing private or undefined function warnings. So I am 👎. :)

wojtekmach:
Agreed with @josevalim

eksperimental:
> Apparently I commented in the wrong place, so I will do it here again:
> 
> This is using many private APIs from Elixir. Code.Typespec, the debug chunk, none of them are public. I also don't believe we should go down this path, it will make things more complex, more expensive, for little to not benefit. Not even Elixir tries to track these when showing private or undefined function warnings. So I am -1. :)

darn! I didnt think that debug_info chunk was not public.

eksperimental:
once #1224 I will update this PR with a middle ground solution that doesn't read from a non public API

wojtekmach:
@eksperimental I initially wrote we could support this feature and that it makes sense to do it for completeness but after reading @josevalim comment I agree that it would add little benefit while increasing complexity, so I wouldn't pursue it further either.

eksperimental:
Well, the implementation was done but the time I posted my comment. but yeah.. it adds complexity.

PR #1224, opened 2020-07-17
Optimize Ref caching system

eksperimental:
Only load and fetch_docs from modules when needed.
When the module is already chached, and we cannot find the ref in the ETS table,
we just define it as :undefined and store it.

### DIFF
--- a/lib/ex_doc/refs.ex
+++ b/lib/ex_doc/refs.ex
@@ -54,18 +54,15 @@ defmodule ExDoc.Refs do
           :undefined
         else
           _ ->
+            kind = elem(ref, 0)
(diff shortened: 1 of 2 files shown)


wojtekmach:
@eksperimental I went ahead in a bit different direction and committed https://github.com/elixir-lang/ex_doc/pull/1225 which I think has a bit simpler underlying design, i.e. doing the optimisation you had in mind was just a matter of these minor changes https://github.com/elixir-lang/ex_doc/commit/c6abc635afad2ed1cd66428cd14cefa3da566e84. Let me know if that looks good or if I missed anything. Thanks!

PR #1226, opened 2020-07-18
Escape escape signature.

eksperimental:
It was breaking HTML validation when function name had special HTML chars that
needs to be escaped. such as `<`


wojtekmach:
Thanks!

PR #1227, opened 2020-07-18
Escape HTML special characters in HTML sidebar entries

eksperimental:
Similar to https://github.com/elixir-lang/ex_doc/pull/1226

This bug was preventing funcions with special chars in their names to be displayed in the sidebar,
such as Kernel.</2


wojtekmach:
I just tried master and it seems it's working fine for me:

image

Can you reproduce it?

wojtekmach:
Nevermind, I see it now. Thanks!

eksperimental:
It was only happening with `<` but not with `<=`, I think the browser was interpreting it as a comment line.

wojtekmach:
Yeah, good call.

PR #1230, opened 2020-07-20
Remove overlapping functions from defaults

eksperimental:
Removes from the :default field in %ExDoc.FunctionNode,
the default functions that overlap functions defined elsewhere,
and therefore printing two HTML elements with the same ID, making it an invalid HTML page.

It takes into consideration the ordered the functions are showed up in the HTML.

With this commit, the HTML documentation for Elixir is valid HTML
according to https://github.com/svenkreiss/html5validator


eksperimental:
A real life example that triggered this was the default of `Supervisor.start_link/3` that overlapped `Supervisor.start_link/2`

josevalim:

eksperimental:
@josevalim Simply amazing how you managed to simplify the code in just a few lines. I didn't know of the existence of `Kernel.update_in/2`

Issue #1219, opened 2020-07-14
ExDoc cannot deal with literal map typespecs :optional and :required fields 


eksperimental:
Can't deal with

```elixir
      | %{required(key_type) => value_type}   # map with required pairs of key_type and value_type
      | %{optional(key_type) => value_type}   # map with optional pairs of key_type and value_type
```

It assumes `required` and òptional` are types

josevalim:
Can you please explain what you mean by "can't deal"? Do you get an error? If so, which one? Is it a warning? A failure to render? Thank you.

eksperimental:
It warns about `t:optional/1` and `t:required\1` not being defined.
this commit in this pending PR solves it.
https://github.com/elixir-lang/ex_doc/pull/1233/commits/16939dfdf6982f7ec50dd776f0d92b5e49551674

josevalim:
Thanks, I see! Maybe we should send a separate PR for it with tests. In particular, I would probably handle them in a separate clause. :)

eksperimental:
Sorry, it does not work in master. It does in that PR that we fine grain the warnings.

wojtekmach:
Fixed in #1233.

PR #1233, opened 2020-07-23
Warning based on refs

eksperimental:
Update warning messages based on Ref's visibility    
    Messages are more accurate now.

    Deal with private types

This PR spawned from https://github.com/elixir-lang/ex_doc/pull/1205 tackling most of the issues addressed there.

### DIFF
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -480,7 +480,9 @@ defmodule ExDoc.Autolink do
         nil
 
       _ ->
-        maybe_warn(ref, config, visibility, %{original_text: original_text})
+        unless kind == :type and name in [:required, :optional] and arity == 1 do
+          maybe_warn(ref, config, visibility, %{original_text: original_text})
+        end
 
         nil
     end
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -71,16 +71,13 @@ defmodule ExDoc.Autolink do
       url = custom_link(attrs, config) ->
         {:a, Keyword.put(attrs, :href, url), inner}
 
-      url = extra_link(attrs, config) ->
-        {:a, Keyword.put(attrs, :href, url), inner}
-
       true ->
         ast
     end
   end
 
   defp walk({:code, attrs, [code]} = ast, config) do
-    if url = url(code, :regular, config) do
+    if url = url(code, :regular_link, config) do
       code = remove_prefix(code)
       {:a, [href: url], [{:code, attrs, [code]}]}
     else
@@ -95,17 +92,23 @@ defmodule ExDoc.Autolink do
   @ref_regex ~r/^`(.+)`$/
 
   defp custom_link(attrs, config) do
-    with {:ok, href} <- Keyword.fetch(attrs, :href),
-         [[_, text]] <- Regex.scan(@ref_regex, href) do
-      url(text, :custom_link, config)
-    else
-      _ -> nil
+    case Keyword.fetch(attrs, :href) do
+      {:ok, href} ->
+        case Regex.scan(@ref_regex, href) do
+          [[_, custom_link]] ->
+            url(custom_link, :custom_link, config)
+
+          [] ->
+            build_extra_link(href, config)
+        end
+
+      _ ->
+        nil
     end
   end
 
-  defp extra_link(attrs, config) do
-    with {:ok, href} <- Keyword.fetch(attrs, :href),
-         uri <- URI.parse(href),
+  defp build_extra_link(link, config) do
+    with uri <- URI.parse(link),
          nil <- uri.host,
          true <- is_binary(uri.path),
          false <- uri.path =~ @ref_regex,
@@ -117,7 +120,7 @@ defmodule ExDoc.Autolink do
         fragment = (uri.fragment && "#" <> uri.fragment) || ""
         HTML.text_to_id(without_ext) <> config.ext <> fragment
       else
-        maybe_warn(nil, config, nil, %{file_path: uri.path, original_text: href})
+        maybe_warn(nil, config, nil, %{file_path: uri.path, original_text: link})
 
         nil
       end
@@ -185,6 +188,7 @@ defmodule ExDoc.Autolink do
     timeout: 0
   ]
 
+  defp url("", _mode, _config), do: nil
   defp url(string = "mix help " <> name, mode, config), do: mix_task(name, string, mode, config)
   defp url(string = "mix " <> name, mode, config), do: mix_task(name, string, mode, config)
 
@@ -196,10 +200,10 @@ defmodule ExDoc.Autolink do
 
           case parse_module_function(rest) do
             {:local, function} ->
-              local_url(kind, function, arity, config, string)
+              local_url(kind, function, arity, config, string, mode: mode)
 
             {:remote, module, function} ->
-              remote_url(kind, module, function, arity, config, string)
+              remote_url({kind, module, function, arity}, config, string, mode: mode)
 
             :error ->
               nil
@@ -308,7 +312,7 @@ defmodule ExDoc.Autolink do
         parts = name |> String.split(".") |> Enum.map(&Macro.camelize/1)
         module = Module.concat([Mix, Tasks | parts])
 
-        {module, module_url(module, :regular, config, string),
+        {module, module_url(module, :regular_link, config, string),
          Refs.get_visibility({:module, module})}
       else
         {nil, nil, :undefined}
@@ -381,7 +385,7 @@ defmodule ExDoc.Autolink do
 
       url =
         if module do
-          remote_url(:type, module, name, arity, config, string)
+          remote_url({:type, module, name, arity}, config, string)
         else
           local_url(:type, name, arity, config, string)
         end
@@ -418,20 +422,24 @@ defmodule ExDoc.Autolink do
 
   ## Internals
 
+  defp module_url(module, _mode, %{current_module: module}, _string) do
+    "#content"
+  end
+
   defp module_url(module, mode, config, string) do
-    if module == config.current_module do
-      "#content"
-    else
-      if Refs.public?({:module, module}) do
+    ref = {:module, module}
+
+    case {mode, Refs.get_visibility(ref)} do
+      {_, :public} ->
         app_module_url(tool(module), module, config)
-      else
-        if mode == :custom_link do
-          ref = {:module, module}
-          maybe_warn(ref, config, Refs.get_visibility(ref), %{original_text: string})
-        end
+
+      {:regular_link, :undefined} ->
+        nil
+
+      {_, visibility} ->
+        maybe_warn(ref, config, visibility, %{original_text: string})
 
         nil
-      end
     end
   end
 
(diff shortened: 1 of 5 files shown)


eksperimental:
Please do not squash the commits

test/ex_doc/formatter/html_test.exs
```diff
@@ -100,12 +100,12 @@ defmodule ExDoc.Formatter.HTMLTest do
...
       end)
 
-    assert output == "warning: index.html redirects to Randomerror.html, which does not exist\n"
+    assert output =~ "warning: index.html redirects to Randomerror.html, which does not exist\n"
     assert File.regular?("#{output_dir()}/index.html")
     assert File.regular?("#{output_dir()}/RandomError.html")
   end
 
-  test "warns on undefined functions" do
+  test "warns on undefined function within the same module" do
```
wojtekmach, line 108: not sure about this test description change, I don't think being within the same module is relevant, is it?

test/ex_doc/refs_test.exs
```diff
@@ -40,7 +40,7 @@ defmodule ExDoc.RefsTest do
     assert Refs.get_visibility({:type, InMemory, :t, 0}) == :undefined
 
     # @typep
-    assert Refs.get_visibility({:type, :sets, :seg, 0}) == :undefined
+    assert Refs.get_visibility({:type, :sets, :seg, 0}) == :hidden
```
wojtekmach, line 43: 👍

lib/ex_doc/autolink.ex
```diff
@@ -389,71 +394,105 @@ defmodule ExDoc.Autolink do
...
+
+      {:regular_link, :undefined} ->
+        nil
+
+      {_, visibility} ->
+        message =
+          if mix_task? do
+            "documentation references \"mix #{module_string}\" but such task is #{visibility}"
+          else
+            "documentation references module \"#{module_string}\" but it is #{visibility}"
```
wojtekmach, line 415: I'm starting to think that _kind_ in "documentation references kind (...)" isn't that useful after all and we could show the prefix (if any) instead:

    documentation references module "Unknown" but it is undefined
    documentation references function "Unknown.foo/1" but it is undefined
    documentation references type "Unknown.t/1" but it is undefined
    documentation references callback "Unknown.handle_foo/1" but it is undefined

we could do:

    documentation references "Unknown" but it is undefined
    documentation references "Unknown.foo/1" but it is undefined
    documentation references "t:Unknown.t/1" but it is undefined
    documentation references "c:Unknown.handle_foo/1" but it is undefined

What's nice is we avoid the function vs macro problem (not that it is a big deal). I think this could potentially simplify some code too, what do you all think? If that sounds reasonable, I'm happy to tackle it later.

I guess the downside, and I think this was the reason it is like that in the first place, is e.g. `t:` is specific to ExDoc, so if we'd eventually support EDoc docs with ExDoc, using the `t:` in the warning would be a bit awkward. But on the flip side, the warning comes from ExDoc so it's reasonable to use conventions that make sense for _it_.

If we do this, I think we'd also simplify the code around Mix tasks, I think what we have on this patch:

    documentation references \"mix ...\" but such task is ..."

is fine, but I'd personally be totally fine with:

    documentation references \"mix ...\" but it is ..."

I don't think we lose much because we have "mix " in the warning already. See the other comment about mix tasks too.

test/ex_doc/autolink_test.exs
```diff
@@ -168,11 +175,37 @@ defmodule ExDoc.AutolinkTest do
...
+      assert captured =~
+               "documentation references module \"LICENSE\" but it is undefined"
+
+      captured =
+        assert_warn(fn ->
+          assert_unchanged(~m"[an unknown task](`mix unknown.task`)")
+        end)
+
+      assert captured =~
+               "documentation references \"mix unknown.task\" but such task is undefined"
```
wojtekmach, line 208: could you also add the following test? `assert_unchanged(~m"[an unknown task](`Mix.Tasks.Unknown`)")`

the warning isn't ideal but that's ok.

test/ex_doc/formatter/html_test.exs
eksperimental: nope. I have removed this in other PR i am submitting after this. I will delete it here.

lib/ex_doc/autolink.ex
eksperimental: I finished before you reviewed this code another PR that refactor the warn and maybe_warn functions. What you suggest could easily be done there. I will submit the PR and we can have this conversation over there

eksperimental: Could you please copy and past this comment over there? https://github.com/elixir-lang/ex_doc/pull/1235
So I can make this issue as resolved. This PR will be based on that one if merged

test/ex_doc/autolink_test.exs
eksperimental: I think we need to store exact string that the user passed, and print it out in the error. Another example is if you call `Elixir.String.upcase/9`. It will be printed out as `String.upcase/9`.
I am submitting another PR that deals with the `Elixir.` namespace, but printing the original string has been left out, so this could potentially be included in the warn and maybe_warn refactoring

wojtekmach: Yes, I too think keeping the original is the way to go because it will aid in finding the broken ref in the docs. Good call.

lib/ex_doc/autolink.ex
```diff
@@ -184,8 +188,9 @@ defmodule ExDoc.Autolink do
     timeout: 0
   ]
 
-  defp url("mix help " <> name, _mode, config), do: mix_task(name, config)
-  defp url("mix " <> name, _mode, config), do: mix_task(name, config)
+  defp url("", _mode, _config), do: nil
```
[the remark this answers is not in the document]
eksperimental: https://spec.commonmark.org/0.29/#example-482
Title and destination of links can be omitted.
While it works because the last clause will catch it, it will save the trip through that clause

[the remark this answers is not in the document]
wojtekmach: I added a raise to that clause and tested with `[foo]()` and the clause wasn’t hit, could you double check?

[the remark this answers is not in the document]
wojtekmach: or don’t check but just remove the clause imho :) I don’t think we need to optimize for extremely unlikely cases and thus keep the code simpler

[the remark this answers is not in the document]
eksperimental: removed!

test/ex_doc/formatter/html_test.exs
eksperimental: fixed  here.

test/ex_doc/autolink_test.exs
```diff
@@ -394,10 +437,27 @@ defmodule ExDoc.AutolinkTest do
...
-    assert captured =~ "documentation references file `Foo Bar.md` but it doesn't exist"
+    assert captured =~ "documentation references file \"Foo Bar.md\" but it doesn't exist"
 
     options = [skip_undefined_reference_warnings_on: ["MyModule"], module_id: "MyModule"]
     assert_unchanged("String.upcase/9", options)
+
+    captured =
+      assert_warn(fn ->
+        opts = [extras: []]
+        assert_unchanged(~m"[bad](`String.upcase/9`)", opts)
```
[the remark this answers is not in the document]
eksperimental: fixed

test/ex_doc/formatter/html_test.exs
```diff
@@ -100,7 +100,7 @@ defmodule ExDoc.Formatter.HTMLTest do
         generate_docs(doc_config(main: "Randomerror"))
       end)
 
-    assert output == "warning: index.html redirects to Randomerror.html, which does not exist\n"
+    assert output =~ "warning: index.html redirects to Randomerror.html, which does not exist\n"
```
wojtekmach, line 103: this change is unnecessary

wojtekmach:
I tested this branch against Elixir docs and got these warnings:

```
warning: documentation references type "Exception.optional/1" but it is undefined
  lib/elixir/lib/exception.ex:15: t:Exception.t/0

warning: documentation references type "URI.optional/1" but it is undefined
  lib/elixir/lib/uri.ex:115: URI.decode_query/2

warning: documentation references type "URI.optional/1" but it is undefined
  lib/elixir/lib/uri.ex:115: URI.decode_query/2

(...)
```

eksperimental:
> lib/elixir/lib/exception.ex:15

It is the last missing piece to be implemented 
https://github.com/elixir-lang/ex_doc/issues/1219

test/ex_doc/formatter/html_test.exs
eksperimental: fixed

test/ex_doc/autolink_test.exs
eksperimental: tackled in https://github.com/elixir-lang/ex_doc/pull/1235/commits/e0a7e493445f6946902e85facfceb4ac914002fa

eksperimental:
Rebased to master.
This is ready for review @wojtekmach
Please do not squash the commits.

test/ex_doc/autolink_test.exs
```diff
@@ -1,5 +1,5 @@
 defmodule ExDoc.AutolinkTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case
```
wojtekmach, line 2: if we test all warnings in that one test we could run it concurrently again, should we do that?

eksperimental: Tests do not run concurrently within the same module, regardless of the `:async` option, so that wouldn't be the limitation, if that is what you mean.
Concurrency was disabled because there are two ExUnit cases that are capturing :stderr, the other one is  `ExDoc.CLITest`

josevalim: It is supported on Elixir v1.11 though so let's leave the option in and push everyone to more recent Elixir versions.

test/ex_doc/autolink_test.exs
```diff
@@ -401,18 +443,16 @@ defmodule ExDoc.AutolinkTest do
   end
 
   defp assert_unchanged(ast_or_text, options \\ []) do
-    assert autolink(ast_or_text, options) == ast(ast_or_text)
+    fun = fn ->
+      assert autolink(ast_or_text, options) == ast(ast_or_text)
+    end
+
+    capture_io(:stderr, fun)
```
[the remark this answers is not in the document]
eksperimental: You are totally right. It was probably before the recent changes you introduced to warn, which is running assert_change now. If you see the test, I run the same AST for warn and for assert_unchanged.
Anyway, I have fixed it.

test/ex_doc/autolink_test.exs
eksperimental: > It is supported on Elixir v1.11 though so let's leave the option in and push everyone to more recent Elixir versions.
@josevalim Sorry, but what is supported in v1.11?

josevalim:
Concurrent captures of stderr.
-- 


*José Valimhttps://dashbit.co/ <https://dashbit.co/>*

eksperimental:
> Concurrent captures of stderr. -- *José Valimhttps://dashbit.co/ <https://dashbit.co/>*

Oh.. I was not aware of. Is this documented somewhere? I couldn't find any reference nor in the ExUnit.CaptureIO docs neither in the CHANGELOG

josevalim:
Maybe it is in 1.10 then?
-- 


*José Valimhttps://dashbit.co/ <https://dashbit.co/>*

eksperimental:
> Maybe it is in 1.10 then? -- *José Valimhttps://dashbit.co/ <https://dashbit.co/>*

is this the PR? https://github.com/elixir-lang/elixir/pull/9527

josevalim:
Yes!

eksperimental:
Sorry, I accidentally closed this, and I am unable to reopen it. Can anybody reopen it?

josevalim:
Apparently I cannot reopen it either.

eksperimental:
I accidendally force pushed to a version that was a the same commit as master, and GitHub automatically closed it.

test/ex_doc/autolink_test.exs
eksperimental: > if we test all warnings in that one test we could run it concurrently again, should we do that?

Do we really have a limitation having different tests emitting a warning since they are in the same test case?

test/ex_doc/autolink_test.exs
```diff
@@ -167,6 +171,22 @@ defmodule ExDoc.AutolinkTest do
...
+      assert warn(~m"[custom text](Unknown)") =~
+               "documentation references file \"Unknown\" but it does not exist"
+
+      assert warn(~m"[custom text](`LICENSE`)", extras: ["LICENSE"]) =~
+               "documentation references module \"LICENSE\" but it is undefined"
+
+      assert warn(~m"[an unknown task](`mix unknown.task`)") =~
+               "documentation references \"mix unknown.task\" but it is undefined"
+
+      assert_unchanged(~m"[an unknown task](`mix unknown.task`)")
```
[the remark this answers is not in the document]
eksperimental: I have mover all the warnings to its own test, regarding the non warning ones. the Elixir ones are only checked in the test named  "special case links" and the Erlang ones are not checked anywhere. I would say it is better to leave them in this test

lib/ex_doc/autolink.ex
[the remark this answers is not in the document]
wojtekmach: ping

lib/ex_doc/autolink.ex
```diff
@@ -442,54 +450,79 @@ defmodule ExDoc.Autolink do
...
-          else
-            app_module_url(tool, module, config) <> fragment(tool, kind, name, arity)
-          end
-      end
-    else
-      if warn? and Refs.public?({:module, module}) do
-        maybe_warn(ref, config, Refs.get_visibility(ref), %{original_text: original_text})
-      end
+          tool ->
+            if module == config.current_module do
```
wojtekmach, line 508:
```suggestion
            if same_module? do
```

lib/ex_doc/autolink.ex
```diff
@@ -442,54 +450,79 @@ defmodule ExDoc.Autolink do
...
+
+      {:type, :hidden} ->
+        nil
+
+      _ ->
+        unless kind == :type and name in [:required, :optional] and arity == 1 do
+          maybe_warn(ref, config, visibility, %{original_text: original_text})
+        end
+
+        nil
```
wojtekmach, line 486:
```suggestion
      # skip `@type %{required(...), optional(...), ...}`
      {:type, visibility} when name in [:required, :optional] and arity == 1 ->
        nil

      _ ->
        maybe_warn(ref, config, visibility, %{original_text: original_text})
        nil
```

lib/ex_doc/refs.ex
```diff
@@ -114,13 +114,17 @@ defmodule ExDoc.Refs do
...
+          {{:module, module}, module_visibility}
+          | to_refs(types(module, [:typep]), module, :type, :hidden)
+        ])
 
       {:error, _} ->
         if Code.ensure_loaded?(module) do
           (to_refs(exports(module), module, :function) ++
              to_refs(callbacks(module), module, :callback) ++
-             to_refs(types(module), module, :type))
+             to_refs(types(module), module, :type) ++
```
wojtekmach, line 126:
```suggestion
             to_refs(types(module, [:type, :opaque]), module, :type) ++
```

lib/ex_doc/refs.ex
```diff
@@ -151,11 +155,11 @@ defmodule ExDoc.Refs do
     end
   end
 
-  defp types(module) do
+  defp types(module, kind_list \\ [:type, :opaque]) do
```
wojtekmach, line 158:
```suggestion
  defp types(module, kind_list) do
```

wojtekmach (commented):
Looks good to me, thank you. I added a few comments. Please also rebase with latest master.

lib/ex_doc/autolink.ex
eksperimental: Much better. Done.

lib/ex_doc/autolink.ex
eksperimental: done

eksperimental:
> Looks good to me, thank you. I added a few comments. Please also rebase with latest master.

All suggestions implemented.  And code has been rebased to master.

lib/ex_doc/autolink.ex
wojtekmach: I don't see this change applied, did you accidentally overwrite your GitHub changes when rebasing and force pushing?

eksperimental: Sorry, since I rebased, i had to force push them. They are up now.

wojtekmach:
Thanks!

eksperimental:
Thank you @wojtekmach for reviewing this one

PR #1234, opened 2020-07-24
Update tooltips to handle new escape values and ignore non-html files.

### LINKED ISSUE #1177: Unexpected bahavior on hovering over a download link
Example [Phoenix 1.5.3 doc](https://hexdocs.pm/phoenix/overview.html): 
> If you would prefer to read these guides as an EPUB [click here!](Phoenix.epub)

### LINKED ISSUE #1232: Bug: Reference popover for Kernel.<>/2 shows the introduction for the Kernel module.
/cc @archdragon

archdragon:
This change contains three fixes:

## 1. Fix displaying tooltips for functions containing special characters (by correctly handling escaped IDs).

Change in how IDs are being formatted/escaped broke some tooltips. I've updated the script to correctly handle escaped values.

image

Resolves #1232 

## 2. Fix linking to non-html files

Tooltips won't try to fire for non-html links.

Resolves #1177 

## 2. Fix linking to module sections

|before|after|
|---|---|
|image|image|

Links to module sections (like "guards" or "functions") were displaying a (completely useless :) ) tooltip window. This will no longer happen.


josevalim:

PR #1235, opened 2020-07-25
Change Autolink.warn/maybe_warn functions so they keep the original string provided in the documentation

eksperimental:
- Change Autolink.warn/maybe_warn functions so they keep the original string provided in the documentation.
- The messages are defined within the warn functions.
- Show original string in warns.
- Warn on custom links with missing Mix tasks.
- Warn on custom links with missing Modules.
- Do not use kind in warnings for type, callback, functions/macros.

### DIFF
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -116,7 +116,7 @@ defmodule ExDoc.Autolink do
         fragment = (uri.fragment && "#" <> uri.fragment) || ""
         HTML.text_to_id(without_ext) <> config.ext <> fragment
       else
-        maybe_warn(nil, config, nil, %{file_path: uri.path})
+        maybe_warn(nil, config, nil, %{file_path: uri.path, original_text: href})
 
         nil
       end
@@ -187,18 +187,18 @@ defmodule ExDoc.Autolink do
   defp url("mix help " <> name, _mode, config), do: mix_task(name, config)
   defp url("mix " <> name, _mode, config), do: mix_task(name, config)
 
-  defp url(code, mode, config) do
-    case Regex.run(~r{^(.+)/(\d+)$}, code) do
+  defp url(string, mode, config) do
+    case Regex.run(~r{^(.+)/(\d+)$}, string) do
       [_, left, right] ->
         with {:ok, arity} <- parse_arity(right) do
           {kind, rest} = kind(left)
 
           case parse_module_function(rest) do
             {:local, function} ->
-              local_url(kind, function, arity, config)
+              local_url(kind, function, arity, config, string)
 
             {:remote, module, function} ->
-              remote_url(kind, module, function, arity, config)
+              remote_url(kind, module, function, arity, config, string)
 
             :error ->
               nil
@@ -209,7 +209,7 @@ defmodule ExDoc.Autolink do
         end
 
       nil ->
-        case parse_module(code, mode) do
+        case parse_module(string, mode) do
           {:module, module} ->
             module_url(module, config)
 
@@ -352,9 +352,9 @@ defmodule ExDoc.Autolink do
 
       url =
         if module do
-          remote_url(:type, module, name, arity, config)
+          remote_url(:type, module, name, arity, config, string)
         else
-          local_url(:type, name, arity, config)
+          local_url(:type, name, arity, config, string)
         end
 
       if url do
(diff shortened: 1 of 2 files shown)


wojtekmach:
(ref: https://github.com/elixir-lang/ex_doc/pull/1233#discussion_r460420065)

I'm starting to think that _kind_ in "documentation references kind (...)" isn't that useful after all and we could show the prefix (if any) instead:

    documentation references module "Unknown" but it is undefined
    documentation references function "Unknown.foo/1" but it is undefined
    documentation references type "Unknown.t/1" but it is undefined
    documentation references callback "Unknown.handle_foo/1" but it is undefined

we could do:

    documentation references "Unknown" but it is undefined
    documentation references "Unknown.foo/1" but it is undefined
    documentation references "t:Unknown.t/1" but it is undefined
    documentation references "c:Unknown.handle_foo/1" but it is undefined

What's nice is we avoid the function vs macro problem (not that it is a big deal). I think this could potentially simplify some code too, what do you all think? If that sounds reasonable, I'm happy to tackle it later.

I guess the downside, and I think this was the reason it is like that in the first place, is e.g. `t:` is specific to ExDoc, so if we'd eventually support EDoc docs with ExDoc, using the `t:` in the warning would be a bit awkward. But on the flip side, the warning comes from ExDoc so it's reasonable to use conventions that make sense for _it_.

If we do this, I think we'd also simplify the code around Mix tasks, I think what we have on this patch:

    documentation references \"mix ...\" but such task is ..."

is fine, but I'd personally be totally fine with:

    documentation references \"mix ...\" but it is ..."

I don't think we lose much because we have "mix " in the warning already. See the other comment about mix tasks too.

eksperimental:
I have tackled both issues https://github.com/elixir-lang/ex_doc/pull/1233#discussion_r460420091 and https://github.com/elixir-lang/ex_doc/pull/1235#issuecomment-663902716
Please have a look @wojtekmach

eksperimental:
ping @wojtekmach

lib/ex_doc/autolink.ex
```diff
@@ -537,4 +530,69 @@ defmodule ExDoc.Autolink do
 
     IO.puts(:stderr, [warning, message, ?\n, stacktrace, ?\n])
   end
+
+  defp warn(
```
wojtekmach, line 534: we're missing a test for this clause

lib/ex_doc/autolink.ex
```diff
@@ -537,4 +530,69 @@ defmodule ExDoc.Autolink do
...
+         %{mix_task: true, original_text: original_text} = metadata
+       ) do
+    message =
+      "documentation references \"#{original_text}\" but such task is" <>
+        format_visibility(visibility, :module)
+
+    warn(message, {file, line}, id, visibility, Map.put(metadata, :message, message))
+  end
+
+  defp warn(
```
wojtekmach, line 548: we're missing test for this too

lib/ex_doc/autolink.ex
```diff
@@ -509,25 +510,17 @@ defmodule ExDoc.Autolink do
...
-  defp warn({kind, module, name, arity}, file, line, id) do
-    message =
-      "documentation references #{kind} #{inspect(module)}.#{name}/#{arity}" <>
-        " but it is undefined or private"
-
-    warn(message, file, line, id)
-  end
-
-  defp warn(message, file, line, id) do
+  defp warn(_ref, {file, line}, id, _visibility, %{message: message}) do
```
wojtekmach, line 523: I think this could be a warn/3:

```suggestion
  defp warn(message, {file, line}, id) do
```

and then all the other warn/5 clauses call it.

lib/ex_doc/autolink.ex
```diff
@@ -116,8 +116,8 @@ defmodule ExDoc.Autolink do
         fragment = (uri.fragment && "#" <> uri.fragment) || ""
         HTML.text_to_id(without_ext) <> config.ext <> fragment
       else
-        message = "documentation references file `#{uri.path}` but it doesn't exist"
-        warn(message, config.file, config.line, config.id)
+        maybe_warn(nil, config, nil, %{file_path: uri.path, original_text: href})
```
wojtekmach, line 119: warning for missing extras is quite different from all the other ones: there's a file_path & original_text, there's neither ref nor visibility, should we call warn/3 directly (basically as before) to remove this indirection?

wojtekmach:
> Refactor Autolink warn and maybe_warn functions

sorry for being pedantic but since we're changing behaviour this is not strictly a _refactoring_. :-) could you update the PR title or description with a bullet list of changes? This will be helpful in adding a changelog entry later.

lib/ex_doc/autolink.ex
eksperimental: Nope. Because maybe_warn checks the files that are excluded in the config, such as `skip_undefined_reference_warnings_on` and the upcoming  `skip_undefined_reference_warnings_for`

lib/ex_doc/autolink.ex
eksperimental: done

lib/ex_doc/autolink.ex
eksperimental: done

eksperimental:
> > Refactor Autolink warn and maybe_warn functions
> 
> sorry for being pedantic but since we're changing behaviour this is not strictly a _refactoring_. :-) could you update the PR title or description with a small bullet list of changes? This will be helpful in adding a changelog entry later.

done
All requests done.

lib/ex_doc/autolink.ex
eksperimental: done

wojtekmach:
I merge this too fast, there's a double warning for Mix tasks:

```
warning: documentation references module "mix foo" but it is undefined
  nofile

warning: documentation references "mix foo" but such task is undefined
  nofile
```

could you take a look?

wojtekmach:
Btw, I just updated the test helper so it catches the duplicate warning.

wojtekmach:
Nvm, fixed in 0646982.

PR #1241, opened 2020-08-08
Do not run concurrent tests when Refs is used

eksperimental:
To avoid race conditions.

### DIFF
--- a/test/ex_doc/refs_test.exs
+++ b/test/ex_doc/refs_test.exs
@@ -1,5 +1,5 @@
 defmodule ExDoc.RefsTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case
   alias ExDoc.Refs
 
   defmodule InMemory do
--- a/test/ex_doc/refs_test.exs
+++ b/test/ex_doc/refs_test.exs
@@ -10,6 +10,10 @@ defmodule ExDoc.RefsTest do
     def foo(), do: :ok
   end
 
+  setup do
+    ExDoc.Refs.clear()
+  end
+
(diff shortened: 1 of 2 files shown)


josevalim:
I would rather run tests as concurrently as we can. What are the race conditions? Can we instead make sure they all work on the same data?

josevalim:
Yeah, one option then is to use `__MODULE__.Foo`, that will remove the race condition and make them easier to debug at the same time. :)

eksperimental:
Sorry for being reiterative, but that approach will fix the existing bug, but it is prone to happen again if we use share module names across different tests files. I think we shouldn't share an ETS table for unrelated tests across different files. It is really hard to track what is going on when an error hits, and passing the same seed doesn't reproduce it.
So it is your call.

josevalim:
If we grow the habit of using `__MODULE__` for custom define modules, then we will reduce the chances considerably, and, if they happen, it would only be within the same file. I value the time spent running tests considerably, so that would be my choice. :) Thanks!

test/ex_doc/refs_test.exs
```diff
@@ -10,6 +10,10 @@ defmodule ExDoc.RefsTest do
     def foo(), do: :ok
   end
 
+  setup do
+    ExDoc.Refs.clear()
+  end
```
eksperimental, line 15: this still can be fixed. Let me know so I can submit a PR

josevalim: Clearing will cause issues though because we can erase something that has just been defined in another test. Ideally we shouldn't ever clear.

PR #1243, opened 2020-08-12
Speed up slow tests

eksperimental:
Do not use String.Unicode module for testing,
which happens to be the heaviest module in the Elixir codebase.


josevalim:

PR #1245, opened 2020-08-15
Escape HTML special characters in HTML sidebar entries (2nd take)

### LINKED ISSUE #1244: Suggestions wrongly deal with escaped chars
If you type in the search box "&" you will get the suggestion of functions that their name has been escaped such as "&lt;/2". Same if you type "gt".

I think we should keep the original name and the escaped name.

eksperimental:
This is a second take to fix what was intented by d26ca71417c40f3e5d9ab49d0933b277c983962d
Superseeds #1227
Closes #1244

### DIFF
--- a/assets/js/autocomplete/display.js
+++ b/assets/js/autocomplete/display.js
@@ -4,6 +4,7 @@
 import $ from 'jquery'
 import autocompleteResultsTemplate from '../templates/autocomplete-suggestions.handlebars'
 import { getSuggestions } from './suggestions'
+import { escapeHtmlEntities } from '../helpers'
 
 // Constants
 // ---------
@@ -16,7 +17,7 @@ function updateSuggestions (term) {
   const template = autocompleteResultsTemplate({
     empty: results.length === 0,
     results: results,
-    term: term
+    term: escapeHtmlEntities(term)
   })
 
   autocompleteElement.html(template)
--- a/assets/js/autocomplete/display.js
+++ b/assets/js/autocomplete/display.js
@@ -4,7 +4,6 @@
 import $ from 'jquery'
 import autocompleteResultsTemplate from '../templates/autocomplete-suggestions.handlebars'
 import { getSuggestions } from './suggestions'
-import { escapeHtmlEntities } from '../helpers'
 
 // Constants
 // ---------
@@ -17,7 +16,7 @@ function updateSuggestions (term) {
   const template = autocompleteResultsTemplate({
     empty: results.length === 0,
     results: results,
-    term: escapeHtmlEntities(term)
+    term: term
   })
 
   autocompleteElement.html(template)
--- a/assets/js/autocomplete/suggestions.js
+++ b/assets/js/autocomplete/suggestions.js
@@ -154,10 +154,11 @@ function highlight (match) {
   const EmOpen = '###EM_OPEN###'
   const EmClose = '###EM_CLOSE###'
 
-  let result = match.input
-  result = result.replace(match, EmOpen + match[0] + EmClose)
-  result = escapeHtmlEntities(result)
-  return result.replace(EmOpen, '<em>').replace(EmClose, '<\/em>')
+  let result = match.input.replace(match, EmOpen + match[0] + EmClose)
+
+  return escapeHtmlEntities(result)
+    .replace(EmOpen, '<em>')
+    .replace(EmClose, '<\/em>')
 }
 
 /**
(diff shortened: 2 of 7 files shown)


eksperimental:
Needs reviewing from JS programmers
/cc @davydog187 @archdragon

assets/js/autocomplete/display.js
```diff
@@ -16,7 +17,7 @@ function updateSuggestions (term) {
   const template = autocompleteResultsTemplate({
     empty: results.length === 0,
     results: results,
-    term: term
+    term: escapeHtmlEntities(term)
```
archdragon, line 20: With the current escaping I would remove this escape and use 

```
<div class="title">"<em>{{term}}</em>"</div>
```

in `autocomplete-suggestions.js` to escape the term

assets/js/autocomplete/suggestions.js
```diff
@@ -151,7 +151,13 @@ function findIn (elements, term, categoryName) {
...
  */
 function highlight (match) {
-  return match.input.replace(match, `<em>${match[0]}</em>`)
+  const EmOpen = '###EM_OPEN###'
+  const EmClose = '###EM_CLOSE###'
+
+  let result = match.input
+  result = result.replace(match, EmOpen + match[0] + EmClose)
+  result = escapeHtmlEntities(result)
+  return result.replace(EmOpen, '<em>').replace(EmClose, '<\/em>')
```
archdragon, line 160: I think this is not needed - I'm not getting any escpaing issues with the default implementation.

assets/js/templates/sidebar-items.handlebars
```diff
@@ -62,7 +62,7 @@
               <ul class="{{group.key}}-list deflist">
                 {{#each group.nodes}}
                   <li>
-                    <a href="{{node.id}}.html#{{anchor}}">{{{id}}}</a>
+                    <a href="{{node.id}}.html#{{anchor}}">{{id}}</a>
```
archdragon, line 65: 👍

eksperimental:
> With the current change on the backend, I think it's enough to just replace:
> 
> `{{{id}}}` with `{{id}}`
> 
> I think with the current backend changes this will take care of ID escaping and it will be enough to fix issue described.
> 
> I've tested just escaping `{{id}}` on my end and all works well then.
> 
> Does that make sense? Or have I missed something?

If you use double curly brackets the then it will be escaped twice.

eksperimental:
> With the current change on the backend, I think it's enough to just replace:
by backend you mean the Elixir backend or JavaScript?

assets/js/autocomplete/suggestions.js
eksperimental: The tricky part is when the search term contains a special html entity, such as `<` or `>`. Could you please try them with that. The changes in this functions is to deal with that edge case.

archdragon:
@eksperimental Sorry, I was not being precise!

I meat elixir changes in the current PR + just escaping id in `JavaScript`.

So to sum up I think that changing
```
 nodes = Enum.map(docs, fn doc -> %{id: h(doc.id), anchor: URI.encode(HTML.link_id(doc))} end)
```
to
```
nodes = Enum.map(docs, fn doc -> %{id: doc.id, anchor: URI.encode(HTML.link_id(doc))} end)
```

and
```
{{{id}}}
```

to

```
{{id}}
```

should be enough. WDYT?

assets/js/autocomplete/suggestions.js
archdragon: I see - I can break it with `">` thought we already account for that, but it's not the case.

image

thanks!

archdragon:
After reading the discussion comments, I see where extra escaping is needed. Adding my +1 :)

eksperimental:
> After reading the discussion comments, I see where extra escaping is needed. Adding my +1 :)

Thank you. Still I am considering one of your suggestions to simplify the code.

assets/js/autocomplete/display.js
eksperimental: This suggestion can be implemented.

eksperimental: done

assets/js/autocomplete/suggestions.js
eksperimental, line 160: @archdragon do you know of a more idiomatic way of doing this in JS?

josevalim:
Absolutely beautiful, thank you! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

eksperimental:
> Absolutely beautiful, thank you! green_heart blue_heart purple_heart yellow_heart heart

You are welcome!

PR #1247, opened 2020-08-16
Fix npm command instructions in README file

eksperimental:
Additionally it appends a slash to the assets folder in the npm commands.

### DIFF
--- a/assets/README.md
+++ b/assets/README.md
@@ -8,12 +8,12 @@ To work on these assets you need to install [Node.js] (version 10) and
 
 ## `npm run` scripts
 
-The following scripts are available
+The following scripts are available from the root folder of the project.
 
 ### `build`
 
 ```bash
-$ npm run --prefix build
+$ npm run --prefix assets build
 ```
 
 This will build a complete production bundle, including JavaScript and CSS.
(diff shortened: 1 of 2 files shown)


josevalim:
Why? In my experience this is not necessary at all. :)

eksperimental:
the commands in the assets/REAMDE.md file are wrong. You can run them. We should either remove the "--prefix" or add "--prefix asset" and run from root

josevalim:
The commands already include `--prefix assets` and that's enough. I don't understand why we are translating them to `--prefix assets/`. :)

eksperimental:
> The commands already include `--prefix assets` and that's enough. I don't understand why we are translating them to `--prefix assets/`. :)

The commands in README omit "assets/" making the commend invalid.

Personally I find useful to refer to folders with a trailing slash, and to files or simple prefixes without it.

josevalim:
Can we keep only the README changes without the / suffix? It is not used anywhere across Phoenix and other projects in the community. Thanks!

eksperimental:
> Can we keep only the README changes without the / suffix? It is not used anywhere across Phoenix and other projects in the community. Thanks!

sure

josevalim:

PR #1248, opened 2020-08-17
Add lint.fix task

eksperimental:
Aditionally, it makes `mix lint` not to format any Elixir code.


josevalim:
Looks great! What if we just call it `fix`? :)

eksperimental:
> Looks great! What if we just call it `fix`? :)

How about `fix` and `check` ?

josevalim:
Both are fine! `lint` may be a bit clearer on the type of checking that will be happening (i.e. it is not a type check or similar).

eksperimental:
> Both are fine! `lint` may be a bit clearer on the type of checking that will be happening (i.e. it is not a type check or similar).

Let' s just lint.fix then.

josevalim:

eksperimental:
@josevalim I was intending to rename it `fix`. I will submit a PR later.

Issue #1192, opened 2020-06-20
Feature Request: support skip_undefined_reference_warnings_for


eksperimental:
I think it will be useful to have an WE need to add an option: `skip_undefined_reference_warnings_for`:
if a function is listed here and it is referenced to, the warning will be ignored.

It addresses https://github.com/elixir-lang/ex_doc/pull/1151#issuecomment-646730079 
where calling `ExUnit.Case.__using__/1`, which is a macro with `@doc false` will emit a warning.

I can submit a PR if agreed on adding this feature.

josevalim:
Sounds good to me too!

eksperimental:
I already have a working version of this, but I'm waiting for https://github.com/elixir-lang/ex_doc/pull/1233 to be merged

PR #1249, opened 2020-08-19
Skip reference warnings to

eksperimental:


test/fixtures/warnings.ex
```diff
@@ -20,3 +20,21 @@ defmodule Warnings do
...
+  @moduledoc """
+  Refer to:
+  - [this file](unknown.md)
+  - [this module](`UknownModule`)
+  - [this function](`CompiledWithDocs.function/0`)
+  - [this callback](`c:CompiledWithDocs.callback/1`)
+  - [this type](`t:CompiledWithDocs.type/2`)
+
+  Refer to with Elixir namespace:
+  - [this module](`Elixir.UknownModule`)
```
maartenvanvliet, line 34:
```suggestion
  - [this module](`Elixir.UnknownModule`)
```

test/fixtures/warnings.ex
```diff
@@ -20,3 +20,21 @@ defmodule Warnings do
   """
   def foo(), do: :ok
 end
+
+defmodule Warnings.Submodule do
+  @moduledoc """
+  Refer to:
+  - [this file](unknown.md)
+  - [this module](`UknownModule`)
```
maartenvanvliet, line 28:
```suggestion
  - [this module](`UnknownModule`)
```

eksperimental: Fixed. Thank you!

eksperimental:
There are two features that can be discussed, but are not intended to be covered by this PR:

1. Whether we want to support Mix tasks. ie. "mix foo", instead of using the Mix.Tasks.TASKNAME module.

2. Whether we want to skip references to a a module and everything inside (functions, types, etc). As of now, listing the Module would ignore the warning for that module in particular, but nothing else.

josevalim:
Do you have any example of when we would use this feature?

eksperimental:
Closes #1192
Addresses #1151 (comment)

eksperimental:
Basically linking to hidden reference (i.e. with doc false), such as in https://github.com/elixir-lang/ex_doc/pull/1151#issuecomment-646730079 

where calling `ExUnit.Case.__using__/1`, which is a macro with `@doc false` and it will emit a warning when referencing it.

josevalim:
@eksperimental oh, I see. I wonder if we can tackle this in some other way. Having to change a completely unrelated config file for a mistake happening in the documentation elsewhere is not very clear. Maybe we can use triple backticks to denote code sections with no syntax highlighting. We would need to change markdown first. For now, I will fix this in Elixir by referring to the function indirectly. Thank you!

PR #1255, opened 2020-08-23
Improve Accessibility

eksperimental:
Most of the suggestions have been done by https://github.com/Khan/tota11y
It passes all the tests, with the exception of one test which I consider is not valid,
and it has been reported here: https://github.com/Khan/tota11y/issues/170

Changes include:
- Font colors have been updated to have enough contrast in day-mode and night-mode.

- The HTML structure has been updated to it only contain the title inside the heading tags (h1, h2), otherwise it will look weird in screen readers.

- The h1 title of the project in the content section has been changed so it looks like:
    Some module <small class="app-vsn">(ProjectName v0.0.1)</small>
	instead of:
		<small class="app-vsn">ProjectName v0.0.1</small> Some module

- Search box makes use of ARIA attributes

- Project name in the sidebar is not a heading anymore.

- Less have been simplified in some parts where code was duplicated in the night mode.

### DIFF
--- a/assets/less/content/general.less
+++ b/assets/less/content/general.less
@@ -18,16 +18,8 @@ h1 {
     margin: 0;
   }
 
-  &.section-heading {
-    margin: 1.5em 0 .5em;
-  }
-
   & small {
     font-weight: 300;
-  }
-
-  a.view-source {
-    font-size: 1.2rem;
   }
 }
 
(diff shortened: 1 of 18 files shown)


josevalim:
Can you please break the PR into smaller ones? Probably one per bullet list. Some notes to have in mind:

* use “less” variables as much as possible
* avoid inline styles for HTML
* for visual changes, please include screenshots for before and after

Thanks!

eksperimental:
All of the changes have already been implemented by the various PRs

PR #1256, opened 2020-08-23
Fix regex for autolink typespecs

eksperimental:
It wouldn't detect typespecs ending in ?
Plus a more specific regex is included for detecting the typespec name.


josevalim:

PR #1261, opened 2020-08-25
Reformat Application (version number)

eksperimental:
Previously the version number would be before Application name.
This affects the readability in screen readers

Additionally the version number in the sidebar is no longer a heading.

### DIFF
--- a/assets/less/sidebar.less
+++ b/assets/less/sidebar.less
@@ -66,6 +66,7 @@
   }
 
   .sidebar-projectVersion {
+    display: block;
     position: relative;
     margin: 0;
     padding: 0;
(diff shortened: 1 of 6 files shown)


josevalim:
@eksperimental does it change how the docs page look like in any way? Or is the visibility the same?

eksperimental:
> @eksperimental does it change how the docs page look like in any way? Or is the visibility the same?

It doesn't (well, it shouldn't) but I found a bug, because I am rebasing from the big PR..

I will submit a fix and screenshots for completeness

eksperimental:
This is how the Heading structure and the h1 title looks like now.

Before
app_vsn_before

After
app_vsn_after

eksperimental:
> @eksperimental does it change how the docs page look like in any way? Or is the visibility the same?

They look exactly the same. So no need for screenshots

josevalim:

PR #1262, opened 2020-08-25
Sort file entries in .build file

eksperimental:


eksperimental:
Mimics files listed by `ls` command.

josevalim:
It is a private file, so what benefit does sorting offer? If we have to sort, basic sorting should be enough.

eksperimental:
Make it more natural for humans looking into it if you are looking for a file.

lib/ex_doc/formatter/html.ex
```diff
@@ -48,7 +48,7 @@ defmodule ExDoc.Formatter.HTML do
         generate_list(nodes_map.modules, nodes_map, config) ++
         generate_list(nodes_map.tasks, nodes_map, config) ++ generate_index(config)
 
-    generate_build(all_files, build)
+    generate_build(Enum.sort_by(all_files, &{Path.dirname(&1), &1}), build)
```
josevalim, line 51:
```suggestion
    generate_build(Enum.sort(all_files), build)
```

josevalim:

PR #1263, opened 2020-08-25
Generate robots.txt

eksperimental:
Be kind to search engines.


josevalim:
Honestly, I don't think it is our job to generate said files. If someone needs it, they can always add it directly to their doc/ directory and we will keep it. :) Thanks for the PR though!

PR #1264, opened 2020-08-25
Set option :javascript_config_path default to nil

eksperimental:
Making the option opt-in, avoids getting a Javascript Error in every single page
when requesting the non-existent docs_config.js file, which is the most common scenario.

### DIFF
--- a/lib/ex_doc/config.ex
+++ b/lib/ex_doc/config.ex
@@ -27,7 +27,7 @@ defmodule ExDoc.Config do
             groups_for_modules: [],
             groups_for_functions: [],
             homepage_url: nil,
-            javascript_config_path: "docs_config.js",
+            javascript_config_path: nil,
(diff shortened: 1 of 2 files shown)


josevalim:
@eksperimental it is enabled by default because Hex uses  it (and most projects are being deployed on Hex).

josevalim:
Thanks for the PR tho!

eksperimental:
I thought Hexdocs was building the docs itself. Well, I guess is does it but to all the time.

I think what we can do is generate and empty docs_config.js to avoid that Javascript error and and ERROR 404 in the logs of the server. I see hexdocs rewrite this file anyway so it wouldn't interfere.

On the other hand it is not clear how this feature works. I create a docs_config.js in my project and  it is not included in the build. We should document this, or we should copy it from the root.

PR #1265, opened 2020-08-27
Accessibility: Restructure headings

eksperimental:
It only holds the title inside the heading,
Everything else, such as links are moved our of the heading.
Everything is contained in a `<div class="section-heading"> `tag.

**A) original****
A_orig

**A) fixed****
A_fixed
***************

**B) original**
B_orig

**B) fixed**
B_fixed
***************

**C) original**
C_orig

**C) fixed**
C_fixed

***************

**D) original**
D_orig

**D) fixed**
D_fixed
***************

**E) original**
E_orig

**E) fixed**
E_fixed

### DIFF
--- a/assets/less/content/general.less
+++ b/assets/less/content/general.less
@@ -8,38 +8,28 @@ h6 {
   font-weight: 700;
   line-height: 1.5em;
   word-wrap: break-word;
+  margin: 1em 0 .5em;
 }
 
 h1 {
   font-size: 2em;
-  margin: 1em 0 .5em;
 
   &.signature {
     margin: 0;
   }
 
-  &.section-heading {
-    margin: 1.5em 0 .5em;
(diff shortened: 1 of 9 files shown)


eksperimental:
It is really hard to make it look the same due to collapsing margins in CSS that now since the heading goes inside a `<div>` it is not possible to make it collapse.

josevalim:
@eksperimental which part doesn't look the same? At least in the screenshots they look equivalent? :)

eksperimental:
> @eksperimental which part doesn't look the same? At least in the screenshots they look equivalent? :)

The look very similar but to exactly the same. It is just the gap above and bellow the headings. Save in your computer original and fixed, and alternate the view. You will see the difference.

josevalim:

PR #1266, opened 2020-08-30
Add aria-label and title attributes to links

eksperimental:
Remove use of sr-only class and elements


josevalim:

PR #1267, opened 2020-08-31
Introduce test.js and test.all tasks

eksperimental:
mix test.js: runs `npm run --prefix assets test`
mix test.all: tries to recreate all tests run by the CI (linting, tests, test/prerelease.sh)

This PR modifies test/prerelease.sh so it can be run locally without geneating
additional files in the project folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ The easiest way to test changes to ExDoc is to locally rebuild the app and its o
   3. If you want to contribute a pull request, please do not add to your commits the files generated in the `assets/` and `formatters/` folders
   4. Run `mix lint.` to check if the Elixir and JavaScript files are properly formatted.
      You can run `mix fix` to let the JavaScript linter and Elixir formatter fix the code automatically before submitting your pull request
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @eksperimental, I think it is fine to rather keep the scripts as is. The CI is there to catch any errors and I would rather not impose a bigger curve for contributors. Thank you!

eksperimental:
Is it possible to not to list it in the README, that way things are simple for new contributors as they are now, but we can still make use of it?
I find it useful to test locally and not to find out that something went wrong after uploading and waiting for the CI to find out

josevalim:
I am worried though that listing those steps in mix.exs is going to feel it is a requirement. You can run those individually if necessary? And things like docs are really not necessary for automated testing, only manual testing (I.e. front end).

eksperimental:
We can run them individually, but every time I want to do that, I need to grep my history because I cannot remember the npm commands. In some way they are a requirement because if the PR doesn't pass the CI, it won't be merged. But we could either leavea comment in mix.exs, or create a proper Mix task.

PR #1269, opened 2020-09-02
Fix bug in assets generation

eksperimental:
It would not pass the test in ExDoc.Formatter.HTMLTest named
"generates in default directory with redirect index.html file"

Since a macro was being used, if there were two assets at the compilation time,
after removing the leftover asset, it would still not pass the test because two assets
were in the compiled code.

### DIFF
--- a/lib/ex_doc/formatter/epub/assets.ex
+++ b/lib/ex_doc/formatter/epub/assets.ex
@@ -1,7 +1,7 @@
 defmodule ExDoc.Formatter.EPUB.Assets do
   @moduledoc false
 
-  defmacrop embed_pattern(pattern) do
+  defp embed_pattern(pattern) do
     ["formatters/epub", pattern]
     |> Path.join()
     |> Path.wildcard()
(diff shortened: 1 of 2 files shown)


josevalim:
It Has to be a macro because ex_doc may be installed as an escript which we won’t have access to disk. So we need to embed their source in the compiled artifacts. Can you please explain how to reproduce the failure?

eksperimental:
It happens all the time when I run `mix build`, a new set of assets will be created. Even if I delete these new assets, the problem persist.

josevalim:
Sorry, it is still not clear what is the bug. How to reproduce it?

eksperimental:
```shell
mix clean

#test will pass
mix test

mix build

git restore formatters/epub/dist formatters/html/dist

# test will fail
mix test --force

#a way to solve it is to run `git clean` after `git restore`
git clean -f formatters/epub/dist formatters/html/dist
mix test --force
```

eksperimental:
and if the after running `git clean` you don't run `mix test ` with the `--force`  argument, it will make developers scratch their heads wondering how on earth it is still failing.

eksperimental:
I am thinking we could add a Mix task to deal with it. It is a common scenario since assets shouldn't be included in the PR.

PR #1270, opened 2020-09-03
Increase contrast in message background for deprecated functions in daymode

eksperimental:
**Before**
before
**************

**After**
after

### DIFF
--- a/assets/less/variables/common.less
+++ b/assets/less/variables/common.less
@@ -8,7 +8,7 @@
 @detailGray: #f7f7f7;
 @groupGray: #dddddd;
 @yellow: #ffff99;
-@deprecated: #fffeed;
+@deprecated: #f5f3d2;
 @darkDeprecated: #dcc100;
 
 // Night mode


josevalim:
Thanks @eksperimental. Can we change the color of the font instead maybe? IIRC, the exact tone of yellow was given to me in a way that matches the rest of the theme and we would need someone to double check if we want to change it. Modifying the font color won't be a big issue though.

eksperimental:
> Thanks @eksperimental. Can we change the color of the font instead maybe? IIRC, the exact tone of yellow was given to me in a way that matches the rest of the theme and we would need someone to double check if we want to change it. Modifying the font color won't be a big issue though.

The problem here is not the contrast between then font and the background color that we are changing. but between this background color and the white background color in the back. They are really difficult to tell them apart, at least for me.

josevalim:
I think that's exactly the point though. The message is already in the first paragraph, this makes it visible to everyone reading those particular docs. At the same time, we don't want to use a very intrusive color, because that will call too much attention to it while reading something else or scrolling through. :)

eksperimental:
I have to disagree. The way I read documentation, specially the Elixir docs that are structured always the same way is by skipping to the important parts. For instance, if I want to learn how a function works, I skip bellow the specs because that is where the docs start, by not using enough contrast it may go unnoticed. Compare how it is implemented in the night mode and even if you skip to the text part you are still called by that deprecated message

PR #1271, opened 2020-09-04
Create empty javascript_config_path file if it does not exist

eksperimental:
This avoids getting an error in the browser calling a file that does not exist.
If the file is not under the output dir, it will not be created.


josevalim:
As mentioned in the previous issue, Hex expects to be no file there in order to plug its own. So you can actually use an empty file to tell Hex to not add its own stuff. If the warning is really a nuisance to some, they can have an alias for mix docs that adds this file afterwards, but we shouldn't do it by default. Thanks!

PR #1273, opened 2020-09-13
Fix Sections arrow

### LINKED ISSUE #1229: "Section" in sidebar should show arrow next to it when hover
The new "Section" sections does not show an arrow next to it. I think it should work consistently as the rest of the tree branches that expand. As of now to achieve the same result we need to click in the "Top" branch

Introduced in https://github.com/elixir-lang/ex_doc/pull/1148

/@cc davydog187

davydog187:
Closes #1229 

Adds an arrow next to `Sections` which links out to the same place as `Top`

Screen Shot 2020-09-13 at 8 32 28 AM

### DIFF
--- a/assets/js/templates/sidebar-items.handlebars
+++ b/assets/js/templates/sidebar-items.handlebars
@@ -36,8 +36,9 @@
         {{else}}
           {{#showSections node}}
             <li class="docs {{#isLocal node.id}}open{{/isLocal}}">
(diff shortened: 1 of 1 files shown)


davydog187:
Possibly the "more correct" place would be to link to the first section. That's a little bit more work, but doable if we want it

josevalim:

PR #1276, opened 2020-09-15
Make tests that capture IO asyncronous

eksperimental:
It reverts the changes introduced in dd799164abacc28b8b54a46c2a21e567b001309e
and makes of a feature introduced in Elixir v1.10 that we can read concurrently from a
captured named device

### DIFF
--- a/test/ex_doc/autolink_test.exs
+++ b/test/ex_doc/autolink_test.exs
@@ -1,5 +1,5 @@
 defmodule ExDoc.AutolinkTest do
-  use ExUnit.Case
+  use ExUnit.Case, async: true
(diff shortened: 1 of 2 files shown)


josevalim:
I don’t think it is a test concern, I don’t think we should push our users to upgrade. We can perhaps reduce concurrency on CI though if the builds are failing due to the reason, but I haven’t seen so far. Thanks!

eksperimental:
Sorry, I thought you meant to push users to upgrade.

josevalim:
I didn’t mean to close. We should definitely add the async true back! 🥇

josevalim:

PR #1290, opened 2020-10-27
Fix autolink to handle URIs with arbitrary scheme part without warning

### LINKED ISSUE #1288: `about` links generate warning
When providing a link to a browser built-in such as `about:blank` or `about:preferences#privacy`, ExDoc produces the following warning:

```
Generating docs...
warning: documentation references file "about:blank" but it does not exist
  guides/overview.md
```

I have been looking for a way to silence this warning. I see that there is an option for `skip_undefined_reference_warnings_on`, but thought that this behavior was unexpected and should not issue a warning.

I have a use case set up in the following repo: https://github.com/sax/issue-cases/tree/main/ex_doc_about_links

Let me know if I can provide any more information!

jonatanklosko:
Fixes #1288.

Currently a URI with no host, but with the scheme part present (like `about:blank`) is treated as a file reference and results in a warning (`documentation references file "about:blank" but it does not exist`).

Checking that both host and scheme parts are absent should do the trick.

### DIFF
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -105,6 +105,7 @@ defmodule ExDoc.Autolink do
 
   defp build_extra_link(link, config) do
     with uri <- URI.parse(link),
+         nil <- uri.scheme,
          nil <- uri.host,
          true <- is_binary(uri.path),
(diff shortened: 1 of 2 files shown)


wojtekmach:
Thank you!

PR #1296, opened 2020-12-04
Indent spec after line break

eksperimental:
Before:
image

After:
image

### DIFF
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -340,6 +340,7 @@ defmodule ExDoc.Autolink do
       |> Macro.to_string()
       |> Code.format_string!(line_length: 80)
       |> IO.iodata_to_binary()
+      |> String.replace("\n", "\n    ")
       |> T.h()
 
     name = typespec_name(ast)
--- a/test/ex_doc/autolink_test.exs
+++ b/test/ex_doc/autolink_test.exs
@@ -360,6 +360,16 @@ defmodule ExDoc.AutolinkTest do
       assert typespec(quote(do: term() < term() :: boolean())) ==
                ~s[<a href="https://hexdocs.pm/elixir/typespecs.html#built-in-types">term</a>() &lt; <a href="https://hexdocs.pm/elixir/typespecs.html#built-in-types">term</a>() :: <a href="https://hexdocs.pm/elixir/typespecs.html#built-in-types">boolean</a>()]
     end
+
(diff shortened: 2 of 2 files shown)


lib/ex_doc/formatter/html/templates.ex
```diff
@@ -30,6 +30,13 @@ defmodule ExDoc.Formatter.HTML.Templates do
     nil
   end
 
+  @doc """
+  Indents specs after line breaks.
+  """
+  def indent_spec(string) do
+    String.replace(string, "\n", "\n    ")
```
josevalim, line 37: Great find. :heart: I think we should potentially indent when we convert the spec to string, and not inside the template. Can you please investigate? We would also need a test. Thank you.

eksperimental: done. There is one issue though is that the lines after the first one may exceed 80 chars, one solution is to set the width to 76 instead of 80.

josevalim:

josevalim:
I will revert this as it was actually correct. Elixir doesn't actually indent when in the formatter, and we are following the same behaviour here.

eksperimental:
I think the situations are different. 

Consider:
```elixir
  @spec ascii_printable?([...], limit) :: boolean()
        when limit: :infinity | pos_integer()
  @spec ascii_printable?([...], limit) :: boolean()
        when limit: :something_else | neg_integer()
  def ascii_printable?(_, _) do
    # ...
  end
```

in the above code you can easily identify each spec visually because:
1. They are prefixed by the `@spec` declaration. 
2. Additionally, while the new line is not indented, the special word "when" is colored

in IEx there the situation is similar:

```
iex(1)> h List.ascii_printable?

                 def ascii_printable?(list, limit \\ :infinity)                 

  @spec ascii_printable?(list(), 0) :: true
  @spec ascii_printable?([], limit) :: true
        when limit: :infinity | pos_integer()
  @spec ascii_printable?([...], limit) :: boolean()
        when limit: :infinity | pos_integer()

```

but in ExDoc the situation is different because:
1. we strip the `@spec` tag
2. we use pre-formatted text

So every line looks like a spec declaration, making it `when`  a spec declaration

I think we either increase the indentation in one level in lines after the first one, or we need to prefix every spec line with a "- " to make it clear what is a spec and what is a line continuing from the line before.

eksperimental:
option number 3: would be to not to use preformatted text, and use bullet lists.

PR #1299, opened 2020-12-05
Drop jQuery and refactor JavaScript codebase

### LINKED ISSUE #1153: Optimize the front end build of ex_doc
# Problem
I noticed that `ex_doc` was using jquery and handlebars, and wondered how large the frontend build of the project was. I was surprised that it was quite big.

```
~/D/e/doc/dist [moduledoc-sidebar-sections] λ du -h *.js
212K	app-6efcc579bb079899793d.js
40K	search_items-39b2a5cbd3.js
8.0K	sidebar_items-342be7f185.js
```

I quickly added [Webpack Bundle Analyzer]( https://github.com/webpack-contrib/webpack-bundle-analyzer) to the project to get more information.

Screen Shot 2020-04-26 at 6 43 49 AM

As you can see, jQuery is taking up a considerable amount of space (84kb parsed), followed by `lunr` and `highlight.js`

# Proposed Solution
So why do we need jQuery in the first place? The reason is that `ex_doc`'s approach to front end is a bit old school. Append and replace containers, basically. First off, [You might not need jQuery](http://youmightnotneedjquery.com/), but I think its more about the approach than whether or not the browser supports some DOM API.

In order to optimize the front end bundle size and get it down by less than a third of what it is now, I suggest:

1. Replace jQuery, Handlebars and lodash.find with [Preact](https://preactjs.com/), which is a lightweight version of React that is well supported and popular in the JS community. At The Outline, we used it for 3 years to power our CMS and parts of our front end
2. The search functionality of `ex_doc` is likely not used on every visit. Regardless, `lunr` should be lazy loaded into the bundle. Luckily, webpack makes this easy to do through [Code Splitting](https://webpack.js.org/guides/lazy-loading/) and dynamic imports
3. SQL, Javascript and other languages are great to highlight, but also can be lazy loaded. 

## Pros
* Smaller bundle size means a faster webpage
* Moving to Preact will make the frontend implementation simpler and hopefully make it easier to contribute to.
* Any design changes to `ex_doc` would be easier to make with a simpler templating system :wink: 

## Cons
* Refactoring is expensive and takes time
* Preact is one choice of many in the frontend world. My experience has shown that it's a good one though!

I would love to move forward with this, but I would like to get a blessing that this is desirable before getting started.

jonatanklosko:
Fixes #1153.

## Motivation

The front-end currently uses jQuery, which was a reasonable choice a long time ago, but since then browsers' API expanded and there's pretty much no reason to use it anymore. After the changes the main bundle dropped from `227 kB` to `131 kB` (`~42%` reduction).

## Changes

I started by just converting all jQuery usages into browser API alternatives and removing the dependency - all that is included in the first commit.

Then, I refactored pretty much the whole JS codebase to improve the readability and structure. I believe I manged to significantly decrease the complexity of several features. I came across numerous bugs (some of them subtle though) and fixed them along the way.

### Bugs fixed

* user having system-wide dark theme would see night mode initially, but if they switched to day mode and refreshed the page they would again be presented with night mode (because we did not store any information when switching to day mode)
* user having the sidebar hidden presses `/` or `s` to search, the sidebar opens, but the search input is not focused
* `Ctrl + Mouse Click` on a sidebar link does not open the page in a new tab
* sometimes elements cause the content to overflow on mobile (e.g. large tables or preformatted text, [like this](https://static.jonatanklosko.com/ex-doc/issue1153/before/license.html)), which causes the sidebar to be always open by default (because the overflowing body counts towards window width)
* pasting into the search input using mouse doesn't trigger autocomplete (because it reacts to `keyup` events rather than `input`)
* autocomplete is unnecessarily triggered twice per every change (reacts both to `keydown` and `keyup` events)
* sidebar gradient goes from alpha `0.01` to `0` instead of `1` to `0`, because of a wrong usage of the `fade` function (compare [this](https://static.jonatanklosko.com/ex-doc/issue1153/before/changelog.html#v0-10-0) to [this](https://static.jonatanklosko.com/ex-doc/issue1153/after/changelog.html#v0-10-0))
* the element selected within sidebar (e.g. a function name) is not always in sync with the actual URL hash (e.g. if the hash changes by clicking on another function under *Summary*)
* when the URL has arity like `/Enum.html#map/2` then switching version redirects to the root page, because we check whether `/:version/2` exists rather than `/:version/Enum.html#map/2` (because of the arity slash)
* within the quick switch modal (`g`) using `Arrow Up` selects the first element item instead of the last one
* hints don't show for modules with empty description (e.g. `ArgumentError` [here](https://hexdocs.pm/elixir/api-reference.html#content))
* multiple modal windows can be open (e.g. by pressing `?` then `g`), which looks weird

### Additional changes

* tooltip requests are now cancelled if the user stops hovering the link before the data is loaded
* always focus the first search autocomplete suggestion by default (i.e. *Search the documentation*), this way it reflects what happens when the user presses enter, and also requires one keystroke less when moving to a specific result
* the search autocomplete is now improved to match across multiple terms, so that "flat map" and "enum map" match just as well as `flat_map` and `enum.map` :shipit:

*Note: while introducing the changes I was very careful not to introduce any bugs, so I was continuously running the build and checking if the given change works as expected. I also extensively used GitHub blame for figuring out the origins of various pieces of code if their purpose was not clear.*

## Preview

To make it easier to verify the changes I made the documentation build [before](https://static.jonatanklosko.com/ex-doc/issue1153/before/readme.html) and [after](https://static.jonatanklosko.com/ex-doc/issue1153/after/readme.html) the changes available online.

## tl;dr

The main bundle size drops by `~42%`, 12 bugs are fixed, 3 additional improvements are in place and the codebase is hopefully easier to read.

Sorry for the huge diff, but that kinda comes from the nature of the task. It's probably gonna be easier to look at the actual files rather than the diff, as there's barely any code that stayed unchanged.


josevalim:
Oh wow, this is massive, thank you ! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

I will just merge this as is and we will fix any bugs as they appear along the way.

I was also thinking about removing highlight.js and push devs to use Elixir's lexer. Would you like to send a PR for that too?

jonatanklosko:
> I was also thinking about removing highlight.js and push devs to use Elixir's lexer. Would you like to send a PR for that too?

Do you mean just removing `highlight.js` and temporarily dropping highlighting for other languages, then to gradually add `makeup` lexers for those languages? Or to first add those lexers and then do the switch?

josevalim:
I would just drop it. I think it will be a bigger incentive for people to write makeup levers.

jonatanklosko:
Alright, makes sense! Will surely send a PR :>

PR #1300, opened 2020-12-06
Remove highlight.js in favour of migration to Makeup

jonatanklosko:
This drops `highlight.js` to eventually use Elixir's lexers with `makeup` to add all the highlighting markers at build time. This means that temporarily some code blocks (e.g. `html`, `json`, `javascript`) won't be highlighted, but this way people have a greater incentive to add `makeup` for other languages.


josevalim:

PR #1301, opened 2020-12-07
Change autolink to return path rather than just hash for the current module

jonatanklosko:
When on a module page, a reference to the same module is autolinked as `#content`. Such URL depends on the context and in some cases may not work as desired. For instance, have a look at the `Kernel` section [here](https://hexdocs.pm/elixir/api-reference.html), as you can see the autolinked `Kernel` resolves to `api-reference.html#content`, rather than `Kernel.html#content`, because the URL doesn't actually contain any path.

### DIFF
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -427,10 +427,6 @@ defmodule ExDoc.Autolink do
 
   ## Internals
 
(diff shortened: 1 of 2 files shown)


josevalim:
Another great PR, thank you! <3

josevalim:

Issue #1295, opened 2020-11-29
Add Next Previous Buttons to theme


lszyba1:
Hello,
Could you add next /previous buttons to a theme in a similar fashion as sphinx (python).

I'm in a middle of reading 
https://hexdocs.pm/phoenix/request_lifecycle.html#another-new-page
But I need to step out of learning, figure out where I'm in a tree of documents, then click on next section. Very distracting when you are following along.

Example of how the "Next Button Should Look":
scroll to bottom of the page to see "next" button:
https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html

Thank you
Lucas

eksperimental:
regardless of introducing the buttons to the interface, the `rel="prev"` and `rel="next"` can be implemented since there is a linear login based on the sidebar tree.

PR #1302, opened 2020-12-07
Add next/previous at the end of extra pages

jonatanklosko:
An initial idea for #1295.

Currently this adds next/previous links at the end of every page from *extras*. I'm not sure if this makes much sense for modules/tasks, as those are rather not read linearly.

You can see the build preview [here](https://static.jonatanklosko.com/ex-doc/issue1295/initial/readme.html). Screenshots for the reference:

**Desktop**

image

**Mobile**

image

### DIFF
--- a/test/ex_doc/formatter/html/templates_test.exs
+++ b/test/ex_doc/formatter/html/templates_test.exs
@@ -463,4 +463,33 @@ defmodule ExDoc.Formatter.HTML.TemplatesTest do
                ~r{<h1>\s*mix task_with_docs\s*<small class=\"app-vsn\">\(Elixir v1.0.1\)</small>}
     end
   end
+
+  describe "bottom actions" do
+    test "does not render next/prev links when the refs are nil" do
+      refs = %{
+        prev: nil,
+        next: nil
+      }
+
+      content = Templates.bottom_actions_template(refs)
+
+      refute content =~ ~r{Previous Page}
+      refute content =~ ~r{Next Page}
+    end
+  end
+
+  test "renders next/prev links when their refs are given" do
+    refs = %{
+      prev: %{path: "Prev.html", title: "Left"},
+      next: %{path: "Next.html", title: "Right"}
+    }
+
+    content = Templates.bottom_actions_template(refs)
+
+    assert content =~
+             ~r{<a href="Prev.html" class="bottom-actions-button" rel="prev">\s*<span class="subheader">\s*← Previous Page\s*</span>\s*<span class="title">\s*Left\s*</span>\s*</a>}
+
+    assert content =~
+             ~r{<a href="Next.html" class="bottom-actions-button" rel="next">\s*<span class="subheader">\s*Next Page →\s*</span>\s*<span class="title">\s*Right\s*</span>\s*</a>}
+  end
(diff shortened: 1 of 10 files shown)


josevalim:
This looks amazing! Thank you.

Just 3 comments:
1. I am afraid the tittle font for the links is too big. For example how would it look like in a project like [Ecto](https://hexdocs.pm/ecto/getting-started.html)?

2. I am wondering if the "continue reading" can be confusing when guides are not related to each other like in exdoc. Maybe only using "next" is enough?

3. We also need tests.

jonatanklosko:
> 1. I am afraid the tittle font for the links is too big. For example how would it look like in a project like Ecto?

Good point, I somehow assumed the titles are kept short, but Ecto is a perfect counterexample here.

> 2. I am wondering if the "continue reading" can be confusing when guides are not related to each other like in exdoc. Maybe only using "next" is enough?

Wasn't sure on that either, this is most suitable for guides, but using a more generic label sounds more reasonable.

If we kept the normal font size that's how it would look (also alternative captions):

image

What do you think?

---

> 3. We also need tests.

Sure, just wanted to see if that's the right direction first. Is `html/templates_test.exs` the right place for those in this case?

jonatanklosko:
I've updated [the preview](https://static.jonatanklosko.com/ex-doc/issue1295/initial/readme.html) with those changes and added tests, let me know if you have any other thoughts!

josevalim:
Should the “next page” be text aligned to the right?

jonatanklosko:
For short names it looks a bit weird, but I don't have a strong opinion on that.

**Aligned right**

Long
image

Short
image

**Aligned left**

Long
image

Short
image

josevalim:
Or the “previous page” can be aligned to the right so both “next page” and “previous page” are side by side in the center?

jonatanklosko:
And here's the "previous page" aligned right. I like this because the alignment is clearly visible, because the text sticks to the edge rather than the arrow.

Long
image

Short
image

josevalim:
Awesome, let’s go with left aligned right then! Once we have tests, we can probably ship it! :D

jonatanklosko:
I've updated the alignment. Do the tests look ok?

test/ex_doc/formatter/html/templates_test.exs
```diff
@@ -463,4 +463,33 @@ defmodule ExDoc.Formatter.HTML.TemplatesTest do
                ~r{<h1>\s*mix task_with_docs\s*<small class=\"app-vsn\">\(Elixir v1.0.1\)</small>}
     end
   end
+
+  describe "bottom actions" do
```
josevalim, line 467: I was hoping more of an integration test, perhaps in html_test, as this one does not really assert we have built the refs properly (which is the complex part!). Do we have two guides in our test suite? It should be enough that generating docs for our existing guides and match on the output.

jonatanklosko: Sure, should've done that before!

josevalim:
Absolutely fantastic, thank you! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #1311, opened 2020-12-26
Fix search click behavior when a suggestion refers the current page

### LINKED ISSUE #1310: Searching and clicking on a function in the same page does not close the suggestion box
Let' s say you are in `doc/elixir/Kernel.SpecialForms.html` and in the search box you type in `import` and you click on "Kernel.SpecialForms.import/2", the suggestion box does not goes away and if you try to hit on `x` icon, it does not work.

Screenshot shows after clicking the import result, the suggestion box is still there, blocking the view of the menu items.
image

jonatanklosko:
Fixes #1310.

I made it so that opening a suggestion in a new tab/window (using `Shift` or `Ctrl`) would keep the autocomplete list open, but otherwise it always gets closed. I also changed the `Enter` behavior to mimic clicking a link rather than submitting a form, because otherwise we always navigate to a new page, even if the URL leads to the same page.

So this should unify clicking on a suggestion and pressing `Enter` to work the same way.

You can see a preview [here](https://static.jonatanklosko.com/ex-doc/issue1310/initial/ExDoc.Markdown.html).

### DIFF
--- a/assets/js/autocomplete/autocomplete-list.js
+++ b/assets/js/autocomplete/autocomplete-list.js
@@ -2,8 +2,8 @@ import autocompleteSuggestionsTemplate from '../handlebars/templates/autocomplet
 import { getSuggestions } from './suggestions'
 import { isBlank, qs } from '../helpers'
 
-const AUTOCOMPLETE_CONTAINER_SELECTOR = '.autocomplete'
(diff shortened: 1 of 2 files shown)


josevalim:

josevalim:

eksperimental:
Thank you so much @jonatanklosko for the fix and for looking into this. It like that it is now unified. There is a little glitch though, I reopened the original issue

PR #1312, opened 2020-12-27
Improve search input blur handling

jonatanklosko:
Follow up for #1311, see [this comment](https://github.com/elixir-lang/ex_doc/issues/1310#issuecomment-751409542).

We rely on the blur event for closing the autocomplete box, but sometimes we have to ignore this event in which case we may end up in a not-easily-closable state. I added a `setTimeout` ensuring that the search gets re-focused when the autocomplete box is still open. While this doesn't seem extremely elegant, I tried several other approaches and none was perfect.

@eksperimental please have a look [here](https://static.jonatanklosko.com/ex-doc/issue1310/followup/ExDoc.Markdown.html) and let me know if it works as expected.

### DIFF
--- a/assets/js/autocomplete/autocomplete-list.js
+++ b/assets/js/autocomplete/autocomplete-list.js
@@ -25,6 +25,13 @@ export function hideAutocompleteList () {
 }
 
 /**
+ * Checks if the search autocomplete list is open.
+ */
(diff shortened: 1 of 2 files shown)


eksperimental:
It works as expected. Thank you very much @jonatanklosko.

Issue #1308, opened 2020-12-25
Linking to function starting with underscore gets links but docs do not exist


eksperimental:
```elixir
defmodule Foo do
  @moduledoc """
  `_foo/1`
  """

  def _foo(foo) do
    foo
  end
end
```

The module docs will link to _foo, but _foo will not show in the generated docs.
A real case happens when you locally link to `__using__/1`

josevalim:
Don’t we warn if you link to something without docs?

eksperimental:
Here' s a more complete example

```elixir
defmodule Foo do
  @moduledoc """
  `_foo/1`, `Foo._foo/1`, `boo/1`, `Foo.boo/1`
  """

  def _foo(foo) do
    foo
  end
end
```
```console
$ mix docs
Generating docs...
warning: documentation references "Foo.boo/1" but it is undefined or private
  lib/foo.ex:2: Foo

View "html" docs at "doc/index.html"
warning: documentation references "Foo.boo/1" but it is undefined or private
  lib/foo.ex:2: Foo

View "epub" docs at "doc/buildable.epub"
```


the docs: only the first two get linked, and the docs do not exist
```
 Foo

_foo/1, Foo._foo/1, boo/1, Foo.boo/1
```

jonatanklosko:
As far as I understand, the problem comes from the fact that underscored functions get `:public` visibility in the `ExDoc.Refs` cache, but they are ignored when generating docs.

---

Autolinking relies on the `visibility` stored in the `ExDoc.Refs` cache:

https://github.com/elixir-lang/ex_doc/blob/15560a59f39d3d6917ee5070adc121a00f1814db/lib/ex_doc/autolink.ex#L476-L480

This visibility is computed here:

https://github.com/elixir-lang/ex_doc/blob/15560a59f39d3d6917ee5070adc121a00f1814db/lib/ex_doc/refs.ex#L107-L110

https://github.com/elixir-lang/ex_doc/blob/15560a59f39d3d6917ee5070adc121a00f1814db/lib/ex_doc/refs.ex#L135-L136

So unless the doc is explicitly hidden we assign `:public` visibility.

---

On the other hand, when generating docs there are more specific criteria a function needs to satisfy:

https://github.com/elixir-lang/ex_doc/blob/15560a59f39d3d6917ee5070adc121a00f1814db/lib/ex_doc/retriever.ex#L262-L267

[...]

PR #1313, opened 2020-12-27
Reimplement ExDoc.Refs.visibility helper

eksperimental:
It corrects the visibility of functions that start with an underscore and that no docs have been defined.

Closes https://github.com/elixir-lang/ex_doc/issues/1308

/cc @jonatanklosko

### DIFF
--- a/lib/ex_doc/refs.ex
+++ b/lib/ex_doc/refs.ex
@@ -100,11 +100,11 @@ defmodule ExDoc.Refs do
   defp fetch_entries(module, result) do
     case result do
       {:docs_v1, _, _, _, module_doc, _, docs} ->
-        module_visibility = visibility({module, module_doc})
+        module_visibility = visibility(module_doc)
 
         for {{kind, name, arity}, _, _, doc, metadata} <- docs do
           ref_kind = to_ref_kind(kind)
-          visibility = visibility({module, module_doc}, {ref_kind, name, doc})
+          visibility = visibility(module_doc, {ref_kind, name, doc})
 
           for arity <- (arity - (metadata[:defaults] || 0))..arity do
             {{ref_kind, module, name, arity}, visibility}
@@ -133,23 +133,24 @@ defmodule ExDoc.Refs do
 
   defp starts_with_underscore?(name), do: hd(Atom.to_charlist(name)) == ?_
 
-  defp visibility({_module, :hidden}),
+  defp visibility(:hidden),
     do: :hidden
 
-  defp visibility({_module, _module_doc}),
+  defp visibility(_module_doc),
     do: :public
 
-  defp visibility({_module, _module_doc}, {kind, _name, _doc})
+  defp visibility(_module_doc, {kind, _name, _doc})
        when kind not in [:callback, :function, :type],
        do: raise(ArgumentError, "Unknown kind #{inspect(kind)}")
 
-  defp visibility({_module, :hidden}, {_kind, _name, _doc}),
+  defp visibility(:hidden, {_kind, _name, _doc}),
     do: :hidden
 
-  defp visibility({_, _}, {_kind, _name, :hidden}),
-    do: :hidden
+  defp visibility(_, {_kind, _name, :hidden}) do
+    :hidden
+  end
 
-  defp visibility({_, _}, {kind, name, doc}) when has_no_docs(doc) do
+  defp visibility(_, {kind, name, doc}) when has_no_docs(doc) do
     cond do
       kind in [:callback, :type] ->
         :public
@@ -162,7 +163,7 @@ defmodule ExDoc.Refs do
     end
   end
 
-  defp visibility({_, _}, {_, _, _}) do
+  defp visibility(_, {_, _, _}) do
     :public
   end
 
--- a/lib/ex_doc/refs.ex
+++ b/lib/ex_doc/refs.ex
@@ -87,8 +87,8 @@ defmodule ExDoc.Refs do
   end
 
   defp fetch({_kind, module, _name, _arity} = ref) do
-    with mod_visibility <- fetch({:module, module}),
-         true <- mod_visibility in [:public, :hidden],
+    with module_visibility <- fetch({:module, module}),
+         true <- module_visibility in [:public, :hidden],
          {:ok, visibility} <- lookup(ref) do
       visibility
     else
@@ -100,17 +100,14 @@ defmodule ExDoc.Refs do
   defp fetch_entries(module, result) do
     case result do
       {:docs_v1, _, _, _, module_doc, _, docs} ->
-        module_visibility = visibility(module_doc)
+        module_visibility = visibility({module, module_doc})
 
         for {{kind, name, arity}, _, _, doc, metadata} <- docs do
-          visibility =
-            case {module_visibility, visibility(doc)} do
-              {:hidden, :public} -> :hidden
-              {_, visibility} -> visibility
-            end
+          ref_kind = to_ref_kind(kind)
+          visibility = visibility({module, module_doc}, {ref_kind, name, doc})
 
           for arity <- (arity - (metadata[:defaults] || 0))..arity do
-            {{kind(kind), module, name, arity}, visibility}
+            {{ref_kind, module, name, arity}, visibility}
           end
         end
         |> List.flatten()
@@ -132,12 +129,46 @@ defmodule ExDoc.Refs do
     end
   end
 
-  defp visibility(:hidden), do: :hidden
-  defp visibility(_), do: :public
+  defguardp has_no_docs(doc) when doc == :none or doc == %{}
 
-  defp kind(:macro), do: :function
-  defp kind(:macrocallback), do: :callback
-  defp kind(other), do: other
+  defp starts_with_underscore?(name), do: hd(Atom.to_charlist(name)) == ?_
+
+  defp visibility({_module, :hidden}),
+    do: :hidden
+
+  defp visibility({_module, _module_doc}),
+    do: :public
+
+  defp visibility({_module, _module_doc}, {kind, _name, _doc})
+       when kind not in [:callback, :function, :type],
+       do: raise(ArgumentError, "Unknown kind #{inspect(kind)}")
+
+  defp visibility({_module, :hidden}, {_kind, _name, _doc}),
+    do: :hidden
+
+  defp visibility({_, _}, {_kind, _name, :hidden}),
+    do: :hidden
+
+  defp visibility({_, _}, {kind, name, doc}) when has_no_docs(doc) do
+    cond do
+      kind in [:callback, :type] ->
+        :public
+
+      kind == :function and starts_with_underscore?(name) ->
+        :hidden
+
+      kind == :function ->
+        :public
+    end
+  end
+
+  defp visibility({_, _}, {_, _, _}) do
+    :public
+  end
+
+  defp to_ref_kind(:macro), do: :function
+  defp to_ref_kind(:macrocallback), do: :callback
+  defp to_ref_kind(other), do: other
 
   defp exports(module) do
     if function_exported?(module, :__info__, 1) do
--- a/lib/ex_doc/refs.ex
+++ b/lib/ex_doc/refs.ex
@@ -146,9 +146,8 @@ defmodule ExDoc.Refs do
   defp visibility(:hidden, {_kind, _name, _doc}),
     do: :hidden
 
-  defp visibility(_, {_kind, _name, :hidden}) do
-    :hidden
-  end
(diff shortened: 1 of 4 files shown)


test/ex_doc/refs_test.exs
```diff
@@ -26,6 +41,33 @@ defmodule ExDoc.RefsTest do
...
+    assert Refs.get_visibility({:function, WithoutModuleDoc, :_foo, 1}) == :hidden
+    assert Refs.get_visibility({:function, WithoutModuleDoc, :foo, 2}) == :hidden
+    assert Refs.get_visibility({:function, WithoutModuleDoc, :_foo, 2}) == :hidden
+    assert Refs.get_visibility({:function, WithoutModuleDoc, :foo, 3}) == :public
+    assert Refs.get_visibility({:function, WithoutModuleDoc, :_foo, 3}) == :public
+    assert Refs.get_visibility({:function, WithoutModuleDoc, :foo, 4}) == :undefined
+    assert Refs.get_visibility({:function, WithoutModuleDoc, :_foo, 4}) == :undefined
+
+    assert Refs.get_visibility({:function, InMemory, :foo, 1}) == :public
+    # assert Refs.get_visibility({:function, InMemory, :_foo, 1}) == :hidden
```
eksperimental, line 63: Is there anything we can do do read the @doc attributes of the docs in this module?

test/support/with_without_module_doc.ex
```diff
@@ -0,0 +1,66 @@
+defmodule WithModuleDoc do
```
eksperimental, line 1: I tried different approaches and this seemed to be the easiest way to have the module avaible in the refs and the autlolink tests

lib/ex_doc/refs.ex
```diff
@@ -132,8 +135,28 @@ defmodule ExDoc.Refs do
...
-  defp visibility(_), do: :public
+  defguardp is_kind(kind) when kind in [:module, :callback, :type, :function]
+  defguardp no_docs?(doc) when doc == :none or (is_map(doc) and map_size(doc) == 0)
+
+  defp visibility(:module, _name, doc) when no_docs?(doc), do: :public
+  defp visibility(:callback, _name, doc) when no_docs?(doc), do: :public
+  defp visibility(:type, _name, doc) when no_docs?(doc), do: :public
+
+  defp visibility(:function, name, doc) when no_docs?(doc) do
+    if hd(Atom.to_charlist(name)) == ?_ do
```
eksperimental, line 146: This could be abstracted into a function, as it is also used in the Retriever module. But the question is where to place it. For now we just have duplicated code.

lib/ex_doc/refs.ex
```diff
@@ -132,8 +135,26 @@ defmodule ExDoc.Refs do
     end
   end
 
-  defp visibility(:hidden), do: :hidden
-  defp visibility(_), do: :public
+  defguardp is_kind(kind) when kind in [:callback, :function, :module, :type]
+  defguardp no_docs?(doc) when doc == :none or (is_map(doc) and map_size(doc) == 0)
```
jonatanklosko, line 139: I believe the second condition could be simplified to `doc == %{}`.

test/ex_doc/autolink_test.exs
```diff
@@ -449,6 +449,32 @@ defmodule ExDoc.AutolinkTest do
     assert_unchanged(~m"`Unknown`")
 
     assert_unchanged(~m"[Blank](about:blank)")
+
+    # warnings with functions starting with an underscore
+    # TODO in a following commit: Split this test into tests inside a describe block
+    assert warn("WithModuleDoc._foo/1",
+             file: "lib/with_module_doc.ex",
+             id: "WithModuleDoc._foo/1"
```
jonatanklosko, line 457: `file` and `id` don't seem relevant here any reason for including these?

test/support/with_without_module_doc.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  @doc false
+  def _foo(x, y), do: {x, y}
+
+  @doc "docs..."
+  def foo(x, y, z), do: {x, y, z}
+
+  @doc "docs..."
+  def _foo(x, y, z), do: {x, y, z}
+end
```
jonatanklosko, line 44: Relying on arity makes it necessary to look into this module to get the idea. Perhaps we could encode the information in function names like `fun_no_doc`, `fun_doc_false`, `fun_with_doc`?

test/ex_doc/refs_test.exs
jonatanklosko: The `Refs` module uses `Code.fetch_docs`, which requires the bytecode to be written to disk ([reference](https://hexdocs.pm/elixir/writing-documentation.html#code-fetch_docs-1)) and that's not the case for in-memory modules. I don't know if there's any way around that.

lib/ex_doc/refs.ex
```diff
@@ -132,8 +135,26 @@ defmodule ExDoc.Refs do
...
+    if hd(Atom.to_charlist(name)) == ?_ do
+      :hidden
+    else
+      :public
+    end
+  end
+
+  defp visibility(kind, _name, _doc) when is_kind(kind) do
+    :public
+  end
```
jonatanklosko, line 157: Is there any case where we would want to behave differently than `ExDoc.Retriever.doc?`? If that's not the case, then perhaps we could extract that boolean `doc?` function and then just convert it to `:public/:hidden` here?

jonatanklosko (commented):
The solution looks good to me, a few tiny comments =)

lib/ex_doc/refs.ex
eksperimental: Implemented. Thank you

test/ex_doc/autolink_test.exs
eksperimental: Removed.

test/support/with_without_module_doc.ex
eksperimental: Done. this is much better now.

lib/ex_doc/refs.ex
eksperimental: This is the only suggestion that I have not implemented. Do you mind sharing the implementation of what you mean?

jonatanklosko: I looked into that and the intent of these functions is in fact different - `ExDoc.Retriever.doc?` is used only to filter which functions/macros should be shown in the doc.

I do however think it would make sense to have a single source of truth regarding which elements get documentation and which are ignored. Currently there are two separate sources:

- the `Retriever` which ignores a given module/function/macro/... based on some rules
- the `Refs` cache that keeps track of module/function/macro/... "visibility"

The question is whether these two should be in line with each other in all cases. The issue hints me that they should, but I may be missing something here.

What I had in mind is a single function that would determine if the given module or function should end up in the generated documentation or rather be ignored. This function could then be used in all the relevant places. [Here](https://github.com/eksperimental/ex_doc/compare/functions_starting_with_underscore...jonatanklosko:pr1313) I implemented such function (`should_have_doc?`) and tried to make it match the current behavior I saw in several different places. I think it removes much redundancy, but someone should confirm that this approach makes sense.

eksperimental: @jonatanklosko  I think covering those issues in this PR it will deviate the purpose too much.
I have taken the ideas of unifying regarding to visibility and put it all in one function.
So now what this PR does is laying the foundation for the logic regarding the visibility of functions.
I think once we merge this we can look into your proposal. What do you think?

Thank you for your contributions

jonatanklosko: Sure, what I suggested was more of a refactoring and can definitely be discussed separately. Your changes surely fix the problem.

test/ex_doc/autolink_test.exs
```diff
@@ -449,6 +449,20 @@ defmodule ExDoc.AutolinkTest do
     assert_unchanged(~m"`Unknown`")
 
     assert_unchanged(~m"[Blank](about:blank)")
+
+    # warnings with functions starting with an underscore
+    # TODO in a following commit: Split this test into tests inside a describe block
```
jonatanklosko, line 454: Other than that, everything looks good to me ^^

eksperimental: That is intentional. I did' t want the reviewing processes be swamped with unrelated changes. I would leave it in this PR, and in a next commit split it into tests.

jonatanklosko: Alright, sounds good!

eksperimental:
Guys, I am planning on going ahead and merge this if nobody has any comments.

lib/ex_doc/refs.ex
```diff
@@ -100,17 +100,14 @@ defmodule ExDoc.Refs do
   defp fetch_entries(module, result) do
     case result do
       {:docs_v1, _, _, _, module_doc, _, docs} ->
-        module_visibility = visibility(module_doc)
+        module_visibility = visibility({module, module_doc})
```
josevalim, line 103: I don't see why we pass the module, we never use it. :)

lib/ex_doc/refs.ex
```diff
@@ -100,17 +100,14 @@ defmodule ExDoc.Refs do
...
 
         for {{kind, name, arity}, _, _, doc, metadata} <- docs do
-          visibility =
-            case {module_visibility, visibility(doc)} do
-              {:hidden, :public} -> :hidden
-              {_, visibility} -> visibility
-            end
+          ref_kind = to_ref_kind(kind)
+          visibility = visibility({module, module_doc}, {ref_kind, name, doc})
```
josevalim, line 107: I think this can be simpler. Something like this:

    if module_visibility == :hidden do
      :hidden
    else
      def_visibility(ref_kind, name, doc)
    end

test/support/with_without_module_doc.ex
josevalim: We don't need to test both autolink and ref test. We already test that autolink warns when something is private, we don't need to add new ones. My suggestion is to move this module to `ref_test` and remove the autolink test.

In a nutshell, autolink should only test the autolink and warning behaviour, it shouldn't really be testing what is public or private, because that's not its responsibility. If we are having a lot of tests around public and private already there, I recommend prunning/simplifying them. :)

lib/ex_doc/refs.ex
eksperimental: I don't think we should have this visibility logic placed in a fetch function. Actually, the if i'm  not mistaken the very same bug that this PR solves is because of this logic being out of the visibility function.

lib/ex_doc/refs.ex
eksperimental: True. fixed

test/support/with_without_module_doc.ex
eksperimental: Got you.
I cannot place the module in the .exs file, because that defeats the purpose of the module which is testing the visibility based on the @doc attributes.
I did my best to simplify the tests in refs_test.exs, please have a look

lib/ex_doc/refs.ex
josevalim: You can put it inside the visibility function if you want. My point is that the logic today is more complicated than it needs to be, because you can work with the module_visibility you already got, instead of computing it over again.

josevalim: But removing the `module` gets a similar result, so it looks good to me! :)

lib/ex_doc/refs.ex
```diff
@@ -132,12 +129,47 @@ defmodule ExDoc.Refs do
...
+  defp visibility(_module_doc, {kind, _name, _doc})
+       when kind not in [:callback, :function, :type],
+       do: raise(ArgumentError, "Unknown kind #{inspect(kind)}")
+
+  defp visibility(:hidden, {_kind, _name, _doc}),
+    do: :hidden
+
+  defp visibility(_, {_kind, _name, :hidden}) do
+    :hidden
+  end
```
josevalim, line 151:
```suggestion
  defp visibility(_, {_kind, _name, :hidden}),
    do: :hidden
```

test/support/with_without_module_doc.ex
josevalim: The tests look great, thanks! Do we still need this @moduledoc now with these function refs? :)

josevalim (approved):

lib/ex_doc/refs.ex
eksperimental: fixed

test/support/with_without_module_doc.ex
eksperimental: Fixed. Thanks

eksperimental:
Thank you @jonatanklosko and @josevalim for your reviews

PR #1321, opened 2021-01-24
Dialyzer

eksperimental:
Use Dialyzer for testing locally ~and in the CI~


josevalim:
Thank you but I am a :-1:. I don't think Dialyzer is worth the ROI. If someone wants to run it locally and provide fixes, that's welcome, but its error messages are hard to follow up to make it part of CI.

eksperimental:
Maybe I can install something like https://github.com/michallepicki/elixir-lang-dialyzer-runs and keep an eye on it.
I will remove the CI part then

eksperimental:
done

mix.exs
```diff
@@ -35,7 +40,8 @@ defmodule ExDoc.Mixfile do
       {:makeup_elixir, "~> 0.14"},
       {:makeup_erlang, "~> 0.1"},
       {:excoveralls, "~> 0.3", only: :test},
-      {:jason, "~> 1.2", only: :test}
+      {:jason, "~> 1.2", only: :test},
+      {:dialyxir, "~> 1.0", only: [:dev, :test], runtime: false}
```
josevalim, line 44:
```suggestion
      {:dialyxir, "~> 1.0", only: [:dialyzer], runtime: false}
```

josevalim (approved):

mix.exs
```diff
@@ -35,7 +39,8 @@ defmodule ExDoc.Mixfile do
       {:makeup_elixir, "~> 0.14"},
       {:makeup_erlang, "~> 0.1"},
       {:excoveralls, "~> 0.3", only: :test},
-      {:jason, "~> 1.2", only: :test}
+      {:jason, "~> 1.2", only: :test},
+      {:dialyxir, "~> 1.0", only: [:dialyzer], runtime: false}
```
josevalim, line 43:
```suggestion
      {:dialyxir, "~> 1.0", only: :dialyzer, runtime: false}
```

josevalim (approved):

eksperimental:
Thank you @josevalim

PR #1323, opened 2021-01-31
fix ast_to_html with empty elements

Zurga:
While creating documentation that contained `<i></i>` inside a `<span>` element, the formatter was not able to cope with this html. See the included test.

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -128,8 +128,7 @@ defmodule ExDoc.Formatter.HTML do
   def ast_to_html(binary) when is_binary(binary), do: Templates.h(binary)
 
   def ast_to_html({tag, attrs, ast, %{verbatim: true}}) do
-    text = (ast != [] && Enum.join(ast, "")) || ""
-    ["<#{tag}", ast_attributes_to_html(attrs), ">", text, "</#{tag}>"]
+    ["<#{tag}", ast_attributes_to_html(attrs), ">", ast, "</#{tag}>"]
   end
 
   def ast_to_html({tag, attrs, ast, %{}}) do
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -128,7 +128,7 @@ defmodule ExDoc.Formatter.HTML do
   def ast_to_html(binary) when is_binary(binary), do: Templates.h(binary)
 
   def ast_to_html({tag, attrs, ast, %{verbatim: true}}) do
-    [text] = ast
+    text = (ast != [] && Enum.join(ast, "")) || ""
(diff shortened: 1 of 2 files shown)


lib/ex_doc/formatter/html.ex
```diff
@@ -128,7 +128,7 @@ defmodule ExDoc.Formatter.HTML do
   def ast_to_html(binary) when is_binary(binary), do: Templates.h(binary)
 
   def ast_to_html({tag, attrs, ast, %{verbatim: true}}) do
-    [text] = ast
+    text = (ast != [] && Enum.join(ast, "")) || ""
```
josevalim, line 131:
```suggestion
    text = Enum.join(ast, "")
```

This will work for one element and no element ASTs. :)

Zurga: Yep!

josevalim (approved):

lib/ex_doc/formatter/html.ex
```diff
@@ -128,7 +128,7 @@ defmodule ExDoc.Formatter.HTML do
   def ast_to_html(binary) when is_binary(binary), do: Templates.h(binary)
 
   def ast_to_html({tag, attrs, ast, %{verbatim: true}}) do
-    [text] = ast
+    text = Enum.join(ast, "")
     ["<#{tag}", ast_attributes_to_html(attrs), ">", text, "</#{tag}>"]
```
wojtekmach, line 132: this function returns an iolist so there's no need to join the elements:

```suggestion
    ["<#{tag}", ast_attributes_to_html(attrs), ">", ast, "</#{tag}>"]
```

test/ex_doc/formatter/html_test.exs
```diff
@@ -527,5 +527,17 @@ defmodule ExDoc.Formatter.HTMLTest do
...
+    test "handles empty elements" do
+      assert """
+             <span>
+             <i></i>
+             </span>
+             """
+             |> ExDoc.Markdown.Earmark.to_ast([])
+             |> HTML.ast_to_html()
+             |> IO.iodata_to_binary() ==
+               "<span><i></i></span>"
```
wojtekmach, line 540: This test passes on master without your change, could you rebase and try again?

I was able to see a failure with such markdown though: `<i></i>`, If you can verify it, could you test that instead ?

Zurga: Ah ok, my test case initially contained two `<i></i>` and it failed. After fixing that I decided to remove one of the elements and it still worked but I did not test that on `master`. I will make the test more explicit.

lib/ex_doc/formatter/html.ex
Zurga: You are right

wojtekmach:
Thank you!

PR #1324, opened 2021-02-02
Introduce ExDoc.DocAST

wojtekmach:


josevalim:

test/ex_doc/doc_ast_test.exs
```diff
@@ -0,0 +1,85 @@
...
+    test "with fun" do
+      markdown = """
+      foo **bar** baz
+      """
+
+      ast = DocAST.parse!(markdown, "text/markdown")
+
+      f = fn
+        {:strong, _, _, _}, string ->
+          String.upcase(string)
```
wojtekmach, line 75: note for posterity: this is basically how we could do auto-linking, we just look for `<code>` nodes, see if they are valid refs, and return `<a href="#{url}">#{string}</a>`!

The only problem is this to_string/2 algorithm is a "postwalk" traversal. It is totally fine for regular code blocks but not for custom links, e.g. we'd unnecessarily first render `` `Bar` `` in `` [`Foo`](`Bar`) ``. So we may not end up using this function in which case I'll remove it later, or maybe there's a prewalk approach to consider. Or we add some special handling for custom links, though I'd rather avoid that. But yeah, I decided to keep this function for now as I believe it's really cool we can do it this way :)

PR #1325, opened 2021-02-02
Rewrite retriever tests to not rely on fixtures

wojtekmach:
I've also added some TODOs, I want to address some as part of the PR and others down the road.


josevalim (approved):
Beautiful!

PR #1326, opened 2021-02-02
Start supporting Erlang

wojtekmach:


josevalim (approved):

test/ex_doc/retriever_test.exs
```diff
@@ -318,6 +320,76 @@ defmodule ExDoc.RetrieverTest do
...
+        name: :function,
+        rendered_doc: nil,
+        signature: "function() -> term()\n",
+        source_path: _,
+        source_url: nil,
+        specs: [],
+        type: :function
+      } = function
+
+      assert DocAST.to_string(function.doc) =~ "<p>function/0 docs.</p>"
```
wojtekmach, line 379: for posterity: we use `=~` instead of `==` because depending on version of edoc you might get:

```html
<p>function/0 docs.</p>
```

or:

```html
<a id="function-0"></a><p>function/0 docs.</p>
```

this all might change when the doc chunk work is finalised so let's keep this relaxed.

PR #1327, opened 2021-02-02
Remove outdated version check

wojtekmach:


eksperimental:
We can also get rid of `ExDoc.Utils.Code`, and rely con `Code.fetch_docs/1`

josevalim (approved):

wojtekmach:
@eksperimental yes but only when we require 1.11 so let’s wait on that a bit.

eksperimental:
> @eksperimental yes but only when we require 1.11 so let’s wait on that a bit.

oops.. you are right

PR #1329, opened 2021-02-04
Refactor fetching callbacks

wojtekmach:


test/test_helper.exs
```diff
@@ -66,25 +67,40 @@ defmodule TestHelper do
...
+    end
+  else
+    def edoc_to_chunk(module) do
+      source_path = module.module_info(:compile)[:source]
+      beam_path = :code.which(module)
+      dir = :filename.dirname(source_path)
+      xml_path = '#{dir}/#{module}.xml'
+      chunk_path = '#{dir}/#{module}.chunk'
+
+      docgen_dir = :code.lib_dir(:erl_docgen)
```
herminiotorres, line 91: maybe use `:code.priv_dir(:erl_docgen)` instead of `:code.lib_dir(:erl_docgen)`?
because on 92 line you pass the `priv` folder path, make sense?

wojtekmach: fixed on master, thx!

lib/ex_doc/retriever.ex
```diff
@@ -382,14 +382,24 @@ defmodule ExDoc.Retriever do
...
+        {:ok, specs} ->
+          specs
+
+        :error ->
+          # gen_statem:state_name/3 is stored as StateName/3 in the docs chunk
+          name = name |> Atom.to_string() |> Macro.underscore() |> String.to_atom()
+          Map.fetch!(module_data.callbacks, {name, arity})
+      end
 
+    {:type, anno, _, _} = hd(specs)
```
herminiotorres, line 397: I have a silly question about the best way to use and references when using Kernel functions... so my question is it's okay to call only the function and avoid doing a full call like `Kernel.hd(specs)`?

wojtekmach: yes, it's totally ok to skip `Kernel.` since it's automatically imported everywhere.

herminiotorres: so the only reason to use a full declaration is when using pipes

lib/ex_doc/retriever.ex
```diff
@@ -382,14 +382,24 @@ defmodule ExDoc.Retriever do
...
-    {:attribute, anno, :callback, {^actual_def, specs}} =
-      Enum.find(module_data.abst_code, &match?({:attribute, _, :callback, {^actual_def, _}}, &1))
+    specs =
+      case Map.fetch(module_data.callbacks, actual_def) do
+        {:ok, specs} ->
+          specs
+
+        :error ->
+          # gen_statem:state_name/3 is stored as StateName/3 in the docs chunk
+          name = name |> Atom.to_string() |> Macro.underscore() |> String.to_atom()
```
josevalim, line 393: Is this a bug on the chunk?

lib/ex_doc/retriever.ex
```diff
@@ -424,15 +434,15 @@ defmodule ExDoc.Retriever do
   # Returns a map of {name, arity} => behaviour.
   defp get_impls(module) do
     for behaviour <- behaviours_implemented_by(module),
-        callback <- callbacks_defined_by(behaviour),
+        callback <- Map.keys(get_callbacks(behaviour)),
```
josevalim, line 437:
```suggestion
        {callback, _} <- get_callbacks(behaviour),
```

?

josevalim (approved):

lib/ex_doc/retriever.ex
wojtekmach: This is complicated.

We have:

```elixir
iex> :gen_statem.behaviour_info(:callbacks)
[
  state_name: 3,
  ...
]
```

And:

```elixir
Code.fetch_docs(:gen_statem) |> elem(6) |> Enum.map(&elem(&1, 0)) |> Enum.filter(&elem(&1, 0) == :callback)
[
  {:callback, :StateName, 3},
  # ...
]
```

And:

```elixir
iex> b :gen_statem
@callback state_name(:enter, oldStateName :: state_name(), data :: data()) ::
            state_enter_result(:state_name)
@callback state_name(event_type(), eventContent :: term(), data :: data()) ::
            event_handler_result(state_name())

# ...
```

And:

```elixir
b :gen_statem.state_name
@callback state_name(:enter, oldStateName :: state_name(), data :: data()) ::
            state_enter_result(:state_name)
@callback state_name(event_type(), eventContent :: term(), data :: data()) ::
            event_handler_result(state_name())
```

Note: no docs, even though they're properly attached to the chunk entry, because we have this discrepancy of `state_name` vs `StateName`.

Maybe it should be `state_name` in the chunk, that'd definitely solve the discrepancy, however looking at https://erlang.org/doc/man/gen_statem.html I think referring to it as `Module:StateName/3` was on purpose, both the  module name and function name vary.

If it makes sense to keep as `StateName` in the chunk, also worth mentioning that we won't be able to document such callback with EDoc given

```erlang
-callback StateName(Foo) -> ok.
```

won't compile. We'd probably need to add support for a `@name` tag to overwrite how it's stored in the chunk.

```erlang
-callback state_name(enter, OldState, Data).
%% @name StateName      (we document callbacks _after_ the definition.)
```

If we keep it as `state_name` in the chunk, maybe we do the reverse, store how the name should be displayed in chunk entry metadata.

lib/ex_doc/retriever.ex
wojtekmach: This is unrelated to pipes, you'd use full declaration when you're calling function from another module. You can skip the module name if the function is local to the current module or was imported using [`import/2`](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#import/2). `Kernel` and `Kernel.SpecialForms` are automatically imported to every module, that's why you can call e.g. `hd(list)` instead of `Kernel.hd(list)`.

lib/ex_doc/retriever.ex
```diff
@@ -379,16 +379,26 @@ defmodule ExDoc.Retriever do
 
   defp get_callback(callback, source, optional_callbacks, module_data) do
     {:docs_v1, _, _, content_type, _, _, _} = module_data.docs
-    {{kind, name, arity}, anno, _, doc, metadata} = callback
+    {{kind, name, arity}, anno, signature, doc, metadata} = callback
     actual_def = actual_def(name, arity, kind)
     doc_line = anno_line(anno)
-    annotations = annotations_from_metadata(metadata)
+    signature = if signature == [], do: nil, else: Enum.join(signature, " ")
```
wojtekmach, line 385: @josevalim I noticed that we were ignore signature from the chunk entry for both callbacks and types so I thought I'd default to it. Was there a particular reason we didn't do it before? If this is the way to go, I'm happy to make sure Elixir is emitting the signature in the chunk.

josevalim: It is on purpose to keep the chunks lean. If there is no signature, we fetch it from the AST. I would keep it as is for now.

PR #1331, opened 2021-02-06
Add --proglang flag

wojtekmach:
This is cherry-picked from #1291 with minor modifications.

Co-authored-by: Radek Szymczyszyn <radoslaw.szymczyszyn@erlang-solutions.com>
Co-authored-by: Wojtek Mach <wojtek@wojtekmach.pl>

### DIFF
--- a/lib/ex_doc.ex
+++ b/lib/ex_doc.ex
@@ -33,6 +33,7 @@ defmodule ExDoc do
     {output, options} = Keyword.pop(options, :output, "./doc")
     {groups_for_modules, options} = Keyword.pop(options, :groups_for_modules, [])
     {nest_modules_by_prefix, options} = Keyword.pop(options, :nest_modules_by_prefix, [])
+    {proglang, options} = Keyword.pop(options, :proglang, :elixir)
 
     {source_url_pattern, options} =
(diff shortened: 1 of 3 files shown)


eksperimental:
I would not recommend to use "proglang", but "programming-language" or "prog-lang" if the former is too long.

wojtekmach:
Good points. I'd like to keep it as is given we use the :proglang name in the code already and revisit the naming before we release it.

Also:

1. on a per-module basis we'd use the language value from the chunk so it's independent from this configuration
2. given 1, we could support project with modules with different languages, here's a proof of concept: http://wojtekmach.pl/docs/opentelemetry_api/otel_ctx.html

So It might we'd use a different name altogether for this, e.g. variant, theme, etc.

eksperimental:
> Good points. I'd like to keep it as is given we use the :proglang name in the code already and revisit the naming before we release it.
> 
> Also:
> 
>     1. on a per-module basis we'd use the language value from the chunk so it's independent from this configuration
> 
>     2. given 1, we could support project with modules with different languages, here's a proof of concept: http://wojtekmach.pl/docs/opentelemetry_api/otel_ctx.html
> 
> 
> So It might we'd use a different name altogether for this, e.g. variant, theme, etc.

That's awesome

Issue #1333, opened 2021-02-08
Erlang support


wojtekmach:
- [x] Basic support
- [x] Render typespecs using erl_pp/erl_prettypr
- [x] Highlight erlang code examples
- [x] Autolink Erlang typespecs
- [x] Rewrite links. E.g. `{:a, [rel: "https://erlang.org/doc/link/seemfa", href: "stdlib:array#new/0", ...}` is rewritten to, based on what docs we generate, an internal link to `array.html#new/0` or external one to erlang.org.
- [x] Warn on invalid refs
- [x] Make sure source_url is correct (e.g. https://hexdocs.pm/hex_core/hex_registry.html#t:private_key/0 isn't.)
- [x] Support `docs.config`
- [ ] Support admonition blocks
- [ ] (UI) Search by `module:fun/arity`
- [x] (UI) Footer link: `` [Erlang programming language](https://erlang.org) ``

TODOs:

- [x] don't crash building diameter docs (https://github.com/erlang/otp/pull/5160)
- [ ] incorrect specs: https://wojtekmach.pl/otp_docs/erts/erlang.html#fun_info/2, https://wojtekmach.pl/otp_docs/erts/erlang.html#monitor/2. We're not handling xml `<type>` tags correctly: https://github.com/erlang/otp/blob/f59eac7b9ce50df08ba583edf5b4d66b7a3b7c7a/erts/doc/src/erlang.xml#L4065:L4068
- [ ] No newline between multiple clauses in a signature: https://wojtekmach.pl/otp_docs/stdlib/gen_statem.html#c:StateName/3
- [ ] no syntax highlighting for _some_ code block: https://wojtekmach.pl/otp_docs/erts/erlang.html#is_map_key/2
- [ ] no support for `<seeguide>`, `<seefile>`, `<seefile>`, etc. https://wojtekmach.pl/otp_docs/kernel/application.html#module-see-also. Not sure how to best handle those, I guess we'd create an extras page for each of them?
- [ ] incorrect link `` [`Memory Allocation`](#content#system_info_allocator) `` in https://wojtekmach.pl/otp_docs/erts/erlang.html#system_info/1
- [ ] incorrectly displaying inline code:
     <img width="644" alt="image" src="https://user-images.githubusercontent.com/76071/131319544-1de79844-5f2c-4d3d-a01e-b8bb57e67044.png">
     (https://wojtekmach.pl/otp_docs/erts/erlang.html#summary)
[...]

PR #1334, opened 2021-02-08
Initial language contract

wojtekmach:
Ref: #1333

### DIFF
--- a/lib/ex_doc/doc_ast.ex
+++ b/lib/ex_doc/doc_ast.ex
@@ -4,6 +4,12 @@ defmodule ExDoc.DocAST do
   alias ExDoc.Markdown
   alias ExDoc.Formatter.HTML.Templates
 
+  @opaque t() :: [ast]
+
+  @typep ast() ::
+           {tag :: atom(), attrs :: [{atom(), binary()}], [ast()], meta :: map()}
+           | binary()
+
   @doc """
   Parse given `doc` according to `doc_format`.
   """
--- a/lib/ex_doc/language/elixir.ex
+++ b/lib/ex_doc/language/elixir.ex
@@ -1,0 +1,151 @@
+defmodule ExDoc.Language.Elixir do
+  @moduledoc false
+
+  @behaviour ExDoc.Language
+
+  @impl true
+  def filter_prefix_pattern(filter_prefix) do
+    if filter_prefix do
+      "Elixir.#{filter_prefix}*.beam"
+    else
+      "*.beam"
+    end
+  end
+
+  @impl true
+  def module_type(module) do
+    cond do
+      function_exported?(module, :__struct__, 0) and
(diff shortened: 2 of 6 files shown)


lib/ex_doc/doc_ast.ex
```diff
@@ -4,6 +4,12 @@ defmodule ExDoc.DocAST do
   alias ExDoc.Markdown
   alias ExDoc.Formatter.HTML.Templates
 
+  @opaque t() :: [ast]
+
+  @typep ast() ::
+           {tag :: atom(), attrs :: [{atom(), binary()}], [ast()], meta :: map()}
+           | binary()
+
```
josevalim, line 12: Should we move this module inside `ExDoc.Language`?

lib/ex_doc/language/elixir.ex
```diff
@@ -0,0 +1,151 @@
...
+
+      function_exported?(module, :behaviour_info, 1) ->
+        :behaviour
+
+      match?("Elixir.Mix.Tasks." <> _, Atom.to_string(module)) ->
+        :task
+
+      true ->
+        :module
+    end
```
josevalim, line 36: Shall we have a `:skip` and get rid of `skip_module?`?

josevalim: Or maybe make it return a tuple and the tuple says if it is shown or hidden by default?

lib/ex_doc/doc_ast.ex
wojtekmach: I wanted to keep the namespaces flatter, doc ast to me is orthogonal to language contract, but at the same time we have this unified doc ast exactly because of language contract. I stil wonder if this DocAST would ever become public, I think it would be nice to refer to `DocAST.t` from `ExDoc.Markdown` contract - you need to know the shape of the thing your callbacks are supposed to return. Of course we could export a separate ExDoc.Markdown.ast type and keep DocAST internal.

So yeah, I think it's fine to make this `ExDoc.Language.DocAST`, we can always change our mind later.

lib/ex_doc/retriever.ex
```diff
@@ -56,11 +59,16 @@ defmodule ExDoc.Retriever do
...
+    cond do
+      language.skip_module?(module) ->
+        []
+
+      docs_chunk = docs_chunk(module) ->
+        generate_node(module, docs_chunk, config)
+
+      true ->
+        []
     end
```
herminiotorres, line 72: On line `64` it is would be necessary? instead of you could do something like this:

```elixir
    cond do
      docs_chunk = docs_chunk(module) when is_tuple(docs_chunk) ->
        generate_node(module, docs_chunk, config)

      true ->
        []
    end
```

wojtekmach:
Closing in favour of #1362.

PR #1335, opened 2021-02-18
Update erlang.org/man URL

eksperimental:
www.erlang.org/man redirects to the erlang.org/man
additionally the prefered links are with https


wojtekmach:
Thanks!

eksperimental:
You are welcome @wojtekmach

PR #1342, opened 2021-03-09
Add function landing page

wojtekmach:
image


josevalim:
Looks great, a couple nits and questions:

1. We should generate one page per name (but not arity). So if there are multiple arities, they will all be shown. The title can be either `Mod.fun/*` or `Mod.fun`.

2. Should we show ExDoc.Markdown as selected on the sidebar?

eksperimental:
What does this feature do?

josevalim:
@eksperimental emits one page per module+function/callback combo to provide better indexing and SEO. For example, if you search for Enum.reduce, you land on Enum, but not on Enum.reduce itself... but sometimes reduce won't even show up because it is too diluted on the page.

josevalim:
> So they will only be accessible to search engines, from the user interface everything remains the same?

yes, we are not planning to link to them.

> https://en.wikipedia.org/wiki/Sitemaps

Hexdocs already generate them!

PR #1347, opened 2021-03-31
Locally link autoimported functions

eksperimental:
If the project is Elixir, functions that link to Kernel and Kernel.SpecialForms now link to the local page, not to hexdocs.pm/elixir/


eksperimental:
This bug was really annoying as it would take me to hexdocs.pm out of the docs in master without even noticing

test/ex_doc/autolink_test.exs
```diff
@@ -107,6 +107,9 @@ defmodule ExDoc.AutolinkTest do
       assert autolink("for/1", apps: [:elixir]) ==
                ~m"[`for/1`](Kernel.SpecialForms.html#for/1)"
 
+      assert autolink("for/1", app: :elixir) ==
+               ~m"[`for/1`](Kernel.SpecialForms.html#for/1)"
```
josevalim, line 111: Are you sure we need a new app: flag? If you see above, we were already using relative links if `apps: [:elixir]`. We just need to make sure the proper apps flag is set when generating Elixir docs.

eksperimental: Thank you. There was a bug somewhere else. Can you please review the PR again?

josevalim:

Issue #1322, opened 2021-01-25
Deal with custom link that links to undefined/hidden module


eksperimental:
```elixir
`Code.Typespec` 

[code typespec](`Code.Typespec`)
```
`Code.Typespec` is a hidden module, in the first example, and ExDoc what it does is to treat  format the text inside the backticks as such generating `<code class="inline">Code.Typespec</code>`

but when using a custom link, the generated code is 
```<a href="`Code.Typespec`">code typespec</a>```

I am not too sure what is the best way to deal with this, probably removing the backticks and linking to Code.Typespec, 
but having the backticks does not seem right

josevalim:
Great find. We should warn and remove the link IMO. :)

eksperimental:
We are currently warning. 
I am not 100% convinced of removing the link because it may hide the error, even though we are warning. A project may generate multiple warnings and this one can get buried there.
But removing it will be consistent  with how we deal in the first example

wojtekmach:
Good find. I think we should definitely remove the link because there's nowhere to link to.

eksperimental:
Alright +1 with that as we don't have any other options

eksperimental:
I will take care of this.

PR #1349, opened 2021-04-01
Deal with custom link that links to undefined/hidden module

eksperimental:
Closes https://github.com/elixir-lang/ex_doc/issues/1322

### DIFF
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -76,13 +76,11 @@ defmodule ExDoc.Autolink do
   end
 
   defp walk({:code, attrs, [code], meta} = ast, config) do
-    case url(code, :regular_link, config) do
-      nil ->
-        ast
-
-      url ->
-        code = remove_prefix(code)
-        {:a, [href: url], [{:code, attrs, [code], meta}], %{}}
+    if url = url(code, :regular_link, config) do
+      code = remove_prefix(code)
+      {:a, [href: url], [{:code, attrs, [code], meta}], %{}}
+    else
+      ast
     end
   end
 
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -91,9 +91,6 @@ defmodule ExDoc.Autolink do
   defp remove_link({:a, _attrs, inner, _meta}),
     do: inner
 
-  defp remove_link(ast),
-    do: ast
-
   @ref_regex ~r/^`(.+)`$/
 
   defp custom_link(attrs, config) do
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -63,27 +63,38 @@ defmodule ExDoc.Autolink do
   end
 
   defp walk({:a, attrs, inner, meta} = ast, config) do
-    cond do
-      url = custom_link(attrs, config) ->
-        {:a, Keyword.put(attrs, :href, url), inner, meta}
+    case custom_link(attrs, config) do
+      :remove_link ->
+        remove_link(ast)
 
-      true ->
+      nil ->
         ast
+
+      url ->
+        {:a, Keyword.put(attrs, :href, url), inner, meta}
     end
   end
 
   defp walk({:code, attrs, [code], meta} = ast, config) do
-    if url = url(code, :regular_link, config) do
-      code = remove_prefix(code)
-      {:a, [href: url], [{:code, attrs, [code], meta}], %{}}
-    else
-      ast
+    case url(code, :regular_link, config) do
+      nil ->
+        ast
+
+      url ->
+        code = remove_prefix(code)
+        {:a, [href: url], [{:code, attrs, [code], meta}], %{}}
     end
   end
 
   defp walk({tag, attrs, ast, meta}, config) do
     {tag, attrs, walk(ast, config), meta}
   end
+
+  defp remove_link({:a, _attrs, inner, _meta}),
+    do: inner
+
+  defp remove_link(ast),
+    do: ast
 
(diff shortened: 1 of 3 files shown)


test/ex_doc/autolink_test.exs
```diff
@@ -449,11 +450,17 @@ defmodule ExDoc.AutolinkTest do
     assert warn(~m"[bad](`String.upcase/9`)", extras: []) =~
              "documentation references \"String.upcase/9\" but it is undefined or private"
 
-    assert warn(~m"[custom text](`Elixir.Unknown`)", []) =~
-             "documentation references module \"Elixir.Unknown\" but it is undefined\n"
+    assert_and_warn(
```
wojtekmach, line 453: could you please revert this change? with this helper we're saving typing a few characters but it is less flexible, for example we can't easily inspect the whole output (which was as simple as `|> IO.inspect` before but now the helper returns a boolean).

fwiw I am the first to admit that I probably went a bit overboard with adding many test helpers (which are saving typing characters but add complexity) so let's not add even more.

test/ex_doc/autolink_test.exs
```diff
@@ -449,11 +450,17 @@ defmodule ExDoc.AutolinkTest do
...
+      "documentation references module \"Elixir.Unknown\" but it is undefined\n"
+    )
 
-    assert warn(~m"[Unknown](`Unknown`)") =~
-             "documentation references module \"Unknown\" but it is undefined"
+    assert_and_warn(
+      ~m"[It is Unknown](`Unknown`)",
+      ["It is Unknown"],
+      "documentation references module \"Unknown\" but it is undefined"
+    )
```
wojtekmach, line 464: which test is testing the changed behaviour?

lib/ex_doc/autolink.ex
```diff
@@ -63,28 +63,39 @@ defmodule ExDoc.Autolink do
...
-      {:a, [href: url], [{:code, attrs, [code], meta}], %{}}
-    else
-      ast
+    case url(code, :regular_link, config) do
+      nil ->
+        ast
+
+      url ->
+        code = remove_prefix(code)
+        {:a, [href: url], [{:code, attrs, [code], meta}], %{}}
```
eksperimental, line 85: this is not changing anything. It will revert it as it was. It was probably a left over from a different approach I was taking

test/ex_doc/formatter/html/templates_test.exs
```diff
@@ -167,9 +167,7 @@ defmodule ExDoc.Formatter.HTML.TemplatesTest do
       content = Templates.sidebar_template(doc_config(), @empty_nodes_map)
 
       assert content =~
-               ~r{<div class="sidebar-header">\s*<div class="sidebar-projectDetails">\s*<a href="#{
-                 homepage_url()
-               }" class="sidebar-projectName">\s*Elixir\s*</a>\s*<strong class="sidebar-projectVersion">\s*v1.0.1\s*</strong>\s*</div>\s*</div>}
+               ~r{<div class="sidebar-header">\s*<div class="sidebar-projectDetails">\s*<a href="#{homepage_url()}" class="sidebar-projectName">\s*Elixir\s*</a>\s*<strong class="sidebar-projectVersion">\s*v1.0.1\s*</strong>\s*</div>\s*</div>}
```
eksperimental, line 170: This got changed when I run `mix format` with elixir 1.12 or master

test/ex_doc/autolink_test.exs
eksperimental: the ones that call assert_and_warn

wojtekmach: I might be missing something but it looked like you changed `assert warn(...)` to `assert_and_warn(...)` with the same inputs so effectively we are effectively not adding additional test coverage?

test/ex_doc/formatter/html/templates_test.exs
wojtekmach: yeah this is unfortunate, please revert this.

test/ex_doc/autolink_test.exs
eksperimental: it is doing an assert between the first and the second argument which checks for the link being stripped

wojtekmach: oh, I see it now, previously we weren't checking what's the generated output. thanks!

lib/ex_doc/autolink.ex
```diff
@@ -85,6 +88,12 @@ defmodule ExDoc.Autolink do
     {tag, attrs, walk(ast, config), meta}
   end
 
+  defp remove_link({:a, _attrs, inner, _meta}),
+    do: inner
+
+  defp remove_link(ast),
+    do: ast
```
wojtekmach, line 95: this clause is not needed, please remove it or add a test that exercises it :)

wojtekmach (approved):
Just one thing left and feel free to merge it afterwards!

lib/ex_doc/autolink.ex
eksperimental: Agreed with that is unneeded, but how would you go about testing it a private function that is called by private functions.

eksperimental: > I wouldn't test a private function!
> 
> I think with our autolink tests we strike a nice enough balance, the tests are sufficiently low level (in that we don't need to spin up the entire exdoc to test something) but at the same time high-level enough that we test the external behaviour, what was the returned doc AST (and potential side effects like warnings), we don't test implementation details like what got called with what, etc.

Sorry for the confusion, when I read your message I read "please remove it AND add a test that exercises it!", that' s why I was a bit puzzled.
I haven't had my morning coffee yet, that's probably it :smile:

PR #1353, opened 2021-04-07
Rename doc variables

eksperimental:
There was a mix of variable rebinding,
and naming everything doc/docs.

This commit adheres to the name definitions in the Code.fetch_docs/1 specs
https://hexdocs.pm/elixir/Code.html#fetch_docs/1

### DIFF
--- a/lib/ex_doc/doc_ast.ex
+++ b/lib/ex_doc/doc_ast.ex
@@ -5,9 +5,9 @@ defmodule ExDoc.DocAST do
   alias ExDoc.Formatter.HTML.Templates
 
   @doc """
-  Parse given `doc` according to `doc_format`.
+  Parses given `doc_content` according to `doc_format`.
   """
-  def parse!(doc, doc_format, options \\ [])
+  def parse!(doc_content, doc_format, options \\ [])
 
   def parse!(markdown, "text/markdown", opts) do
     parse_markdown(markdown, opts)
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -218,19 +218,19 @@ defmodule ExDoc.Retriever do
   ## Function helpers
 
   defp get_docs(%{type: type, docs: docs} = module_data, source, config) do
-    {:docs_v1, _, _, _, _, _, docs} = docs
+    {:docs_v1, _, _, _, _, _, doc_elements} = docs
 
     groups_for_functions =
       Enum.map(config.groups_for_functions, fn {group, filter} ->
         {Atom.to_string(group), filter}
       end) ++ [{"Functions", fn _ -> true end}]
 
-    function_docs =
-      for doc <- docs, doc?(doc, type) do
-        get_function(doc, source, module_data, groups_for_functions)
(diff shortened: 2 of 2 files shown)


lib/ex_doc/doc_ast.ex
```diff
@@ -5,9 +5,9 @@ defmodule ExDoc.DocAST do
   alias ExDoc.Formatter.HTML.Templates
 
   @doc """
-  Parse given `doc` according to `doc_format`.
+  Parses given `doc_content_binary` according to `doc_format`.
   """
-  def parse!(doc, doc_format, options \\ [])
+  def parse!(doc_content_binary, doc_format, options \\ [])
```
wojtekmach, line 10: it's not a binary for `"application/erlang+html"` :)

eksperimental: should we name it `doc_content_binary_or_ast` ?

wojtekmach: how about just doc_content?

eksperimental: > how about just doc_content?

because doc_content according  the specs is the map that contains this string for Elixir docs.

`doc_content: %{optional(binary()) => binary()} | :none | :hidden,`

eksperimental: we can just leave `doc` and explain in the docs what `doc` could be, either an erlang ast or a string containing markdown

eksperimental: Something like this.

```elixir
  @doc """
  Parses given `doc` according to `doc_format`.

  When `doc_format` is "text/markdown", `doc` is expeceted to be a binary.
  When `doc_format` is "application/erlang+html", `doc` is expeceted to be an Erlang AST.
  """
  def parse!(doc, doc_format, options \\ [])
```

wojtekmach: A friendly reminder that this is a private module, we don’t need to add comprehensive docs

eksperimental: Yes, I know but I am working on a fix and everthing is named "doc" and "docs" and  it is hard to grasp what is what.

eksperimental: I am reverting this change

lib/ex_doc/retriever.ex
```diff
@@ -156,8 +156,8 @@ defmodule ExDoc.Retriever do
     {first in ?a..?z, name, arity}
   end
 
-  defp doc_ast(format, %{"en" => doc}, options),
-    do: DocAST.parse!(doc, format, options)
+  defp doc_ast(format, %{"en" => doc_content_binary}, options),
+    do: DocAST.parse!(doc_content_binary, format, options)
```
wojtekmach, line 160:
```suggestion
  defp doc_ast(format, %{"en" => doc_content}, options),
    do: DocAST.parse!(doc_content, format, options)
```

wojtekmach (approved):

wojtekmach:
Thanks!

PR #1361, opened 2021-05-06
Add ellipsis to more sections

dnsbty:
It looks like we moved the ellipsis usage to be more specific in 6650cf1176330af380855a896769bfbb4adaa2a2 and I think we need to expand the places where ellipsis are used. For example, the docs for ExUnit currently look like this:

image

So this change makes it look like this:

image

I think this provides better affordance to let users know that there is more to the title than just what they can see. It also may be worth considering removing the `overflow: hidden` to let the lines spill over and have the whole thing be visible, but this change works within the parameters of the current design.

### DIFF
--- a/assets/less/sidebar.less
+++ b/assets/less/sidebar.less
@@ -253,7 +253,7 @@
       }
 
       a.expand,
-      .sections-list a {
+      .sections-list > li {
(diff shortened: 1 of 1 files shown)


josevalim:

PR #1362, opened 2021-05-11
Add language contract

wojtekmach:
Ref: #1333

For a demo, see: https://hexdocs.pm/hex_core/hex_tarball.html#functions

Note, for now we render Erlang type specs without linking. Support for those would be in a future PR.

### DIFF
--- a/lib/ex_doc/cli.ex
+++ b/lib/ex_doc/cli.ex
@@ -23,6 +23,7 @@ defmodule ExDoc.CLI do
         ],
         switches: [
           language: :string,
+          proglang: :atom,
           source_ref: :string,
           version: :boolean
         ]
--- a/lib/ex_doc/cli.ex
+++ b/lib/ex_doc/cli.ex
@@ -23,7 +23,6 @@ defmodule ExDoc.CLI do
         ],
         switches: [
           language: :string,
-          proglang: :atom,
           source_ref: :string,
           version: :boolean
         ]
--- a/lib/ex_doc/language.ex
+++ b/lib/ex_doc/language.ex
@@ -7,6 +7,8 @@ defmodule ExDoc.Language do
 
   @callback module_data(module()) :: map()
 
+  @callback function_data(entry :: tuple(), module_data :: map()) :: map()
+
   def get(:elixir), do: ExDoc.Language.Elixir
   def get(:erlang), do: ExDoc.Language.Erlang
 end
--- a/lib/ex_doc/language.ex
+++ b/lib/ex_doc/language.ex
@@ -1,8 +1,6 @@
 defmodule ExDoc.Language do
   @moduledoc false
 
-  @callback id() :: atom()
-
   @callback module_data(module()) :: map()
 
   @callback function_data(entry :: tuple(), module_data :: map()) :: map()
--- a/lib/ex_doc/language.ex
+++ b/lib/ex_doc/language.ex
@@ -3,6 +3,8 @@ defmodule ExDoc.Language do
 
   @callback id() :: atom()
 
+  @callback filter_prefix_pattern(String.t() | nil) :: String.t()
+
   def get(:elixir), do: ExDoc.Language.Elixir
   def get(:erlang), do: ExDoc.Language.Erlang
 end
--- a/lib/ex_doc/language.ex
+++ b/lib/ex_doc/language.ex
@@ -1,0 +1,8 @@
+defmodule ExDoc.Language do
+  @moduledoc false
+
+  @callback id() :: atom()
+
+  def get(:elixir), do: ExDoc.Language.Elixir
+  def get(:erlang), do: ExDoc.Language.Erlang
+end
--- a/lib/ex_doc/language.ex
+++ b/lib/ex_doc/language.ex
@@ -11,6 +11,8 @@ defmodule ExDoc.Language do
 
   @callback callback_data(entry :: tuple(), module_data :: map()) :: map()
 
+  @callback type_data(entry :: tuple(), spec :: term(), module_data :: map()) :: map()
+
   def get(:elixir), do: ExDoc.Language.Elixir
   def get(:erlang), do: ExDoc.Language.Erlang
 end
--- a/lib/ex_doc/language.ex
+++ b/lib/ex_doc/language.ex
@@ -1,15 +1,76 @@
 defmodule ExDoc.Language do
   @moduledoc false
 
+  @doc """
+  Returns a map with module information.
+
+  The map has the following keys:
+
+    * `:id`
+
+    * `:title`
+
+    * `:type`
+
+    * `:skip` - whether module should be skipped from generating the docs
+
+    * `:extra_callback_types` - a list of types that are considered callbacks
+
+  """
   @callback module_data(module()) :: map()
 
+  @doc """
+  Returns a map with function information.
+
+  The map has the following keys:
+
+    * `:doc_fallback` - if set, a 0-arity function that returns DocAST which
+       will be used as fallback to empty docs on the function node
+
+    * `:extra_annotations`
+
+    * `:line_override` - if set, overrides the line where the code is located
+
+    * `:specs` - a list of specs that will be later formatted by `c:typespec/2`
+
+  """
   @callback function_data(entry :: tuple(), module_data :: map()) :: map()
 
+  @doc """
+  Returns a map with callback information.
+
+  The map has the following keys:
+
+    * `:actual_def` - `{name, arity}` of how the callback is actually represented
+      in abstract code
+
+    * `:line` - if set, returns the line where the code is located
+
+    * `:signature_fallback` - if set, a 0-arity function that returns signature which
+      which will be used as a fallback to empty signature on the callback node
+
+    * `:specs` - a list of specs that will be later formatted by `c:typespec/2`
+
+  """
   @callback callback_data(entry :: tuple(), module_data :: map()) :: map()
 
+  @doc """
+  Returns a map with type information.
+
+  The map has the following keys:
+
+    * `:spec` - a spec that will be later formatted by `c:typespec/2`
+
+    * `:signature_fallback` - if set, a 0-arity function that returns signature which
+      which will be used as a fallback to empty signature on the callback node
+
+  """
   @callback type_data(entry :: tuple(), spec :: term(), module_data :: map()) :: map()
 
-  @callback typespec(spec :: term(), opts :: keyword()) :: term()
+  @doc """
+  Formats typespecs.
+  """
+  @callback typespec(spec :: term(), opts :: keyword()) :: iodata()
 
   def get(:elixir), do: ExDoc.Language.Elixir
   def get(:erlang), do: ExDoc.Language.Erlang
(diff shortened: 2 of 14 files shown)


josevalim:
@wojtekmach in the example, the code snippets do not have syntax highlighting. is there something we can do about it? 🤔

lib/ex_doc/cli.ex
```diff
@@ -23,7 +23,6 @@ defmodule ExDoc.CLI do
         ],
         switches: [
           language: :string,
-          proglang: :atom,
```
josevalim, line 26: Shouldn't we still be able to pass it as an argument from the CLI? 🤔

lib/ex_doc/retriever.ex
```diff
@@ -58,22 +58,21 @@ defmodule ExDoc.Retriever do
...
-      nil -> {nil, nil}
-      prefix -> {String.trim_leading(title, prefix <> "."), prefix}
-    end
-  end
-
-  # Special case required for Elixir
+  # TODO: see if we can do something about it after all?
+  # Special case required for Elixir - we don't know it's Elixir until we read
+  # the chunk so we can't have ExDoc.Language.Elixir specific behaviour.
   defp docs_chunk(:elixir_bootstrap), do: false
```
josevalim, line 76: I think it is fine to do nothing about this as I think both of them are optimizations?

josevalim (commented):
Wow, awesome work! I have dropped some minor comments.

Before we move forward, we should probably document the language contract better. I am also thinking we should restructure the unit tests. We have already discussed a couple times how the current tests are non-ideal and I think the retriever refactor can be a perfect opportunity to handle those (it can be a separate PR).

Finally, there is some stuff in the ExDoc (nesting and grouping) that is Elixir specific, we will want to address that too (but it is fine if it is done as part of another PR). :)

lib/ex_doc/cli.ex
wojtekmach: yup, I think I made this change by mistake, will revert.

it brings an interesting question though, what the proglang is actually used for, and perhaps it is a bit overloaded already. I could even group the contract callbacks into 3 groups:

- `@callback id()` - this is used for the "theme", picking the correct CSS, etc. This is a **project-wide** setting
- `@callback filter_prefix_pattern()` - `"Elixir.#{filter_prefix}*.beam"` for Elixir, `"#{filter_prefix}*.beam"` for Erlang. This is a **project-wide** setting
- remaining callbacks are **per-module**

so in a way, the `--proglang` flag only matters for the first two callbacks.

I think it is fine to move forward with things as is but just adding some more context around this.

josevalim: That's a good point, because it raises a question about mixed projects and other configuration, like group nesting. For now, I think we need to at least change how filtering works:

1. First get all modules from disk
2. And then, based on the language, filter the module

Effectively moving it from a project config to a module config. WDYT?

wojtekmach: sounds good, thanks. I'll leave the `filter_prefix_pattern` callback as is, I'll just change how it is being used internally.

wojtekmach: I removed the `filter_prefix_pattern/1` callback in 49d3fcd0fa50a15c51ad4f2ef63bae6b0d580d4c

in fact, I removed the `id/0` callback too in b4d1fcfd545d670f5b684dcc5d4e6a38a7395f1c.

Now, `proglang` is the project-wide option for CSS and such, and `language` (as in ExDoc.Language) is per-module.

Thanks for the feedback! <3

wojtekmach:
> in the example, the code snippets do not have syntax highlighting. is there something we can do about it? 

good catch. I noticed that Edoc parses:

    @doc
    ```
    foo().
    '''

into the following erlang+html:

```elixir
{:pre, [], ["  foo()"]}
```

so we'd need to handle that on a per-language basis. (For Elixir, we have `<pre><code>...</code></pre>`.)

I'll do that in the next PR.

There's a workaround in the meantime:

```
%% @doc
%%
%% Examples
%%
%% <pre><code class="erlang">sets:new().</code></pre>
```

renders:

image

and it looks decent in the shell too:

image


image

lib/ex_doc/language.ex
```diff
@@ -0,0 +1,77 @@
...
+  @doc """
+  Returns a map with module information.
+
+  The map has the following keys:
+
+    * `:id`
+
+    * `:title`
+
+    * `:type`
```
josevalim, line 13: Let's add a small description to those too?

lib/ex_doc/language.ex
```diff
@@ -0,0 +1,77 @@
...
+
+    * `:skip` - whether module should be skipped from generating the docs
+
+    * `:extra_callback_types` - a list of types that are considered callbacks
+
+  """
+  @callback module_data(module()) :: map()
```
josevalim, line 20: Instead `map()` should return a proper spec for the keys above?

lib/ex_doc/language/elixir.ex
```diff
@@ -0,0 +1,240 @@
...
+    }
+  end
+
+  @impl true
+  def callback_data(entry, module_data) do
+    {{kind, name, arity}, _anno, _signature, _doc, _metadata} = entry
+    actual_def = actual_def(name, arity, kind)
+
+    specs =
+      case Map.fetch(module_data.callbacks, actual_def) do
```
josevalim, line 53: So it seems the module_data we receive in callbacks is a superset of the one we return. If so we should document the input module_data too. :)

lib/ex_doc/retriever.ex
```diff
@@ -58,25 +57,29 @@ defmodule ExDoc.Retriever do
...
 
-  defp nesting_info(title, prefixes) do
-    prefixes
-    |> Enum.find(&String.starts_with?(title, &1 <> "."))
-    |> case do
-      nil -> {nil, nil}
-      prefix -> {String.trim_leading(title, prefix <> "."), prefix}
+  defp maybe_skip(module_data, filter_prefix) do
+    if filter_prefix && not String.starts_with?(module_data.id, filter_prefix) do
+      %{module_data | skip: true}
```
josevalim, line 77: Nice!

lib/ex_doc/language/elixir.ex
```diff
@@ -0,0 +1,240 @@
...
+
+    %{
+      doc_fallback: fn ->
+        impl = Map.fetch(module_data.impls, actual_def)
+
+        callback_doc_ast(name, arity, impl) ||
+          delegate_doc_ast(metadata[:delegate_to])
+      end,
+      extra_annotations: extra_annotations,
+      line_override: find_function_line(module_data, actual_def),
```
josevalim, line 42: Do we need to call it override? Instead, should we simply say `line` is optional? And if one is not set, we default to one? Or rather, maybe we should be explicit and always ask for a `line`?

The other thing is that we have two entries: doc_line and source_line. I assume this one is source_line?

lib/ex_doc/language/elixir.ex
josevalim: Or is this actually `module_node`?

lib/ex_doc/nodes.ex
```diff
@@ -36,7 +37,8 @@ defmodule ExDoc.ModuleNode do
           typespecs: list(),
           source_path: nil | String.t(),
           source_url: nil | String.t(),
-          type: nil | :module | :exception | :protocol | :impl | :behaviour | :task
+          type: nil | :module | :exception | :protocol | :impl | :behaviour | :task,
```
josevalim, line 40: We should probably generalize this to be `nil | atom()`?

lib/ex_doc/retriever.ex
```diff
@@ -370,36 +318,31 @@ defmodule ExDoc.Retriever do
...
-          specs = Enum.map(specs, &Code.Typespec.spec_to_quoted(name, &1))
-          signature = signature || get_typespec_signature(hd(specs), arity)
-          {specs, line, signature}
-
-        :error ->
-          {[], doc_line, signature || "#{name}/#{arity}"}
-      end
+    signature =
+      signature ||
+        (callback_data.signature_fallback && callback_data.signature_fallback.()) ||
```
josevalim, line 340: Make it default to a function that returns nil so you can do:

```suggestion
        callback_data.signature_fallback.() ||
```

lib/ex_doc/retriever.ex
```diff
@@ -499,61 +445,11 @@ defmodule ExDoc.Retriever do
...
-  defp to_var([{:->, _, _} | _], _), do: {:function, [], nil}
-  defp to_var({:<<>>, _, _}, _), do: {:binary, [], nil}
-  defp to_var({:%{}, _, _}, _), do: {:map, [], nil}
-  defp to_var({:{}, _, _}, _), do: {:tuple, [], nil}
-  defp to_var({_, _}, _), do: {:tuple, [], nil}
-  defp to_var(integer, _) when is_integer(integer), do: {:integer, [], nil}
-  defp to_var(float, _) when is_integer(float), do: {:float, [], nil}
-  defp to_var(list, _) when is_list(list), do: {:list, [], nil}
-  defp to_var(atom, _) when is_atom(atom), do: {:atom, [], nil}
-  defp to_var(_, position), do: {:"arg#{position}", [], nil}
```
josevalim, line 540: Such a beautiful clean up.

lib/ex_doc/retriever.ex
```diff
@@ -145,60 +140,58 @@ defmodule ExDoc.Retriever do
...
+      title: extra.title,
       name: module,
-      type: get_type(module),
+      language: language,
+      type: extra.type,
+      skip: extra.skip,
       specs: get_specs(module),
       impls: get_impls(module),
       callbacks: get_callbacks(module),
       abst_code: get_abstract_code(module),
```
josevalim, line 189: We should probably move this to the language callbacks... i think Elixir and Erlang are the only two languages that benefit from looking at the abstract code. LFE, Gleam, etc either won't have it or it will be inaccurate.

test/ex_doc/retriever_test.exs
```diff
@@ -1,6 +1,6 @@
 defmodule ExDoc.RetrieverTest do
```
josevalim, line 1: We should leave a TODO to break this in two: Elixir and Erlang specific tests.

lib/ex_doc/language/elixir.ex
wojtekmach: For now I'll just rename this to `line`. About `find_module_line`, good point too, you're right that most likely on Erlang & Elixir will rely on it, so perhaps we should move that share functionality between each other, I'll add a TODO.

lib/ex_doc/retriever.ex
wojtekmach: Good call

lib/ex_doc/language/elixir.ex
wojtekmach: oops, it's neither! I added type specs.

tedhexaflow:
Hello @wojtekmach 
Just noticed a minor issue while checking on the demo.
I believed you might want to have a look and created an issue. https://github.com/elixir-lang/ex_doc/issues/1366

PR #1364, opened 2021-05-21
Deal with remote types in signatures

eksperimental:
It deals with signature for remote types, such as in:
`@callback callback1(GenServer.options()) :: :ok
`
it generates the function signature
`callback1(options)`

while previously it would generate
`callback1(arg1)`

### DIFF
--- a/test/ex_doc/retriever_test.exs
+++ b/test/ex_doc/retriever_test.exs
@@ -250,6 +250,19 @@ defmodule ExDoc.RetrieverTest do
       assert upcase.doc == ExDoc.Markdown.to_ast("See `String.upcase/1`.")
     end
 
+    test "signatures", c do
+      elixirc(c, ~S"""
+      defmodule Signatures do
+        @callback remote(GenServer.options()) :: :ok
+      end
+      """)
+
+      [mod] = Retriever.docs_from_modules([Signatures], %ExDoc.Config{})
+      [remote] = mod.docs
+
+      assert remote.signature == "remote(options)"
+    end
+
     test "Mix tasks", c do
       elixirc(c, ~S"""
       defmodule Mix.Tasks.MyTask do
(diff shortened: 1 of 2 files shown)


whatyouhide (approved):
Nice catch. 💟

eksperimental:
> Nice catch. heart_decoration

Thanks. I am creating a test for the signatures, since I found several cases that we are not covering which will be covered in a future PR, so it is better to have a test of its own for these signatures. I have rebased the commit.

test/ex_doc/retriever_test.exs
```diff
@@ -250,6 +250,19 @@ defmodule ExDoc.RetrieverTest do
       assert upcase.doc == ExDoc.Markdown.to_ast("See `String.upcase/1`.")
     end
 
+    test "signatures", c do
```
wojtekmach, line 253: I'm not sure if we should have this as a separate test or merge this with an existing one, e.g.:

https://github.com/elixir-lang/ex_doc/blob/47ddfd27f1172f43e5de49d4ea9708eae997da8e/test/ex_doc/retriever_test.exs#L15

wdyt @josevalim?

wojtekmach (approved):

test/ex_doc/retriever_test.exs
eksperimental: I am working on a different PR where I am testing aiming to test all the Typespecs described in the Typespecs page. As I have found that some were not covered. So these will be added here as well.

wojtekmach: actually I'm keen to merge this as is, is that ok? I plan to move tests around after #1365 1362 lands.

eksperimental: Sure. Please go ahead.

PR #1365, opened 2021-05-23
Do not filter out EarmarkParser options

linkdd:
# Actual Behavior

Options passed to the `EarmarkParser` Markdown Processor are filtered out in order to provide default values.

# Desired Behavior

Let the user pass extra options to `EarmarkParser`.

# Changes

Use `Keyword.merge` with the default values as first argument and the user supplied values as second argument.

### DIFF
--- a/lib/ex_doc/markdown/earmark.ex
+++ b/lib/ex_doc/markdown/earmark.ex
@@ -28,13 +28,13 @@ defmodule ExDoc.Markdown.Earmark do
   @impl true
   def to_ast(text, opts) do
     options = [
-      gfm: Keyword.get(opts, :gfm, true),
-      line: Keyword.get(opts, :line, 1),
-      file: Keyword.get(opts, :file, "nofile"),
-      breaks: Keyword.get(opts, :breaks, false),
-      smartypants: Keyword.get(opts, :smartypants, false),
+      gfm: true,
+      line: 1,
+      file: "nofile",
+      breaks: false,
+      smartypants: false,
       pure_links: true
-    ]
+    ] |> Keyword.merge(opts)
 
     case EarmarkParser.as_ast(text, options) do
       {:ok, ast, messages} ->


lib/ex_doc/markdown.ex
```diff
@@ -52,8 +52,14 @@ defmodule ExDoc.Markdown do
   @doc """
   Changes the markdown processor globally.
   """
-  def put_markdown_processor(processor, options \\ []) do
+  def put_markdown_processor(processor) do
     Application.put_env(:ex_doc, @markdown_processor_key, processor)
```
josevalim, line 56: Instead of having two keys and functions, let’s store it as a tuple {processor, options}. WDYT?

linkdd: The `:markdown_processor_options` can be supplied without a `:markdown_processor` key (relying on the default implementation, `ExDoc.Markdown.Earmark` in this case).

What value should `{processor, options}` have if there is no `processor` set?

The current behavior is to discover the `processor` when no value is set:

`get_markdown_processor/0`:

https://github.com/linkdd/ex_doc/blob/patch-1/lib/ex_doc/markdown.ex#L40-L50

`find_markdown_processor/0`:

https://github.com/linkdd/ex_doc/blob/patch-1/lib/ex_doc/markdown.ex#L66-L70

This way, I'm sure I'm not breaking anything 😄

josevalim: The options are specific to a processor anyway, so passing them with the processor is the way to go.

lib/ex_doc/markdown.ex
```diff
@@ -18,22 +18,20 @@ defmodule ExDoc.Markdown do
   """
   @callback to_ast(String.t(), Keyword.t()) :: term()
 
-  @doc """
-  Returns true if all dependencies necessary are available.
-  """
-  @callback available?() :: boolean()
```
josevalim, line 24: Please don’t remove this callback. :)

lib/ex_doc/markdown.ex
josevalim: To be clear, we could make
The markdown_processor to be a tuple or an atom. If a tuple, the second element is options. :)

linkdd: I opted to always set a tuple but with `processor` set to nil, and if it is nil, we try to resolve it.

josevalim: And we remove markdown_processor_options since it isn’t used anyway. :)

linkdd: So `markdown_processor: {nil, options}` in the Mix configuration instead of 2 options ?

josevalim: Yes, but I don’t want to allow nil. If we change the default, then your options may not work with the new default, so better to be explicit.

linkdd: So, the configuration would be one of:

 - `markdown_processor: ExDoc.Markdown.Earmark`
 - `markdown_processor: [module: ExDoc.Markdown.Earmark, options: ...]`
 - `markdown_processor: [options: ...]`

josevalim: It should be module() or {module(), keyword()}.

linkdd: Ok done 🙂

lib/ex_doc/markdown.ex
```diff
@@ -33,21 +33,32 @@ defmodule ExDoc.Markdown do
...
   end
 
   @doc """
   Gets the current markdown processor set globally.
   """
   def get_markdown_processor do
-    case Application.fetch_env(:ex_doc, @markdown_processor_key) do
-      {:ok, processor} ->
-        processor
+    {processor, options} = case Application.fetch_env(:ex_doc, @markdown_processor_key) do
```
josevalim, line 44: Sorry, one last nitpick: let's normalize it when we write (on put_markdown_process) and not when we read. It is generally preferred to normalize data when we store it, instead of every time we write. :)

lib/ex_doc/markdown.ex
```diff
@@ -33,29 +33,33 @@ defmodule ExDoc.Markdown do
...
   end
 
   @doc """
   Changes the markdown processor globally.
   """
-  def put_markdown_processor(processor) do
-    Application.put_env(:ex_doc, @markdown_processor_key, processor)
+  def put_markdown_processor(processor) when is_atom(processor) do
+    put_markdown_processor({processor, []})
+  end
```
josevalim, line 60:
```suggestion
  end

```

lib/ex_doc/markdown/earmark.ex
```diff
@@ -28,13 +28,13 @@ defmodule ExDoc.Markdown.Earmark do
...
-      breaks: Keyword.get(opts, :breaks, false),
-      smartypants: Keyword.get(opts, :smartypants, false),
+      gfm: true,
+      line: 1,
+      file: "nofile",
+      breaks: false,
+      smartypants: false,
       pure_links: true
-    ]
+    ] |> Keyword.merge(opts)
```
josevalim, line 37:
```suggestion
    ]

    options = Keyword.merge(options, opts)
```

josevalim:
@linkdd, please update the tests and we should be good to go! Thanks!

linkdd:
Will do tomorrow, 11pm here time for bed 🛌

Thank you for your suggestions!

linkdd:
@josevalim sorry for the delay, I fixed the unit test and added a new one.

josevalim:

PR #1373, opened 2021-06-21
Add autolinking for Erlang docs

wojtekmach:


wojtekmach:
I noticed some performance issues when generating Elixir docs using this branch. I have another idea for a smoother transition that would hopefully more easily uncover those, so I'm gonna go ahead and close this for now.

PR #1376, opened 2021-07-02
ExDoc.Language.Erlang: Autolink typespecs

wojtekmach:
Demo:

http://wojtekmach.pl/docs/telemetry/telemetry.html#execute/3

image

### DIFF
--- a/lib/ex_doc/language/erlang.ex
+++ b/lib/ex_doc/language/erlang.ex
@@ -79,25 +79,21 @@ defmodule ExDoc.Language.Erlang do
     nil
   end
 
-  def autolink_spec(attribute, _opts) do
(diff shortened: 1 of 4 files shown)


lib/ex_doc/language/erlang.ex
```diff
@@ -272,6 +268,147 @@ defmodule ExDoc.Language.Erlang do
     "#t:#{name}/#{arity}"
   end
 
+  # Traverses quoted and formatted string of the typespec AST, replacing refs with links.
```
wojtekmach, line 271: the plan is to re-use this algorithm for Elixir in a separate PR.

josevalim (approved):

PR #1378, opened 2021-07-03
Add "Using ExDoc with Erlang projects" to README

wojtekmach:
Hopefully this will be made easier soon by tighter integration in rebar3/rebar3_hex.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ For example, here are some acceptable values:
 
 To use ExDoc with Erlang projects you need to do the following:
 
-1. Use OTP 24+
+1. Use Erlang/OTP 24+
 
 2. Add the following to your `rebar.config`. This instructs `edoc` to generate doc chunks
    instead of HTML docs:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -119,6 +119,51 @@ For example, here are some acceptable values:
     GITHUB_USER     => elixir-lang
     GITHUB_REPO     => ecto
 
+## Using ExDoc with Erlang projects
+
+To use ExDoc with Erlang projects you need to do the following:
+
+1. Use OTP 24+
```
josevalim, line 126:
```suggestion
1. Use Erlang/OTP 24+
```

README.md
```diff
@@ -119,6 +119,51 @@ For example, here are some acceptable values:
...
+
+   Replace `<app>` with the name of your app.
+
+3. Install ExDoc escript:
+
+   ```bash
+   $ mix escript.install github elixir-lang/ex_doc
+   ```
+
+4. Generate docs:
```
josevalim, line 146: We need to say that, if the escript path is not in their system paths, they will need to point to it directly.

PR #1386, opened 2021-08-25
Add support for LiveBook's `.livemd` Markdown files

samharnack:


josevalim:
Thank you @samharnack! One thing we want to do with .livemd files is to also include the original, so we can easily import them into Livebook. Can you please look into this? @wojtekmach should we also include the source .md files?

samharnack:
Will do! @josevalim do you have a preference/suggestion of where the source files should be copied or is the root fine?

test/ex_doc/formatter/html_test.exs
```diff
@@ -43,7 +43,8 @@ defmodule ExDoc.Formatter.HTMLTest do
         "test/fixtures/LICENSE",
         "test/fixtures/PlainText.txt",
         "test/fixtures/PlainTextFiles.md",
-        "test/fixtures/README.md"
+        "test/fixtures/README.md",
+        "test/fixtures/LiveBookFile.livemd"
```
wojtekmach, line 47:
```suggestion
        "test/fixtures/LivebookFile.livemd"
```

test/fixtures/LiveBookFile.livemd
```diff
@@ -0,0 +1,3 @@
+# LiveBook Files
```
wojtekmach, line 1:
```suggestion
# Livebook Files
```

wojtekmach:
I'm not sure I follow, in the context of livebook what is the difference between `.livemd` and `.md` and what do we need on the ExDoc side of things? My understanding is that when we work on our livebook, the file is persisted as `.livemd`. And when export, we are nudged towards calling it `.livemd` too.

josevalim:
For .livemd files, you want to import them into Livebook, so we need to
keep a copy of the original and link to it. For now, keeping the original
is fine until we figure out how to link it. Then the question is if we also
want to keep the originals for .md.

For keeping them, I think they can have the exact same path as the html
one, we just change the extension. WDYT?
-- 


*José Valimhttps://dashbit.co/ <https://dashbit.co/>*

samharnack:
IMO including the `.md` might get noisy so I've just copied the `.livemd` files. I'm happy to update the PR to include the `.md` files if thats the desired direction.

wojtekmach:
ok, I think I get it. Say we include `ecto.livemd` in Ecto docs. We already include for example the changelog. You are proposing to have all of these, right?

- https://hexdocs.pm/ecto/ecto.livemd
- https://hexdocs.pm/ecto/ecto.html
- https://hexdocs.pm/ecto/changelog.md
- https://hexdocs.pm/ecto/changelog.html

if so, sounds good to me.

lib/ex_doc/formatter/html.ex
```diff
@@ -203,27 +203,42 @@ defmodule ExDoc.Formatter.HTML do
...
+  end
 
-      if File.regular?(output) do
-        IO.puts(:stderr, "warning: file #{Path.relative_to_cwd(output)} already exists")
-      end
+  defp copy_extras(config, extras) do
+    for %{source_path: source_path} <- extras,
+        filename = Path.basename(source_path),
+        extension_name(filename) == ".livemd" do
+      output = "#{config.output}/#{filename}"
```
josevalim, line 236: I don't think we want the original filename, but we want the new .html file name. This is because we want to make it easy to find the .livemd from the HTML one. So I think this should be this:

```suggestion
    for %{source_path: source_path, id: id} <- extras,
        ext = extension_name(source_path),
        ext == ".livemd" do
      output = "#{config.output}/#{id}.#{ext}"
```

Maybe we should also have a test with a non-conventional .livemd filename?

josevalim:
Btw, for full disclosure, we will be soon launching a Livebook website and we want to add a "Run in Livebook" badge. Clicking on the badge will point to the .livemd file, which you will be able to preview and run in your favorite Livebook instance (it can be local or cloud provided). The idea is to include this badge on all Livebooks generated as part of ExDoc. :)

lib/ex_doc/formatter/html.ex
samharnack: Makes sense, I'll get the tests added. For clarity, when you say "non-conventional" do you mean with spaces and special characters?

josevalim: Yes, exactly!

Issue #1339, opened 2021-03-04
Sidebar gets expanded on windows resize


eksperimental:
If I collapse the sidebar and I resize the window, the sidebar gets expanded every time.

This can clearly be seen when the sidebar is collapsed and you choose to go Full screen mode and back to normal mode.

devtayls:
I have a pull request up for this issue:

https://github.com/elixir-lang/ex_doc/pull/1394

PR #1394, opened 2021-09-12
fix(sidebar): track user pref for sidebar state

devtayls:
**Issue:** https://github.com/elixir-lang/ex_doc/issues/1339

Resolves the issue with the state of the sidebar forcefully changing when the screen size updates. 

Now, the sidebar will track the preference of the user. 

When the preference is closed --> the sidebar will remain closed on window resize

When the preference is open (or no_pref) --> the sidebar will toggle according to screen width on resize

### DIFF
--- a/assets/js/sidebar/sidebar-drawer.js
+++ b/assets/js/sidebar/sidebar-drawer.js
@@ -7,6 +7,12 @@ const ANIMATION_DURATION = 300
 const SIDEBAR_TOGGLE_SELECTOR = '.sidebar-toggle'
 const CONTENT_SELECTOR = '.content'
 
+const userPref = {
+  CLOSED: 'closed',
+  OPEN: 'open',
+  NO_PREF: 'no_pref'
+}
+
 const SIDEBAR_CLASS = {
   opened: 'sidebar-opened',
   opening: 'sidebar-opening',
@@ -20,7 +26,9 @@ const state = {
   // Keep track of the current timeout to clear it if needed
   togglingTimeout: null,
   // Record window width on resize to update sidebar state only when it actually changes
-  lastWindowWidth: window.innerWidth
+  lastWindowWidth: window.innerWidth,
+  // No_PREF is defaults to OPEN behavior
+  sidebarPreference: userPref.NO_PREF
 }
 
 /**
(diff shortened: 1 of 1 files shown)


assets/js/sidebar/sidebar-drawer.js
```diff
@@ -135,3 +157,21 @@ function closeSidebarIfSmallScreen () {
...
+/**
+ * Track the sidebar preference for the user
+ */
+function setPreference () {
+  switch (state.sidebarPreference) {
+    case userPref.OPEN:
+      state.sidebarPreference = userPref.CLOSED
+      break
+    case userPref.CLOSED:
+      state.sidebarPreference = userPref.OPEN
```
josevalim, line 170: Is this correct? If the user preference is closed, we are setting it to open? 🤔

devtayls: The control logic is correct. The function is called when the `click` event tied to the toggle button fires. So whenever the user changes the state of the sidebar with the toggle button, we'll update their preference. 

> So, if you're preference _was_ closed we'll set it _open_
> 
> And if you're preference _was_ open, we'll set it to _closed_
> 
> And if you had _no_pref_, we'll set the pref to the state of the sidebar. 

I've attached a short video showing the state changes while working with the sidebar

https://user-images.githubusercontent.com/16793430/133001792-dbec7ca4-bf61-4c12-987b-7b5e0b4f7d60.MP4

devtayls: would it be better if I called it `togglePreference()` instead of `setPreference()`

josevalim: Beautiful video, thank you! 😍

josevalim:

assets/js/sidebar/sidebar-drawer.js
josevalim: > would it be better if I called it togglePreference() instead of setPreference()

Ah, good idea! I just merged the PR though, but feel free to submit an update. :)

devtayls: will do

devtayls:
Thanks @josevalim I really appreciate your hard work. :)

PR #1398, opened 2021-09-27
Do not check :private on visibility in ExDoc.Language.Erlang.do_url

wingyplus:
The dialyzer report on file erlang.ex in line 364 that
the variable visibility cannot be a :private because
typespec declared it as :hidden or :public or :undefined.
Fixes it by removing :private on visibility check in do_url
function.

### DIFF
--- a/lib/ex_doc/language/erlang.ex
+++ b/lib/ex_doc/language/erlang.ex
@@ -361,7 +361,7 @@ defmodule ExDoc.Language.Erlang do
 
     # TODO: type with content = %{} in otp xml is marked as :hidden, it should be :public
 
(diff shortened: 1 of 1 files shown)


wingyplus:
> 
> 
> The spec is correct, since there is not such a thing as a private visibility from the ExDoc point of view. What should be corrected is the `erlang.ex` file
> 
> It should be changed to:
> `if visibility == :public or (visibility == :hidden and elem(ref, 0) == :type) do`

Thanks for suggestion. If there's no `:private` in visibility, it's make senses to change `erlang.ex` to match only hidden. And do we need to

eksperimental:
Thank you @wingyplus

PR #1412, opened 2021-10-21
Add --warnings-as-errors flag for non-zero exit code

### LINKED ISSUE #1294: Feature request: Warnings as Errors
Hi,

it would be great to have a switch to treat warnings as errors. One could use that in automated builds.

Cheers, Jan Niklas

### LINKED ISSUE #1411: Support --strict flag that fails with non-zero exit code for warnings
To ensure docs don't get merged with invalid links or syntax that fails to parse it would be great to add support for a `--strict` flag that fails the `mix docs` task with a non-zero exit code so that way we can create a job in CI to indicate an issue with doc generation before it merges.

kdawgwilk:
Closes #1411 
Closes #1294

### DIFF
--- a/lib/ex_doc/application.ex
+++ b/lib/ex_doc/application.ex
@@ -17,6 +17,11 @@ defmodule ExDoc.Application do
         match?("makeup_" <> _, Atom.to_string(app)),
         do: Application.ensure_all_started(app)
 
-    Supervisor.start_link([ExDoc.Refs], strategy: :one_for_one)
+    children = [
+      ExDoc.Refs,
+      ExDoc.WarningCounter
+    ]
+
+    Supervisor.start_link(children, strategy: :one_for_one)
   end
 end
--- a/lib/ex_doc/cli.ex
+++ b/lib/ex_doc/cli.ex
@@ -29,14 +29,25 @@ defmodule ExDoc.CLI do
           source_ref: :string,
           version: :boolean,
           formatter: :keep,
-          quiet: :boolean
+          quiet: :boolean,
+          warnings_as_errors: :boolean
         ]
       )
 
-    if List.keymember?(opts, :version, 0) do
-      print_version()
-    else
-      generate(args, opts, generator)
+    cond do
+      List.keymember?(opts, :version, 0) ->
+        print_version()
+
+      opts[:warnings_as_errors] == true and ExDoc.WarningCounter.count() > 0 ->
+        IO.puts(
+          :stderr,
+          "Doc generation failed due to warnings while using the --warnings-as-errors option"
+        )
+
+        exit({:shutdown, ExDoc.WarningCounter.count()})
+
+      true ->
+        generate(args, opts, generator)
     end
   end
 
@@ -159,29 +170,30 @@ defmodule ExDoc.CLI do
       ex_doc "Project" "1.0.0" "_build/dev/lib/project/ebin" -c "docs.exs"
 
     Options:
-      PROJECT             Project name
-      VERSION             Version number
-      BEAMS               Path to compiled beam files
-      -n, --canonical     Indicate the preferred URL with rel="canonical" link element
-      -c, --config        Give configuration through a file instead of a command line.
-                          See "Custom config" section below for more information.
-      -f, --formatter     Docs formatter to use (html or epub), default: html and epub
-      -p, --homepage-url  URL to link to for the site name
-          --paths         Prepends the given path to Erlang code path. The path might contain a glob
-                          pattern but in that case, remember to quote it: --paths "_build/dev/lib/*/ebin".
-                          This option can be given multiple times
-          --language      Identify the primary language of the documents, its value must be
-                          a valid [BCP 47](https://tools.ietf.org/html/bcp47) language tag, default: "en"
-      -l, --logo          Path to the image logo of the project (only PNG or JPEG accepted)
-                          The image size will be 64x64 and copied to the assets directory
-      -m, --main          The entry-point page in docs, default: "api-reference"
-          --package       Hex package name
-          --proglang      The project's programming language, default: "elixir"
-          --source-ref    Branch/commit/tag used for source link inference, default: "master"
-      -u, --source-url    URL to the source code
-      -o, --output        Path to output docs, default: "doc"
-      -v, --version       Print ExDoc version
-      -q, --quiet         Only output warnings and errors
+      PROJECT              Project name
+      VERSION              Version number
+      BEAMS                Path to compiled beam files
+      -n, --canonical      Indicate the preferred URL with rel="canonical" link element
+      -c, --config         Give configuration through a file instead of a command line.
+                           See "Custom config" section below for more information.
+      -f, --formatter      Docs formatter to use (html or epub), default: html and epub
+      -p, --homepage-url   URL to link to for the site name
+          --paths          Prepends the given path to Erlang code path. The path might contain a glob
+                           pattern but in that case, remember to quote it: --paths "_build/dev/lib/*/ebin".
+                           This option can be given multiple times
+          --language       Identify the primary language of the documents, its value must be
+                           a valid [BCP 47](https://tools.ietf.org/html/bcp47) language tag, default: "en"
+      -l, --logo           Path to the image logo of the project (only PNG or JPEG accepted)
+                           The image size will be 64x64 and copied to the assets directory
+      -m, --main           The entry-point page in docs, default: "api-reference"
+          --package        Hex package name
+          --proglang       The project's programming language, default: "elixir"
+          --source-ref     Branch/commit/tag used for source link inference, default: "master"
+      -u, --source-url     URL to the source code
+      -o, --output         Path to output docs, default: "doc"
+      -v, --version        Print ExDoc version
+      -q, --quiet          Only output warnings and errors
+      --warnings-as-errors Exit with non-zero status if doc generation has one or more warnings
 
     ## Custom config
 
--- a/lib/ex_doc/warning_counter.ex
+++ b/lib/ex_doc/warning_counter.ex
@@ -1,0 +1,21 @@
+defmodule ExDoc.WarningCounter do
+  @moduledoc false
+
+  use Agent
+
+  def start_link(_opts) do
+    Agent.start_link(fn -> 0 end, name: __MODULE__)
+  end
+
+  def count do
+    Agent.get(__MODULE__, & &1)
+  end
+
+  def increment do
+    Agent.update(__MODULE__, &(&1 + 1))
+  end
+
+  def reset do
+    Agent.update(__MODULE__, fn _ -> 0 end)
+  end
+end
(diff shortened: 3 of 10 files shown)


kdawgwilk:
Are we okay with exiting after the first round of warnings are encountered?  Or do we want to spit out all of the warnings before we exit?  The easier approach would be to exit early but it may be frustrating to:
- run it
- see it fail
- fix warnings
- run it again
- see more warnings
- repeat
But due to this being a new feature having the early exit is still an improvement and could be improved later to show all warnings.

josevalim:
Ideally we would not fail early. We would have a process in ExDoc's supervision tree (it can be an Agent) which will store if warnings were emitted. If so, you abort at the end of the `--warnings-as-errors` flag is given. :)

lib/mix/tasks/docs.ex
```diff
@@ -349,6 +352,16 @@ defmodule Mix.Tasks.Docs do
     for formatter <- get_formatters(options) do
       index = generator.(project, version, Keyword.put(options, :formatter, formatter))
       Mix.shell().info([:green, "View #{inspect(formatter)} docs at #{inspect(index)}"])
+
+      if List.keymember?(options, :warnings_as_errors, 0) and ExDoc.WarningCounter.count() > 0 do
```
eksperimental, line 356: I think it is clearer in intention to use `Keyword.has_key?/2` since options is a keyword.

```suggestion
      if Keyword.has_key?(options, :warnings_as_errors) and ExDoc.WarningCounter.count() > 0 do
```

lib/ex_doc/cli.ex
```diff
@@ -35,6 +36,11 @@ defmodule ExDoc.CLI do
     else
       generate(args, opts, generator)
     end
+
+    if List.keymember?(opts, :warnings_as_errors, 0) and ExDoc.WarningCounter.count() > 0 do
```
wojtekmach, line 40: we should check for the value:

```suggestion
    if opts[:warnings_as_errors] == true and ExDoc.WarningCounter.count() > 0 do
```

because, though extremely unlikely, someone could set it to false and then it wouldn't work:

```elixir
iex> OptionParser.parse!(~w(--warnings-as-errors=false), strict: [warnings_as_errors: :boolean])
{[warnings_as_errors: false], []}
```

lib/ex_doc/application.ex
```diff
@@ -17,6 +17,6 @@ defmodule ExDoc.Application do
         match?("makeup_" <> _, Atom.to_string(app)),
         do: Application.ensure_all_started(app)
 
-    Supervisor.start_link([ExDoc.Refs], strategy: :one_for_one)
+    Supervisor.start_link([ExDoc.Refs, ExDoc.WarningCounter], strategy: :one_for_one)
```
wojtekmach, line 20:
```suggestion
    children = [
      ExDoc.Refs,
      ExDoc.WarningCounter
    ]

    Supervisor.start_link(children, strategy: :one_for_one)
```

lib/ex_doc/cli.ex
eksperimental, line 40: I think it is clearer in intention to use `Keyword.has_key?/2` since opts is a keyword.

```suggestion
    if Keyword.has_key?(opts, :warnings_as_errors) and ExDoc.WarningCounter.count() > 0 do
```

lib/ex_doc/warning_counter.ex
```diff
@@ -0,0 +1,19 @@
+defmodule ExDoc.WarningCounter do
+  @moduledoc false
+
+  @name __MODULE__
+
+  use Agent
+
+  def start_link(_ \\ []) do
+    Agent.start_link(fn -> 0 end, name: @name)
```
eksperimental, line 9:
```suggestion
  use Agent

  def start_link(_initial_value) do
    Agent.start_link(fn -> 0 end, name: __MODULE__)
```

lib/ex_doc/cli.ex
```diff
@@ -35,6 +36,11 @@ defmodule ExDoc.CLI do
...
+      IO.warn("Doc generation failed due to warnings while using the --warnings-as-errors option")
```
wojtekmach, line 41: I don't think it should be printed as a warning (`warning: `) so I'd consider just a simple IO.puts.

```suggestion
      IO.puts("Doc generation failed due to warnings while using the --warnings-as-errors option")
```

wojtekmach (commented):
Looks good to me, just a couple minor comments!

wojtekmach:
@kdawgwilk ping! I think this is pretty close, feel free to change from a draft to ready for review so we can merge it soon. (Of course no rush!)

kdawgwilk:
Struggling to figure out how to write some good tests for this, other than that I think it is mostly ready to go

kdawgwilk:
@wojtekmach I think I have enough tests now just having a hard time getting them to behave with the exit code, have 1 left that is still failing because there are no actual warnings where I am expecting a warning so I can assert it exits as expected.  Not sure how to use the fixtures to get a doc warning when I run from the tests.  After thats figured out this should be good to go

kdawgwilk:
Just figured it out now so this should be ready for review!

test/ex_doc/cli_test.exs
```diff
@@ -35,6 +35,41 @@ defmodule ExDoc.CLITest do
...
+    test "exits with 1 with warnings" do
+      ExDoc.WarningCounter.increment()
+
+      fun = fn ->
+        io =
+          capture_io(:stderr, fn ->
+            {[{"ExDoc", "1.2.3", opts}, _], _io} =
+              run(["ExDoc", "1.2.3", @ebin, "--warnings-as-errors"])
+
+            assert Keyword.get(opts, :warnings_as_errors)
```
Billzabob, line 60: I'm pretty sure that this test never actually hits this `assert` because of the `exit`. If I throw a `IO.puts` right above it, I never actually see it.

test/ex_doc/cli_test.exs
```diff
@@ -35,6 +35,41 @@ defmodule ExDoc.CLITest do
...
+          end)
+
+        assert io =~
```
Billzabob, line 63: Same thing with this assertion

test/ex_doc/cli_test.exs
kdawgwilk: Yeah your right 😞  I wonder how I can capture_io and the exit in the same go?

Billzabob: I'm not sure if there's a way to capture the output of a function that `exit`s. I believe this would be hitting [this guy](https://github.com/elixir-lang/elixir/blob/1e4ed32eda7957d469860e0143b5e5021538b477/lib/ex_unit/lib/ex_unit/capture_io.ex#L270-L282) which throws away the IO if the function throws.

milmazz:
@kdawgwilk This looks good to me. Do you need help with something in particular? or in your opinion are we good to go?

test/mix/tasks/docs_test.exs
```diff
@@ -181,4 +181,8 @@ defmodule Mix.Tasks.DocsTest do
       ] = run([], app: :umbrella, apps_path: "apps/", docs: [ignore_apps: [:foo]])
     end)
   end
+
+  test "accepts warnings_as_errors in :warnings_as_errors" do
+    assert catch_exit(run([], app: :ex_doc, docs: [warnings_as_errors: true])) == {:shutdown, 1}
```
milmazz, line 186: 🤔 Individually, this test should fail, it's assuming that the state of the `WarningCounter` is greater than zero when is not the case.

test/ex_doc/cli_test.exs
```diff
@@ -35,6 +35,41 @@ defmodule ExDoc.CLITest do
...
+               {"ExDoc", "1.2.3",
+                [formatter: "html", apps: [:ex_doc], source_beam: @ebin, warnings_as_errors: true]}
+
+      assert epub ==
+               {"ExDoc", "1.2.3",
+                [formatter: "epub", apps: [:ex_doc], source_beam: @ebin, warnings_as_errors: true]}
+    end
+
+    test "exits with 1 with warnings" do
+      ExDoc.WarningCounter.increment()
```
milmazz, line 52: This test is running in _async_ mode, and you're modifying a global state here that could affect other tests, which is what happening with `test/mix/tasks/docs_test.exs`

milmazz (changes requested):
Found a minor big in the _unit tests_. HTH

warmwaffles:
Would love to see this feature.

test/ex_doc/cli_test.exs
eksperimental: This works:

```elixir
      fun = fn ->
        run(["ExDoc", "1.2.3", @ebin, "--warnings-as-errors"])
      end

      io =
        capture_io(:stderr, fn ->
          assert catch_exit(fun.()) == {:shutdown, 1}
        end)

      assert io =~
               "Doc generation failed due to warnings while using the --warnings-as-errors option\n"
```

test/ex_doc/cli_test.exs
eksperimental: What would be the simplest way to solve this?
I have tried some but it requires a large refactoring.

test/ex_doc/cli_test.exs
kdawgwilk: @eksperimental Thanks! just pushed that up

test/ex_doc/cli_test.exs
kdawgwilk: If I do a `ExDoc.WarningCounter.reset()` on the tests that use this global state would that work?

test/ex_doc/cli_test.exs
eksperimental: I guess this is is already resolved.

test/ex_doc/cli_test.exs
eksperimental: Awesome! Let me/us know if you need help with anything.

test/ex_doc/cli_test.exs
eksperimental: > If I do a `ExDoc.WarningCounter.reset()` on the tests that use this global state would that work?

the issue is that since they are in separate files they could be executed at the same time.
If you execute reset in one file, it will be resetting the ones in the other tests. The only guarantee is that if you use `async: false` is that the tests within the same file are not going to be running concurrently.

kdawgwilk: Sounds like this is going to be more work than I was expecting to get added 😞  I started a new job and am no longer working with elixir day to day so I likely won't be finishing this up anymore.  If someone else wants to pick it up feel free.

kdawgwilk:
No longer working with Elixir day to day 😢  so I won't be able to prioritize this for awhile unfortunately

eksperimental:
> No longer working with Elixir day to day cry so I won't be able to prioritize this for awhile unfortunately

No worries @kdawgwilk, I can jump in from where you left off.

test/ex_doc/cli_test.exs
eksperimental: > Sounds like this is going to be more work than I was expecting to get added disappointed I started a new job and am no longer working with elixir day to day so I likely won't be finishing this up anymore. If someone else wants to pick it up feel free.

Sure. No worries. I was already working with your PR so I'm familiar with it.
Thanks a lot for the PR, it is almost there.

eksperimental:
So let's define how do we want to deal with the global states of the warnings.

Issue #1419, opened 2021-11-05
Generating documentation for 'private' modules


mat-hek:
Hello!

Currently, the standard way of documenting the code seems to be to add comments for private stuff and docs for public stuff. When it comes to private functions, it seems reasonable, as comments usually don't make much sense without code. Things get more complicated when a 'private' codebase grows within a project, new modules are created and naturally, there's an API between them. I think it would be beneficial to be able to explore that API just like the public one:
- It could help maintainers keep the private API clear and consistent
- It could help newcomers reason about the project and contribute
- It could help anyone find out the information flow within the project
- It could improve the process of adding/requesting a new public API
- The docs could be available in IEx/code suggestion tools

The downside of this approach could be that more people could potentially start relying on private APIs since they would be more accessible, so any solution should address that.

A way it can be (partially) solved now is to make the private API public, but move it to a separate [group](https://hexdocs.pm/ex_doc/Mix.Tasks.Docs.html#module-groups). There are several problems with that though:
- I haven't seen anyone doing it, while I believe we should have a standard for that across the community
- It doesn't show clearly enough that the API is private and may change at any time (especially in IEx). One would have to at least state that manually in each moduledoc
- Not sure how code suggestion tools handle private modules, but they don't even have a chance to do so in this case
- ExDoc won't warn when using/referring to a private function/type in the public API
- And probably more that I haven't thought of

If it seems reasonable, I thought that the moduledoc could support an option like `api: :private`, so we could have

```elixir
defmodule My.Module do
  @moduledoc """
  Blah blah
  """
  @moduledoc api: :private

  ...
```

I'm curious what you think about it

josevalim:
Hi @mat-hek! I added in the main branch support for a custom function in `filter_modules`. So now you can do something like this in your mix.exs:

    filter_modules: if System.get_env("EX_DOC_INCLUDE_PRIVATE") do
      fn _, _ -> true end
    else
      fn _module, metadata -> metadata[:api] != :private end
    end

Now, whenever you want to include the private modules, you can set the environment variable.

mat-hek:
Hi @josevalim, thanks. Would it be possible to make the metadata visible in the docs? Now when I set EX_DOC_INCLUDE_PRIVATE, there's no option to see which modules are private and which are not.

josevalim:
@mat-hek for now, you can combine that with `:group_for_modules`:

```elixir
group_for_modules: [
  Private: & &1[:api] == :private
]
```

[...]

PR #1430, opened 2021-11-12
Added optional module annotations

Qizot:
As a follow-up to this issue #1419. This PR adds a possibility to specify additional module's annotations that will be displayed right next to `View Source` icon. It can help to distinguish between different kinds of modules e.g. private ones when we decide to include them in documentation for internal purposes but make it obvious that they are private. In the end the annotation can be whatever we want.

I didn't know if I was to regenerate build files but I guess it happens during deployment?

I decided to go with the following pattern:

```elixir
defmodule InternalModule do
  @moduledoc """
  This documentation is for internal purposes...
  """
  @moduledoc annotations: :private
end
```

The example UI looks like this:
image

### DIFF
--- a/assets/less/content/general.less
+++ b/assets/less/content/general.less
@@ -75,6 +75,10 @@ a.livebook-badge {
   font-weight: normal;
 }
 
+h1 .note {
+  float: right;
+}
+
 blockquote {
   font-style: italic;
   margin: .5em 0;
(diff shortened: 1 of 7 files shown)


josevalim:
Thank you @Qizot! Can you please rename the metadata field to `:tags`? Internally we can still call it `:annotations` in `ExDoc.ModuleNode`, but I think from the user perspective `:tags` would be a better fit. Thank you!

Qizot:
Done. One thing though, should we put some documentation somewhere that this options is available? I saw in elixir's documentation that we just mention the existence of metadata in couple of tags like `@moudledoc` or `@doc` and some of its keys but nothing more.

josevalim:
Yeah, I was thinking the same, I will push some README docs soon after.

josevalim:

PR #1435, opened 2021-11-15
Document how to render Mermaid graphs

### LINKED ISSUE #1434: Trimmed whitespace in embedded HTML breaks Mermaid.js diagrams
I am trying to use [Mermaid.js](https://mermaid-js.github.io/mermaid/#/) to add state diagrams to my documentation. Why Mermaid.js? Mostly because it looks easier to maintain than what I'm currently doing with ASCII art and exporting PNGs from another tool. 

There is a good blog post on this at 
https://fireproofsocks.medium.com/enhancing-elixir-documentation-with-mermaid-charts-340fad5c6426. To summarize the post, adding a diagram to a `moduledoc` looks like this:

```elixir
@moduledoc """
My documentation goes here...
<div class="mermaid">
graph TD;
classDef server fill:#D0B441,stroke:#AD9121,stroke-width:1px;
classDef topic fill:#B5ADDF,stroke:#312378,stroke-width:1px;
classDef db fill:#9E74BE,stroke:#4E1C74,stroke-width:1px;T1(TopicA):::topic --> G1{{GenServerA}}:::server;
T1(TopicA):::topic --> G2{{GenServerB}}:::server;
G2{{GenServerB}}:::server --> T2(TopicB):::topic;
T2(TopicB):::topic ==> DB[("Storage#nbsp;")]:::db;
</div>
"""
```

Note the semicolons at the end of the line. They're needed since the generated documentation trims all whitespace in the `div` so the diagram specification is one long line. This almost works. It turns out that the Mermaid.js parser requires new lines on some state diagram statements (like `direction LR`). Check out [this example](https://mermaid-js.github.io/mermaid-live-editor/edit#eyJjb2RlIjoic3RhdGVEaWFncmFtLXYyXG4gICAgWypdIC0tPiBTdGlsbFxuICAgIFN0aWxsIC0tPiBbKl1cbiAgICBTdGlsbCAtLT4gTW92aW5nXG4gICAgTW92aW5nIC0tPiBTdGlsbFxuICAgIE1vdmluZyAtLT4gQ3Jhc2hcbiAgICBDcmFzaCAtLT4gWypdXG4gICAgICAgICAgICAiLCJtZXJtYWlkIjoie1xuICBcInRoZW1lXCI6IFwiZGVmYXVsdFwiXG59IiwidXBkYXRlRWRpdG9yIjpmYWxzZSwiYXV0b1N5bmMiOnRydWUsInVwZGF0ZURpYWdyYW0iOmZhbHNlfQ) for what a state diagram spec looks like.

One fix to this would be to turn off the code that trims whitespace in the embedded HTML. I could not find an option to do this. It naively seems like this should be an easy thing to turn off, but I didn't find the code from my brief skim through the ex_doc source.

Before I dig deeper, I was wondering if this was a reasonable approach or whether support for this already existed and it was just a matter of decorating the HTML properly.

jonatanklosko:
Closes #1434.

@fhunleth please give it a try and let me know if it works for you.

FTR initially I tried a slightly simpler solution, which was to first replace all `<pre>` tags with `<div class="mermaid">` and then run the automatic rendering:

```js
document.addEventListener("DOMContentLoaded", function () {
  for (const codeEl of document.querySelectorAll("pre code.mermaid")) {
    const preEl = codeEl.parentElement;
    const graphDefinition = codeEl.textContent;
    const graphEl = document.createElement("div");
    graphEl.classList.add("mermaid");
    graphEl.textContent = graphDefinition;
    preEl.insertAdjacentElement("afterend", graphEl);
    preEl.remove();
  }
  mermaid.initialize({ startOnLoad: true });
});
```

Then I tried a more manual approach, which I included in this PR. The primary difference is when the graph definition is invalid. When rendering manually we keep the definition in a `<pre>` tag so it can be read and there is an error in the console. On the other hand, when using the auto rendering the element gets replaced with a huge "failed to render" image and there's nothing useful in the console.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,35 @@ Other objects you may want to render in a special manner are code snippets. For 
 </script>
 ```
 
-For more details and configuration options see the [vega/vega-embed](https://github.com/vega/vega-embed).
+For more details and configuration options see [vega/vega-embed](https://github.com/vega/vega-embed).
+
+### Rendering Mermaid graphs
+
+Similarly to the example above, if your Markdown includes Mermaid graph specification in `mermaid` code snippets, you can do:
+
+```html
+<script src="https://cdn.jsdelivr.net/npm/mermaid@8.13.3/dist/mermaid.min.js"></script>
+<script>
+  document.addEventListener("DOMContentLoaded", function () {
+    mermaid.initialize({ startOnLoad: false });
+    let id = 0;
+    for (const codeEl of document.querySelectorAll("pre code.mermaid")) {
+      const preEl = codeEl.parentElement;
+      const graphDefinition = codeEl.textContent;
+      const graphEl = document.createElement("div");
+      const graphId = "mermaid-graph-" + id++;
+      mermaid.render(graphId, graphDefinition, function (svgSource, bindListeners) {
+        graphEl.innerHTML = svgSource;
+        bindListeners && bindListeners(graphEl);
+        preEl.insertAdjacentElement("afterend", graphEl);
+        preEl.remove();
+      });
+    }
+  });
+</script>
+```
+
+For more details and configuration options see the [Mermaid usage docs](https://mermaid-js.github.io/mermaid/#/usage).
 
 ## Contributing
 


josevalim:

PR #1445, opened 2021-12-06
add support for themed img

### LINKED ISSUE #1405: Some documentation assets don't look great in darkmode
Example: https://hexdocs.pm/floki/readme.html  see the logo in darkmode. Screenshot:
Screen Shot 2021-10-15 at 5 35 29 PM

I don't know what the best path here is. Either a nudge to authors that they should check their logos in darkmode or if ex_doc supported darkmode/daymode specific logos.

yordis:
closes #1405

### DIFF
--- a/assets/less/content.less
+++ b/assets/less/content.less
@@ -13,3 +13,7 @@
   @import './content/footer';
   @import './content/bottom-actions.less';
 }
+
+body:not(.night-mode) .content-inner img[src*="#gh-dark-mode-only"] {
+  display: none;
+}
--- a/assets/less/epub.less
+++ b/assets/less/epub.less
@@ -17,6 +17,10 @@ nav > ol {
   }
 }
 
+img[src*="#gh-dark-mode-only"] {
+  display: none;
+}
+
 .title-container {
   text-align: center;
 }
--- a/assets/less/night/content/general.less
+++ b/assets/less/night/content/general.less
@@ -33,3 +33,7 @@ a.view-source {
 .summary h2 a {
   text-decoration: none;
 }
+
+img[src*="#gh-light-mode-only"] {
+  display: none;
+}


yordis:
@josevalim how do you test these types of things locally? I pointed my project to use `ex_doc` with `path: "..."` config, but I may be doing something wrong since I don't see the code.

Is that the way to go?

jonatanklosko:
@yordis you can run `mix docs` directly in the `ex_doc` repo and then open `doc/index.html`. README is included in the docs, so you can temporarily put content there to verify if it is rendered as expected.

yordis:
@jonatanklosko yeah, I think I figured it out, it works

https://user-images.githubusercontent.com/4237280/144942291-4591876c-60aa-46f3-afa2-dcb0be0b8278.mov

yordis:
Last-mile, dealing with the `epub`

https://user-images.githubusercontent.com/4237280/144942523-a99c50f3-5ad2-4080-8cee-d8bb1d840d21.mov

assets/less/content/general.less
```diff
@@ -165,6 +165,10 @@ img {
   max-width: 100%;
 }
 
+img[src*="#gh-dark-mode-only"] {
```
jonatanklosko, line 168: Alternatively we could do:

```
body:not(.night-mode) img[src*="#gh-dark-mode-only"] {
  display: none;
}
```

This way we don't need to explicitly do `display: inline` to revert this, which is good because the image may potentially have different `display`.

Maybe it would make sense to have both of these together as `body:not(.night-mode) img` and `body.night-mode img` in a single file?

yordis: > Maybe it would make sense to have both of these together as body:not(.night-mode) img and body.night-mode img in a single file?

I am following the existing pattern, worth noting that this is only inside the `.content-inner`

josevalim:
For EPUB we should always hide the dark mode one.

yordis:
@josevalim is there a CSS or something? Sorry, I actually don't know how to do that, I googled about it, but I am getting lost the more I click links.

jonatanklosko:
@yordis there are separate styles for EPUB in `assets/less/epub.less` :)

yordis:
Thank you all (specially you @jonatanklosko ❤️ ) ready to CR

assets/less/epub.less
```diff
@@ -17,6 +17,10 @@ nav > ol {
   }
 }
 
+img[src*="#gh-light-mode-only"] {
```
jonatanklosko, line 20: Shouldn't we keep the light one for EPUB and hide the dark one instead?

yordis: 🤦🏻 ...

assets/less/content.less
```diff
@@ -13,3 +13,7 @@
   @import './content/footer';
   @import './content/bottom-actions.less';
 }
+
+body:not(.night-mode) .content-inner img[src*="#gh-dark-mode-only"] {
```
jonatanklosko, line 17: Actually now I wonder if we can do it in `general.less` with a fancy selector like `body:not(.night-mode) & img[src*="#gh-dark-mode-only"]`, but either is fine ^^

jonatanklosko (approved):
Looks good to me!

josevalim:

PR #1447, opened 2021-12-18
Use config.proglang in epub head template

starbelly:
This PR fixes a bug where the head template for epub had elixir hard coded for the style sheet resulting in a crash if the proglang differed (i.e., erlang).


test/ex_doc/formatter/epub_test.exs
```diff
@@ -91,6 +91,11 @@ defmodule ExDoc.Formatter.EPUBTest do
   test "generates an EPUB file in the default directory" do
     generate_docs(doc_config())
     assert File.regular?("#{output_dir()}/#{doc_config()[:project]}.epub")
+
+    File.rm!("#{output_dir()}/#{doc_config()[:project]}.epub")
+
```
josevalim, line 96: Can we use a separate test instead? :D

starbelly: Sure :)

josevalim:

PR #1449, opened 2021-12-20
Add --quiet switch to ExDoc.CLI

starbelly:


test/ex_doc/cli_test.exs
```diff
@@ -150,4 +149,27 @@ defmodule ExDoc.CLITest do
       File.rm!("test.config")
     end
   end
+
+  describe "verborsity" do
+    test "without --quiet" do
```
starbelly, line 154: Hmm, run is not working as I expected to, bad test. Maybe we don't even want this test 🤔

josevalim:

PR #1450, opened 2021-12-21
Remove deprecated timeout option for EarmarkParser

starbelly:


josevalim:
I would keep it for a while just so we don't impose a very recent earmark version on users. :)

starbelly:
> I would keep it for a while just so we don't impose a very recent earmark version on users. :)

That's fair. Closing this then.

eksperimental:
> > I would keep it for a while just so we don't impose a very recent earmark version on users. :)
> 
> That's fair. Closing this then.

Maybe add a PR with a TODO comment to remove it when certain version of EarMark is required.

starbelly:
> > > I would keep it for a while just so we don't impose a very recent earmark version on users. :)
> > 
> > 
> > That's fair. Closing this then.
> 
> Maybe add a PR with a TODO comment to remove it when certain version of EarMark is required. If the PR can include the source that documents that the feature has been deprecated it will be helpful too.

Will do.

PR #1453, opened 2021-12-29
Update README version to 0.26

shahryarjb:


josevalim:
That version will get the latest 0.26, 0.27, and so on. So it is fine to keep it as is and then we don't have to update always!

eksperimental:
@josevalim I am not getting ExDoc updated.

```
$ mix deps.get
Resolving Hex dependencies...
Dependency resolution completed:
Unchanged:
  ...
  ex_doc 0.26.0
  ...
All dependencies are up to date
```

while latest ex_doc released version is `0.27.3` and my deps config in mix.exs is `{:ex_doc, "~> 0.26", only: :dev, runtime: false}`.
What could it be?

josevalim:
It has to be `mix deps.update ex_doc`.

eksperimental:
Jeez, how could I missed that! Thank you

Issue #1438, opened 2021-11-22
Trap focus on settings modal


LostKobrakai:
The newly introduced settings modal doesn't trap focus, which would be best practice for accessibility.

josevalim:
Can you please send a PR? :)

josevalim:
Ping. :)

LostKobrakai:
I'll try to get to it at the end of the week :)

PR #1454, opened 2021-12-30
Implement focus trapping on modals

LostKobrakai:
Closes https://github.com/elixir-lang/ex_doc/issues/1438

While implementing I noticed that a lot of buttons/links in ex_doc lack proper focus styling. Should this be another issue?

### DIFF
--- a/assets/js/modal.js
+++ b/assets/js/modal.js
@@ -43,15 +43,16 @@ function renderModal () {
  */
 function trapFocus (event) {
   if (state.ignoreFocusChanges) return
-  var modal = qs(MODAL_SELECTOR)
-  console.log(event.target)
+  const modal = qs(MODAL_SELECTOR)
   if (modal.contains(event.target)) {
     state.lastFocus = event.target
   } else {
     state.ignoreFocusChanges = true
-    firstFocusableDescendant(modal).focus()
-    if (state.lastFocus == document.activeElement) {
+    const firstFocusable = firstFocusableDescendant(modal)
+    if (state.lastFocus === firstFocusable) {
       lastFocusableDescendant(modal).focus()
+    } else {
+      firstFocusable.focus()
     }
     state.ignoreFocusChanges = false
     state.lastFocus = document.activeElement
@@ -73,7 +74,7 @@ function lastFocusableDescendant(element) {
  * @param {{ title: String, body: String }} attrs
  */
 export function openModal ({ title, body }) {
-  state.prevFocus = typeof document !== 'undefined' && document.activeElement
+  state.prevFocus = document.activeElement
   document.addEventListener('focus', trapFocus, true)
 
   qs(MODAL_TITLE_SELECTOR).innerHTML = title
--- a/assets/js/modal.js
+++ b/assets/js/modal.js
@@ -5,7 +5,11 @@ const MODAL_SELECTOR = '#modal'
 const MODAL_CLOSE_BUTTON_SELECTOR = '#modal .modal-close'
 const MODAL_TITLE_SELECTOR = '#modal .modal-title'
 const MODAL_BODY_SELECTOR = '#modal .modal-body'
-let prevFocus = null
+const state = {
+  prevFocus: null,
+  lastFocus: null,
+  ignoreFocusChanges: false
+} 
 
 /**
  * Initializes modal layout.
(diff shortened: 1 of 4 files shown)


assets/js/modal.js
```diff
@@ -31,12 +32,36 @@ function renderModal () {
   })
 }
 
+/**
+ * Trap focus in modal
+ * Only called on open modals
+ */
+function trapFocus (e) {
```
jonatanklosko, line 39: I think a more reliable way would be to let the focus work normally, but have a `focus` listener that moves the focus back to the modal once it falls outside of it. [This page](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html) looks like a reference example and the interesting bit is [here](https://github.com/w3c/aria-practices/blob/3dd5a2d3af89adc1249bc76106eefce0e44a5de5/examples/dialog-modal/js/dialog.js#L288-L310) :)

assets/js/modal.js
```diff
@@ -5,6 +5,7 @@ const MODAL_SELECTOR = '#modal'
 const MODAL_CLOSE_BUTTON_SELECTOR = '#modal .modal-close'
 const MODAL_TITLE_SELECTOR = '#modal .modal-title'
 const MODAL_BODY_SELECTOR = '#modal .modal-body'
+let prevFocus = null
```
jonatanklosko, line 8: Just a convention, but it's neat to group the mutable state in a single object, as in:

https://github.com/elixir-lang/ex_doc/blob/f5c0304f81c914c89ac6920c0c0d4110bbc6b5db/assets/js/sidebar/sidebar-drawer.js#L25-L32

assets/js/modal.js
LostKobrakai: I tried this, but even though it looks more bullet prove it seems to be broken on safari (15). For some reason `event.target` on the `focus` event seems to always point to `.search-input`.

jonatanklosko: Interesting, does the example from the original page work for you in safari or same problem?

LostKobrakai: The previous implementation does indeed work, but I can surely see the issue in trying to reimplement focus handling.

jonatanklosko: I meant the ["Add Delivery Address
" example](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html#ex_label) :D

assets/js/modal.js
```diff
@@ -31,12 +36,85 @@ function renderModal () {
...
+ * Trap focus in modal
+ * Only called on open modals
+ */
+function trapFocus (event) {
+  if (state.ignoreFocusChanges) return
+  var modal = qs(MODAL_SELECTOR)
+  if (modal.contains(event.target)) {
+    state.lastFocus = event.target
+  } else {
+    focusFirstDescendant(modal)
```
jonatanklosko, line 49: I believe we could simplify this a bit by finding focusable elements on our own, something like this:

```js
const FOCUSABLE_SELECTOR = 'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'

function firstFocusableDescendant(element) {
  return element.querySelector(FOCUSABLE_SELECTOR)
}

function lastFocusableDescendant(element) {
  const elements = element.querySelectorAll(FOCUSABLE_SELECTOR)
  return elements[elements.length - 1]
}
```

assets/js/modal.js
LostKobrakai: I can't reproduce it on the "Add deliver address" modal, but on the second level modal when clicking on "Verify address, where focus jumps between the first and the last targetable element. Possibly because the main page does not include form elements on its own.

jonatanklosko: So in our case if you manually focus an element (let's say the checkbox), then hitting Tab moves the focus to the search form?

LostKobrakai: It at least tries to (`event.target` points to the search), then the event handler comes in and either moves the focus to the first or last focusable element in the modal (the close button or the live book url input).

jonatanklosko: Could it be [this](https://stackoverflow.com/a/1914496)?

LostKobrakai: 😩 that's it. In the German translation it's even worse because it sounds more about being about highlighting things, rather than targeting. Sorry for the noice, but I'm not using safari again for more than a few days.

jonatanklosko: No worries, thanks for confirming! Now we know what causes this behaviour and I think it's fine to stick to the current approach :)

assets/js/modal.js
LostKobrakai: This is better than iterating childNodes, but I guess the `attemptFocus` would still be useful when things are hidden via css, but still targetable (see the fix I made to the hidden liveview url input).

LostKobrakai: I guess we can ignore that case given we control when things are hidden. I'll change it to your proposed implementation.

jonatanklosko: We could get all focusable elements and find the first for which `attemptFocus` succeeds (and if that's the last element we would reverse the elements and do the same). But going the simpler route sounds fine as you said!

assets/js/modal.js
```diff
@@ -31,12 +37,45 @@ function renderModal () {
...
+  var modal = qs(MODAL_SELECTOR)
+  console.log(event.target)
+  if (modal.contains(event.target)) {
+    state.lastFocus = event.target
+  } else {
+    state.ignoreFocusChanges = true
+    firstFocusableDescendant(modal).focus()
+    if (state.lastFocus == document.activeElement) {
+      lastFocusableDescendant(modal).focus()
+    }
```
jonatanklosko, line 55:
```suggestion
    const firstFocusable = firstFocusableDescendant(modal)
    if (state.lastFocus === firstFocusable) {
      lastFocusableDescendant(modal).focus()
    } else {
      firstFocusable.focus()
    }
```

assets/js/modal.js
```diff
@@ -31,12 +37,45 @@ function renderModal () {
...
 }
 
+/**
+ * Trap focus in modal
+ * Only called on open modals
+ */
+function trapFocus (event) {
+  if (state.ignoreFocusChanges) return
+  var modal = qs(MODAL_SELECTOR)
+  console.log(event.target)
```
jonatanklosko, line 47:
```suggestion

```

assets/js/modal.js
```diff
@@ -31,12 +37,45 @@ function renderModal () {
   })
 }
 
+/**
+ * Trap focus in modal
+ * Only called on open modals
+ */
+function trapFocus (event) {
+  if (state.ignoreFocusChanges) return
+  var modal = qs(MODAL_SELECTOR)
```
jonatanklosko, line 46:
```suggestion
  const modal = qs(MODAL_SELECTOR)
```

assets/js/modal.js
```diff
@@ -31,12 +37,45 @@ function renderModal () {
...
+  return elements[elements.length - 1]
+}
+
 /**
  * Shows modal with the given content.
  *
  * @param {{ title: String, body: String }} attrs
  */
 export function openModal ({ title, body }) {
+  state.prevFocus = typeof document !== 'undefined' && document.activeElement
```
jonatanklosko, line 76: Any specific reason for the document check?

jonatanklosko:
A couple nitpicks, other than that looks great!

jonatanklosko (approved):
Thanks :ship:

josevalim:

Issue #602, opened 2016-09-21
Remove or use normal font for the "Specs" heading


josevalim:
Today it is just repetitive and it gives the impression the current section, including the notes about the function definition, are all part of a section named "Specs".

josevalim:
@eksperimental should we remove it or should we use normal font?

eksperimental:
If we remove it we should do something like we do with the functions that has the icon `</>`, and I 'm not sure if it will be clear that those are the specs.

If we leave it, we need to change the font, as that font in normal weight looks bad.

Do we have other examples of documentation where the specs are shown. i would like to see how other people are tackling this issue

josevalim:
> If we leave it, we need to change the font, as that font in normal weight looks bad.

We would use the same found used in the regular text. That's what Erlang docs do too.

josevalim:
> Do we have other examples of documentation where the specs are shown. i would like to see how other people are tackling this issue

http://erlang.org/doc/man/lists.html#all-2

eksperimental:
Playing with it, i think it's better to get rid of the colored background, 
![ex_doc_specs](https://cloud.githubusercontent.com/assets/9133420/18755224/86d0b4a2-8115-11e6-844f-498260207dd2.png)

"Specs" uses same font like the spec code.

(it's just a sketch, i'm horrible with design) :-p

josevalim:
@eksperimental I think Specs should either use the same font as all other headings (how it works today) OR it should be the regular text font. I don't think we should introduce yet another font/demarcation/heading. Removing the background also looks neat to me but I will delegate the decision to @dignifiedquire :)

eksperimental:
Regular font, looks wrong (size too big)

this is regular font, size: 0.90em
![image](https://cloud.githubusercontent.com/assets/9133420/18755803/bb614a5e-8117-11e6-9e9f-295900555736.png)

HEADING FONT (bold)
![image](https://cloud.githubusercontent.com/assets/9133420/18755993/63c3480a-8118-11e6-8e6d-f0919293cea5.png)

HEADING FONT (normal weight)
![image](https://cloud.githubusercontent.com/assets/9133420/18756030/8b1a5e66-8118-11e6-94f1-61e4bf0caace.png)

I would go with the last one.

bbugh:
[...]

PR #1455, opened 2022-01-03
Replace Specs heading

Odaeus:
See issue #602.

As well as replacing the heading I removed the padding as it seemed excessive, but I'm happy to restore that if you like.

Original:
CleanShot 2022-01-03 at 18 44 22

New (without padding):
CleanShot 2022-01-03 at 18 45 13
--
CleanShot 2022-01-03 at 18 44 57

With padding:
CleanShot 2022-01-03 at 18 46 22
--
CleanShot 2022-01-03 at 18 44 41

### DIFF
--- a/lib/ex_doc/formatter/html/templates.ex
+++ b/lib/ex_doc/formatter/html/templates.ex
@@ -33,19 +33,23 @@ defmodule ExDoc.Formatter.HTML.Templates do
   @doc """
   Get the attribute string used to define the spec of the given `node`.
   """
-  def get_spec_attribute(node)
+  def get_spec_attribute(module, node) do
+    node
+    |> get_spec_attribute()
+    |> module.language.format_attribute()
+  end
 
   def get_spec_attribute(%ExDoc.TypeNode{} = node) do
     if "opaque" in node.annotations do
-      "@opaque"
+      "opaque"
     else
-      "@type"
+      "type"
     end
   end
 
-  def get_spec_attribute(%ExDoc.FunctionNode{type: :callback}), do: "@callback"
-  def get_spec_attribute(%ExDoc.FunctionNode{type: :macrocallback}), do: "@macrocallback"
-  def get_spec_attribute(%ExDoc.FunctionNode{}), do: "@spec"
+  def get_spec_attribute(%ExDoc.FunctionNode{type: :callback}), do: "callback"
+  def get_spec_attribute(%ExDoc.FunctionNode{type: :macrocallback}), do: "macrocallback"
+  def get_spec_attribute(%ExDoc.FunctionNode{}), do: "spec"
 
   @doc """
   Get defaults clauses.
@@ -328,7 +332,7 @@ defmodule ExDoc.Formatter.HTML.Templates do
   end
 
   templates = [
-    detail_template: [:node, :_module],
+    detail_template: [:node, :module],
     footer_template: [:config, :node],
     head_template: [:config, :page],
     module_template: [:config, :module, :summary, :nodes_map],
--- a/lib/ex_doc/formatter/html/templates.ex
+++ b/lib/ex_doc/formatter/html/templates.ex
@@ -31,25 +31,11 @@ defmodule ExDoc.Formatter.HTML.Templates do
   end
 
   @doc """
-  Get the attribute string used to define the spec of the given `node`.
+  Format the attribute type used to define the spec of the given `node`.
   """
-  def get_spec_attribute(module, node) do
-    node
-    |> get_spec_attribute()
-    |> module.language.format_attribute()
+  def format_spec_attribute(module, node) do
+    module.language.format_spec_attribute(node)
   end
-
-  def get_spec_attribute(%ExDoc.TypeNode{} = node) do
-    if "opaque" in node.annotations do
-      "opaque"
-    else
-      "type"
-    end
-  end
-
-  def get_spec_attribute(%ExDoc.FunctionNode{type: :callback}), do: "callback"
-  def get_spec_attribute(%ExDoc.FunctionNode{type: :macrocallback}), do: "macrocallback"
-  def get_spec_attribute(%ExDoc.FunctionNode{}), do: "spec"
 
   @doc """
   Get defaults clauses.
--- a/lib/ex_doc/formatter/html/templates.ex
+++ b/lib/ex_doc/formatter/html/templates.ex
@@ -31,6 +31,23 @@ defmodule ExDoc.Formatter.HTML.Templates do
   end
 
   @doc """
+  Get the attribute string used to define the spec of the given `node`.
+  """
+  def get_spec_attribute(node)
+
+  def get_spec_attribute(%ExDoc.TypeNode{} = node) do
+    if "opaque" in node.annotations do
+      "@opaque"
+    else
+      "@type"
+    end
+  end
+
+  def get_spec_attribute(%ExDoc.FunctionNode{type: :callback}), do: "@callback"
+  def get_spec_attribute(%ExDoc.FunctionNode{type: :macrocallback}), do: "@macrocallback"
+  def get_spec_attribute(%ExDoc.FunctionNode{}), do: "@spec"
+
+  @doc """
   Get defaults clauses.
   """
   def get_defaults(%{defaults: defaults}) do
--- a/lib/ex_doc/language/erlang.ex
+++ b/lib/ex_doc/language/erlang.ex
@@ -163,6 +163,9 @@ defmodule ExDoc.Language.Erlang do
     }
   end
 
+  @impl true
+  def format_attribute(attribute) when is_binary(attribute), do: "-" <> attribute
+
   ## Shared between Erlang & Elixir
 
   @doc false
--- a/lib/ex_doc/language/erlang.ex
+++ b/lib/ex_doc/language/erlang.ex
@@ -164,7 +164,9 @@ defmodule ExDoc.Language.Erlang do
   end
 
   @impl true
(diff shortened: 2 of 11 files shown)


wojtekmach:
Types should be rendered with `@type`. If missing, please also add a test how opaque type is rendered.

eksperimental:
I prefer the one with the padding. It makes it stand out from the rest of the docs that follow.

josevalim:
No padding unless we are adding border around it, to make consistent elsewhere.

eksperimental:
> No padding unless we are adding border around it, to make consistent elsewhere.

Note that the current specs are padded and have no border around them.

eksperimental:
I'm fine with either way.

josevalim:
Sorry, I was not clear enough. The previous had the header, to give some context about it. If we pad it now, it will look like code, but not quite like code, and not enough context around it. Although I think making it look like code is worse? 🤔

Odaeus:
@wojtekmach good point, thanks!

Should opaque types be rendered with `@opaque` or `@type`?
CleanShot 2022-01-03 at 21 14 02@2x

I only need to consider `@type`, `@spec`, and `@opaque` right?

josevalim:
There are also `@callback`s and `@macrocallback`s, for example: https://hexdocs.pm/ecto/Ecto.Repo.html

Odaeus:
I've added support for all the mentioned attribute strings (and `@opaque`) in a new commit (which can be squashed if wanted).

Callbacks:
CleanShot 2022-01-03 at 23 40 56@2x

josevalim:
Oh, we need to check how those look for Erlang. I think `mix docs —language erlang` would generate Erlang ones. And we need to see if the gray color works well on Dark Mode (press `n` to enable it).

Odaeus:
The contrast ratio on the attribute in night mode was too low (using Chromium's built-in analyser) so I've set it to a night mode gray and it's happy now. (Thank you to whoever set up these vars!)

Before:
CleanShot 2022-01-04 at 17 09 23

After:
CleanShot 2022-01-04 at 17 24 25

I need a pointer on how to generate Erlang examples please. @josevalim 's suggestion of `mix docs --language erlang` doesn't seem to do anything (`--language` flag is for human languages?). I see there are some tmp test directories created by the `ExDoc.Retriever.ErlangTest` but they are empty unlike the Elixir tmp outputs.

wojtekmach:
besides what José said, if you're generating the docs with the escript, `ex_doc ... --proglang erlang` would work too.

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -30,6 +30,23 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
+  @doc """
+  Get the attribute string used to define the spec of the given `node`.
+  """
+  def get_spec_attribute(node)
+
+  def get_spec_attribute(%ExDoc.TypeNode{} = node) do
+    if "opaque" in node.annotations do
+      "@opaque"
+    else
+      "@type"
```
wojtekmach, line 42: `@type` and friends is Elixir specific which means it won't be appropriate for e.g. Erlang. We try to keep language-specific functionality in the corresponding `ExDoc.Language.*` module so let's see if we can make the change there.

Odaeus:
Thanks for the config setting, I didn't realise it was working because the output looks identical... except with red highlights instead of blue! I will work on adjusting the output for Erlang support.

Odaeus:
I've added a new commit on top so you can see the extraction I made to support both languages as suggested. But I'm lost on how to demonstrate or test the Erlang side. I'm not sure what effect `proglang` actually has given all the code `mix build` has is still Elixir, so it still outputs "@spec" and so. I wanted to add tests but I can't find any that test Erlang output, only the "autolinker" and AST generation. I feel like generating new Erlang output tests is out of scope for this change, but am I missing something?

wojtekmach:
> I wanted to add tests but I can't find any that test Erlang output, only the "autolinker" and AST generation.

Yup. I just added a basic test here: https://github.com/elixir-lang/ex_doc/pull/1514/files. Once this lands, could you rebase your PR on top and adjust the test accordingly? Thanks!

cc @josevalim

wojtekmach:
btw, `proglang` is a bit of misnomer. On a per-module basis we ignore that value, instead we render the docs for a given module based on the language the module is compiled with. `proglang` is really about the style we render the whole docs with.

Odaeus:
Thanks so much @wojtekmach! I approached trying to do the same thing a few times but kept getting stuck.

wojtekmach:
the PR landed, please rebase. Thank you for working on this! <3

Odaeus:
@wojtekmach I've rebased and amended your [new test slightly](https://github.com/elixir-lang/ex_doc/pull/1455/files#diff-ecc03effb0b4e2511eb39d41e8548c458dc2743eb975b3114a02338298bab0c3R29) to test the new output.

I hit an [issue](https://github.com/elixir-lang/ex_doc/issues/1515) with running the tests but it seems isolated to my system.

wojtekmach:
@Odaeus thanks, fixed on main!

test/ex_doc/formatter/html/erlang_test.exs
```diff
@@ -24,9 +24,9 @@ defmodule ExDoc.Formatter.HTML.ErlangTest do
 
     doc = generate_docs(c)
 
-    assert [_] = Floki.find(doc, "pre:fl-contains('foo(atom()) -> atom().')")
+    assert "-spec foo(atom()) -> atom()." = doc |> Floki.find("pre:fl-contains('foo(atom())')") |> Floki.text()
```
wojtekmach, line 27: nice!

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -30,6 +30,27 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
+  def get_spec_attribute(%ExDoc.TypeNode{} = node) do
+    if "opaque" in node.annotations do
+      "opaque"
+    else
+      "type"
+    end
+  end
+
+  def get_spec_attribute(%ExDoc.FunctionNode{type: :callback}), do: "callback"
+  def get_spec_attribute(%ExDoc.FunctionNode{type: :macrocallback}), do: "macrocallback"
```
wojtekmach, line 51: macrocallback is Elixir specific so ideally it wouldn't be in a language-agnostic module like the html formatter.

The new `format_attribute` callback make sense within the existing code structure where we store the spec ast _without_ the _kind_ (kind being basically `@type`, `-type`, etc). However, if we change the internals to do store the kind in ast, we wouldn't need the callback. In fact if we do so, we wouldn't need the logic that trims the kind in Erlang: https://github.com/elixir-lang/ex_doc/blob/main/lib/ex_doc/language/erlang.ex#L588:L589.

Odaeus: The best I can do so far is to move these functions into the language modules so that `get_spec_attribute(module, node)` becomes:

```elixir
def get_spec_attribute(module, node) do
  module.language.get_spec_attribute(node)
end
```

And those respective functions return either "@type" or "-type" as appropriate. With `:macrocallback` not present in the Erlang module of course. Let me know if I should update the PR with this change.

I looked into including the original text in the AST but kept hitting issues, so would need some hints for that. I'm not sure how you would want to keep the kind from Erlang while still having it separated so that the output can be wrapped in the formatting span element?

wojtekmach: > I'm not sure how you would want to keep the kind from Erlang while still having it separated so that the output can be wrapped in the formatting span element?

yeah, good point. I didn't realise we need the kind separate so we can render it with appropriate css class.

Your snippet looks good, I'd just rename it to `format_spec_attribute` (both function & callback)

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -30,6 +30,27 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
+    |> get_spec_attribute()
+    |> module.language.format_attribute()
+  end
+
+  def get_spec_attribute(%ExDoc.TypeNode{} = node) do
+    if "opaque" in node.annotations do
+      "opaque"
+    else
+      "type"
+    end
```
wojtekmach, line 47: I believe this should work, wdyt?

```suggestion
    Atom.to_string(node.type)
```

Odaeus: Huh, funny, I could have sworn I got `type: :type` previously. Yes this works!

Odaeus:
@wojtekmach updates made with a commit on top (22ab1c9a64c33bc945a8774362f5d76a6ad0df59). Let me know if I should squash away the in-between commits.

lib/ex_doc/language/erlang.ex
```diff
@@ -163,6 +163,18 @@ defmodule ExDoc.Language.Erlang do
...
 
+  @impl true
+  def format_spec_attribute(%ExDoc.TypeNode{} = node) do
+    if "opaque" in node.annotations do
+      "-opaque"
+    else
+      "-type"
+    end
+  end
+
```
wojtekmach, line 174:
```suggestion
  def format_spec_attribute(%ExDoc.TypeNode{type: type}), do: "-#{type}"
```

wojtekmach (approved):
LGTM, just one tiny comment left.

wojtekmach:
No need to squash yourself, we'll do it when merging.

Odaeus:
Just checking in to make sure you're not waiting on me for anything that I've forgotten. 😄

wojtekmach (approved):

wojtekmach:
Thank you!

PR #1456, opened 2022-01-05
Redesign sidebar

cristineguadelupe:
Initial sidebar redesign including a new header and a few navigation improvements

**New dark mode**
Screen Shot 2022-01-05 at 16 02 18

**New light mode**
Screen Shot 2022-01-05 at 16 01 49

**Erlang versions**
Screen Shot 2022-01-05 at 15 59 08

Screen Shot 2022-01-05 at 15 59 29


eksperimental:
Hi @cristineguadelupe,
welcome and thank you for your contributions.
Could you please share the same screenshots for the current version of ExDoc, so it is easier for us to see what changes. 
Thanks a lot for the PR.

eksperimental:
> The current version can be found here: https://hexdocs.pm/ecto/Ecto.html

Yes I know. But due to Operating system, web browser, screen size (which I don't share 3 out of 3 from the screenshot taken) it is hard to compare what changes due to PR and what changes due to the system's setup.

cristineguadelupe:
Hey @eksperimental! Thank you!
I have updated the main comment with screenshots showing the new color palette.

Before
Screen Shot 2022-01-05 at 16 13 56

After
Screen Shot 2022-01-05 at 16 14 04

assets/less/icons.less
```diff
@@ -9,6 +9,17 @@
   font-style: normal;
 }
 
+@font-face {
+  font-family: 'remixicon';
```
jonatanklosko, line 13: We could use the `remixicon` npm package and import the corresponding .css file, less files to maintain :)

Sidenote, we currently load fonts form CDN, but I guess for offline usage it's ok if the font doesn't load, while we still want the icons to be there.

josevalim: Would npm generate files with only the icons we use?

jonatanklosko: Nah, it's just that we wouldn't need to embed the files manually.

cristineguadelupe: The remix is quite large. I think that for our case where we use a few icons, doing it manually fits us well.

jonatanklosko: Oh you're including only the subset, ignore me then!

josevalim:
I will go ahead and merge this. To make sure we are all on the same page, the following work will happen in the future:

1. We will remove icomoon font in favor of remix icon everywhere
2. We will change the navigation so clicking on the sidebar items navigates by default
3. We will unify all items in the footer under a settings page

josevalim:

jonatanklosko:
Absolutely fantastic! :shipit:

PR #1467, opened 2022-01-10
Fix navigation

cristineguadelupe:


assets/less/sidebar.less
```diff
@@ -242,7 +242,7 @@
         margin-left: 10px;
       }
 
-      a.expand {
+      a.expand > span {
```
jonatanklosko, line 245: Maybe include the `.icon-expand` class for clarity?

cristineguadelupe: Good idea!

assets/js/sidebar/sidebar-list.js
```diff
@@ -58,17 +58,22 @@ function renderSidebarNodeList (nodesByType, type) {
...
+      }
+
+      // Clear the previous current section
+      if (previousSection) {
+        previousSection.classList.remove('current-section')
       }
 
       if (anchor.matches('.expand')) {
-        const listItem = target.closest('li')
         listItem.classList.add('open')
```
jonatanklosko, line 77: We shouldn't expand the item if we are navigating to another page (so that we don't expand and refresh immediately). A check like `anchor.pathname === window.location.pathname` should help :)

jonatanklosko:
@josevalim did you indent to include `nest_modules_by_prefix: [ExDoc]` in the configuration or is it a leftover?

assets/js/sidebar/sidebar-list.js
cristineguadelupe: Let's wait for @josevalim to confirm, but as far as I remember, we should now!

jonatanklosko (approved):
:shipit:

josevalim:

PR #1471, opened 2022-01-11
Check if settings selector element exists

timothyvanderaerden:
On some pages like `search.html` or `404.html` the following error is thrown: `Uncaught TypeError: Object(...)(...) is null` which can be seen in the console. This is because the settings button isn't declared on those pages.

This pull request adds a check to see if the `SETTINGS_LINK_SELECTOR` exists on the document before adding an event-listener to the element.

An alternate approach would be to add the settings button on those pages as well, I'm not sure which is the desired approach.

### DIFF
--- a/assets/js/settings.js
+++ b/assets/js/settings.js
@@ -31,9 +31,11 @@ export function initialize () {
 }
 
 function addEventListeners () {
-  qs(SETTINGS_LINK_SELECTOR).addEventListener('click', event => {
-    openSettingsModal()
-  })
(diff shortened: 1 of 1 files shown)


timothyvanderaerden:
I don't see a reason to not have the settings button on those pages as well, unless there is a specific reasoning behind it?

josevalim:
I don't think so! Would you like to update or send a separate PR?

josevalim:
I will push several fixes to address the License and the lack of Settings soon, thank you!

PR #1473, opened 2022-01-11
Add settings button to all pages

timothyvanderaerden:
This adds the settings button on all pages, I'm not sure if this is the best approach since the button is defined in both `settings_button_template.eex` and `search-results.handlebars`.


josevalim:

PR #1475, opened 2022-01-12
Add case insensitive quick switch search match

simonprev:
If we search for `Ecto` in the global search, `hex.pm` returns packages that matches `ecto`. The JavaScript filtering should also show those packages.

I also made an assumption that all `hex.pm` packages are lowercase so when you search `Ecto` and press `Enter`, it now redirects to [`https://hex.pm/ecto`](https://hexdocs.pm/ecto/) instead of a `404` error for https://hexdocs.pm/Ecto.


josevalim:

PR #1478, opened 2022-01-13
Remove background from sidebar header

wmnnd:
This PR removes the background from the sidebar "header". Previously the header section had the same background color as the search field above.

Removing the background makes the design cleaner and removes potential confusion from using the same color as background for a container element and for an interactive search field.

## Updated design
Screenshot 2022-01-13 at 21-42-05 Naming Conventions — Elixir v1 14 0-dev

## Before
Screenshot 2022-01-13 at 21-42-34 Naming Conventions — Elixir v1 14 0-dev

### DIFF
--- a/assets/less/sidebar.less
+++ b/assets/less/sidebar.less
@@ -37,8 +37,6 @@
 
   .sidebar-header {
     margin: 12px;
-    border-radius: 4px;
-    background-color: @bgMediumGray;
   }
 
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @wmnnd, I went through this with the designer and we preferred to keep the background. Thanks for the PR though! (the changes to the headers you mentioned on Twitter were approved, in case you are wondering!).

Issue #1480, opened 2022-01-14
Custom scrollbar in `sidebar`


moogle19:
The current scrollbar looks kind of out of place (especially when using the light theme in Chrome).
In my opinion it would be nicer to use a themed scrollbar for the `sidebar`.

For WebKit based browsers this could be achieved with `::-webkit-scrollbar` and Firefox has `scrollbar-color`.

I add the following to the generated css for testing (The colors were just for testing):

```
.sidebar {
   scrollbar-color: #8e7ce6 #0d1829;
}

.sidebar ::-webkit-scrollbar {
  width: 10px;
}

.sidebar ::-webkit-scrollbar-track {
  background: #0d1829;
}

.sidebar ::-webkit-scrollbar-thumb {
  background: #8e7ce6c0;
	border-radius: 10px;
	border: 2px solid #0d1829;
}

.sidebar ::-webkit-scrollbar-thumb:hover {
  background: #8e7ce6ff;
}
```

And here are some before / after screenshots: 

Chrome:

<img width="465" alt="Screenshot 2022-01-14 at 14 07 40" src="https://user-images.githubusercontent.com/792046/149520778-ff964004-40c1-49b5-b1a0-fa4a8b8e083f.png">
 
<img width="469" alt="Screenshot 2022-01-14 at 14 08 01" src="https://user-images.githubusercontent.com/792046/149520789-0b087377-5a99-4a8c-ad2d-ef13023a09de.png">

Firefox: 

<img width="458" alt="Screenshot 2022-01-14 at 14 08 20" src="https://user-images.githubusercontent.com/792046/149520796-8f69d5c0-d504-4acd-8e72-c5bf548ee565.png">

<img width="477" alt="Screenshot 2022-01-14 at 14 08 34" src="https://user-images.githubusercontent.com/792046/149520802-1d78b516-01bf-4a90-ae10-7c43af564428.png">

josevalim:
Hi @moogle19, agreed. However, I think the color purple is a bit too strong. I think we should keep the purple for highlighting. Incidentally, I think the default Firefox color is quite good. Maybe we can settle on a gray color, quite similar to what Firefox is doing? WDYT? Do you want to send a PR?

moogle19:
@josevalim 
Yes, the purple was just for demonstration purposes 😄.

I'll take the default Firefox colors and make a PR for that.

PR #1481, opened 2022-01-14
Add custom scrollbar for sidebar

moogle19:
Added a custom scrollbar to the sidebar as mentioned in #1480 .

I tried to avoid adding new colors and used the existing ones, but I am not sure if the hover color in Chrome is the best choice.
Since Firefox doesn't support changing the color on hover, it maybe would be better to make it static in Chrome as well?

Firefox:

Screenshot 2022-01-14 at 14 43 01

Chrome:

Normal:
Screenshot 2022-01-14 at 14 43 27

Hovered:
Screenshot 2022-01-14 at 14 43 39

### DIFF
--- a/assets/less/sidebar.less
+++ b/assets/less/sidebar.less
@@ -5,6 +5,7 @@
   background: @darkGray;
   color: @lightestGray;
   overflow: hidden;
+  scrollbar-color: @mediumGray @darkGray;
 
   .gradient {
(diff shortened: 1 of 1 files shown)


josevalim:
>  Since Firefox doesn't support changing the color on hover, it maybe would be better to make it static in Chrome as well?

Yes, let's make it static on both for consistency!

moogle19:
@josevalim Done!

josevalim:

Issue #1443, opened 2021-12-06
Add an ability to sync night mode with system preferences


mkaput:
Currently, night mode is controlled by a local storage variable, and it must be manually enabled/disabled. Some people tend to extensively use the system settings for controlling the light/dark mode depending on lightning conditions, and this does not play well with the local storage approach ex_doc uses. Ideally, it would be best if there was an `Sync with system` option, just like for example GitHub does. Even better, it would be awesome to let it be the default option.

Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

josevalim:
A pull request to implement this would be welcome. IIRC, we already consult the system preference, but then we store that in the cookie.

yordis:
If you want to observe the theme changes

```js
window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', e => {
    const systemTheme = e.matches ? "dark" : "light";
});
```

josevalim:
Closing this for now but a PR is welcome if you want to improve this!

mkaput:
If anyone would like to work on this, then please welcome! Sorry for not mentioning this earlier, but I do not have free bandwidth.

PR #1491, opened 2022-01-16
chore: add stale workflow

yordis:
@josevalim I am not sure if you manually check things, and potentially close them or do something about it.

This could help with the topic, I think. Feel free to close it or whatever.

Related to #1443


josevalim:
Thank you. It is fine to do it by hand as we don’t have such a high influx of PRs. :)

PR #1505, opened 2022-02-03
Accessibility - general improvement 

cristineguadelupe:
This PR solves some accessibility issues without modifying any behavior and with minor visual changes.

Light theme - More contrast for inline links and light grays

CleanShot 2022-02-03 at 17 29 37@2x
CleanShot 2022-02-03 at 17 30 11@2x

Dark theme - More contrast for footer text

CleanShot 2022-02-03 at 17 32 42@2x
CleanShot 2022-02-03 at 17 32 53@2x


josevalim:

PR #1513, opened 2022-02-12
Add padding at top to #targets

cw789:
This should theoretically add a padding at top to the function #targets.
I find it uncommon to start reading at top without a padding after using #links.

But I could not test it jet. Windows does not load the webpack.js...

PS: To test this within ex_doc you'll need to have a really small window (pages are to short).

### DIFF
--- a/assets/less/content/functions.less
+++ b/assets/less/content/functions.less
@@ -9,6 +9,8 @@
 }
 
 .detail {
+  padding-top: 1.0em;
+
   &:target .detail-header {
     animation-duration: 0.55s;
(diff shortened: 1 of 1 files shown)


cw789:
Ok. Probably I'm just wrong. I'm no expert.
No, I've no example which would say this should be it as it is more common.
So it's probably more a personal opinion of myself.

Feel free to close.
By the way thanks for reconsidering the sidebar navigation. It's great now.

josevalim:
Glad to hear you enjoyed it!

PR #1529, opened 2022-03-06
support for `source_url_pattern` in config being a function

mayel:
👋 Having a use case of generating one set of docs for a project + some of its dependencies, this allows providing a function instead of a single pattern, so we can link to the source within each dep's repo.

Always open to suggestions about how else I could go about this...

### DIFF
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -7,7 +7,7 @@ defmodule ExDoc.Retriever do
     defexception [:message]
   end
 
-  alias ExDoc.{DocAST, GroupMatcher, Refs}
+  alias ExDoc.{DocAST, GroupMatcher, Refs, Utils}
   alias ExDoc.Retriever.Error
 
   @doc """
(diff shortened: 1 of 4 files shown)


lib/ex_doc/retriever.ex
```diff
@@ -330,8 +330,19 @@ defmodule ExDoc.Retriever do
   defp source_link(%{path: _, url: nil}, _line), do: nil
 
   defp source_link(source, line) do
-    source_url = Regex.replace(~r/%{path}/, source.url, source.path)
-    Regex.replace(~r/%{line}/, source_url, to_string(line))
+    source_url_pattern(source.url, source.path, to_string(line))
+  end
+
+  def source_url_pattern(source_url_pattern, path, line) do
```
josevalim, line 336: Can you please move this to `ExDoc.Utils`? Add a basic unit test and we are good to go!

mayel:
@josevalim done :)

josevalim:
Please run `mix format` so CI is happy? :)

mayel:
CI now passing

josevalim:

PR #1530, opened 2022-03-06
fix sorting when a list of dirs was provided

mayel:
When generating docs for a list of paths rather than a single one, sorting was being done per-path rather than over the entire list of modules to include.


josevalim:

Issue #1533, opened 2022-03-11
night-mode The blockquote has a wrong contrast for the code tag.


paridin:
Currently 
![Screen Shot 2022-03-11 at 13 09 10](https://user-images.githubusercontent.com/5170054/157947047-7e89b052-6a77-4ba2-9087-343b27c72fa4.png)

Proposed
![Screen Shot 2022-03-11 at 13 38 31](https://user-images.githubusercontent.com/5170054/157947112-5299c27f-331e-4ed4-bd42-a737bfc1b139.png)

The proposal is based on the current color to keep the same theme feeling.

![image](https://user-images.githubusercontent.com/5170054/157948982-5e693e7f-56e0-4727-96f8-15158601a4fe.png)

If it looks good, I will do the PR.

josevalim:
It has been fixed in master already. :) thanks for offering tho!

paridin:
I tried master before the issue. I'm seeing this.

![image](https://user-images.githubusercontent.com/5170054/157951320-9907ca7b-22f6-4f9e-baff-571038168e20.png)

josevalim:
Oh, so I misunderstood it by something else. Reopening. PRs welcome!

PR #1534, opened 2022-03-11
Night-Theme color for blockquote.

paridin:
Update the `blockquote` colors to keep them in the same context as the theme. #1533

### DIFF
--- a/assets/less/night/night.less
+++ b/assets/less/night/night.less
@@ -14,7 +14,7 @@ body.dark {
   }
 
   blockquote {
-    border-left: 3px solid rgb(68, 68, 76);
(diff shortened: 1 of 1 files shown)


paridin:
Doubt, about the `formatters` directory,  how is updated? I could not find how it is recreating.

josevalim:
Right, so this is what I meant. The blockquote should not have the same background as the `code` tag. The background of the code tag itself has been fixed on 1b8fb0dc7180cb09b3b0a2265ce7093de39b1f82. :) You will also see that the blockquote in light mode also does have the same background as `code`. :)

josevalim:
In any case, thank you for the PR!

Issue #1539, opened 2022-03-23
Unable to build Elixir Docs


eksperimental:
Running both up-to-date branches Elixir and ExDoc.

```console
---
==> eex (compile)
==> mix (compile)
Generated mix app
==> ex_unit (compile)
Generated ex_unit app
==> logger (compile)
Generated logger app
Generated eex app
==> iex (compile)
Generated iex app
==> ex_doc (elixir)
CANONICAL=main/ bin/elixir ../ex_doc/bin/ex_doc "Elixir" "1.14.0-dev" "lib/elixir/ebin" --main "Kernel" --source-url "https://github.com/elixir-lang/elixir" --source-ref "beb6b4737b2ce07bce0e8798caa5ce51cae2dd19"  --output doc/elixir --canonical "https://hexdocs.pm/elixir/main/" --homepage-url "https://elixir-lang.org/docs.html" --formatter "html" --config "lib/elixir/docs.exs"

11:37:53.644 [notice] Application makeup_erlang exited: :stopped

11:37:53.655 [notice] Application makeup_elixir exited: :stopped

11:37:53.656 [notice] Application makeup exited: :stopped

11:37:53.656 [notice] Application nimble_parsec exited: :stopped

11:37:53.656 [notice] Application earmark_parser exited: :stopped

11:37:53.656 [notice] Application crypto exited: :stopped

11:37:53.656 [notice] Application eex exited: :stopped
** (MatchError) no match of right hand side value: {:error, {:makeup_html, {'no such file or directory', 'makeup_html.app'}}}
    (ex_doc 0.28.3) lib/ex_doc/cli.ex:8: ExDoc.CLI.main/2
    (elixir 1.14.0-dev) lib/code.ex:1228: Code.require_file/2
```

eksperimental:
It works now. Thank you.

Question: How come CI was passing on Elixir?

eksperimental:
I see building docs is part of the release.yml, but not part of the CI.yml.
I wonder whether we should test the building of Elixir Docs, with the master version of ExDoc.
With the introduction of the `--warnings-as-errors` flags in ExDoc we could guarantee that the Elixir Docs are generating no warnings.

josevalim:
@eksperimental I like that. Can you please submit a PR?

eksperimental:
@josevalim Should we wait until #1412  is merged?
I found a way to exit with an error in Linux, so it could still be done I guess.

josevalim:
[...]

PR #1541, opened 2022-03-28
CI: Build Elixir docs

eksperimental:
Related discussion: https://github.com/elixir-lang/ex_doc/issues/1539#issuecomment-1079763510


josevalim:
Thank you @eksperimental! Given that we already check this on the Elixir side and Elixie tends to be more actively committed, I think having it there is enough but we can revisit later if necessary. ❤️

Issue #1518, opened 2022-02-20
Include page titles in the search autocomplete


josevalim:

ByeongUkChoi:
Isn't the page title already included in the search autocomplete?

josevalim:
They appear on search but not in the autocomplete afaik?

ByeongUkChoi:
Sorry, I didn't understand. Aren't those the titles of the page shown in the screenshot?

<img width="600" alt="image" src="https://user-images.githubusercontent.com/43606066/159145471-0efe9124-e9d6-4b67-ae5a-2561cdcfe47a.png">

josevalim:
Oh, by "pages" I mean the "Extras" / "Guides" / etc. You can customize how they are called, so in Ecto's case, it is "Guides". For example, I can't autocomplete for "Getting Started" in Ecto: https://hexdocs.pm/ecto/getting-started.html

And i think both the "Pages" title and the subtitles for each section are already available, we just don't autocomplete them.

PR #1543, opened 2022-03-30
Include page child titles in search autocomplete

ByeongUkChoi:
https://github.com/elixir-lang/ex_doc/issues/1518
Add extra child(headers) node in autocomplete :)


josevalim:
Hi @ByeongUkChoi! I was actually thinking about this after your PR and I thought it is probably best for now to not include this. That's because module docs also have subtitles and maybe we should index those too... but I am also worried it is too much, so I think it is best to wait for now and gather feedback on the feature you already implemented. :)

In any case, thanks so much for the PR and we can reopen this later if necessary!

Issue #1552, opened 2022-04-14
Javascript Script libraries' licenses are missing in doc builds


eksperimental:
app-DIGEST.js.LICENSE.txt is missing from doc builds, `dist/app-DIGEST.js` mentions:

> /*! For license information please see app-a123a6cc33ae3464293e.js.LICENSE.txt */

PR #1553, opened 2022-04-16
Include Javascript Script libraries in doc build

timothyvanderaerden:
Not sure if `*.txt` is to ambitious, otherwise something like below could be a better solution.

```elixir
dist_license(), do: embed_pattern("dist/*.LICENSE.txt")
```

Ref: #1552


josevalim:
Let's go with dist_license(), it is a great suggestion. thank you!

josevalim:

PR #1557, opened 2022-04-18
Optimize html/epub tests by making them asyncronous by using tmp_dir tag

eksperimental:


wojtekmach:
Can we use ExUnit `@tag :tmp_dir` instead? We should be on a recent enough Elixir version to be able to use it.

eksperimental:
This is ready for reviewing @wojtekmach

josevalim:

PR #1576, opened 2022-05-25
Add exception for functions called record/*

### LINKED ISSUE #1575: Error processing erlang source when function is called record
An issue was reported on rebar3_ex_doc (https://github.com/starbelly/rebar3_ex_doc/issues/26). Specifically if a function is called record and more specifically has an accompanying spec an edge is hit. I came up with a quick fix (as seen in the issue), but this is probably not a good approach (i.e., may cause other issues). 

We specifically hit here : https://github.com/elixir-lang/ex_doc/blob/935f2dac019d0bdcb470e68fc8b077e4e27bfe15/lib/ex_doc/language/erlang.ex#L463 but we should be hitting the catch all in the `cond` I believe. 

My quick fix was to adjust the this line to : 

` name in [:"::", :when, :%{}, :{}, :|, :->, :record] and acc != [] ->` 

Of course, I don't know this code that well atm, and the above is based on an assumption as to what's going on in the code, and assumptions are bad :)

starbelly:
Closes #1575 

This commit attempts to resolve a case where there are specs defined referring to a function called record/*

### DIFF
--- a/lib/ex_doc/language/erlang.ex
+++ b/lib/ex_doc/language/erlang.ex
@@ -460,7 +460,7 @@ defmodule ExDoc.Language.Erlang do
               arity = length(args)
 
               cond do
-                name in [:record] and acc != [] ->
+                name == :record and acc != [] ->
                   {ast, acc}
 
                 name in [:"::", :when, :%{}, :{}, :|, :->] ->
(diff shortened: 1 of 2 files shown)


josevalim:

PR #1579, opened 2022-06-16
Fix loading serialized index from sessionStorage

timothyvanderaerden:
I noticed that lunr throws an error when it tries to load the serialized index from the `sessionStorage`. This causes the index to be generated upon each new search entry, which can have a performance impact for large projects.

Lunr throws the error because `elixirTokenSplitter` added to the pipeline during the index creation but it isn't registered when the index is loaded from the `sessionStorage`.

The following error is thrown when searching:
```
Failed to load index:  Error: Cannot load unregistered function: elixirTokenSplitter
```


josevalim:

PR #1580, opened 2022-06-25
Improve admonition block example

timothyvanderaerden:


timothyvanderaerden:
That makes sense! I will change the PR with the suggested changes, it makes the example more clear.

josevalim:

PR #1583, opened 2022-06-28
QuickSearch feature implementation

igorsegallafa:
image
image


josevalim:
At the moment we don't plan to add another mechanism for searching besides the bar on the top left. Thank you!

PR #756, opened 2017-07-18
fix: remove extra term() argument at start of macro spec

Eiji7:
fixes: #755

Issue #755, opened 2017-07-17
Extra term() as first argument for all macros in @spec.


Eiji7:
Here you can see [original post](https://elixirforum.com/t/exdoc-why-i-get-always-term-as-a-type-of-first-argument-in-spec-for-any-macro/6959?u=eiji) on Elixir forum.
I looked deeper and found: #166 PR.

@ScrimpyCat asked:
> The macro specs include a first argument of type term(), I'm not sure if it's best for that to be stripped out of the displayed spec or not. So I just left it be.

and @josevalim answered:
> Thank you! We should definitely fix the spec name (by removing the "MACRO-" prefix) and remove the first argument. :)

I do not see any `MACRO-` prefix, so this part looks ok for me, but I can see `term()` as first argument, so it should be corrected as José said.

Eiji7:
@josevalim Looks like exactly same happen for `@macrocallback`. What's interesting there was a test for `term()` as first argument for this case, but I think it's inconsistent with `defmacro`. WDYT?

PR #1590, opened 2022-08-18
fix: remove extra term() argument at start of macrocallback spec

Eiji7:
related: #755 and #756
ping @josevalim

### DIFF
--- a/lib/ex_doc/language/elixir.ex
+++ b/lib/ex_doc/language/elixir.ex
@@ -109,14 +109,14 @@ defmodule ExDoc.Language.Elixir do
       if actual_def in module_data.private.optional_callbacks, do: ["optional"], else: []
 
     specs =
-      case {kind, Map.fetch(module_data.private.callbacks, actual_def)} do
-        {:macrocallback, {:ok, specs}} ->
+      case module_data.private.callbacks do
+        %{^actual_def => specs} when kind == :macro_callback ->
           Enum.map(specs, &remove_callback_term/1)
 
-        {:callback, {:ok, specs}} ->
+        %{^actual_def => specs} ->
           specs
 
-        {_kind, :error} ->
+        %{} ->
           []
       end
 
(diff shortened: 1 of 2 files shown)


lib/ex_doc/language/elixir.ex
```diff
@@ -109,11 +109,14 @@ defmodule ExDoc.Language.Elixir do
       if actual_def in module_data.private.optional_callbacks, do: ["optional"], else: []
 
     specs =
-      case Map.fetch(module_data.private.callbacks, actual_def) do
-        {:ok, specs} ->
+      case {kind, Map.fetch(module_data.private.callbacks, actual_def)} do
+        {:macrocallback, {:ok, specs}} ->
+          Enum.map(specs, &remove_callback_term/1)
+
+        {:callback, {:ok, specs}} ->
```
wojtekmach, line 116:
```suggestion
      case Map.fetch(module_data.private.callbacks, actual_def) do
        if kind == :macrocallback do
          Enum.map(specs, &remove_callback_term/1)
        else
          specs
        end
```

wojtekmach (commented):
Nice catch!

lib/ex_doc/language/elixir.ex
Eiji7: @wojtekmach I was also curious about "preferred style" in core team. That was also my first version, but I was thinkining that I'm adding unnecessary indention, so I took a look at other `case` in this file …

I found similar code (variable + function call in tuple) in lines 770 and 835 (or 759 and 824 before change). Before commit those I would like to ask about consistency with mentioned lines if you don't mind. :slightly_smiling_face:

lib/ex_doc/language/elixir.ex
```diff
@@ -109,11 +109,14 @@ defmodule ExDoc.Language.Elixir do
       if actual_def in module_data.private.optional_callbacks, do: ["optional"], else: []
 
     specs =
-      case Map.fetch(module_data.private.callbacks, actual_def) do
-        {:ok, specs} ->
+      case module_data.private.callbacks do
+        %{^actual_def => specs} when kind == :macro_callback ->
```
josevalim, line 113:
```suggestion
        %{^actual_def => specs} when kind == :macrocallback ->
```

josevalim:

lib/ex_doc/language/elixir.ex
Eiji7: José Valim already fixed that much better with `when`. :+1: :heart:

PR #1595, opened 2022-08-28
Add cheatmd support

yordis:


test/fixtures/cheatsheets.cheatmd
```diff
@@ -0,0 +1,763 @@
+---
+title: Elixir
+category: Elixir
+layout: 2017/sheet
+tags: [New]
+updated: 2018-07-04
+weight: -10
+---
```
yordis, line 8: Do we suppose to do something with this?

josevalim: No, we won't have preludes I think. We can discard it :)

test/fixtures/cheatsheets.cheatmd
```diff
@@ -0,0 +1,754 @@
...
+  def extensions, do: ["json"]
+end
+```
+
+See: [Module](http://elixir-lang.org/docs/stable/elixir/Module.html)
+
+## References
+{: .-one-column}
+
+- [Learn Elixir in Y minutes](https://learnxinyminutes.com/docs/elixir/)
```
josevalim, line 754: I think we can reduce the size of this cheatsheet. :)

lib/ex_doc/formatter/html.ex
```diff
@@ -543,4 +552,33 @@ defmodule ExDoc.Formatter.HTML do
         {base, opts[:filename] || text_to_id(Path.rootname(base))}
     end)
   end
+
+  defp sectionize(list), do: sectionize(list, [])
```
josevalim, line 556: Let's move this function to the `Markdown` module, WDYT?

test/fixtures/cheatsheets.cheatmd
josevalim: I think we can keep it small mostly so it is easier for us to test it. :)

lib/ex_doc/formatter/html.ex
yordis: I thought on doing that, wasn't sure about the intention of such a module.

josevalim:

Issue #1542, opened 2022-03-28
Move to esbuild


josevalim:
We will probably need npm still, for handlebars, but it should streamline things considerably.

sigu:
I can get started in this, what is the procedure?

josevalim:
The next step would be to submit a PR that replaces webpack by esbuild. Please let me know if you have any questions. :)

osbre:
Is this a complete webpack removal or adding [esbuild-loader](https://github.com/privatenumber/esbuild-loader) to speed it up?

DavidOliver:
@sigu, it sounds like you're taking this, but let me know if not as I'd be happy to have a go in between other bits over the next few days.

@josevalim, I wondered if this would be a good time to remove Less for CSS. I tend to use plain CSS now as browser support for custom properties (variables) is good.

josevalim:
Ah, I forgot about less. I think we can definitely consider removing it. Perhaps before moving to esbuild. Do you want to analyze how doable doing so is? Thanks!

DavidOliver:
Sure - will do at some point this week hopefully, unless @sigu chooses to look at that part before I get to it. :)

sigu:
Hello @DavidOliver I will have a look at how doable it is in the next few hours and share my findings. You can also have a look in parallel

cw789:
Removing less is in my opinion a good idea.

DavidOliver:
> Hello @DavidOliver I will have a look at how doable it is in the next few hours and share my findings. You can also have a look in parallel

Hi @sigu. What's the state of play with this your end? I started to have a quick look through the less files and saw that some reworking would be required with regards to nested selectors and content variables. I'd be happy to take a look at handling the switch to plain CSS next week if you'd like?

DavidOliver:
I've been away but have now made a start on removing Less.

DavidOliver:
@josevalim, while I'm going through all the CSS, should I remove styles and assets that were added to support now-out-of-scope(?) browsers such as IE and old versions of current browsers? E.g., removing all font formats except for woff2.

josevalim:
Sounds good to me!

moogle19:
@DavidOliver 
I played around with the `lessc` command line tool and got a working version with less completly removed (see https://github.com/moogle19/ex_doc/tree/less_less). It didn't produce the most beautiful css but could be a good starting point.

DavidOliver:
[...]

PR #1597, opened 2022-09-02
Remove Less CSS

DavidOliver:
As discussed in #1542.

I've tested all the features I can find using ExDoc's own docs as built locally with `mix build`, but I think it would be good if one or two people with more experience could try it out. I've created blockquotes/admonition blocks of all types to test those colours, and I'm not aware of anything that's likely to be incorrect, but everything has necessarily been touched styling-wise.

One thing I haven't tested is the version dropdown menu, as this isn't present in the local build by default. But I expect the styling to remain as previously.

This commit removes the Less-related npm packages, and updates the remaining npm packages according to existing versions set in `package.json`.

A few notes on specific aspects follow.

## Theme colours

As CSS custom properties' values can change depending on scope, such as `body.dark`, we no longer have to specify different variable names in `night/*` includes, with their repeated selectors. All such files have been removed, and the main styles cover both themes by using one set of custom property names.

All custom properties that need to change depending on theme are set in the `theme-{light|dark}.css` files in the `custom-props` folder (`makeup` colours currently excepted), mostly by referring to other custom properties set in `common.css`:

Screenshot from 2022-09-02 17-25-11

The only colours whose values are specified in the theme files are the admonition block colours, as they seem to be one-offs.

## ePub 'content-inner' class (template modified)

In plain CSS, we can't choose to/not to nest `@import`s in the `.content-inner` class, so the 'content' styles' selectors now always include the `.content-inner` class to create descendent selectors. This meant that the content styles which are included in the ePub styles weren't taking effect, so I've added the `content-inner` class to the ePub template's `body` tag.

`lib/ex_doc/formatter/epub/templates/head_template.eex`:

```
<body class="content-inner">
```

If this is deemed unacceptable, we can probably ensure that all 'content' style selectors are specific enough even without the `content-inner` class, but the above template modification seemed safer for now.

## “Button” underline decoration (template modified)

Button elements that look like normal links, and that should be underlined, now rely on the `a-main` class being present in the HTML. (Just the Search button in the footer, I believe.)

`lib/ex_doc/formatter/html/templates/footer_template.eex`:

```
<button class="a-main ...">
  Search
</button>
```

## no-underline links

The inline code/command text in links with the `no-underline` class were previously shown underlined in dark mode (but not in light mode). In order to match the light mode and to simplify the CSS, such links are now displayed in the 'main' colour (purple or red) with no underline.

### Light (unchanged)
image

### Dark: before
image

### Dark: after
image

## Keyboard shortcut key colours

Similarly, to simplify the CSS and (I think) improve the styling, the keyboard shortcut key colours have been tweaked.

### Dark (unchanged)
image

### Light: before
image

### Light: after
image


josevalim:
This is amazing work, thank you! Two questions:

1. Regarding content-inner

> so I've added the content-inner class to the ePub template's body tag.

I am not sure I follow what is happening here. Do we need to add them to the body? Could we have a div inside the body instead? Or maybe we should add `class="epub"` to the body instead?

2. Did you give an Erlang project a try too? Such as github.com/beam-telemetry/telemetry?

DavidOliver:
1. For example:

    `_epub.css`:

    ```css
    @import 'content/code.css';
    ```

    `content/code.css` example selectors:

    ```css
    .content-inner :is(a.no-underline, pre a) { }
    .content-inner code { }
    ```

    I expect we could add a div inside the epub body, but the class on the body seemed convenient and unobtrusive.

2. No, not yet, but will do now you've mentioned it. :)

josevalim:
Should we go then with:

```
<div class="content-outer">
  <div id="content" class="content-inner">
```

to be consistent with the HTML one? or will that mess up other stuff?

DavidOliver:
To my way of thinking, as we're not including styles that apply to `content-outer` in the epub output, and the epub layout is much simpler, not requiring anything outside the (inner) content area, it makes sense to stick with the one, inner class. But I'm happy to go with whichever option you'd prefer.

As far as I know, any of the options would work without issue.

josevalim:
Beautiful! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

If there is anything pending on the Erlang side, another PR is welcome!

DavidOliver:
Thanks! I'll test with that Erlang project's docs soon.

DavidOliver:
I've not done anything with Erlang before, and I'm afraid I'm struggling to use ExDoc with telemetry.

I have Erlang 25 and rebar 3 installed via asdf.

I thought I'd first try with whatever version of ExDoc 'rebar3_ex_doc' uses. In `~/.config/rebar3/rebar.config`:

```
{plugins, [rebar3_ex_doc]}.
```

```
david@sanguinalis:/m/projects/telemetry$ rebar3 plugins list
--- Global plugins ---
rebar3_ex_doc (0.2.13)

david@sanguinalis:/m/projects/telemetry$ rebar3 ex_doc
===> Verifying dependencies...
===> Analyzing applications...
===> Compiling telemetry
===> Running edoc for telemetry
===> An unknown error occurred generating docs config. Run with DIAGNOSTICS=1 for more details.

david@sanguinalis:/m/projects/telemetry$ DIAGNOSTICS=1 rebar3 ex_doc
===> Verifying dependencies...
===> Analyzing applications...
===> Compiling telemetry
===> Running edoc for telemetry
===> An unknown error occurred generating docs config. Run with DIAGNOSTICS=1 for more details.
```

I suppose I should switch to trying to use the current version of ExDoc with Less removed anyway, but I don't know how to go about doing that. Are you able to give me any pointers, please?

josevalim:
Hi @DavidOliver! For simplicity I added `mix docs --proglang erlang`, which you can run in ExDoc itself. I tried it and it looks good, but a double check is appreciated :)

DavidOliver:
Nice - thanks! Yes, I can see red instead of purple where expected.

PR #1602, opened 2022-09-06
Prepare CSS for esbuild

DavidOliver:
Re. #1542. This prepares the CSS for building with esbuild (barring the normalize `@import`), while still working with the current webpack build.

- `@import` at-rules moved to the top and simplified where appropriate.
- Note on removing '~' from `normalize` `@import` added.


josevalim:

Issue #1554, opened 2022-04-17
Feature to store sidebare state across reloads


cw789:
Is it desired to store the sidebar state in ~~`localStorage` or~~ `sessionStorage`?

For me if I go into my docs "reading mode".
Means I close the sidebar and read page by page using the "Next Page →" link on the bottom of each page.
It is annoying if on every next page the sidebar pops up "opened" again.

I could contribute this feature when the dev environment is ready to be used on latest node again.
I'm to lazy to install old node version for small contributions.

josevalim:
This is tricky because it is stored Per Domain and also we don’t want necessarily for a closed side bar to stick when you come back to hexdocs a week later. So my  thought at the moment is no. :)

cw789:
I understand. But if we use `sessionStorage` it would only be valid for your Browser tab session.
Means sidebar would only be still closed if you use the same Browser tab a week later (just hitting the peoples never closing tabs - and finding them a week later again ;).

josevalim:
Ah, so that works for me!

cw789:
Fine. Then keep it open.

josevalim:
> I could contribute this feature when the dev environment is ready to be used on latest node again.

Btw, a PR that bumps to latest node is welcome (i.e. a PR that makes it work for you). Otherwise no one might submit this feature.

cw789:
I know, but first I've tried to get rid of `less` and switch to `esbuild`... Then my time ended.
The setup here has grown quite some complexity.

josevalim:
I think it is easier first to just update to latest node, instead of doing a full migration. :)

cw789:
So I've really tried now to update `ex_doc` to be compatible with latest Node.
But I'm sorry I've not managed to be successful.

Doing an immediate update of webpack 4 to webpack 5 to make it work (for now) seems to much energy for now for me.

cw789:
I've spent bit of time again. Downgraded Node to LTS v16 on my machine - no success.
So it's either that the setup does not work on Windows or does not work with latest Node v16 either.

Then I've tried to ditch webpack with esbuild - no success.
The handlebars plubin I've found for esbuild doesn't seem to work.

This is not any critic. This is just a public log what I've already tried.

DavidOliver:
> I know, but first I've tried to get rid of `less` and switch to `esbuild`... Then my time ended.
> The setup here has grown quite some complexity.

[...]

PR #1606, opened 2022-09-16
Save sidebar state per session

cw789:
Means sidebar stays open or close if you navigate.
For example using previous / next page navigation.

Close https://github.com/elixir-lang/ex_doc/issues/1554

Additionally I needed to fix a CSS animation.
By doing so I moved the sidebar button 7px more from left & top into the page.
Found it unwanted to have a different height position between closed / opened sidebar.

### DIFF
--- a/assets/css/layout.css
+++ b/assets/css/layout.css
@@ -39,11 +39,11 @@ body {
 .sidebar-button {
   position: fixed;
   z-index: 99;
-  left: 7px;
+  left: 0;
   top: 7px;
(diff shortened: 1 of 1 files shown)


josevalim:
Can you please revert the button changes? I am pretty sure it was intentional in the design. :)

josevalim:

josevalim:
Unfortunately this introduces a bug:

1. If I close the sidebar
2. Then I open it
3. And then make the screen small, the sidebar no longer collapses automatically

I think we should only store the closed state. Can you please send a pull request? I will revert this PR meanwhile.

josevalim:
Thank you!

PR #1608, opened 2022-09-16
Static hosting of google fonts

### LINKED ISSUE #1607: Static hosting of google fonts
ExDoc states:

- Generates HTML documentation accessible online and offline

Yes it is offline usable but not that pretty without the fonts.

The other thing about using google fonts is the GDPR topic in some European countries - if you host our ExDoc docs in those countries.
At least in Germany it is problematic - in Austria it's currently also a topic.

Are we open to add the fonts directly to the assets?
I could take care of it.

cw789:
To be discussed

Close #1607


josevalim:
Yes, let's go ahead. GDPR is a strong enough argument to me.

jonatanklosko:
Not sure if it works well with esbuild, but there's a `@fontsource/*` npm package for every font from google fonts, like [`@fontsource/merriweather
`](https://www.npmjs.com/package/@fontsource/merriweather), which simplifies it into a single import.

josevalim:
I don’t think we should be Latin only because someone can write documentation in Chinese, right?

But I am fine with getting rid of woff.

josevalim:
Thank you! I am thinking we should revert to your original commit and host them ourselves as we don't need woff. Thoughts?

josevalim:
Ah, maybe your original commit is latin only?

jonatanklosko:
As far as I understand https://github.com/elixir-lang/ex_doc/pull/1608/commits/a888a90404c2f2e6974fbf7a9b32f154a4103aac got rid of woff, but I don't see it in the final diff for some reason.

josevalim:

PR #1609, opened 2022-09-18
Readable yellow color for Dark mod

shahryarjb:
Hi, I think this color for yellow color can be warmer and readable in both dark mod and light mod

## Before changing

As you see, we can't read the text in dark mod

---
## After changing

Changing the color and format is based on the taste of each person.
**I did not separate the colors because a variable was created only for the color yellow.**

### DIFF
--- a/assets/css/custom-props/common.css
+++ b/assets/css/custom-props/common.css
@@ -26,8 +26,8 @@
   --coldGray:                hsl(240,   5%,  46%      );
   --coldGrayDark:            hsl(240,   5%,  28%      );
   --coldGrayDim:             hsl(240,   5%,  18%      );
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @shahryarjb!

We don't need to use the same color for both. We can keep the light theme as is, and update the dark mode only. Can you please amend?

Thank you!

shahryarjb:
Hello sir @josevalim,
I added a new color and set it for dark mod. I prefer not to change the other `yellow` colors, because they are used in another files, and maybe we have conflict, so I think it is better to add new color as `yellowDark`

Please see the new changes and If you have any suggestion please let me know I can do it immediately.

josevalim:

PR #1611, opened 2022-09-20
Distinguish output code blocks

jonatanklosko:
When rendering notebook contents with outputs, it is not obvious which blocks are outputs, especially that they are highlighted as regular code. This adds a specialized styling for output blocks.

image

### DIFF
--- a/lib/ex_doc/markdown/earmark.ex
+++ b/lib/ex_doc/markdown/earmark.ex
@@ -72,6 +72,7 @@ defmodule ExDoc.Markdown.Earmark do
     []
   end
 
+  # We are matching on Livebook outputs here, because we prune comments at this point
   defp fixup_list(
          [
            {:comment, _, [~s/ livebook:{"output":true} /], %{comment: true}},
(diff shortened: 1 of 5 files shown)


lib/ex_doc/markdown/earmark.ex
```diff
@@ -72,6 +72,19 @@ defmodule ExDoc.Markdown.Earmark do
...
+           {:comment, _, [~s/ livebook:{"output":true} /], %{comment: true}},
+           {"pre", pre_attrs, [{"code", code_attrs, [source], code_meta}], pre_meta}
+           | ast
+         ],
+         acc
+       ) do
+    code_attrs = Enum.reject(code_attrs, &match?({"class", _}, &1))
+    new_code = {"code", [{"class", "output"} | code_attrs], [source], code_meta}
+    fixup_list([{"pre", pre_attrs, [new_code], pre_meta} | ast], acc)
+  end
```
jonatanklosko, line 86: I'm not sure if the Livebook matching shouldn't happen somewhere upstream, but here we reject all comments, so it is the only place given the current approach.

jonatanklosko:
@josevalim updated and added a test :)

image

josevalim:

PR #1617, opened 2022-10-05
Support Erlang Extras

the-mikedavis:
This change supports generating extras with Erlang syntax. Inline-code blocks are parsed as Elixir but written as Erlang, so `:telemetry.execute/3` is correctly autolinked and re-written as `telemetry:execute/3`.

Connects https://github.com/elixir-lang/ex_doc/issues/1333

### DIFF
--- a/test/ex_doc/language/erlang_test.exs
+++ b/test/ex_doc/language/erlang_test.exs
@@ -161,6 +161,57 @@ defmodule ExDoc.Language.ErlangTest do
     end
   end
 
+  describe "autolink_doc/2 for extras" do
+    test "function", c do
+      assert autolink_extra("`foo:foo/0`", c) ==
+               ~s|<a href="foo.html#foo/0"><code class="inline">foo:foo/0</code></a>|
+    end
+
+    test "OTP function", c do
+      assert autolink_extra("`lists:reverse/1`", c) ==
+               ~s|<a href="https://www.erlang.org/doc/man/lists.html#reverse-1"><code class="inline">lists:reverse/1</code></a>|
+    end
+
+    test "type", c do
+      assert autolink_extra("`t:bar:t()`", c) ==
+               ~s|<a href="bar.html#t:t/0"><code class="inline">bar:t()</code></a>|
+    end
+
+    test "OTP type", c do
+      assert autolink_extra("`t:array:array()`", c) ==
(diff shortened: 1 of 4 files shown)


the-mikedavis:
Was this what you were thinking with ExDoc syntax @josevalim? Re-reading your comment I just realized you might have meant that we should use `telemetry:execute/3` / `t:telemetry:event_name()` / `c:gen_server:init/1` rather than `:telemetry.execute/3` / `t::telemetry.event_name/0` / `c::gen_server.init/1` 😅

josevalim:
> Re-reading your comment I just realized you might have meant that we should use telemetry:execute/3 / t:telemetry:event_name() / c:gen_server:init/1

Yeah, that's what I meant :D

test/ex_doc/language/erlang_test.exs
```diff
@@ -161,6 +161,59 @@ defmodule ExDoc.Language.ErlangTest do
...
+                        ~s|<code class="inline">t:bad:bad/0</code>|
+             end) =~ ~s|references type "bad:bad()" but it is undefined or private|
+    end
+
+    test "bad module", c do
+      assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
+               assert autolink_extra("`does_not_exist`", c) ==
+                        ~s|<code class="inline">does_not_exist</code>|
+             end) =~ ~s|references module "does_not_exist" but it is undefined|
+    end
```
the-mikedavis, line 214: Is emitting warnings for unknown modules too verbose? Code like <code>\`true\`</code> would trigger a warning

josevalim: Yes. :(

the-mikedavis:
I think that the CI failure is spurious - I can't reproduce it locally - and seems to be unrelated to this change.

josevalim (approved):
LGTM! :heart: Asking for a second look from @wojtekmach too!

wojtekmach:
Thank you!

josevalim:
@the-mikedavis can you please submit some documentation to rebar_ex_doc (if you haven't yet) about how to use this feature from Erlang? Thank you! :heart:

v0.29.1 will be out in a couple minutes with it.

the-mikedavis:
Sweet! I'll work on a PR to show this off rebar3_ex_doc 👍

PR #1619, opened 2022-10-19
Save sidebar state per session

cw789:
Only restore close state.
Sidebar will still auto close on window resize.

As discussed in https://github.com/elixir-lang/ex_doc/pull/1606#issuecomment-1283887744.


josevalim:

Issue #1452, opened 2021-12-28
Earmark GFM Footnote Return to Article Link Text Rendered Incorrectly


jwilger:
When parsing a markdown document that contains footnotes with the option `footnotes: true` passed to Earmark, the link that is inserted at the end of the footnote by Earmark has the link text set to `["&#x21A9;"]` (https://github.com/RobertDober/earmark_parser/blob/5797a53a1d60003e5d0038c326bfd1e8b129c85f/lib/earmark_parser/ast/renderer/footnote_renderer.ex#L31).

When rendering the documentation as HTML via ex_doc, this link text is escaped, replacing the `&` with `&amp;`, so that the literal `&#x21A9;` is displayed in the text rather than the HTML entity that was intended.

It looks like ex_doc can take a `verbatim: true` option when rendering a tag (https://github.com/elixir-lang/ex_doc/blob/7dbc93742f812f9d4af9e55f4bd3cc4e89ed61dd/lib/ex_doc/doc_ast.ex#L49) which seems like it would address this issue, but I'm not entirely sure how to get ex_doc to use that option in this case. Part of me wonders if that option shouldn't just be used for all text that is coming from the markdown processor, thus making any escaping of HTML entities the job of the markdown processor itself rather than the templating modules of ex_doc, but I'm not familiar enough with the ins and outs of the project to know if that would be a good idea or not.

josevalim:
@jwilger looking at the earmark code you linked, it is not set `verbatim: true` to that particular node, so maybe that's what we should do? Can you please send a PR to earmark_parser and see if it works?

jwilger:
In addition to looking into @josevalim's comment above, there's a somewhat kludgy workaround using CSS for anyone else who runs into this; just add the following to your `mix.exs`:

```elixir
def project do
  [
    # ...
    docs: [
      # ...
      before_closing_head_tag: &before_closing_head_tag/1
    ]
  ]
end

defp before_closing_head_tag(:html) do
  """
  <style>
    a.reversefootnote {
      display: inline-block;
      text-indent: -9999px;
      line-height: 0;
    }

    a.reversefootnote:after {
      content: ' ↩'; /* or any other text you want */
      text-indent: 0;
      display: block;
      line-height: initial;
    }
  </style>
  """
end

defp before_closing_head_tag(_), do: ""
```

josevalim:
Closing this, as I believe it is not an ExDoc issue!

marick:
[...]

PR #1623, opened 2022-10-26
Enable Earmark footnotes by default

oskarkook:
GFM has footnotes ([link](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#footnotes)), which currently aren't supported by default with ExDoc. Since GFM is already enabled by default, I think it would make sense to enable footnotes by default as well.

See also: [documentation on footnotes in Earmark repository](https://github.com/robertdober/earmark_parser#footnotes).


josevalim:
Can you provide screenshots of how they are rendered? I believe we had them at some point but disabled due to rendering issues.

oskarkook:
Yeah, I can actually reproduce https://github.com/elixir-lang/ex_doc/issues/1452

I don't have enough context on the issue to fix it, so I think I'll close this for the time being.

PR #1624, opened 2022-10-26
Add tooltip to functions on sidebar

### LINKED ISSUE #1614: Include signatures in sidebar items
I think it will improve the user experience to include the function signature for every function, type, guard, callback, etc in the side bar.

This could be added to the `title` attribute in the anchor element, and it would look like this when you hover it:

image

The only drawback is that it could be distracting and detrimental, but it only shows after a second or so of leaving the cursor idle.

vinibrsl:
This commit adds a tooltip to the sidebar with the function signature. This improves UX when listing functions with same name but different arities. It also works for types, guards, and callbacks.

Screen Shot 2022-10-26 at 19 05 20

Closes #1614


josevalim:

PR #1625, opened 2022-10-27
Remove Coveralls leftovers

antedeguemon:
Hey there! 

The [coveralls](https://hex.pm/packages/coveralls) library was removed from ex_doc by 649c21e, but it was kept as the default coverage tool - which caused `mix test --cover` to fail. 😿 

This PR does two things:
1. Removes Coveralls leftovers from `mix.exs` - so `mix test --cover` works again
2. Removes old coveralls.io badge from README - the coverage is pretty outdated since the last build is from 2019


josevalim:

PR #1627, opened 2022-11-17
Add optional function annotations

maartenvanvliet:
This PR adds the possibility to specify additional function annotations that will be rendered by ExDoc. This is at the function level what https://github.com/elixir-lang/ex_doc/pull/1430 does for modules.

My use case is that a bunch of functions are generated at compile time and I want to mark them as such using ```@doc tags: [:generated]```

This would be the result:
Screenshot 2022-11-17 at 14 08 56

### DIFF
--- a/lib/ex_doc/config.ex
+++ b/lib/ex_doc/config.ex
@@ -6,8 +6,10 @@ defmodule ExDoc.Config do
   def filter_modules(_module, _metadata), do: true
   def before_closing_head_tag(_), do: ""
   def before_closing_body_tag(_), do: ""
+  def annotations_for_docs(_), do: []
 
-  defstruct api_reference: true,
+  defstruct annotations_for_docs: &__MODULE__.annotations_for_docs/1,
+            api_reference: true,
             apps: [],
             assets: nil,
             authors: nil,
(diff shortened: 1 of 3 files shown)


josevalim:
Thanks @maartenvanvliet! Unfortunately I am a bit reserved about using `:tags` for this, given that tags in EDoc mean something else.

One thing we could do is to introduce a `annotations_for_functions`, which receives the metadata documentation and return a list. In the same way `groups_for_functions` and similar also handle functions.

maartenvanvliet:
I've added the `annotations_for_functions` config option. It's invoked with the metadata. It's not invoked for callbacks and types though, that could be added but then the function would probably need a new name :)

josevalim:
Apparently `groups_for_functions` is called for callbacks, so it may be a good idea to do the same here.

We can either keep it `annotations_for_functions` or rename it to `annotations_for_entries` or `annotations_for_docs` (given they are both declared with `@doc`). Any suggestions?

Then I will rename `groups_for_functions` later.

josevalim:
Oh, and please rebase could I made some changes to prepare the unification. :D Sorry!˜

maartenvanvliet:
I've gone with `annotations_for_docs`

lib/ex_doc/config.ex
```diff
@@ -44,6 +46,7 @@ defmodule ExDoc.Config do
             version: nil
 
   @type t :: %__MODULE__{
+          annotations_for_docs: (term() -> list()),
```
josevalim, line 49:
```suggestion
          annotations_for_docs: (map() -> list()),
```

josevalim:

PR #1628, opened 2022-11-18
Initial pass at improving print.css

zolrath:
Hello!
I know there is quite a bit of work to be done for Cheatsheets to be viably printed but there is some low hanging fruit as far as printability goes for ExDoc as a whole.

There's a bit of dupl

I'm split on a few things as we can set CSS to force page breaks and cause codeblocks to shift to resist being split across multiple pages etc, but very interested in knowing what direction you'd like to head with `print.css`!

# Left Spacing + Show Source + Settings
Currently there's a large `300px` gap on the left side. We can remove this spacing as well as the Show Source and Settings buttons.
## Before
image
## After
image

# Admonition Blocks + Code Borders
The background colors for admonition blocks disappear so we can make them visible again by bringing in borders.
The code blocks and inline code borders were a very low contrast which lets them disappear but also makes distinguishing them difficult, the contrast has been increased for visibility.

## Before
image
## After
image

# Tooltip Removed
The bottom of every page displayed a toast, in my case displaying Browser changed theme to "light". This has been hidden.
## Before
image
## After
image

# Code Wrap
Not all code blocks force wrapping. This is fine when viewed on a computer but it's fairly hard to scroll to the right once printed.
All code blocks now wrap.
## Before
image
## After
image

# Remove footer information
The footer contains information related to ExDoc versioning etc, but it doesn't contain information like the version of the package that the docs were printed for. Given their limited use in printed form, I've removed them but would be interested in them containing some information which might allow a user to see that their printed documentation is outdated.

## Before
image
## After
image

### DIFF
--- a/assets/css/print.css
+++ b/assets/css/print.css
@@ -22,4 +22,12 @@
   #toast {
     display: none;
   }
+
+  a.view-source {
+    display: none;
+  }
+
+  button.display-settings {
+    display: none;
+  }
 }
--- a/assets/css/print.css
+++ b/assets/css/print.css
@@ -11,6 +11,8 @@
   .content {
     padding-left: 0;
     overflow: visible;
+    left: 0px;
+    width: 100%;
   }
 
   .summary-row {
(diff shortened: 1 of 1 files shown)


josevalim:
This is great! About the footer, maybe we can keep the "Build with ExDoc (v0.29.0)" on the footer of every page?

josevalim:

PR #1629, opened 2022-11-18
Basic Cheatsheet print.css

zolrath:
Hey hey! This attempts to make the printed version appear more like a low-ink version of the one column 1000px screen version of the Cheatsheet.

By wrapping all of the existing cheatsheet css in `@media screen` and adding a `@media print` section for the cheatsheet we have fewer styles to overwrite for printability. It halves the amount the columns displayed: `.list-4` displays 2, `.list-6` displays 3, etc. I've also brought back the Built using ExDoc portion of the footer string.

Styling for print could make further departures from the screen version for clarity but this is a large difference in usability from the existing print styles.

Two PDFs of the [ExDoc Cheatsheet](https://hexdocs.pm/ex_doc/cheatsheet.html) for comparison:

[Cheatsheet User Guide - Old.pdf](https://github.com/elixir-lang/ex_doc/files/10040549/Cheatsheet.User.Guide.ExDoc.v0.29.0.-.Old.Print.pdf)

[Cheatsheet User Guide - New.pdf](https://github.com/elixir-lang/ex_doc/files/10040598/Cheatsheet.User.Guide.ExDoc.v0.29.0.-.New.Print.pdf)

## ExDoc Cheatsheet Screen Version
image

## Original Print
image

## New Print
image
image

### DIFF
--- a/assets/css/_html.css
+++ b/assets/css/_html.css
@@ -33,6 +33,7 @@
 @import 'toast.css';
 @import 'screen-reader.css';
 @import 'print.css';
+@import 'print-cheatsheet.css';
 @import 'makeup.css';
 
 
--- a/assets/css/content/cheatsheet.css
+++ b/assets/css/content/cheatsheet.css
@@ -1,283 +1,285 @@
-.page-cheatmd .content-inner {
-  max-width: 1200px;
-}
+@media screen {
+  .page-cheatmd .content-inner {
+    max-width: 1200px;
+  }
 
-/* h1 styling */
+  /* h1 styling */
 
-.page-cheatmd h1 {
-  margin-bottom: 1em;
-}
+  .page-cheatmd h1 {
+    margin-bottom: 1em;
+  }
 
-/* h2 styling */
+  /* h2 styling */
 
-.page-cheatmd h2 {
-  margin: 1em 0;
-  column-span: all;
-  padding-left: 3px;
-  color: var(--gray700);
-  font-weight: 500;
-}
+  .page-cheatmd h2 {
+    margin: 1em 0;
+    column-span: all;
+    padding-left: 3px;
+    color: var(--gray700);
+    font-weight: 500;
+  }
 
-.page-cheatmd.dark h2 {
-  color: var(--gray200);
-}
+  .page-cheatmd.dark h2 {
+    color: var(--gray200);
+  }
 
-/* h3 styling */
+  /* h3 styling */
 
-.page-cheatmd h3 {
-  white-space: nowrap;
-  overflow: hidden;
-  margin: 0 0 1em;
-  padding-left: 5px;
-  color: var(--main);
-  font-weight: 400;
-}
+  .page-cheatmd h3 {
+    white-space: nowrap;
+    overflow: hidden;
+    margin: 0 0 1em;
+    padding-left: 5px;
+    color: var(--main);
+    font-weight: 400;
+  }
 
-.page-cheatmd section.h3 {
-  min-width: 300px;
-  margin: 0 0 2em 0;
-  break-inside: avoid;
-  -webkit-column-break-inside: avoid;
-}
+  .page-cheatmd section.h3 {
+    min-width: 300px;
+    margin: 0 0 2em 0;
+    break-inside: avoid;
+    -webkit-column-break-inside: avoid;
+  }
 
-.page-cheatmd h3::after {
-  margin-left: 24px;
-  content: "";
-  vertical-align: middle;
-  display: inline-block;
-  width: 100%;
-  height: 1px;
-  background: linear-gradient(
-    to right,
-    rgba(116, 95, 181, 0.2),
-    transparent 80%
-  );
-}
+  .page-cheatmd h3::after {
+    margin-left: 24px;
+    content: "";
+    vertical-align: middle;
+    display: inline-block;
+    width: 100%;
+    height: 1px;
+    background: linear-gradient(
+      to right,
+      rgba(116, 95, 181, 0.2),
+      transparent 80%
+    );
+  }
 
-/* h4 styling */
+  /* h4 styling */
 
-.page-cheatmd h4 {
-  display: block;
-  margin: 0;
-  padding: 0.25em 1.5em;
-  font-weight: 400;
-  background: var(--gray100);
-  color: #567;
-  border: solid 1px 1px 0 1px var(--gray100);
-}
+  .page-cheatmd h4 {
+    display: block;
+    margin: 0;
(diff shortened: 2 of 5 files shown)


josevalim:
Hi @zolrath!

This is definitely an improvement but I wonder if we could actually leverage more space by printing on horizontal rather than vertical? Ideally we would even remove as much margin as we possibly can and break the page itself in two columns. For example, if we have this:

```
# Page

## One

## Two

## Three

## Four
```

It would be awesome if we render it in A4/letter horizontal like this:

```
# Page

## One         ## Three

## Two         ## Four
```

zolrath:
Sounds good @josevalim, that's doable! 

We can set the page print orientation with
```css
@media print {
  @page {
    size: landscape;
  }
}
```

The only problem I see is that all .css files are imported in `_html.css`.
This means that the `@page` rule would apply to all ExDoc pages, causing the standard documentation to be landscape as well. 

Am I missing something that already exists that would let us add the `@page` rule to only Cheatsheets?

zolrath:
These styles still apply landscape to every page, but here's a pass at forcing the Cheatsheet to landscape so we can see what it will end up looking like. 

It still could use some additional styling but I'd like to make sure this is the layout you're imagining!


For reference, here's the [original css in Landscape.pdf](https://github.com/elixir-lang/ex_doc/files/10045777/Cheatsheet.User.Guide.-.Old.Landscape.pdf)

We've got a couple options for avoiding page breaks; if we avoid page breaks within the h2 elements it keeps topics grouped together but at the cost of making a new page more often.

[New - Don't Avoid Pagebreaks .pdf](https://github.com/elixir-lang/ex_doc/files/10045778/Cheatsheet.User.Guide.-.Landscape.-.No.H2.Avoid.pdf)
[New - With H2 Avoid Pagebreaks.pdf](https://github.com/elixir-lang/ex_doc/files/10045782/Cheatsheet.User.Guide.-.Landscape.-.With.H2.Avoid.pdf)

josevalim:
That looks great! The only last change I would do is to decrease the size a little bit. I printed one version here and it is quite large, I think we can scale everything down a bit (if at all possible).

> These styles still apply landscape to every page, but here's a pass at forcing the Cheatsheet to landscape so we can see what it will end up looking like.

But we can use `@media print` and do this change only on printed pages, right?

zolrath:
> The only last change I would do is to decrease the size a little bit. 

I'll do some experiments with decreasing the font size a few points and push an update.

> But we can use `@media print` and do this change only on printed pages, right?

Yes, forcing landscape only applies to `@media print`!  
I mean that it also forces that orientation when printing other ExDoc pages as well. The other pages definitely seem more suited for portrait printing from what I can tell.

If we could inject the `@media print { @page { size: landscape; } }` rule only on the Cheatsheet pages it would allow normal pages to still be printed normally.

josevalim:
> If we could inject the @media print { @page { size: landscape; } } rule only on the Cheatsheet pages it would allow normal pages to still be printed normally.

Yes, we can do it here:

https://github.com/elixir-lang/ex_doc/blob/main/lib/ex_doc/formatter/html/templates/head_template.eex#L13

Something like this:

```
<%= if page.type == :cheatmd do %>
  ...
<% end %>
```

:)

zolrath:
@josevalim Sounds good! I've updated the PR to add the landscape styling in the template on `.cheatmd` files and reduced the font size.

Comparing the two, the smaller size definitely feels more normal when holding the page, though the larger size was easier to read with the paper on my desk. I suppose it depends on how you'd expect to use them!

[Cheatsheet User Guide - Reduced Size.pdf](https://github.com/elixir-lang/ex_doc/files/10046789/Cheatsheet.User.Guide.-.Reduced.Size.pdf)

josevalim:
Hi @zolrath!

I think the structure is perfect, what is left is to clean up the borders a bit. I believe the border is too dark and sometimes it is duplicated. Here are my suggestions:

1. Remove lines from the h3, h4, or make it very light
2. Remove borders on the left/right of tables and lists and brighten them
3. Make borders in paragraphs in the same color as code?

The idea is to undo less of the original layout in those cases. WDYT?

zolrath:
Sounds good! I've lightened all the borders to be the same as the code blocks and lightened the H3 line.

> 2\. Remove borders on the left/right of tables and lists and brighten them

Hm, is this what you mean by removing the left/right borders? If I'm looking at the dark mode version of cheatsheets I assume everything should be contained in a box, but if I look at the light mode version the contrast is low enough that the boxes disappear a bit more.

image
[No left-right borders.pdf](https://github.com/elixir-lang/ex_doc/files/10048043/no.left-right.borders.pdf)

josevalim:
Yeah, you are right. We should keep the borders on tables, but I think lists look good without them.

two last changes:

1. Please remove the bottom border of lists (if they don’t have anything coming after)
2. please remove the horizontal line followinf headers so we take a look?


Thank you! ❤️

zolrath:
@josevalim Excellent! I've removed the border from the last li of a list and produced two h3 variants.

This should be the final round of questions, I think!

### H3 Line
1. Completely remove the H3 line
image
[Without H3 Lines.pdf](https://github.com/elixir-lang/ex_doc/files/10048098/Without.H3.Lines.pdf)

3. Greatly lighten the H3 line. It still barely shows up during printing but doesn't steal focus.
image
[Light H3 Line.pdf](https://github.com/elixir-lang/ex_doc/files/10048100/Light.H3.Line.pdf)

### Table Cell borders
You can see that I've brought back the table row borders above, but the cells themselves still don't have a border.
This matches the screen styles but I was uncertain if you wanted this or the old behavior of full borders around every cell.

Thanks for the rapid feedback!

josevalim:

josevalim:
Thank you @zolrath! I have merged this and removed the h3 line. :)

I think I have spotted only two places to fix:

1. On Firefox, in the last page, we have this tiny overlap on h2+h3:

Screenshot 2022-11-20 at 11 41 01

2. The bottom line of h4 is still overlapping with the top line of whatever comes after. Is that fixable?

zolrath:
@josevalim Oh hm, the firefox issues are interesting.
Using the Firefox Inspector with the "Toggle print media simulation for the page" button active shows:

## H2 + H3
### Preview
image
### Print
image

## Table Header
### Preview
image
### Print
image

## Table Cell
### Preview
image
### Print
image

Not sure why Firefox in `print.css` preview mode seems to get everything right but turns around and gets it wrong when you actually hit print, I'll have to do some exploration.

josevalim:
Oh, so it is Firefox specific? So don’t spend too much time on it I think. I will release a new version tomorrow! Awesome work!

PR #1632, opened 2022-11-23
Add watch mode to asset build process

DavidOliver:
To help with working on the assets.

When in watch mode, docs are rebuilt after each assets rebuild. I've used [mix commands](https://github.com/elixir-lang/ex_doc/pull/1632/files#diff-bcfc80f32c23faeb609b83289a81441e9c957c11adf7e9ddaef23e7647cf2d69R152-R153), triggered with `child_process.execSync()`. I now wonder if we can detect whether Mix or Rebar is in use and alter the commands as necessary to support both?

HTML and ePub rebuilds are done independently of each other, depending on which assets are changed.

A browser/ePub reader refresh/reload is currently still required, but at least we no longer have to run the `mix build` command for every iteration. Perhaps a mean of auto-refreshing the browser could be added later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,15 +291,15 @@ For more details and configuration options see the [Mermaid usage docs](https://
 
 The easiest way to test changes to ExDoc is to locally rebuild the app and its own documentation:
 
-  1. Run `mix setup` to install all dependencies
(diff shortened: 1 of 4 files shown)


josevalim:
Beautiful!

PR #1638, opened 2022-12-18
Tighten up cheatsheet spacing on small screens and fix h3 line wrapping

zachallaun:
Hi there!

I happened to be browsing some of the new [Ecto cheatsheets](https://hexdocs.pm/ecto/associations.html) on my phone and noticed a couple of things:

- H3s were not wrapping, but instead extending off of the edge of the page
- General excess of whitespace which contributed to some unnecessary/strange wrapping and horizontal scrolling

Decided to take a crack at improving these things slightly and this PR is the result!

---

This PR introduces two CSS variables scoped to `.page-cheatmd`, `--horizontal-space` and `--vertical-space`, defined based on the common usage throughout the stylesheet, and then provides tighter spacing for screens under 600px wide (only chosen because that breakpoint was already being used elsewhere in the file).

It additionally decreases the vertical spacing in some situations on larger screens, but I believe it does so in a way that is more consistent between elements. There are also a handful of special case spacing that I've left alone.

Finally, this PR updates the `mix setup` alias to run `mkdir -p tmp/handlebars` -- I noticed upon first running `mix build` that I was getting an error because this directory didn't exist.

### DIFF
--- a/assets/css/content/cheatsheet.css
+++ b/assets/css/content/cheatsheet.css
@@ -1,4 +1,16 @@
 @media screen {
+  .page-cheatmd {
+    --horizontal-space: 1.5em;
+    --vertical-space: 1em;
+  }
+
+  @media (max-width: 600px) {
+    .page-cheatmd {
+      --horizontal-space: 1em;
+      --vertical-space: .75em;
+    }
+  }
+
   .page-cheatmd .content-inner {
     max-width: 1200px;
   }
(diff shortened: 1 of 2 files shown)


zachallaun:
Before            |  After
:-------------------------:|:-------------------------:
before-full  |  after-full
before-mobile-1 | after-mobile-1
before-mobile-2 | after-mobile-2
before-mobile-long-h3 | after-mobile-long-h3

josevalim:

PR #1640, opened 2022-12-29
Change format for tooltip header

### LINKED ISSUE #1639: Pop-ups (previews) would ideally also use special formatting for code
I think it would be better if we had code formatting in the pop-ups (previews) shown when hovering a link to a function.

See in the image below how "value" and "fun" are written as if they were just text, which of course we can still read, but is somewhat confusing when everywhere else code entities are written in a monospaced font with a special background.

image

solar05:
Fixes #1639 

Result screen for light theme:
image

And for dark:
image


josevalim:

PR #1642, opened 2022-12-31
Add media type for .license files for epub

fhunleth:
This fixes an error when creating epub docs when an assets directory
contains `.license` files. These files are used to specify license and
copyright information to binary files like JPEGs and PNGs. See the
the [REUSE Specification](https://reuse.software) for an example that
documents the use of these files.


josevalim:

PR #1649, opened 2023-01-13
Use full-width for tabs and use variable for border-radius

moogle19:
I updated the tabs in the sidebar to use all available space.

### 2 Tabs
Before:
Screenshot 2023-01-13 at 14 31 13
After:
Screenshot 2023-01-13 at 14 33 02

### 3 Tabs
Before
Screenshot 2023-01-13 at 14 30 54
After:
Screenshot 2023-01-13 at 14 39 00

I also introduced a variable for the `border-radius` to make it's value consistent throughout all files. Since there were some with 3px and some with 4px I am not quite sure which is the "correct" one. I did set it to 4px for now.

### DIFF
--- a/assets/css/custom-props/common.css
+++ b/assets/css/custom-props/common.css
@@ -2,6 +2,7 @@
   /* Layout & Whitespace */
   --content-width: 949px;
   --content-gutter: 60px;
+  --border-radius: 4px;
 
   /* Font Families */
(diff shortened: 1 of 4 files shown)


josevalim:
We don't plan to go ahead with this change for now but if you want to submit a PR for the border radius it would be welcome! Thank you!

PR #1650, opened 2023-01-13
Prevent content-inner overflow (unwanted horizontal scrolling)

DavidOliver:
The issue was showing up on the cheatsheet example page.

### DIFF
--- a/assets/css/layout.css
+++ b/assets/css/layout.css
@@ -145,6 +145,7 @@ body.search-focused .sidebar-search .search-close-button {
   max-width: var(--content-width);
   margin: 0 auto;
   padding: 3px var(--content-gutter);
+  overflow: hidden;
 }
 
 .content-inner:focus {
--- a/assets/css/layout.css
+++ b/assets/css/layout.css
@@ -145,7 +145,7 @@ body.search-focused .sidebar-search .search-close-button {
   max-width: var(--content-width);
   margin: 0 auto;
   padding: 3px var(--content-gutter);
-  overflow: hidden;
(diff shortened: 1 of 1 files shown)


josevalim:
Couldn't we end up hiding something that we wouldn't want? :D

DavidOliver:
As long as all relevant types of elements have a max-width of 100% (or less), which I should check, I don't think hidden/clipped content would be an issue.

I spent some time trying to work out which elements were causing the extra width. I think I managed to narrow it down to a couple of sections by deleting each in turn, but couldn't find any too-wide elements while inspecting, so I eventually resorted to the setting of overflow.

Leave it with me and I'll take another look. I wonder if avoiding an explicit width on the content area would help...

josevalim:
I am also ok with merging this. I guess showing a sidebar is as much as a bug as clipping, and both would only be found visually anyway.

josevalim:
One last question: should it be overflow-y?

DavidOliver:
If one of x and y, x I believe, as it was an unwanted horizontal scrollbar that was appearing. I'll give that a go. :)

josevalim:
Er, yes!

DavidOliver:
`overflow-x` seems to be working as expected.

[`img` does have a `max-width` of `100%`](https://github.com/elixir-lang/ex_doc/blob/ff7f70dcdddc685b8592d228b7f994e7b876ceb2/assets/css/content/general.css#L182), and text in `code` elements scrolls horizontally when necessary, so this change shouldn't affect those kinds of content.

josevalim:

PR #1659, opened 2023-02-03
Erlang: Skip unexported types

the-mikedavis:
EDoc includes the Doc chunks for unexported types, so they are not skipped by the comprehension filter `content != :hidden` in `ExDoc.Retriever`. So unexported types end up being documented by ExDoc but they are hidden in EDoc-generated docs.

This change checks which types are exported by inspecting the abstract code for `export_type` attributes (similar to how `Code.Typespec.fetch_types/1` works) and skips any unexported types.

Connects #1333

### DIFF
--- a/lib/ex_doc/language.ex
+++ b/lib/ex_doc/language.ex
@@ -92,7 +92,7 @@ defmodule ExDoc.Language do
               }
 
   @doc """
-  Returns a map with type information.
+  Returns a map with type information or the atom `:skip`.
 
   The map has the following keys:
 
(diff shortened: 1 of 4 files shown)


josevalim:
Thanks @the-mikedavis!

@erszcz, what do you think? Is this something we should do or should EDoc mark those as private in the chunk?

erszcz:
I think it would be better to handle upstream in EDoc, so that it's taken care of for `shell_docs`, too. I'll prepare a fix. Thanks for raising this!

the-mikedavis:
Sounds good, thank you!

PR #1660, opened 2023-02-11
Remove superfluous import of common custom properties from Erlang CSS

DavidOliver:
`common.css` is imported from `_html.css` and `_epub.css`.


josevalim:
Huh... it seems we were importing common.css more commonly than expected...

josevalim:

DavidOliver:
Yes :) My fault. I must have inadvertently left it there after a reworking.

PR #1662, opened 2023-02-16
Don't attempt parsing code blocks that don't look like modules

### LINKED ISSUE #1661: Codeblocks that contain ? throws a warning
Hi, I am getting this warning when generating a module doc with the following snippet in it. I think there's two issues here. 

1. `nofile` when there was indeed a file that it was in and no additional context. This made it pretty tricky to track down where it was coming from (just started generating docs for the project). 
2. Not being able to use `?` inside code blocks. Is this a bug or something by design?

Warning
```
warning: found ? followed by code point 0x20 (space), please use ?\s instead
  nofile:1:15
```

Snippet of docs
```md
    - `FOR UPDATE OF ? SKIP LOCKED` is used to ...
 ```

### PR DESCRIPTION
**Pull request by wojtekmach**
Closes #1661.

### DIFF
--- a/lib/ex_doc/language/elixir.ex
+++ b/lib/ex_doc/language/elixir.ex
@@ -611,10 +611,14 @@ defmodule ExDoc.Language.Elixir do
   end
 
   defp parse_module(<<first>> <> _ = string, _mode) when first in ?A..?Z do
-    do_parse_module(string)
+    if string =~ ~r/[a-z0-9_]+/u do
+      do_parse_module(string)
+    else
+      :error
+    end
   end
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/ex_doc/language/elixir.ex [hunk 1]
```diff
@@ -611,10 +611,14 @@ defmodule ExDoc.Language.Elixir do
   end
 
   defp parse_module(<<first>> <> _ = string, _mode) when first in ?A..?Z do
-    do_parse_module(string)
+    if string =~ ~r/[a-z0-9_]+/u do
```
josevalim, line 614: Shouldn't we consider `.` and stuff like `HTML` where everything is upcase?

josevalim (approved):

PR #1665, opened 2023-02-23
Update Cheatsheet print styles

DavidOliver:
Main changes:

- all background color removed in order to avoid potential photocopying issues;
- more color from the main styles' text preserved in order to keep more character;
- H3 horizontal rules preserved, at 2px instead of 1px to differentiate from 1px tables and code boxes;
- paragraphs' and lists' borders removed, and tables' left/right borders removed, in order to lessen the "boxiness" (this allows for the H3 rules to make more visual sense).

The "screen" media query is removed from the main cheatsheet styles, allowing them to apply to the printed version in order to reduce repetition of setting font, color, etc.

Increases use of .content-inner selector namespace to be safe.

## Remaining to do

The horizontal rules next to H3 text are achieved via background, and so only appear when the 'print backgrounds' option is selected. We should see if we can convert to non-background content.

### Before

image

### After

image

### DIFF
--- a/assets/css/content/cheatsheet.css
+++ b/assets/css/content/cheatsheet.css
@@ -1,297 +1,297 @@
-@media screen {
-  .page-cheatmd {
-    --horizontal-space: 1.5em;
-    --vertical-space: 1em;
-  }
+.page-cheatmd .content-inner {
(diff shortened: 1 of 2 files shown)


josevalim:

PR #1666, opened 2023-02-23
Fix visited link color in admonition blocks

zachallaun:
Visited link styles were not being overridden by blockquotes, which caused visited links to appear with very low contrast.

Before:

block-link-visited-before

After:

block-link-visited-after


josevalim:

Issue #1668, opened 2023-03-02
Very low contrast in sidebar


cdegroot:
![image](https://user-images.githubusercontent.com/84811/222322344-e73faab3-ed66-405d-8ecc-fe8a5f254cba.png)

This is a screenshot from Linux/KDE/Firefox. It might be the font rendering, but that contrast is low to the point of it being unreadable to me (some details [here](https://github.com/livebook-dev/livebook/issues/238#issuecomment-1451222828)). 

I've experimented with a styling browser plugin and making things just white-on-<whatever the background color is> with making the currently highlighted entries bolder works very nicely - a much better contrast than the marginal current one (I think it doesn't hit AA) and still looks pleasant. 

Ideally, light mode would be "actually light mode" with a dark-on-light sidebar, but that would be a large change. I'm willing to come up with a color scheme if this would be an acceptable solution although I'm not a designer and, frankly, given that this is code documentation and not a marketing/PR site, I'd end up with lots of shades of grey suitably spaced apart and call it a day :)

(I usually like to submit PRs over Issues but this probably warrants some discussion before making changes).

josevalim:
There is definitely a font issue, so I would try to address that first. Any ideas why the font rendering is so faint on Linux?

cdegroot:
Well... yes and no for the font issue, of course. If you design with a low-ish contrast and a very thin font, you're walking near the edge and you shouldn't be surprised if you stumble and fall over it :-). I'm suggesting to move well away from the ledge instead of doing a deep dive into font rendering issues.

I'm not sure whether it's the font, the copy of the webfont in particular, the Linux/X11 rendering engine, or the settings on my machines (I'm seeing this both on my laptop, a FHD 15" panel, and my desktop, a 4K 32" panel). I do know that I can work with pretty much everything else and that a true light mode would fix this cross-platform and increase ex_doc's accessibility score. I did try changing to a locally installed font on my main box (Lucida Sans, I think), and that got indeed rendered a bit better. 

Anyway, back to what I think is the better solution; not suggesting this is any good, but I did some tinkering using a styling browser plugin, and something like

![image](https://user-images.githubusercontent.com/84811/222331321-1bcf388c-866f-4cc8-8d43-2730db08f758.png)

[...]

PR #1672, opened 2023-03-04
Make sidebar CSS use semantic vars

cdegroot:
Closes https://github.com/elixir-lang/ex_doc/issues/1668

I decided to do a relatively clean implementation of the sidebar, keeping the color scheme with neutral greys plus the language accent colors. Most work was making all the CSS for the sidebar use semantically-named variables which I think is a good idea anyway. 

Disclaimer: I'm not a designer, which is mainly why I stuck with the existing color definitions. I do think that I created a reasonably faithful "inverse" of the current dark-mode-only sidebar, though. More than happy to see tweaks, of course.

Note: the highlighted item still doesn't pop out due to the too-thin rendering of Lato on Linux/X11. However, in this mode it only impacts how much it stands out, not the readability itself so it is less of an issue.

image

### DIFF
--- a/assets/css/custom-props/_erlang.css
+++ b/assets/css/custom-props/_erlang.css
@@ -2,6 +2,7 @@
   --main:               hsl(0, 100%, 64%);
   --main-darkened-10:   hsl(0, 100%, 54%);
   --main-darkened-20:   hsl(0, 100%, 44%);
+  --main-darkened-50:   hsl(0, 100%, 14%);
   --main-lightened-05:  hsl(0, 100%, 69%);
   --main-lightened-10:  hsl(0, 100%, 74%);
 }
--- a/assets/css/custom-props/theme-light.css
+++ b/assets/css/custom-props/theme-light.css
@@ -59,25 +59,25 @@
 
   --success:                     var(--green);
 
+  --mark: red; /* TODO removeme */
+
   --sidebarButtonBackground:     linear-gradient(180deg,
                                                  var(--gray900) 20%,
                                                  var(--gray900-opacity-50) 70%,
                                                  var(--gray900-opacity-0) 100%);
-  --sidebarAccentMain:           var(--gray50);
-  --sidebarBackground:           var(--gray800);
+  --sidebarAccentMain:           var(--gray800);
+  --sidebarBackground:           var(--gray50);
   --sidebarGradient:             linear-gradient(var(--sidebarBackground),
                                                  var(gray800-opacity-0));
-  --sidebarHeader:               var(--gray700);
-  --sidebarMuted:                var(--gray300);
-  --sidebarHover:                var(--white);
+  --sidebarHeader:               var(--gray100);
+  --sidebarMuted:                var(--gray600);
+  --sidebarHover:                var(--main-lightened-10);
   --sidebarScrollbarThumb:       var(--coldGray);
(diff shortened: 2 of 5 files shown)


josevalim:
Hi @cdegroot! This is a 1-by-1 translation, right? So this looks good to me. There is just one TODO that you added and we can ship it?

cdegroot:
Is there a todo? I missed that then :). Will check later today and clean that up. And give it a review in daylight to see whether colors want tweaking.

cdegroot:
Fixed todo. And yes, it's an 1-1 translation. I followed a two step process:
* Move all the colors in sidebar.css to semantically named variables in the dark mode theme css
* Copy the resulting list of variables over to the light mode theme css and tweak until things look right. 

I kept the "language accent colors" so Elixir and Erlang still have their purple/orange accents.

Even if this results in follow-up discussions/tweaks, it should now be much simpler to work on. 

Screenshot after the tweak of commit 17f989b

image

assets/css/custom-props/_erlang.css
```diff
@@ -2,6 +2,7 @@
   --main:               hsl(0, 100%, 64%);
   --main-darkened-10:   hsl(0, 100%, 54%);
   --main-darkened-20:   hsl(0, 100%, 44%);
+  --main-darkened-50:   hsl(0, 100%, 14%);
```
josevalim, line 5: This does not seem to be used anywhere.

assets/css/custom-props/theme-light.css
```diff
@@ -59,6 +59,23 @@
 
   --success:                     var(--green);
 
-  --sidebarButton:               var(--gray50);
-  --sidebarButtonBackground:     linear-gradient(180deg, var(--white) 20%, var(--white-opacity-50) 70%, var(--white-opacity-0) 100%);
+  --sidebarButtonBackground:     linear-gradient(180deg,
+                                                 var(--gray900) 20%,
+                                                 var(--gray900-opacity-50) 70%,
+                                                 var(--gray900-opacity-0) 100%);
```
josevalim, line 65: It seems you changed the color compared to the previous theme.

assets/css/sidebar.css
```diff
@@ -442,15 +442,15 @@
...
+  color: var(--sidebarHover);
 }
 
 .sidebar-button:is(:active, :hover, :focus) {
   outline: none;
 }
 
 .sidebar-button {
-  color: var(--sidebarButton);
+  color: var(--sidebarAccentMain);
```
josevalim, line 453: This was `--gray50` for light mode but now it is `--gray800`.

josevalim (changes requested):
Thank you for the PR. I did an initial pass and it seems the values are changing in some occasions. Can you please double check all vars to guarantee there is no change in behaviour?

josevalim:
Sorry, I got confused. I thought this was only an extraction of the sidebar into variables for easier change by plugins. But it is a new sidebar color altogether. We don't plan to merge this, see https://github.com/elixir-lang/ex_doc/issues/1668.

cdegroot:
So is it acceptable if I just push the first commit? That makes it easier to override styling using things like the Stylus browser plugin but has no visible changed.

josevalim:
Yes, I would be glad to accept that, as long as no changes in behaviour!

PR #1673, opened 2023-03-04
Lato => Open Sans, closes #1668

### LINKED ISSUE #1668: Very low contrast in sidebar

josevalim:
@DavidOliver can you let me know if this gives a non-insignificant improvement? :)


DavidOliver:
:) See #1675.

PR #1677, opened 2023-03-09
Improve consistency of type rendering weight across OSes

### LINKED ISSUE #1675: Text rendering: weight and consistency across OSes
@josevalim, further to #1668, I've tested in Safari in Mac OS (Safari 16, Mac OS Ventura), and using the following CSS does seem to make all text thinner, contrary to what you found the other day:

```css
body {
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
}
```

(I find that I often need to empty the `_build` directory for newly-built assets to be included in newly-generated docs; perhaps this prevented you from seeing the change?)

This results in type rendering being very closely matched in Mac OS and Linux. Left Safari/Mac OS with new props, right Linux (with new props, but unaffected and so as before):

image

Now that I have both rendering the same/similarly, I should be able to adjust the weights of text in the sidebar (light to regular), and some other places as appropriate, to bring both into line with the current rendering in Mac OS.

One thing to note before I proceed, however, is that the serif font used for the main text (Merriweather, 400/normal) does not have a 500 weight that we can move up to - only 700/bold, which is not an option as that does not render well in either Mac OS or Linux. So the main text would look slightly thinner than you've been used to in Mac OS:

image

Is this acceptable to you? I imagine it may initially feel a little thin to you, but I believe this rendering to be a little clearer and nearer a normal, not semi-bold, weight.

Although applying the above approach across the board is my preference to maximise simplicity and consistency, if you don't like the lighter serif font for the main text, we could limit the new properties to the sidebar and allow the main text to be different (but readable) across OSes.

DavidOliver:
Fixes #1675

As the sidebar type is thinned in Mac OS via the alternative anti-aliasing, its weight can be increased from light to normal to avoid overly thin type in Linux and Windows.

The thin weight is preserved/used in places where differentiation from normal/bold is warranted.

Additionally, the normal/400 weight font was not being loaded, which resulted in type set to normal being rendered light. We're now loading light/300, normal/400 and bold/700 for the Lato sans serif font.

# Screenshots

All Firefox/Ubuntu, before followed by after. (@josevalim to make sure he's happy with Safari/WebKit in Mac OS.)

## Light: Sidebar

Screen Shot 2023-03-09 at 15 25 32
Screen Shot 2023-03-09 at 15 14 25

## Light: Settings

Screen Shot 2023-03-09 at 15 26 02
Screen Shot 2023-03-09 at 15 14 48

## Dark: Sidebar

(Same as light theme sidebar.)

Screen Shot 2023-03-09 at 15 26 08
Screen Shot 2023-03-09 at 15 15 03

## Dark: Settings

Screen Shot 2023-03-09 at 15 26 11
Screen Shot 2023-03-09 at 15 15 08

### DIFF
--- a/assets/css/_html.css
+++ b/assets/css/_html.css
@@ -1,4 +1,5 @@
 @import '@fontsource/lato/300.css';
+@import '@fontsource/lato/400.css';
 @import '@fontsource/lato/700.css';
 @import '@fontsource/merriweather/300.css';
 @import '@fontsource/merriweather/300-italic.css';
@@ -39,4 +40,10 @@
 body:not(.dark) .content-inner img[src*="#gh-dark-mode-only"],
 body.dark .content-inner img[src*="#gh-light-mode-only"] {
   display: none;
+}
+
+/* Improves consistency of type rendering across OSes; results in thinner type in Mac OS */
+body {
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
 }
--- a/assets/css/_html.css
+++ b/assets/css/_html.css
@@ -41,9 +41,3 @@ body:not(.dark) .content-inner img[src*="#gh-dark-mode-only"],
 body.dark .content-inner img[src*="#gh-light-mode-only"] {
   display: none;
 }
-
-/* Improves consistency of type rendering across OSes; results in thinner type in Mac OS */
-body {
-  -webkit-font-smoothing: antialiased;
-  -moz-osx-font-smoothing: grayscale;
-}
--- a/assets/css/settings.css
+++ b/assets/css/settings.css
@@ -43,6 +43,7 @@
 
 #settings-modal-content .switch-button-container > div > p {
   font-size: 14px;
+  font-weight: 300;
   line-height: 1.4;
   margin: 0;
   padding-bottom: 6px;
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -124,7 +124,6 @@
 
 .sidebar .sidebar-listNav :is(li, li a) {
   text-transform: uppercase;
-  font-weight: 300;
   font-size: 14px;
   color: var(--sidebarMuted);
 }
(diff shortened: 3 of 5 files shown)


josevalim:
Thank you @DavidOliver ! 

Yeah, unfortunately the regular text appears to thin on Mac now :( So we have two options:

1. Change the font for the regular text
2. Check if the user agent is a Mac and load heavier fonts if not

josevalim:
I think second option may be simpler. If Mac, we set font-weight in the sidebar to 300, otherwise 400, and we don't modify font smoothing in any case.

DavidOliver:
Screenshots in Safari/Mac OS for reference follow.

Before:

Screen Shot 2023-03-09 at 17 06 01

After:

Screen Shot 2023-03-09 at 17 06 24

I'm also having doubts about grayscale anti-aliasing now. The subpixel anti-aliasing (the default, as before) seems to result in more refined tracking/spacing between letters, as well as better letter shapes, in Mac OS.

It seems a shame to resort to UA string sniffing, but maybe that's the only way we're going to get the results we'd like. It looks like we should check whether or not 'Macintosh' is in the string.

So:

- No anti-aliasing modification.
- Load both 300 and 400 Lato font weights. (Both are used.)
- Font weight choices set with Mac OS in mind by default.
- If 'Macintosh' is not present in UA string, switch certain instances of Lato text, such as sidebar links, from 300 to 400.

Does that look right to you?

josevalim:
Yes! I am also fine to be safe and render with heavy font by default and convert to 300 if MacOS. I think we will only need this change on the sidebar. Can you think about any other place we may need to change?

Thank you! :heart:

josevalim:
If we can use the Elixir website as representative:

image

Which reminds me we likely want to detect for iOS as well.

DavidOliver:
Ha! About 50/50 Apple/others then. Thanks for the info.

I think iOS is covered by 'Macintosh', but I'll check.

DavidOliver:
New approach as discussed pushed.

Safari/Mac OS:

Screen Shot 2023-03-11 at 12 57 52

Firefox/Linux (representing non-Apple):

image

josevalim:
Amazing, thank you so much! :heart:

PR #1679, opened 2023-03-11
Fix precedence of switches vs config

starbelly:
Currently config takes precedence over switches. This was discussed in build and packaging working group and it was concluded this is unexpected behaviour. This PR rectifies that.


josevalim:

PR #1680, opened 2023-03-13
Propagate :since metadata from modules

whatyouhide:
`:since` was showing for functions, types, and callbacks, but not for modules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ ExDoc supports metadata keys in your documentation.
 
 In Elixir, you can add metadata to modules and functions.
 
-For a module, use `@moduledoc`:
+For a module, use `@moduledoc`, which is equivalent to adding the annotation to everything inside the module (functions, macros, callbacks, types):
 
 ```elixir
 @moduledoc since: "1.10.0"
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -37,6 +37,10 @@
   font-weight: normal;
 }
 
+.content-inner h1 .note {
+  font-family: var(--monoFontFamily);
+}
+
 .content-inner h2 {
   font-size: 1.6em;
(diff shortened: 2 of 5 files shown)


josevalim:
Can you please provide screenshots of how it looks like? IIRC we didn't have a placement for it.

whatyouhide:
Will do in a few hours when at a computer, but yes it turns our we do 😬, and it looks good too 😁 I tested locally.

whatyouhide:
@josevalim see screenshot 👇 

CleanShot 2023-03-13 at 17 47 51@2x

josevalim:
I think the font is not matching the font we have in usual signatures, no?

whatyouhide:
@josevalim I don't know 😄 Should it be monospaced?

CleanShot 2023-03-13 at 19 55 19@2x

josevalim:
I think so, yes!

whatyouhide:
@josevalim new commit should do it:

CleanShot 2023-03-13 at 20 14 50@2x

josevalim:
Can you also please adjust the baseline? The fonts must all start at the same place from the bottom.

Btw, this is what I meant we had no place for it. There is no design for them, so we need to take into account all of the design considerations that happen elsewhere on the page.

whatyouhide:
@josevalim got it now. If I align the baseline with the module name, it looks pretty bad because it's then not aligned with the `</>` and gear icons anymore 🤔 An alternative is to put this on a separate line, before the moduledoc starts.

josevalim:
You can also love the icons baseline as well. The since and icons should have the same baseline anyway as the title anyway. :)

whatyouhide:
@josevalim done, and the implementation is really simple as well. Ready for another review.

josevalim:

PR #1682, opened 2023-03-16
Keep language class on livebook output code block

jonatanklosko:
Currently we are omitting the information if the output code block contains special content like mermaid graph.


josevalim:

PR #1686, opened 2023-04-07
Bug fix: properly show the modules in the left pane when using external libraries (such as KaTeX)

woodward:
Hi!

I added the latest ex_doc and KaTeX to a project recently, and ran into problems; this pull request is a fix for the problems that I encountered.  Basically, once I had added KaTeX to the project, the modules would no longer show up in the left-hand pane.  If one opens up the JS console, you see an error about an event listener that is attempting to be added to an HTML element with ID "toast", but this element doesn't yet exist on the page at that time (it does later).  So the fix is simple; only attempt to add the event listener if the HTML element `toast` already exists. 

I created a sample repo here which shows the problem (along with the fix):  

https://github.com/woodward/shows_ex_doc_error

Note that the bug was introduced in ExDoc 0.28.6 (i.e., 0.28.5 works fine). If you clone this sample repo, generate the docs (via `mix docs`) and then open `doc/index.html` you should see some math displayed in the main module docs.  If you can also see the module names in the left-hand pane, then the bug is fixed (which it is with the commits in this fork).

You can see the KaTeX setup in the mix file, and you can also try out various versions of ex_doc by commenting out the deps in the mix file (including this fork).

Please give me a holler if you run into any questions. 

-- Greg


woodward:
Thank you for merging in my pull request, @josevalim .  I'm looking at your [subsequent commit](https://github.com/elixir-lang/ex_doc/commit/2c581239e0989fdc91e3c83b4ce28ef4fe0adda6) ; thank you for a much better solution than my fix!  I realized at the time I made my commit that my "fix" was really just a band-aid so that I could get past the problem and to get my KaTeX to render (and the modules to show properly), but that it most likely was not the best nor the correct approach (as it left toast in an un-toasted state 😄 ).  The correct solution required a knowledge of the internals of ExDoc that I didn't possess, though.  Thank you for taking the time to improve my band-aid fix and to fix it properly!

-- Greg

josevalim:

PR #1687, opened 2023-04-08
Update the README - bump KaTeX version from 0.13.19 to 0.16.4, and update KaTeX config

woodward:
Hi:

This commit just updates the ExDoc README for KaTeX.  The KaTeX version is bumped from 0.13.19 to 0.16.4 (latest), and a couple of additional things are added to the config. The KaTeX fonts are loaded, a plugin which lets you copy/paste TeX from the rendered browser page was added ([to see how it works, see this web page](https://upupming.site/katex-copytex/)), and the minimal configuration for the auto-render plugin was added.

Thanks!

-- Greg

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,18 +265,18 @@ If you write TeX-style math in your Markdown, such as `$\sum_{i}^{N} x_i$`, it e
 ```html
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.16.4/dist/katex.min.css" integrity="sha384-vKruj+a13U8yHIkAyGgK1J3ArTLzrFGBbBc0tDp4ad/EyewESeXE/Iv67Aj8gKZ0" crossorigin="anonymous">
 <script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.4/dist/katex.min.js" integrity="sha384-PwRUT/YqbnEjkZO0zZxNqcxACrXe+j766U2amXcgMg5457rve2Y7I6ZJSm2A0mS4" crossorigin="anonymous"></script>
-<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.4/dist/contrib/auto-render.min.js" integrity="sha384-+VBxd3r6XgURycqtZ117nYw44OOcIax56Z4dCRWbxyPt0Koah1uHoK0o4+/RRE05" crossorigin="anonymous"></script>
 
-<script>
-  document.addEventListener("DOMContentLoaded", function () {
-    renderMathInElement(document.body, {
-      delimiters: [
+<link href="https://cdn.jsdelivr.net/npm/katex-copytex@1.0.2/dist/katex-copytex.min.css" rel="stylesheet" type="text/css">
+<script src="https://cdn.jsdelivr.net/npm/katex-copytex@1.0.2/dist/katex-copytex.min.js" crossorigin="anonymous"></script>
+
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.4/dist/contrib/auto-render.min.js" integrity="sha384-+VBxd3r6XgURycqtZ117nYw44OOcIax56Z4dCRWbxyPt0Koah1uHoK0o4+/RRE05" crossorigin="anonymous"
+  onload="renderMathInElement(document.body, {
+    delimiters: [
         {left: '$$', right: '$$', display: true},
         {left: '$', right: '$', display: false},
-      ],
-      throwOnError : false
-    });
-  });
+    ],
+    throwOnError : false
+  });"></script>
 </script>
 ```
 `
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ If you write TeX-style math in your Markdown, such as `$\sum_{i}^{N} x_i$`, it e
   });"></script>
 </script>
 ```
-`
+
 For more details and configuration options, see the [KaTeX Auto-render Extension](https://katex.org/docs/autorender.html).
 
 ### Rendering Vega-Lite plots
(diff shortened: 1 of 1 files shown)


josevalim:
Do we need to load the fonts? How does it work currently without listing them?

woodward:
That is a great question, @josevalim ; I had pondered whether to include them or not.  I thought they might be necessary as I saw those lines listed when [viewing the source on the KaTeX home page](view-source:https://katex.org/).  However, perhaps they are used by the webfont loader plugin listed on line 33 there. I deleted the font lines from the KaTeX `ex_doc config`, cleared out my browser cache, and regenerated the docs for my [sample repo](https://github.com/woodward/shows_ex_doc_error).  It seems to render the math equations ok, and then when I view a math element in the developer tools, I see it references the KaTeX fonts, so I guess they got pulled down ok.  So I've added a commit to this pull request that deletes the font lines -- good call on your part.

The other item which might be viewed as not completely necessary for a minimal setup is the copytex plugin, and I'm happy to delete those lines if you'd prefer a minimal configuration.  It sure is nice as a "quality of life" thing, though, as it's great to be able to copy/paste from the rendered math in the browser and to get the actual TeX rather than a character version of the font files.  Again, though, I'm happy to remove it if you'd prefer a more minimal config in the docs (and I can always just include it in my own project).  It's your call as to which way you'd prefer.

-- Greg

README.md
```diff
@@ -263,11 +263,26 @@ docs: [
...
-<script defer src="https://cdn.jsdelivr.net/npm/katex@0.13.19/dist/contrib/auto-render.min.js" integrity="sha384-+XBljXPPiv+OzfbB3cVmLHf4hdUFHlWNZN5spNQ7rmHTXpd7WvJum6fIACpNNfIR" crossorigin="anonymous" onload="renderMathInElement(document.body);"></script>
-```
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.16.4/dist/katex.min.css" crossorigin="anonymous">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex-copytex@latest/dist/katex-copytex.min.css" crossorigin="anonymous"> 
+
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.4/dist/katex.min.js" crossorigin="anonymous"></script>
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.4/dist/contrib/copy-tex.min.js" crossorigin="anonymous"></script>
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.4/dist/contrib/auto-render.min.js" integrity="sha384-+VBxd3r6XgURycqtZ117nYw44OOcIax56Z4dCRWbxyPt0Koah1uHoK0o4+/RRE05" crossorigin="anonymous"></script>
 
+<script>
```
josevalim, line 273: I think we should still use the onload as before. That should be enough!

woodward: ok, @josevalim , I removed the copytex plugin.  I also added in the integry sha's for katex.min.css and katex.min.js as they were not present, and seem like a good thing to have.  Thanks!

-- Greg

josevalim: I think copytex was fine. :) The pending biggest question is if we should revert to `onload=` as before .

woodward: ahhh, I see; I misinterpreted your comment, @josevalim !  I've added back in copytex (which admittedly is super handy to have! 😄 ) and switched back to using `onload=`  I had changed it to the current form (using `document.addEventListener("DOMContentLoaded", ...`) just so that it matched up with the other sections in the README (about Mermaid and Vega-Lite), and I was assuming that `onload=` and `document.addEventListener("DOMContentLoaded", ...` were functionally equivalent (but perhaps they're not?).

-- Greg

woodward: Note that I'm happy to go with either `onload` or `document.addEventListener("DOMContentLoaded"`, whichever you prefer (I thought you meant to go with `onload`, but now that I'm re-reading your comment I'm not 100% sure that's what you meant).  I also pinned the versions on the copytex plugin just to hopefully avoid future issues if things change and break (thinking back to the problems we had a few years ago with phoenix & npm before it switched over to using esbuild).

josevalim: Let’s go with onload cause that’s what Katex recommend on their website. The closer the better :)

woodward: roger that; the current version in my fork is using the `onload`.  Agreed; the closer with the KaTeX docs the better.  😄

josevalim: Sorry, do we need to pass all arguments? Why can’t it be like before? :)

woodward: you're correct; I think I had added in the arguments at one point when I was debugging why it wasn't rendering properly, and then erroneously assumed they were necessary since at some point it suddenly started working. I've removed the arguments and it seems to still work fine. So they're gone now; good catch.

woodward: oops, now I see why I had added in the args for `renderMathInElement`.  It's necessary for the single dollar sign case (i.e., an equation that's inline).  For example, if you have a paragraph where you say $ F = M A $ - this only renders properly if you have the args to `renderMathInElement`.  And it seems with KaTeX that if you add in the arg for $, then you have to also add in the arg for $$.

README.md
```diff
@@ -263,9 +263,21 @@ docs: [
...
+<link href="https://cdn.jsdelivr.net/npm/katex-copytex@1.0.2/dist/katex-copytex.min.css" rel="stylesheet" type="text/css">
+<script src="https://cdn.jsdelivr.net/npm/katex-copytex@1.0.2/dist/katex-copytex.min.js" crossorigin="anonymous"></script>
+
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.4/dist/contrib/auto-render.min.js" integrity="sha384-+VBxd3r6XgURycqtZ117nYw44OOcIax56Z4dCRWbxyPt0Koah1uHoK0o4+/RRE05" crossorigin="anonymous"
+  onload="renderMathInElement(document.body, {
+    delimiters: [
+      {left: '$$', right: '$$', display: true},
+      {left: '$', right: '$', display: false},
+    ],
+    throwOnError : false
```
josevalim, line 278:
```suggestion
    ]
```

josevalim:

PR #1703, opened 2023-05-15
Add Jump to Latest Button

Ccasper1224:
Adds a "Jump to Latest" button to the sidebar that takes you to the most recent docs.

image

On Hover
image

### DIFF
--- a/assets/js/handlebars/templates/versions-dropdown.handlebars
+++ b/assets/js/handlebars/templates/versions-dropdown.handlebars
@@ -6,4 +6,7 @@
       </option>
     {{/each}}
   </select>
+  {{#unless nodes.0.isCurrentVersion}}
+    <button type="button" class="sidebar-projectLatestVersionButton" value="{{nodes.0.version}}">Jump to Latest</button>
+  {{/unless}}
 </form>
--- a/assets/js/handlebars/templates/versions-dropdown.handlebars
+++ b/assets/js/handlebars/templates/versions-dropdown.handlebars
@@ -6,7 +6,11 @@
       </option>
     {{/each}}
   </select>
-  {{#unless nodes.0.isCurrentVersion}}
-    <button type="button" class="sidebar-projectLatestVersionButton" value="{{nodes.0.version}}">Jump to Latest</button>
-  {{/unless}}
+  {{#each nodes}}
+    {{#if latest}}
+      {{#unless isCurrentVersion}}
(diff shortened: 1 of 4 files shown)


assets/js/handlebars/templates/versions-dropdown.handlebars
```diff
@@ -6,4 +6,7 @@
       </option>
     {{/each}}
   </select>
+  {{#unless nodes.0.isCurrentVersion}}
+    <button type="button" class="sidebar-projectLatestVersionButton" value="{{nodes.0.version}}">Jump to Latest</button>
+  {{/unless}}
```
wojtekmach, line 11: I believe there is an edge case that would work unexpectedly. Say we used this to render Elixir 1.14.4 docs. When there will be 1.15.0-rc.0 it would be at the top of the list, `nodes.0`, but it should _not_ be considered _latest_.

This would definitely be an issue for Postgrex:

image

There's an old issue for hex docs around this https://github.com/hexpm/hexdocs/issues/10, btw.

Ccasper1224: Oof, Did not realize that.

Would it be alright to add a `latestVersionNode` or `latestVersionUrl` variable to `docs_config.js` to provide the relevant information?
Would need to update hexdocs to add that variable as well, but thats easy enough.

That would allow explicit selection of which docs are latest, while still allowing pre-releases in the list.

wojtekmach: Variable sounds good to me. Alternatively we could add `"latest":true` pair like this; before:

```js
[
  {"url":"https://hexdocs.pm/foo/1.0.0","version":"v1.0.0"},
  {"url":"https://hexdocs.pm/foo/0.1.0","version":"v0.1.0"}
]
```

after:

```js
[
  {"url":"https://hexdocs.pm/foo/1.0.0","version":"v1.0.0","latest":true},
  {"url":"https://hexdocs.pm/foo/0.1.0","version":"v0.1.0"}
]
```

The advantage is if this key is not defined, for packages which haven't been published recently, we can gracefully handle that as opposed to getting undefined variable error in JS. But maybe it's fine like you proposed. cc @josevalim @ericmj

josevalim:
Closing this for now, as it requires changes upstream first. Once that is in place, I will be glad to reopen this.

PR #1705, opened 2023-05-16
Add doc markdown to search_items.json

wojtekmach:
This is step 1 of search_items.json improvements. In extras we extract sections (`<h2>` headers) but we don't do that in `@moduledoc/@doc` yet. That's what we will do in step 2.


lib/ex_doc/formatter/html/search_items.ex
```diff
@@ -2,64 +2,85 @@ defmodule ExDoc.Formatter.HTML.SearchItems do
...
     )
   end
 
   defp encode(ref, title, type, doc) do
-    %{ref: URI.encode(ref), title: title, type: type, doc: clean_doc(doc)}
+    %{
+      ref: URI.encode(ref),
+      title: title,
+      type: type,
+      doc: clean_doc(doc)
```
josevalim, line 75: We should only clean this doc for HTML. For Markdown, let's keep everything as is?

josevalim: Maybe we should even move this to the language bits? They both have to return the "html_doc" and the "text_doc/search_doc"?

wojtekmach: it's rare but markdown may contain html tags and we shouldn't render them on the search page.

josevalim: Hrm... I am worried about removing something we should not when we strip the white-space. Maybe we strip only the tags for Markdown?

josevalim (approved):

lib/ex_doc/formatter/html/search_items.ex
wojtekmach: Agreed, for markdown we just clean tags from markdown, for Erlang we strip tags & clean whitespace from _html_.

I'll look into moving this behaviour to the language contract later, after removing duplication from sidebar/search generation.

PR #1707, opened 2023-05-21
Safety check before running File.rm_rf! in doc gen

### LINKED ISSUE #1555: ExDoc could potentially wipe out any directory
While working on improving the `make docs`  command in Elixir, I accidentally discovered this bug that deleted my entire Elixir repo locally loosing all local branches. I don't know if there could be any implication of anybody running ExDoc, but it could potentially wipe out the entire disk in a remote or a local user machine.

When `--output`  is provided to the ex_doc command, and it is empty, it will deleted  `--output ""` `--output ~` will wipe out your home dir.
My case was that `--output` was set to `"${VAR}"`  and `VAR` didn't get set properly and was evaluated to an empty string, so the project root directory was gone.

Additionally, I think we should review the use of File.rm_rf! across the project.
https://github.com/elixir-lang/ex_doc/search?q=File.rm_rf%21

These are the lines causing such behaviour.
https://github.com/elixir-lang/ex_doc/blob/6ba40a7383f87727c49c1fc51ddd5cf192ef3e7a/lib/ex_doc/formatter/html.ex#L159-L160

This removing of the output directory is intended to clean out old documentation copies. If a .build file in found, only files listed in .build are removed.

viniciusmuller:
This pull requests adds a safety check before running `Fille.rm_rf!` when creating docs.

Now it will only allow users to select as output either:
- Unexisting directories (which will be created by ex_doc)
- Directories already managed by ex_doc (have the `.ex_doc` file)

One I think inevitable outcome of this is that this is a breaking change, users with existing `docs` folders that try to run it in a new version will receive the error message.

Maybe we can add a line in the error saying `Run "rm -rf docs" if you are seeing this after upgrading ex_doc`

Closes #1555

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -147,18 +147,38 @@ defmodule ExDoc.Formatter.HTML do
   end
 
   defp output_setup(build, config) do
-    if File.exists?(build) do
-      build
-      |> File.read!()
-      |> String.split("\n", trim: true)
-      |> Enum.map(&Path.join(config.output, &1))
-      |> Enum.each(&File.rm/1)
+    safety_path = Path.join(config.output, ".ex_doc")
 
-      File.rm(build)
-    else
-      File.rm_rf!(config.output)
-      File.mkdir_p!(config.output)
+    cond do
+      File.exists?(build) ->
+        build
+        |> File.read!()
+        |> String.split("\n", trim: true)
+        |> Enum.map(&Path.join(config.output, &1))
+        |> Enum.each(&File.rm/1)
+
+        File.rm(build)
+
+      File.exists?(safety_path) ->
+        File.rm_rf!(config.output)
+        mkdir_output!(config.output)
+
+      not File.exists?(config.output) ->
+        mkdir_output!(config.output)
+
+      true ->
+        raise """
+        ex_doc cannot output to #{config.output}:
+        Directory already exists and is not managed by ex_doc.
+
+        Try giving an unexisting directory as `--output`.
+        """
     end
+  end
+
+  defp mkdir_output!(path) do
+    File.mkdir_p!(path)
+    File.touch!(Path.join(path, ".ex_doc"))
   end
 
   defp generate_build(files, build) do
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -16,7 +16,7 @@ defmodule ExDoc.Formatter.HTML do
     config = %{config | output: Path.expand(config.output)}
 
     build = Path.join(config.output, ".build")
-    setup_output(config, &cleanup_output_dir/2, &create_output_dir/2)
+    setup_output(config.output, &cleanup_output_dir(&1, config), &create_output_dir(&1, config))
 
     project_nodes = render_all(project_nodes, ".html", config, [])
     extras = build_extras(config, ".html")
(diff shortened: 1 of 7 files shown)


lib/ex_doc/formatter/html.ex
```diff
@@ -147,20 +147,40 @@ defmodule ExDoc.Formatter.HTML do
...
+
+      true ->
+        raise """
+        ex_doc cannot output to #{config.output}:
+        Directory already exists and is not managed by ex_doc.
+
+        Try giving an unexisting directory as `--output`.
+        """
     end
   end
```
josevalim, line 177: Is there a way we could share this functionality? We can either:

1. Move it outside ExDoc.Formatter.HTML and call it earlier (it depends if we can also move the creation of .ex_doc itself higher in the call stack)

2. Make `setup_output` public and invoke it from EPUB (as it does with other HTML functions)

Thoughts?

viniciusmuller: That's a tricky question, I decided to leave them like this intially because I could find enough justifications to separate them, because there are some differences in their logic.

For example, in the EPUB example, we use the original `config.output` value before `normalize_config` appends the project name to it, and the mkdir calls are different in both the HTML and EPUB versions.

Although the error message  could certainly be shared between both, to avoid future inconsistencies.

One approach that might be possible would be making `setup_output` public and then passing the function that does directory and `.ex_doc` file creation as callback, and them on the HTML case, just passing both `path` and `parent` as the same value, what do you think?

josevalim: I think that's worth trying. Do it as additional comments so we revert if necessary?

viniciusmuller: Done, I think it's considerably better now @josevalim

lib/ex_doc/formatter/html.ex
```diff
@@ -146,18 +146,46 @@ defmodule ExDoc.Formatter.HTML do
...
+      not File.exists?(root) ->
+        create.(root, config)
+
+      true ->
+        raise """
+        ex_doc cannot output to #{root}:
+        Directory already exists and is not managed by ex_doc.
+
+        Try giving an unexisting directory as `--output`.
+        """
```
josevalim, line 166: What do you think about this one?

```suggestion
      File.ls(root) == {:ok, []} ->
        :ok

      true ->
        IO.warn(
          "ExDoc is outputting to an existing directory. " <>
            "Beware documentation output may be mixed with other entries"
        )
```

Instead of raising, we warn, but only if the root directory is empty?

lib/ex_doc/formatter/html.ex
```diff
@@ -146,18 +146,46 @@ defmodule ExDoc.Formatter.HTML do
     |> ExDoc.DocAST.highlight(language, opts)
   end
 
-  defp output_setup(build, config) do
+  def setup_output(config, cleanup, create, root \\ nil) do
```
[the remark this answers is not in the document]
viniciusmuller: It is currently that way because in the EPUB example, `normalize_config` appends something to the path, so that `config.output` is not the root directory.

https://github.com/elixir-lang/ex_doc/blob/b66861bb60173b056022e95662cf72275eba18a8/lib/ex_doc/formatter/epub.ex#L13

But thinking a bit better, I think it is possible to just call `setup_output` before `normalize_config`

viniciusmuller:
Is there anything that still needs to be done in this PR?

josevalim:

PR #1708, opened 2023-05-22
Allow multiple extra files with the same name

### LINKED ISSUE #1550: [Improvement|Bug?] Multiple README files from multiple directories
I have the following config

```elixir
extras: [
        "README.md",
        "CHANGELOG.md",
        "docs/references/README.md",
        "docs/how-to/README.md",
        "docs/explanations/README.md",
      ],
```

As you can see they are multiple `README.md` files because I have multiple directories.

When I generate the docs I got the following:

Screen Shot 2022-04-11 at 5 11 02 AM

Notice that multiple things have the `selected` state, and the URL is adding a hash at the end.

### Expected Behavior

I would expect such files to have different paths and the selection to work properly.

viniciusmuller:
This pull request adds a check to disambiguate the id generated ids if any of them are duplicated.

This already solves the issue, but I still have the following questions:
- Can we try to generate the ids considering the full filepaths (maybe parent directory), so that we don't need to disambiguate them later?
  - I think this is possible, but probably there will be edge cases
- I couldn't find any great way to test it, does someone have an idea?
   - I've tried adding multiple extra README.md files to the fixture `foo` app and then capturing IO/mix output, no success
- Are there any safe heuristics we could follow in order to avoid having `readme-1`, `readme-2`, etc in the URLs? Maybe using the groups or file paths?

Closes #1550

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -355,11 +355,7 @@ defmodule ExDoc.Formatter.HTML do
 
     extras
     |> Enum.map_reduce(1, fn extra, idx ->
-      if ids_count[extra.id] > 1 do 
-        {disambiguate_id(extra, idx), idx + 1}
-      else
-        {extra, idx}
-      end
+      if ids_count[extra.id] > 1, do: {disambiguate_id(extra, idx), idx + 1}, else: {extra, idx}
     end)
     |> elem(0)
     |> Enum.sort_by(fn extra -> GroupMatcher.group_index(groups, extra.group) end)
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -354,9 +354,14 @@ defmodule ExDoc.Formatter.HTML do
     ids_count = Enum.reduce(extras, %{}, &Map.update(&2, &1.id, 1, fn c -> c + 1 end))
 
     extras
-    |> Enum.with_index(fn extra, idx ->
-      if ids_count[extra.id] > 1, do: disambiguate_id(extra, idx + 1), else: extra
+    |> Enum.map_reduce(1, fn extra, idx ->
+      if ids_count[extra.id] > 1 do 
+        {disambiguate_id(extra, idx), idx + 1}
+      else
+        {extra, idx}
+      end
     end)
+    |> elem(0)
     |> Enum.sort_by(fn extra -> GroupMatcher.group_index(groups, extra.group) end)
   end
 
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -343,13 +343,25 @@ defmodule ExDoc.Formatter.HTML do
       skip_undefined_reference_warnings_on: config.skip_undefined_reference_warnings_on
     ]
 
-    config.extras
-    |> Task.async_stream(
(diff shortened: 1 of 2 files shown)


wojtekmach:
@viniciusmuller here's a test snippet you can use:

```elixir
    test "multiple extras with the same name", c do
      File.mkdir_p!("#{c.tmp_dir}/foo")

      File.write!("#{c.tmp_dir}/foo/README.md", """
      # README foo
      """)

      File.mkdir_p!("#{c.tmp_dir}/bar")

      File.write!("#{c.tmp_dir}/bar/README.md", """
      # README bar
      """)

      config =
        Keyword.replace!(doc_config(c), :extras, [
          "#{c.tmp_dir}/foo/README.md",
          "#{c.tmp_dir}/bar/README.md"
        ])

      generate_docs(config)

      doc = EasyHTML.parse!(File.read!("#{c.tmp_dir}/html/readme.html"))["#content"]
      assert to_string(doc["h1 > span"]) == "README foo"
    end
```

note the test currently warns:

> `warning: file tmp/ExDoc.Formatter.HTMLTest/test-generates-extras-multiple-extras-with-the-same-name-dc7fa2eb/html/readme.html already exists`

and intentionally fails. I'm checking the contents of readme.html and it happens to be "README bar" but we actually should have readme-1.html and readme-2.html.

viniciusmuller:
Thanks @wojtekmach! I've adapted the test to check to contents of both `foo` and `bar` READMEs and it already helped me to catch a bug (counter starting at zero instead of one)!

wojtekmach (approved):

lib/ex_doc/formatter/html.ex
```diff
@@ -343,15 +343,27 @@ defmodule ExDoc.Formatter.HTML do
...
+        &build_extra(&1, groups, language, autolink_opts, source_url_pattern),
+        timeout: :infinity
+      )
+      |> Enum.map(&elem(&1, 1))
+
+    ids_count = Enum.reduce(extras, %{}, &Map.update(&2, &1.id, 1, fn c -> c + 1 end))
+
+    extras
+    |> Enum.with_index(fn extra, idx ->
+      if ids_count[extra.id] > 1, do: disambiguate_id(extra, idx + 1), else: extra
```
josevalim, line 358: This version of `with_index` is not supported in earlier Elixir versions. Maybe implement this with a `Enum.map_reduce` instead?

wojtekmach: ExDoc requires Elixir 1.11, I think we're good. :)

viniciusmuller: not sure if I got it, shall I use other function than `with_index`, there are some other places in the codebase that seems to use it just fine, like

https://github.com/elixir-lang/ex_doc/blob/b66861bb60173b056022e95662cf72275eba18a8/lib/ex_doc/formatter/epub.ex#LL91C18-L91C18

But indeed the tests for 1.11 are failing

josevalim: This particular version of `with_index` does not work, so you need to use something else. I would do seomthing like this:
```
Enum.map_reduce(extras, 1, fn extra, idx ->
  if ids_count[extra.id] > 1, do: {disam(...), idx + 1}, else: {extra, idx}
end)
|> elem(0)
```

josevalim:
Awesome PRs!

PR #1713, opened 2023-05-31
Create .exdoc file after warning about using an existing directory 

viniciusmuller:
Currently, if an user tries to set `--output` to a directory that already contains files, ex_doc will emit a warning about that and create the files anyway, but it never creates the `.exdoc` file, meaning every time `mix docs` is run on that directory it will display the same warnings.

This pull requests calls the create callback after emitting the warning, so that in subsequent invocations no warnings will be emitted.


viniciusmuller:
I've added this check to the tests that I've added when implementing this feature, I think there's no reason to duplicate the entire test just to test this property.

viniciusmuller:
Not sure why this test failing, I've just ran the test suite locally 15 times in a row and it passed them all consistently :thinking:

test/ex_doc/formatter/epub_io_test.exs
```diff
@@ -0,0 +1,55 @@
...
+             generate_docs(new_config)
+           end) =~ "ExDoc is outputting to an existing directory"
+  end
+
+  defp generate_docs(config) do
+    config = Keyword.put_new(config, :skip_undefined_reference_warnings_on, ["Warnings"])
+    ExDoc.generate_docs(config[:project], config[:version], config)
+  end
+
+  defp doc_config(%{tmp_dir: tmp_dir} = _context) do
```
viniciusmuller, line 43: Should we create a helper module that provides the `generate_docs` and `doc_config` functions for both the non-io and io versions of the formatters test consume them from a centralized source?

test/ex_doc/formatter/epub_io_test.exs
```diff
@@ -0,0 +1,55 @@
+defmodule ExDoc.Formatter.EPUBIOTest do
```
wojtekmach, line 1: I think we can have a much smaller and simpler test.

If we avoid fixtures then we don't need to worry about warnings and any of that either. In fact, it's totally fine if we point ExDoc to a directory with no .beam files (the directory doesn't even have to exist.) 

Instead of refuting that a given warning was emitted, let's assert there were no warnings.

How about this?

```elixir
defmodule ExDoc.Formatter.EPUBIOTest do
  use ExUnit.Case, async: false

  @moduletag :tmp_dir

  test "does not warn if writing to an empty existing directory", %{tmp_dir: tmp_dir} do
    File.mkdir!("#{tmp_dir}/doc")

    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
             generate_docs(tmp_dir)
           end) == ""
  end

  test "warns if trying to write into existing directory with files", %{tmp_dir: tmp_dir} do
    File.mkdir!("#{tmp_dir}/doc")
    File.touch!("#{tmp_dir}/doc/foo.txt")

    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
             generate_docs(tmp_dir)
           end) =~ "ExDoc is outputting to an existing directory"

    # don't warn twice
    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
             generate_docs(tmp_dir)
           end) == ""
  end

  defp generate_docs(tmp_dir) do
    config = [
      app: :foo,
      formatter: "epub",
      output: "#{tmp_dir}/doc",
      source_beam: "#{tmp_dir}/ebin"
    ]

    ExDoc.generate_docs("Foo", "1.0.0", config)
  end
end
```

viniciusmuller: Thanks for the suggestions! I've applied them and those tests are indeed much cleaner now.

test/ex_doc/formatter/epub_io_test.exs
```diff
@@ -0,0 +1,38 @@
...
+  use ExUnit.Case, async: false
+
+  @moduletag :tmp_dir
+
+  test "succeeds if trying to write into an empty existing directory", %{tmp_dir: tmp_dir} do
+    File.mkdir!("#{tmp_dir}/doc")
+
+    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
+             generate_docs(tmp_dir)
+           end) =~ ""
```
wojtekmach, line 11: _any_ string would pass this assertion be we want to make sure we didn't print any warnings!

```suggestion
           end) == ""
```

test/ex_doc/formatter/epub_io_test.exs
```diff
@@ -0,0 +1,38 @@
...
+    File.touch!("#{tmp_dir}/doc/foo.txt")
+
+    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
+             generate_docs(tmp_dir)
+           end) =~ "ExDoc is outputting to an existing directory"
+
+    # Warn only once
+    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
+             generate_docs(tmp_dir)
+           end) =~ ""
```
wojtekmach, line 25:
```suggestion
           end) == ""
```

test/ex_doc/formatter/html_io_test.exs
```diff
@@ -0,0 +1,38 @@
...
+  use ExUnit.Case, async: false
+
+  @moduletag :tmp_dir
+
+  test "succeeds if trying to write into an empty existing directory", %{tmp_dir: tmp_dir} do
+    File.mkdir!("#{tmp_dir}/doc")
+
+    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
+             generate_docs(tmp_dir)
+           end) =~ ""
```
wojtekmach, line 11:
```suggestion
           end) == ""
```

test/ex_doc/formatter/html_io_test.exs
```diff
@@ -0,0 +1,38 @@
...
+    File.touch!("#{tmp_dir}/doc/foo.txt")
+
+    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
+             generate_docs(tmp_dir)
+           end) =~ "ExDoc is outputting to an existing directory"
+
+    # Warn only once
+    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
+             generate_docs(tmp_dir)
+           end) =~ ""
```
wojtekmach, line 25:
```suggestion
           end) == ""
```

wojtekmach (approved):

test/ex_doc/formatter/epub_io_test.exs
viniciusmuller: My bad here, forgot to change the operator

PR #1716, opened 2023-05-31
Distinct deprecated autocomplete suggestions

### LINKED ISSUE #1268: Search suggestions in sidebar should mention if they are deprecated
I use the sidebar for a quick search to remind me what is the module a functions belongs to. For instance if I type "cwd", I got System.cwd as a suggestion, but I doesn't mention that is is deprecated. So I guess it should be added.

viniciusmuller:
This patch shows a `(deprecated)` tag when autocompleting deprecated nodes.

Preview:
deprecated_on_search_suggestions

Closes #1268

### DIFF
--- a/assets/css/autocomplete.css
+++ b/assets/css/autocomplete.css
@@ -29,6 +29,11 @@
   text-decoration: none;
 }
 
+.autocomplete-suggestion-title-wrapper {
+  display: flex;
+  flex-wrap: wrap;
+}
+
 .autocomplete-suggestion:hover,
 .autocomplete-suggestion.selected {
   background-color: var(--gray600);
@@ -55,5 +60,5 @@
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
-  width: 100%;
+  max-width: 100%;
 }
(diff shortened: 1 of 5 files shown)


josevalim:
Looks good to me. Does anyone have any suggestions styling wise? Should we try to call more attention to it (I thought so originally but I think calling more attention to the deprecated ones is actually the wrong thing to do?)?

sabiwara:
>  Does anyone have any suggestions styling wise? 

I think it is quite common to see deprecated functions displayed as strikethrough in IDEs for instance (vscode screenshot below), this makes it clear while avoiding making it standout more. Could be an idea?

Screenshot 2023-06-01 at 9 09 27

josevalim:
I like this! @viniciusmuller, can you please try this version out?

viniciusmuller:
Here's how it looks like with the latest commit, I like it
image

josevalim:
Show we do both the styling and the (deprecated) note? Or just the styling is enough?

wojtekmach:
I'd go with just strikethrough, otherwise we call too much attention to it. And if someone is curious they can follow the autocomplete suggestion. Something else we could do is on autocomplete suggestion to the popover thing and show it's deprecated _there_ but it's a separate discussion and it might be a bit too much.

josevalim:

PR #1717, opened 2023-06-05
Add source URL to nav link list

Nezteb:
This started as a question in the Elixir Slack: https://elixir-lang.slack.com/archives/C0U07G7FA/p1685743433113699

Essentially, myself and a few others would like an easier way to go from a Hexdocs project page to its source/git repo.

So, I've added this as part of the sidebar list nav, which is now scrollable in case there are too many items to fit.
Screenshot 2023-06-05 at 10 45 41 AM

I'm open to other location suggestions. 😄 

cc @lawik @axelson


josevalim:
We have a bunch of links at the footer at the page, so it could be added there. You can also access the source in the top right of each page, so that also works as a reference. :)

Nezteb:
I actually was going to use the footer at first, but someone on Slack suggested somewhere closer to the top.

> You can also access the source in the top right of each page

This isn't always the case, I've noticed. Sometimes specific pages don't have the code links, like the API reference page: https://hexdocs.pm/phoenix/api-reference.html

Some projects don't have the source icon in the top right: 
- https://hexdocs.pm/pigeon/getting-started.html
- https://hexdocs.pm/bamboo/readme.html
- https://hexdocs.pm/pushex/readme.html

The above do have source links in their modules pages, but that's usually it. That being said, I'm guessing that's because they were generated with older versions of `ex_doc`? In that case, even my MR won't fix that. 😅 

I guess my main issue could be resolved by adding the same icon link to the "API reference" page, which I can add!

josevalim:
>  my main issue could be resolved by adding the same icon link to the "API reference" page, which I can add!

exactly! Let’s do this then!

josevalim:

PR #1722, opened 2023-06-17
Warn on invalid references in links

### LINKED ISSUE #1572: Warn on invalid format inside backticks in links
[old Erlang child specification](`t:supervisor.child_spec()`)

will generate 

    <a href="`t:supervisor.child_spec()`">old Erlang child specification</a>

I find myself making this kind of mistakes, and ExDoc not warnings about it.
I think if Exdoc ends up generating a link that starts with a backtick, it should emit a warning.

~Well. Actually i think the format should be supported to avoid this kind of mistake, and render it without the backticks.~

viniciusmuller:
This pull request adds a check to warn and remove a link if it is a reference (has backticks) and could not be parsed as a reference.

That means links such as  ``[supervisor docs](`t:supervisor.child_spec/0`)`` will now warn on build and not emit a broken link rather than failing silently.j

Closes #1572

### DIFF
--- a/lib/ex_doc/language/elixir.ex
+++ b/lib/ex_doc/language/elixir.ex
@@ -550,8 +550,13 @@ defmodule ExDoc.Language.Elixir do
     timeout: 0
   ]
 
-  defp url(string = "mix help " <> name, mode, config), do: mix_task(name, string, mode, config)
-  defp url(string = "mix " <> name, mode, config), do: mix_task(name, string, mode, config)
+  defp url(string = "mix help " <> name, mode, config) do
+    name |> mix_task(string, mode, config) |> maybe_remove_link(mode)
+  end
+
+  defp url(string = "mix " <> name, mode, config) do
+    name |> mix_task(string, mode, config) |> maybe_remove_link(mode)
+  end
 
   defp url(string, mode, config) do
     case Regex.run(~r{^(.+)/(\d+)$}, string) do
@@ -721,11 +726,9 @@ defmodule ExDoc.Language.Elixir do
         mix_task: true,
         original_text: string
       })
-
-      :remove_link
-    else
-      url
     end
+
+    url
   end
 
   defp safe_format_string!(string) do
--- a/lib/ex_doc/language/elixir.ex
+++ b/lib/ex_doc/language/elixir.ex
@@ -447,7 +447,7 @@ defmodule ExDoc.Language.Elixir do
           [[_, custom_link]] ->
             custom_link
             |> url(:custom_link, config)
-            |> remove_and_warn_if_invalid(href, config)
+            |> remove_and_warn_if_invalid(custom_link, config)
 
           [] ->
             build_extra_link(href, config)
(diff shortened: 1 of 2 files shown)


test/ex_doc/language/elixir_test.exs
```diff
@@ -32,6 +32,20 @@ defmodule ExDoc.Language.ElixirTest do
...
+        assert autolink_doc(~m"[text](`fakefunction`)") == ~m"text"
+      end)
+
+      assert_invalid_ref_warning("`t:supervisor.child_spec/0`", fn ->
+        assert autolink_doc(~m"[text](`t:supervisor.child_spec/0`)") == ~m"text"
+      end)
+
+      assert_invalid_ref_warning("`mix foobar`", fn ->
+        assert autolink_doc(~m"[text](`mix foobar`)") == ~m"text"
+      end)
```
wojtekmach, line 46: we're testing this scenario below on line 516-518. How are these different? warn/1 ensures warning in the form of `documentation references (...)`, assert_invalid_ref_warning/1 in the form of `found invalid reference ...`.  How come both tests are correct? Do we emit two warnings? Also note we tent to use `"` instead of `` ` `` in warning messages.

For the new warning my suggestion would be to have `documentation referenced "ref" but it is invalid` for consistency even though maybe it doesn't read perfectly for these types of warnings. This way we can use warn/1 test helper. Also nitpick

Could you also move these assertions to the one big warning test? I think it's helpful to see all possible warnings in one place.

lib/ex_doc/language/elixir.ex
```diff
@@ -444,15 +444,29 @@ defmodule ExDoc.Language.Elixir do
...
         end
 
       _ ->
         nil
     end
   end
 
+  defp maybe_remove_invalid_ref(result, href) do
+    if is_nil(result) and String.match?(href, ~r/`.*`/) do
+      IO.warn("found invalid reference: #{href}")
```
wojtekmach, line 463: Could you look at the other reference warnings code and include file:line in this one too? The line isn’t perfect for these because it corresponds to where doc block starts, not where the warning happens inside the block, but that’s ok for now.

test/ex_doc/language/elixir_test.exs
viniciusmuller: Hi @wojtekmach! Currently there were two warnings being emitted for those cases.

I agree we can update the warning message to be more consistent and emita single warning when we have invalid references.

josevalim (approved):

test/ex_doc/language/elixir_test.exs
```diff
@@ -460,6 +460,22 @@ defmodule ExDoc.Language.ElixirTest do
 
     warn(~m"`t:Calendar.date/9`")
 
+    assert warn(fn ->
+             assert autolink_doc(~m"[text](`fakefunction`)") == ~m"text"
+           end) =~ ~s[but it is invalid]
```
wojtekmach, line 465: could you include the reported reference in the warning string? I.e. this assertion should be:

```elixir
(...) =~ ~s["`fakefunction`" but it is invalid]
```

but I think we actually want:

```elixir
(...) =~ ~s["fakefunction" but it is invalid]
```

btw it's fine if we include more warning in the assertion, i.e.:

```elixir
(...) =~ ~s[documentation references "fakefunction" but it is invalid]
```

because we have some logic around including additional words like module, function, etc so stricter assertion can be helpful here.

wojtekmach (approved):
Just one last comment.

test/ex_doc/language/elixir_test.exs
viniciusmuller: No problems! I've updated the code to not output the backticks on warning and improved the assertions :)

wojtekmach:
Thank you!

PR #1723, opened 2023-06-27
mermaid.js to 10.2.3 (async invocation) and make it respect color theme

serpent213:
Color theme is only evaluated on (re)load. I don't see an easy way to hook into the (minified) SettingsStore.


josevalim:

PR #1729, opened 2023-07-04
Rename `searchNodes` to `searchData`

wojtekmach:
- Rename `searchNodes` to `searchData`
- Add content-type based on proglang

### DIFF
--- a/assets/js/search-page.js
+++ b/assets/js/search-page.js
@@ -1,4 +1,4 @@
-/* globals searchNodes */
+/* globals searchData */
 
 import lunr from 'lunr'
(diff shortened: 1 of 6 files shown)


wojtekmach:
I briefly considered this structure:

```js
{
  "items": [
    {"content-type": "text/markdown", ...}.
    {"content-type": "text/plain", ...}.
  ]
}
```

that is, we have a content-type on a per-node (module, function, extra, etc) basis, given it more explicitly supports projects with both "Elixir" and "Erlang" parts (the latter currently not using markdown) but given the community seems going towards markdown for everything, such approach might be moot, i.e. top-level content-type is good enough.

Issue #1732, opened 2023-07-06
Elixir ExDocs Search index too big for SessionStorage


jeregrine:
When you visit the HexDocs page for an Elixir (or any large hex docs) such as https://hexdocs.pm/elixir/1.15.2/search.html?q=hi the console error log shows the following error:
```
html-JDI3AVDD.js:2 Failed to save index:  DOMException: Failed to execute 'setItem' on 'Storage': Setting the value of 'index:Elixir v1.15.2' exceeded the quota.
    at Ui (https://hexdocs.pm/elixir/1.15.2/dist/html-JDI3AVDD.js:2:39805)
    at $i (https://hexdocs.pm/elixir/1.15.2/dist/html-JDI3AVDD.js:2:39572)
    at zi (https://hexdocs.pm/elixir/1.15.2/dist/html-JDI3AVDD.js:2:39047)
    at Ut (https://hexdocs.pm/elixir/1.15.2/dist/html-JDI3AVDD.js:2:38986)
    at https://hexdocs.pm/elixir/1.15.2/dist/html-JDI3AVDD.js:2:53681
    at at (https://hexdocs.pm/elixir/1.15.2/dist/html-JDI3AVDD.js:2:27572)
    at https://hexdocs.pm/elixir/1.15.2/dist/html-JDI3AVDD.js:2:53613
    at https://hexdocs.pm/elixir/1.15.2/dist/html-JDI3AVDD.js:2:53705
```
This is because the lunr index exceeds 5mb when we try and store it in the session here https://github.com/elixir-lang/ex_doc/blob/main/assets/js/search-page.js#L78

the page still works cause you can search the lunr index in memory, and every subsequent search forces a reload (and index rebuild) anyways. 

Not sure if this is a fix or ignore kinda bug but I figured I'd report it. :)

josevalim:
I believe we started handling this case at some point, as otherwise it would crash. Perhaps we can partition or compress the lunr index somehow: https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API

jeregrine:
Another one that is extremely broken is Scholar https://hexdocs.pm/scholar/search.html?q=hiu

I'm not 100% usre whats going on but the json it produces is 7mb gzippped and  a *mess*
![Screenshot 2023-07-07 at 11 41 36 AM](https://github.com/elixir-lang/ex_doc/assets/100886/a3317ce9-1eb3-42c8-a8ce-3663c21e0982)

[...]

PR #1735, opened 2023-07-11
Compresses and Decompresses the index for ExDoc

jeregrine:
Successfully stores Elixir Hex Docs in Session storage

Screenshot 2023-07-11 at 10 49 44 AM

This looks more simple then it was since DecompressionStream/CompressionStream's are fairly new, but well supported outside of IE. https://caniuse.com/mdn-api_compressionstream


jeregrine:
Partially fixes https://github.com/elixir-lang/ex_doc/issues/1732

assets/js/search-page.js
```diff
@@ -74,13 +76,56 @@ function loadIndex () {
 
 function saveIndex (index) {
   try {
-    const serializedIndex = JSON.stringify(index)
-    sessionStorage.setItem(indexStorageKey(), serializedIndex)
+    const serializedIndex = compress(index).then((compressedIndex) => {
+      sessionStorage.setItem(indexStorageKey(), compressedIndex)
```
enewbury, line 79: Do we want the try catch in the callback rather than the top level flow?

assets/js/search-page.js
```diff
@@ -72,15 +73,56 @@ function loadIndex () {
...
+  const len = binary_string.length;
+  const bytes = new Uint8Array(new ArrayBuffer(len));
+  for (let i = 0; i < len; i++) {
+    bytes[i] = binary_string.charCodeAt(i);
+  }
+  return bytes;
+}
+
 function indexStorageKey () {
   return `index:${getProjectNameAndVersion()}`
```
josevalim, line 127:
```suggestion
  return `idv2:${getProjectNameAndVersion()}`
```

Let's change the key just to ensure we don't accidentally load a past index through compression.

josevalim:

PR #1741, opened 2023-07-18
Fix generation of `.livemd` paths to handle omitted '.html'

josecfreittas:
Some HTTP servers redirect and omit the `.html` extension when you go to an html page, hosting a generated doc on one of these makes the “Run in Livebook” button break because it doesn’t add the `.livemd` extension to the link. This PR fixes it making the `.html` optional in the replacement.


josevalim:

PR #1748, opened 2023-08-20
Remove sidebar gradient

angelikatyborska:
When I checked out the project to work on something else, I noticed an invalid CSS variable, `var(gray800-opacity-0)` (missing leading `--` in front of the name).

After digging around, I realized that, if fixed, the sidebar gradient would be a gradient from gray800 to gray800-opacity-0, and it would be rendered inside of the sidebar that already has a solid background color of gray800. That means the gradient would not be visible at all. Its positioning styles also mean that the gradient element never took up any space.

I decided it makes no sense to fix this gradient element that would be invisible anyway and doesn't affect the flow of other elements, and instead I'm deleting it.


josevalim:

PR #1754, opened 2023-08-22
Add Devbox environment config

DavidOliver:
[Devbox](https://www.jetpack.io/devbox/) uses [Nix](https://nixos.org/guides/how-nix-works), a purely functional package manager, and provides a simple way to take advantage of that without requiring learning of the Nix language.

I'm suggesting adding its two `devbox.(json|lock)` files for optional usage because:

- I use it and the config being present in the main branch would mean I don't have to switch branches and subsequently cherry-pick to contribute :smiley: 
- it provides a simple, declarative way of specifying and using versions (Elixir, Node, and Chrome for the JS tests) across MacOS, Linux and Windows without having to otherwise install them or deal with images.

[Installation](https://www.jetpack.io/devbox/docs/installing_devbox/) is a curl command. Then, to use in a project with the devbox files:

```plain
$ devbox shell
```

The specified versions of Elixir, Node and Chrome are installed/activated, and the mix setup command is called so that Elixir and Node dependencies are also installed if they're not already present.

Versions currently specified:

- Elixir 1.15.4
- Node 18.16.1
- Chrome 115.0.5790.170 (latest available for Nix)

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -1,0 +1,16 @@
+{
+  "packages": [
+    "nodejs@18.16.1",
+    "elixir_1_15@latest",
+    "google-chrome@latest"
+  ],
+  "shell": {
+    "init_hook": [
+      "export ERL_AFLAGS='-kernel shell_history enabled'",
+      "mix setup"
+    ]
+  },
+  "nixpkgs": {
+    "commit": "f80ac848e3d6f0c12c52758c0f25c10c97ca3b62"
+  }
+}
--- a/devbox.lock
+++ b/devbox.lock
@@ -1,0 +1,20 @@
+{
+  "lockfile_version": "1",
+  "packages": {
+    "elixir_1_15@latest": {
+      "last_modified": "2023-07-23T03:35:12Z",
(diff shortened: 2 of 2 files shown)


wojtekmach:
Devbox looks like an interesting and useful tool. That being said I’m probably not gonna use it on this particular project because I tend to use Elixir main while developing ExDoc, making sure Elixir docs are not broken, testing new features, etc. So I’m personally -1 on including it. But I can totally see how it could remove some barrier to contributing.

whatyouhide:
I’m not sure this is an issue that is shared by many devs working on ex_doc. 

> I use it and the config being present in the main branch would mean I don't have to switch branches and subsequently cherry-pick to contribute

Can't you put this file in `.git/info/exclude` in your repo, so that you don't have to check it out? 😬

josevalim:
Closing this for now. I agree with @wojtekmach that we may not use it, we often test against Elixir main, but please try @whatyouhide's suggestion and, if that doesn't work, I am fine with updating our `.gitignore` to include it. Thank you!

DavidOliver:
Thanks for the consideration and replies. :)

I wanted the Devbox config to be version-controlled, so I've gone with [a parent git repo which has ExDoc as a submodule](https://github.com/DavidOliver/ex_doc-env).

Issue #1663, opened 2023-02-16
Proposal: Autolink with `[code]`


wojtekmach:
Currently we autolink expressions like `` `String` ``, `` `String.upcase/1` ``, etc. The obvious problem is auto linking "Erlang" modules, e.g.: `` `:file` ``, which we don't do, because we don't know if the user intended to link to a module or just wrote the atom `:file`. We can today solve this by writing `` [`:file`](`:file`) `` but it's verbose as well as when any other Markdown parser sees it, it will output garbage.

Instead of implicitly treat `` `code` `` as potential links, I'd like to propose supporting an explicit syntax: `` [`code`] ``. This is what [Rust does](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html). See https://github.com/erlef/build-and-packaging-wg/issues/60#issuecomment-1347292105. By having explicit syntax, the tooling will know the user intended to write a code reference and thus can output warnings, etc. This will have much more predictable result on non-ExDoc: GitHub, etc would render `` [`:file`] `` as `[<code>:file</code>]` which I think is very reasonable.

Here's a proof of concept: https://gist.github.com/ef18199a4283d9a59b9fdfb571a4e251.

Per https://github.com/erlef/build-and-packaging-wg/issues/60, there are conversations about introducing another EEP that would make Markdown officially supported document format for Erlang. I think there's appetite for supporting the syntax I'm proposing there. If it would be accepted there then I think there'd be a strong reason to support it in Elixir ecosystem too.

josevalim:
I am fine with supporting this but only after we agree on the format for Erlang so we don't end up with three distinct ways in the future. :)

Eiji7:
Interesting! How about change it by using `sigil`-like syntax and modifiers?

For example:
1. default `module` or `mfa` (without optional modifier): <code>\~\`String\`</code> or <code>\~\`String.upcase/1\`</code> or <code>\~\`:file\`</code>
2. same `module` or `mfa` (with `m` modifier): <code>\~\`String\`m</code> or <code>\~\`String.upcase/1\`m</code> or <code>\~\`:file\`m</code>
3. Elixir `forum` topic: <code>\~\`53184\`f</code> or <code>\~\`elixir-v1-14-3-released\`f</code><br/>**Note**: both `elixirforum.com/t/:id` and `elixirforum.com/t/:slug` redirects to `elixirforum.com/t/:slug/:id`
[...]

PR #1759, opened 2023-09-02
Add skip_code_autolink_to option

### LINKED ISSUE #1750: `:skip_undefined_reference_warnings_on` with modules in pages
I have an extra page, say `pages/my-page.md`. In it, I mention `MyModule`, which is now a private module.

I don't want to warn on it. If I pass

```elixir
skip_undefined_reference_warnings_on: ["MyModule"]
```

I still get the warning. The only way I found to silence it is to pass `"pages/my-page.md"` altogether, which is not what I want because it silences all warnings. When I inspected ex_doc, the `config.id` and `config.module_id` were both `nil`.

This might be a documentation problem, or a missing feature, or a bug. Not really sure. I'd love to help out more but I’m a lacking time at the moment 😉

viniciusmuller:
Adds a new option `:skip_code_autolink_to` that given a list of strings, skips term autolinking if it matches one of the strings.

I think `Enum.any?/2` will do fine, but if needed we can convert the list of strings given by the user to a MapSet and use it internally as well.

Closes #1750


lib/ex_doc/language/elixir.ex
```diff
@@ -559,6 +559,14 @@ defmodule ExDoc.Language.Elixir do
   end
 
   defp url(string, mode, config) do
+    if Enum.any?(config.skip_code_autolink_to, &(&1 == string)) do
```
whatyouhide, line 562: Should we allow regexes here as well?

josevalim: We can add if necessary later on? Unless you have an immediate need for it?

whatyouhide: I deprecated a module, so I'd like to do `~r/MyMod(\..*)?/` basically instead of having to list the references to the module alone and then to all the functions I mention. Unless I'm reading the feature wrong 😉

wojtekmach (approved):
The PR looks good to me.

I'm not even sorry for sounding like a broken radio, we really need to opt-in to auto linking by writing `` [`Foo.bar`] `` (#1663), not opt-out.

@whatyouhide fyi: I do the above explicitly in my extras. It is annoying to type because I have to maintain links manually, e.g. in my README.md I have:


```markdown
* Follows redirects (via [`redirect`] step.)

(...)

[`retry`]: https://hexdocs.pm/req/Req.Steps.html#retry/1
```

on the flip side, my not-so-auto links work on _both_ https://hexdocs.pm/req/readme.html AND https://github.com/wojtekmach/req/blob/main/README.md.

so if you'd adopt it, the need for this PR disappears. But ngl, it's annoying to manually keep them, so +1 for the PR.

lib/ex_doc/language/elixir.ex
wojtekmach: not sure if this would ever come up but the above regex will also catch `MyNamespace.MyMod`. So maybe it should be: `~r/^MyMod(\..*)?/`. but then we'd still warn on callbacks and types so I suppose `~r/^([bt]:)?MyMod(\..*)?/`.

I think we can add the regex support but it needs to be used carefully. but yeah I'd rather have users spell out things explicitly to keep the code explicit too. how annoying is this gonna be for you? is this for a changelog? If so consider just cutting off old changelog entries, e.g. write:

The CHANGELOG for v1.15 and previous releases can be found [in the v1.15] branch

or something like that.

josevalim: Maybe we just make it a function and allow you to pass anything?

wojtekmach: Nice, +1 from me.

josevalim: I will merge this and we can make both `skip_` options accept a function too.

josevalim:

PR #1761, opened 2023-09-07
Warn if typespec references filtered module

### LINKED ISSUE #1744: Using `filter_modules` doesn't produce warning referencing `hidden or private` docs
Using `filter_modules` in the `mix.exs` doesn't produce warnings indicating undocumented code reference.

Using `@moduledoc false`:
1. Code:
```elixir
defmodule Foo do
  @moduledoc false
  # "Private" module

  @type foo :: integer()
end


defmodule Bar do
  @moduledoc """
  "Public" module
  """
  alias Foo

  @spec bar(Foo.foo()) :: nil
  def bar(num) do
    IO.inspect(num)
    nil
  end
end
```

2. Generates warring while trying to generate docs to notify users, that they are using type reference unavailable in the documentation:
```console
$ mix docs
Generating docs...
warning: documentation references type "Foo.foo()" but it is hidden or private
  lib/membrane_template.ex:16: Bar.bar/1

View "html" docs at "doc/index.html"
```

Using `filter_modules`:
1. `mix.exs`
```elixir
defmodule Mixfile do
  use Mix.Project

  def project do
    [
...
      # docs
      name: "Example",
      source_url: @github_url,
      docs: docs()
    ]
  end

...

  defp docs do
    [
      main: "readme",
      extras: ["README.md", "LICENSE"],
      formatters: ["html"],
      source_ref: "v#{@version}",
      filter_modules: ~r/Bar/,
    ]
  end
end
```

2. Code
```elixir
defmodule Foo do
  @moduledoc """
  "Private" module excluded with `filter_modules`
  """

  @type foo :: integer()
end


defmodule Bar do
  @moduledoc """
  "Public" module
  """
  alias Foo

  @spec bar(Foo.foo()) :: nil
  def bar(num) do
    IO.inspect(num)
    nil
  end
end
```

3. Doesn't generate any warring, despite `Foo.foo()` being unavailable in `docs`
```console
$ mix docs       
Generating docs...
View "html" docs at "doc/index.html"
```

4. Generated docs with `Foo.foo()` reference:
Screenshot 2023-07-27 at 11 49 43

viniciusmuller:
This PR updates the retriever to also return a list of filtered modules, and updates the code to warn if we find a typespec that references a type from a filtered module.

Closes #1744

### DIFF
--- a/test/ex_doc/language/elixir_test.exs
+++ b/test/ex_doc/language/elixir_test.exs
@@ -439,6 +439,16 @@ defmodule ExDoc.Language.ElixirTest do
     warn(~m"`c:InMemory.unknown/0`")
   end
 
+  test "warning if typespec references filtered module" do
+    ExDoc.Refs.insert([
+      {{:module, AutolinkTest.Keep}, :public},
+      {{:function, AutolinkTest.Filtered}, :public},
+      {{:type, AutolinkTest.Filtered, :type, 0}, :public}
+    ])
+
+    # TODO: testing
+  end
+
   test "warnings" do
     ExDoc.Refs.insert([
       {{:module, AutolinkTest.Foo}, :public},
--- a/lib/ex_doc.ex
+++ b/lib/ex_doc.ex
@@ -21,8 +21,8 @@ defmodule ExDoc do
       ExDoc.Markdown.put_markdown_processor(processor)
     end
 
-    {module_nodes, _filtered_nodes} = config.retriever.docs_from_dir(config.source_beam, config)
-    find_formatter(config.formatter).run(module_nodes, config)
+    {module_nodes, filtered_nodes} = config.retriever.docs_from_dir(config.source_beam, config)
+    find_formatter(config.formatter).run({module_nodes, filtered_nodes}, config)
   end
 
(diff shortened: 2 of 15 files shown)


test/ex_doc/language/elixir_test.exs
```diff
@@ -439,6 +439,16 @@ defmodule ExDoc.Language.ElixirTest do
...
   end
 
+  test "warning if typespec references filtered module" do
+    ExDoc.Refs.insert([
+      {{:module, AutolinkTest.Keep}, :public},
+      {{:function, AutolinkTest.Filtered}, :public},
+      {{:type, AutolinkTest.Filtered, :type, 0}, :public}
+    ])
+
+    # TODO: testing
```
viniciusmuller, line 449: I would like some guidance here, I was able to check the fix with a local project but I'm not sure what would be the best way to test this.

Also, maybe we should create a new test case for this, as it's a bit more complex than the other cases that are based only on autolink

josevalim:

viniciusmuller:
Are we going to add the test or it's not necessary for this? :eyes:

PR #1763, opened 2023-09-09
Autolink support for handling remote types in records

starbelly:
This PR is in reference to and resolves https://github.com/starbelly/rebar3_ex_doc/issues/66

### DIFF
--- a/lib/ex_doc/language/erlang.ex
+++ b/lib/ex_doc/language/erlang.ex
@@ -572,14 +572,6 @@ defmodule ExDoc.Language.Erlang do
             {{:., _, [module, name]}, _, args}, acc ->
               {{:t, [], args}, [{pp({module, name}), {module, name, length(args)}} | acc]}
 
-            ## remote type module.type/0
-            {:., _, [module, name]} = ast, acc ->
-              {ast, [{pp({module, name}), {module, name, 0}} | acc]}
-
-            ## remote type module.type/0
-            {_local_name, {{:., _, [module, name]}, _, args}} = ast, acc ->
-              {ast, [{pp({module, name}), {module, name, length(args)}} | acc]}
-
             {name, _, _}, acc when name in [:<<>>, :..] ->
               {nil, acc}
 
@@ -590,6 +582,11 @@ defmodule ExDoc.Language.Erlang do
             # fun() (spec_to_quoted expands it to (... -> any())
             {:->, _, [[{name, _, _}], {:any, _, _}]}, acc when name == :... ->
               {nil, acc}
+
+            # record{type :: remote:type/arity}
+            {:field_type, _, [type, {{:., _, [remote_mod, remote_type]}, _, args}]}, acc ->
+              {type,
+               [{pp({remote_mod, remote_type}), {remote_mod, remote_type, length(args)}} | acc]}
 
             # #{x :: t()}
             {:field_type, _, [name, type]}, acc when is_atom(name) ->
--- a/lib/ex_doc/language/erlang.ex
+++ b/lib/ex_doc/language/erlang.ex
@@ -572,6 +572,10 @@ defmodule ExDoc.Language.Erlang do
             {{:., _, [module, name]}, _, args}, acc ->
               {{:t, [], args}, [{pp({module, name}), {module, name, length(args)}} | acc]}
 
+            ## type module.type/0
+            {:., _, [module, name]} = ast, acc ->
+              {ast, [{pp({module, name}), {module, name, 0}} | acc]}
+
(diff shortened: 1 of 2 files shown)


starbelly:
Opened as draft, I need to flesh out the commit message.

paulo-ferraz-oliveira:
I've tried this solution out and it fixes the linked issue 👍

lib/ex_doc/language/erlang.ex
```diff
@@ -572,6 +572,10 @@ defmodule ExDoc.Language.Erlang do
             {{:., _, [module, name]}, _, args}, acc ->
               {{:t, [], args}, [{pp({module, name}), {module, name, length(args)}} | acc]}
 
+            ## type module.type/0
+            {:., _, [module, name]} = ast, acc ->
+              {ast, [{pp({module, name}), {module, name, 0}} | acc]}
```
starbelly, line 577: I'm going to look at this again because we make an assumption here, which may be fine, but we're assuming that the next element in the tuple is going to be an empty list, which isn't great. We should perhaps catch all this at field type or possibly do a look behind and look ahead.

lib/ex_doc/language/erlang.ex
```diff
@@ -572,6 +572,14 @@ defmodule ExDoc.Language.Erlang do
             {{:., _, [module, name]}, _, args}, acc ->
               {{:t, [], args}, [{pp({module, name}), {module, name, length(args)}} | acc]}
 
+            ## remote type module.type/0
+            {:., _, [module, name]} = ast, acc ->
+              {ast, [{pp({module, name}), {module, name, 0}} | acc]}
+
+            ## remote type module.type/0
```
starbelly, line 579: This also fixes the case where the following would not be linked : 

```
          {:type, 41, :field_type,
            [
              {:atom, 41, :authorization_endpoint},
              {:remote_type, 41,
               [{:atom, 41, :uri_string}, {:atom, 41, :uri_string}, []]}
            ]}
```

The other clause handles the case where there's a union (i.e., `uri_string:string() | undefined`)

test/ex_doc/formatter/html/erlang_test.exs
```diff
@@ -29,15 +29,33 @@ defmodule ExDoc.Formatter.HTML.ErlangTest do
...
+        Eh = <<"eh?">>,
+        #rec{k=Eh}.
     """)
 
-    doc = generate_docs(c)
+    refute ExUnit.CaptureIO.capture_io(:stderr, fn ->
+             send(self(), {:doc_path, generate_docs(c)})
+           end) =~ "inconsistency, please submit bug"
+
+    assert_received {:doc_path, doc}
```
wojtekmach, line 47: there's no need for this, if the warning is printed we'll (eventually) see it when running tests locally

```suggestion
    doc = generate_docs(c)
```

wojtekmach: I tried cherry-picking just this test on main and it works. I'd like to see it fail before committing the patch. Could you double-check? Perhaps Elixir or OTP version matters? What's yours?

test/ex_doc/formatter/html/erlang_test.exs
```diff
@@ -29,15 +29,33 @@ defmodule ExDoc.Formatter.HTML.ErlangTest do
...
     %% t/0 type.
+
+    -record(rec, {k :: undefined}).
+
+    -type my_tea() :: #rec{k :: uri_string:uri_string() | undefined}.
+
+    -spec baz() -> my_tea().
+      baz() ->
+        Eh = <<"eh?">>,
+        #rec{k=Eh}.
```
wojtekmach, line 40: let's keep it shorter

```suggestion
    -record(rec, {k :: term}).

    -type t2() :: #rec{k :: uri_string:uri_string() | undefined}.
```

lib/ex_doc/language/erlang.ex
wojtekmach, line 579: > `remote type module.type/0`

if I'm reading this correctly this is not rarity 0, please update the comment.

starbelly:
@wojtekmach  You wanted to test manually, something went wrong which I tried to make the tests prettier :)  If you cherry pick 98d9b69e5f6da7e7a8e0ab830f002866bb11a456 that should do the trick. I'll look at revisiting this again today, as the solution still seems funky.

starbelly:
@wojtekmach This is ready for review with a more "proper" clause in the erlang language module to cover this. As far as making source code for test shorter, we need at least two keys in the record to trigger to trigger this FWIW.

wojtekmach:
Thank you!

Issue #1765, opened 2023-09-10
Proposal: Narrow search results by type


zachdaniel:
In this very poor mockup, I show an example UI that would allow for selecting the types of things you would like to search. I would imagine that by default they would all be checked, so we could also add a little "only" button next to each one to instantly select only that one. 

![CleanShot 2023-09-10 at 14 55 36@2x](https://github.com/elixir-lang/ex_doc/assets/5722339/1fab3741-938c-4961-afbb-6b68ead86e8f)

This could be put behind a configuration of some kind, but to me it sounds generally useful enough to be done automatically. If this is something we would be interested in I'm happy to take a stab at it.

zachdaniel:
For some context, I'm trying to find ways to get hex docs to cover the cases that we need to (ideally) get rid of ash-hq.org. Multi-package search is one major issue, but we have some strategies to help with that (leave our ash-hq multi-package search up, but have it link to hex. The other major issue is search ability of DSLs within hex. Worst case scenario we can't get rid of ash-hq.org, but will still have improved experience for those users who prefer to use hex :)

josevalim:
Hi Zach! We are exploring multi package search, which will require a UI redesign, then I think it is a good time to revamp this discussion.

something we could do right now though is to allow you to specific a type via keywords, something like: `change type:function`, but that’s not very discoverable. If you want to explore this, a PR is welcome.

zachdaniel:
Keywords sound like a good idea, and potentially are a good first solution, as any UI can be built on top of them and ultimately end up modifying the search string, even if it is hidden from the user. I will look at it now 👍

zachdaniel:
It also looks like `lunr` is used under the hood which already has keyword search. It may be as simple as exposing the key in some way :)

josevalim:
Closing in favor of PR.

zachdaniel:
@josevalim I'm not sure if this is something that I've screwed up or not, but if I point `ex_doc` at the main branch on the repo, I get some kind of strange javascript error that prevents me from scrolling the sidebar. I can't see any new js errors in the console either. Just letting you know.

zachdaniel:
Nope :) Happens on commits before you merged the changes from my PR, so this must just be a side effect of trying to use ex_doc from GitHub. I will wait for a release :)

josevalim:
Related to #1773 perhaps? Can you post your git sha in there and how you can reproduce the failure?

zachdaniel:
[...]

PR #1766, opened 2023-09-10
improvement: add `type` field to the docs for reference in search

zachdaniel:
The type is already surfaced in search results, so I don't think this is a significant change. Something I noticed is that if you do `type:function foo` you are actually getting search results for "type:function OR foo" which is pretty unexpected in that context. Might want to do `and` for field matches and `OR` for the rest. Hard to say! I wasn't sure if I should include artifacts from `mix build` or anything like that in this PR. Happy to make any changes necessary :)

### DIFF
--- a/assets/js/handlebars/templates/search-results.handlebars
+++ b/assets/js/handlebars/templates/search-results.handlebars
@@ -40,7 +40,7 @@
     <li>Use <code>*</code> anywhere (such as <code>fo*</code>) as wildcard</li>
     <li>Use <code>+</code> before a word (such as <code>+foo</code>) to make its presence required</li>
     <li>Use <code>-</code> before a word (such as <code>-foo</code>) to make its absence required</li>
-    <li>Use <code>:</code> to search on a particular field (such as <code>field:word</code>). The available fields are <code>title</code> and <code>doc</code></li>
+    <li>Use <code>:</code> to search on a particular field (such as <code>field:word</code>). The available fields are <code>title</code>, <code>doc</code> and <code>type</code></li>
     <li>Use <code>WORD^NUMBER</code> (such as <code>foo^2</code>) to boost the given word</li>
     <li>Use <code>WORD~NUMBER</code> (such as <code>foo~2</code>) to do a search with edit distance on word</li>
   </ul>
--- a/assets/js/handlebars/templates/search-results.handlebars
+++ b/assets/js/handlebars/templates/search-results.handlebars
@@ -36,11 +36,11 @@
   <p>The search functionality is full-text based. Here are some tips:</p>
 
   <ul>
-    <li>Multiple words (such as <code>foo bar</code>) are searched as <code>OR</code></li>
+    <li>Multiple words (such as <code>foo bar</code>) are searched as <code>AND</code></li>
     <li>Use <code>*</code> anywhere (such as <code>fo*</code>) as wildcard</li>
-    <li>Use <code>+</code> before a word (such as <code>+foo</code>) to make its presence required</li>
+    <li>Use <code>+</code> before a word to make its presence optional. For example <code>+this +or_that </code></li>
     <li>Use <code>-</code> before a word (such as <code>-foo</code>) to make its absence required</li>
-    <li>Use <code>:</code> to search on a particular field (such as <code>field:word</code>). The available fields are <code>title</code>, <code>doc</code> and <code>type</code></li>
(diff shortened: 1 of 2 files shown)


zachdaniel:
This would effectively cover the need we have for this in Ash for #1765 in the short-to-medium term. I've got separate logic that modifies the search data to add searchable deep links to DSLs

josevalim:
@zachdaniel I think having those be OR is very confusing. Is that an easy change to make? Could you please explore?

josevalim:
Yeah, fine by me as well.

zachdaniel:
I'm open to this, but what I would have to do is do some kind of regex to pull out the field filters, and then apply them manually after the filter is run, or use something other than lunr. We can also advise users to use this syntax:

```
+type:function +ast
```

Which actually would result in a conjunction of the two. I could put this into the documentation string we show when you get to the search page. Just need to keep in mind that there will be no way to handle things like `contains ast AND (type == foo or type == bar)`.  See this issue: https://github.com/olivernn/lunr.js/issues/353

I could likely figure out a way to make it *always* be a conjunction, via somehow altering the search terms at some point in the pipeline (essentially making it as if the user put a `+` before each thing).

josevalim:
Yeah, I would look into lunr.Builder.searchPipeline and figure out if there is something to do to convert them all to AND. I think even regular query terms should be `AND`.

zachdaniel:
So I wasn't able to get the metadata I needed in the build pipeline for some reason, but I was able to manually take over the search building step and call into their queryString builder, and alter the results. While I was in there, I saw an opportunity to retain the old behavior by changing the definition of `+`. The new behavior now treats `+` as optional and the bare word as required. While `+` is a bit confusing to do that, there isn't a good way for me to choose a new prefix to do that work unfortunately. This still feels like an improvement to me, and makes things like `type:function reduce` behave very well.

assets/js/handlebars/templates/search-results.handlebars
```diff
@@ -36,11 +36,11 @@
   <p>The search functionality is full-text based. Here are some tips:</p>
 
   <ul>
-    <li>Multiple words (such as <code>foo bar</code>) are searched as <code>OR</code></li>
+    <li>Multiple words (such as <code>foo bar</code>) are searched as <code>AND</code></li>
     <li>Use <code>*</code> anywhere (such as <code>fo*</code>) as wildcard</li>
-    <li>Use <code>+</code> before a word (such as <code>+foo</code>) to make its presence required</li>
+    <li>Use <code>+</code> before a word to make its presence optional. For example <code>+this +or_that </code></li>
```
josevalim, line 41: It is a bit weird to use `+` for optional but I don't think we have another option, right? I am sorry for the back and forth on this but it may be better to just tell people to use `+type:function`. Unsure.

josevalim: In any case, it is a matter of picking the first commit or both. So just let me know your final thoughts and I will take it from here. Thank you.

zachdaniel: I think it is a bit weird, yes :) however, I think it's a bit weird either way honestly. It's an unconventional search experience regardless. I think I'm partial to making the more useful behavior the default (AND) especially given how unlikely I think users are to need `OR`. I understand the reservations though, whatever you decide will work for me in the end. As part of our (slow) migration back to hex I'll also be adding a "how to use this documentation" guide.

zachdaniel: Will do 👍

zachdaniel: Poll is here: https://twitter.com/ZachSDaniel1/status/1701590059391070711

It looks like it's relatively split, but leaning towards the new behavior introduced here. Its up for a week, I'm in no rush :)

josevalim: Sweet, thank you! Wake me up when ~september~ the pool ends.

zachdaniel: Okay, so it's still running, but it is almost *exactly* split which is...wild. I don't really care either way 😆. Probably better to err on the side of no change? The first commit to me is more important, allowing to filter on type. We're adding custom stuff to the search items with `type:dsl`, so that will help users search DSLs which is one of the main use cases of ash_hq.

josevalim: Yeah, I think no change and no additional code is best at the end.

josevalim:
Merged manually, thank you!

PR #1767, opened 2023-09-11
Fix invalid output markup for “hover link” headings which contain their own links

DavidOliver:
Avoids nested anchors.

Also:

- restores cheatsheet h3 hover link icons;
- slightly increases contrast of main link underlining.

---

As discussed [here](https://github.com/elixir-lang/ex_doc/pull/1752).


josevalim:

PR #1760, opened 2023-09-06
Screen reader accessibility

angelikatyborska:
Resolves #1756

Task list:

- Missing landmark elements: `<nav>`, `<main>` ✅  https://github.com/elixir-lang/ex_doc/pull/1760/commits/c9d58cebf349a372f7cf9982b8c5f757ef391518
- Expand menu button is missing `aria-controls` and `aria-expanded` properties  ✅ https://github.com/elixir-lang/ex_doc/pull/1760/commits/132ccb2b7563c365ae67224ffafe1830e9d448b0
- Search input: `<p>` is not a valid child of `<label>`  ✅ https://github.com/elixir-lang/ex_doc/pull/1760/commits/17adec572f5d5d35206f8ee2bfd403991fe0e0f4
- Version select:
	- Lacks a label.  ✅ https://github.com/elixir-lang/ex_doc/pull/1760/commits/86194b2d47f9c8d9c90922db7b62674a97885c45
	- Decorative character "\25bc" in version select is being read as "down pointing black pointer".  ✅ https://github.com/elixir-lang/ex_doc/pull/1760/commits/86194b2d47f9c8d9c90922db7b62674a97885c45
- Expand buttons in navigation links are missing `aria-controls` and `aria-expanded` properties. ✅ https://github.com/elixir-lang/ex_doc/pull/1760/commits/b65fe77252b61cbeb561064535d5bb6d686701e1
- Consider using aria-current for links visually styled as "active". ✅  https://github.com/elixir-lang/ex_doc/pull/1760/commits/56c0cf1bd93fa737cb4409fde25ff7ef132dff18
- "Pages / modules / mix tasks" lack feedback after clicking, rewrite using buttons and role=tablist, use aria-selected ✅ https://github.com/elixir-lang/ex_doc/pull/1760/commits/c868c2a8274cf7f0bee244677d964115bf723337
- Settings modal:
	- Close button lacks a label, being read as "times button". ✅ https://github.com/elixir-lang/ex_doc/pull/1760/commits/e4e895324af0d9a1471d83d278d34a661810fd56
	- Theme select lacks a label. ✅ I was mistaken here, it has a label already
- Copy button for code snippets:
	- Lacks aural feedback after successful copying, consider `aria-live` on the "copied!" text. ✅  https://github.com/elixir-lang/ex_doc/pull/1760/commits/7be1ee0701c0af6fb2e1bf0dd9f4349fc9d174e6
[...]

PR #1770, opened 2023-09-13
Fix tabpanel role invalid on a list element

angelikatyborska:
A follow up to https://github.com/elixir-lang/ex_doc/pull/1760


josevalim:

PR #1780, opened 2023-09-19
Accept a function in :deps option

pnezis:
When you are having private packages not published in hex there is no easy way to set the docs base url. You can use `:deps` but this has the following limitations:

- It is not applied recursively, so you need to define in the `:deps` non direct mix dependencies.
- For big projects with multiple internal dependencies it becomes difficult to maintain, unless you write custom linter tasks.
- If the base url of a dependency changes you need to update all packages depending on it.

`:base_url` allows setting the docs base url on the package itself.

Marking it as draft, since tests are missing

### DIFF
--- a/lib/mix/tasks/docs.ex
+++ b/lib/mix/tasks/docs.ex
@@ -93,9 +93,43 @@ defmodule Mix.Tasks.Docs do
       the "assets" directory in the output path under the name "cover" and the
       appropriate extension. This option has no effect when using the "html" formatter.
 
-    * `:deps` - A keyword list application names and their documentation URL.
-      ExDoc will by default include all dependencies and assume they are hosted on
-      HexDocs. This can be overridden by your own values. Example: `[plug: "https://myserver/plug/"]`
+    * `:deps` - A keyword list with application names and their documentation URL
+      or a function for generating such a list dynamically. ExDoc will by default include
+      all dependencies and assume they are hosted on HexDocs. This can be overridden by
+      your own values.
+
+      Using a keyword list:
+      
+      ```elixir 
+      docs: [
+        deps: [plug: "https://myserver/plug/"]
+      ]
+      ```
+
+      Alternatively you can use a function to generate this list dynamically. For example
+      you could read a custom project option from your dependencies and update the deps
+      list if this is set:
+
+      ```elixir
+      docs: [
+        deps: fn -> deps_base_urls() end
+      ]
+
+      defp deps_base_urls do
+        for {app, path} <- Mix.Project.deps_paths(),
+            _ = Application.load(app),
+            base_url = base_url(app, path),
+            not is_nil(base_url) do
+          {app, base_url}
+        end
+      end
+
+      defp base_url(app, path) do
+        Mix.Project.in_project(app, path, fn _ ->
+          get_in(Mix.Project.config(), [:docs, :base_url])
+        end)
+      end
+      ``` 
 
     * `:extra_section` - String that defines the section title of the additional
       Markdown and plain text pages; default: "PAGES". Example: "GUIDES"
@@ -482,7 +516,7 @@ defmodule Mix.Tasks.Docs do
   end
 
   defp normalize_deps(options) do
-    user_deps = Keyword.get(options, :deps, [])
+    user_deps = user_deps(options[:deps])
 
     deps =
       for {app, doc} <- Keyword.merge(get_deps(), user_deps),
@@ -492,6 +526,10 @@ defmodule Mix.Tasks.Docs do
 
     Keyword.put(options, :deps, deps)
   end
+
+  defp user_deps(nil), do: []
+  defp user_deps(deps) when is_function(deps, 0), do: deps.()
+  defp user_deps(deps), do: deps
 
   defp get_deps do
     for {key, _} <- Mix.Project.deps_paths(),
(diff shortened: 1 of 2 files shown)


josevalim:
The problem is that it will also break links to Elixir and any other public Hex package. :(

pnezis:
> The problem is that it will also break links to Elixir and any other public Hex package. :(

Why will the links break? This is used only if set, the default behaviour does not change.

josevalim:
I mean, in your own docs. Then you can link or refer to any public package.

josevalim:
I mean, in your own docs. If you set base_url then you can't link or refer to any public package.

pnezis:
> I mean, in your own docs. If you set base_url then you can't link or refer to any public package.

Probably I am missing something but from the code I see that the base_url is applied only on the modules of the given dependencies. The options struct has a `deps` key which for a sample project has the following:

```
 [
  formatters: ["html"],
  deps: [
    bar: "https://hexdocs.pm/earmark_parser/bar/0.1.0",
    baz: "https://hexdocs.pm/earmark_parser/baz/0.1.0",
    earmark_parser: "https://hexdocs.pm/earmark_parser/1.4.35/",
    ex_doc: "https://hexdocs.pm/ex_doc/0.30.6/",
    makeup: "https://hexdocs.pm/makeup/1.1.0/",
    makeup_elixir: "https://hexdocs.pm/makeup_elixir/0.16.1/",
    makeup_erlang: "https://hexdocs.pm/makeup_erlang/0.1.2/",
    nimble_parsec: "https://hexdocs.pm/nimble_parsec/1.3.1/"
  ],
  apps: [:foo],
  source_beam: "/Users/panagiotisnezis/code/oss/ex_doc_test/foo/_build/dev/lib/foo/ebin",
  formatter: "html",
  proglang: :elixir
]
```

by setting `base_url` to `bar` and `baz` you modify it to:

```elixir
 [
  formatters: ["html"],
  deps: [
    bar: "https://my.server/bar",
    baz: "https://my.server/baz",
    earmark_parser: "https://hexdocs.pm/earmark_parser/1.4.35/",
    ex_doc: "https://hexdocs.pm/ex_doc/0.30.6/",
    makeup: "https://hexdocs.pm/makeup/1.1.0/",
    makeup_elixir: "https://hexdocs.pm/makeup_elixir/0.16.1/",
    makeup_erlang: "https://hexdocs.pm/makeup_erlang/0.1.2/",
    nimble_parsec: "https://hexdocs.pm/nimble_parsec/1.3.1/"
  ],
  apps: [:foo],
  source_beam: "/Users/panagiotisnezis/code/oss/ex_doc_test/foo/_build/dev/lib/foo/ebin",
  formatter: "html",
  proglang: :elixir
]
```

which is the exact same thing you would get if you set 

```
deps: [
   bar: "https://my.server/bar",
   baz: "https://my.server/baz"
]
```

in the docs options of `foo`

pnezis:
I did a quick test, links seem fine:

image

image

josevalim:
I see, I misunderstood the code, sorry. :) You are actually reaching into the dependency to get its `base_url`! I thought you were using base_url to set the base for all deps.

However, I am not sure how much I like the current approach either. I can see unexpected scenarios. For example, will this work if the dependency is a rebar dependency or a makefile?

We could store this configuration in the application environment but it sounds problematic too. It is a pity Erlang does not allow us to add custom keys to the application specification, because that would be perfect here. Maybe a compromise is to allow `deps: fn dep -> ... end` to be a function. This way you can implement whatever logic you want to get a custom URL. WDYT?

josevalim:
One of such logic could be the code you have here or reading a file within the dependency that is specific to your org.

pnezis:
Making `deps` support also a function would be perfectly fine (the only "drawback" is that I will have to add this function to all internal projects mix.exs, but It is fine).

I can update the PR accordingly if you are ok with this. 

Any idea about how to test this (without using fixtures)?

pnezis:
> However, I am not sure how much I like the current approach either. I can see unexpected scenarios. For example, will this work if the dependency is a rebar dependency or a makefile?

I have never user a rebar or a makefile dependency, but I will try it out of curiosity. How are these handled by the current implementation of `normalize_deps`?

josevalim:
The issue is Mix.Project.in_project. So the current code does not have to worry about it.

pnezis:
@josevalim `Mix.Project.in_project`'s `fun` will be invoked with a [`nil` if not a `Mix` project](https://github.com/elixir-lang/elixir/blob/e66e6efb13fc1db958dbc2cc357c1be56b70fcf6/lib/mix/lib/mix/project.ex#L908C5-L908C5)  and a default config will be returned so I guess this will work in these cases as well (tested locally with hackney as a dependency).

Making `deps` accepting a function as well is perfectly fine. WDYT?

pnezis:
@josevalim updated the MR, also added the initial implementation as an example in the docs.

lib/mix/tasks/docs.ex
```diff
@@ -93,9 +93,43 @@ defmodule Mix.Tasks.Docs do
...
+      
+      ```elixir 
+      docs: [
+        deps: [plug: "https://myserver/plug/"]
+      ]
+      ```
+
+      Alternatively you can use a function to generate this list dynamically. For example
+      you could read a custom project option from your dependencies and update the deps
+      list if this is set:
```
josevalim, line 111: That's not what I had in mind... but this is better. :)

Actually, you can wrap the whole docs generation in a function, which means we don't need to support functions here. You can do all of the below today by doing this:

```
docs: fn ->
  [deps: deps_base_urls()]
end
```

So I think we can close this? :)

pnezis: Awesome I had missed this, yes we can close this :)

PR #1783, opened 2023-09-25
Use backticks consistently in `mix docs` docs

joshprice:


lib/mix/tasks/docs.ex
```diff
@@ -10,11 +10,11 @@ defmodule Mix.Tasks.Docs do
...
     * `--canonical`, `-n` - Indicate the preferred URL with
-      rel="canonical" link element, defaults to no canonical path
+      `rel="canonical"` link element, defaults to no canonical path
 
-    * `--formatter`, `-f` - Which formatters to use, "html" or
-      "epub". This option can be given more than once. By default,
-      both html and epub are generated.
+    * `--formatter`, `-f` - Which formatters to use, `"html"` or
+      `"epub"`. This option can be given more than once. By default,
+      both `html` and `epub` are generated.
```
wojtekmach, line 17: since these are CLI flags I think we can drop string quotes.

```suggestion
    * `--formatter`, `-f` - Which formatters to use, `html` or
      `epub`. This option can be given more than once. By default,
      both `html` and `epub` are generated.
```

lib/mix/tasks/docs.ex
```diff
@@ -24,8 +24,8 @@ defmodule Mix.Tasks.Docs do
     * `--output`, `-o` - Output directory for the generated
       docs, default: `"doc"`
 
-    * `--proglang` - Chooses the main programming language: "elixir"
-      or "erlang"
+    * `--proglang` - Chooses the main programming language: `"elixir"`
+      or `"erlang"`
```
wojtekmach, line 28:
```suggestion
    * `--proglang` - Chooses the main programming language: `elixir`
      or `erlang`
```

wojtekmach (approved):

wojtekmach:
Thank you!

PR #1787, opened 2023-10-03
Partially fix abstract Erlang types (`{@type ...}`)

paulo-ferraz-oliveira:
`@type` is deprecated but [still documented](https://www.erlang.org/doc/apps/edoc/chapter.html): you get a warning upon documentation generation.

This surfaced at https://github.com/starbelly/rebar3_ex_doc/issues/38.

Code out in the wild may contain references like the ones we include in the test.

This pull request attempts at solving a basic case for those.

EDoc seems broken for some types of `@type` (or at least not generating useful information), but at least this commit prevents a crash.

While discussing the solution with @starbelly we didn't know where to put the normalisation code. It might be good here, or it might belong in the `doc_ast.ex` module.

### DIFF
--- a/lib/ex_doc/language/erlang.ex
+++ b/lib/ex_doc/language/erlang.ex
@@ -215,6 +215,24 @@ defmodule ExDoc.Language.Erlang do
     binary
   end
 
+  defp walk_doc({:code, attrs, [{:a, a_attrs, [a_inner], _a_meta} | _] = a, meta}, _config) do
+    {arity, extra} =
+      case :lists.keydelete(:a, 1, a) do
+        [] ->
+          {0, ""}
+
+        [extra] ->
+          # a_meta is not broken down, so we have to parse stuff like (X, Y)
+          {extra |> String.split(",") |> Enum.count(), extra}
+      end
+
+    type = type_from_fragment(:otp, a_attrs[:href])
+    url = fragment(:ex_doc, :type, type, arity)
+    text = a_inner <> extra
+
+    {:a, [href: url], {:code, attrs, [text], meta}, %{}}
+  end
+
   defp walk_doc({:code, attrs, [code], meta} = ast, config) do
     {text, url} =
       case parse_autolink(code) do
@@ -455,6 +473,8 @@ defmodule ExDoc.Language.Erlang do
   defp fragment(:ex_doc, :type, name, arity) do
     "#t:#{name}/#{arity}"
   end
+
+  defp type_from_fragment(:otp, "#type-" <> name), do: name
 
   defp kind("c:" <> rest), do: {:callback, rest}
   defp kind("t:" <> rest), do: {:type, rest}
--- a/test/ex_doc/language/erlang_test.exs
+++ b/test/ex_doc/language/erlang_test.exs
@@ -124,6 +124,27 @@ defmodule ExDoc.Language.ErlangTest do
                ~s|<a href="https://www.erlang.org/doc/man/array.html#type-array"><code>array:array()</code></a>|
     end
 
+    test "abstract types - description", c do
+      # if type exists:
+      # <code><a href="#type-myList">myList</a>(X)</code>
+      assert autolink_doc("{@type myList(X). A special kind of lists ...}", c) ==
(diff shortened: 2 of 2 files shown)


lib/ex_doc/language/erlang.ex
```diff
@@ -215,6 +215,24 @@ defmodule ExDoc.Language.Erlang do
...
+  defp walk_doc({:code, attrs, [{:a, a_attrs, [a_inner], _a_meta} | _] = a, meta}, _config) do
+    {arity, extra} =
+      case :lists.keydelete(:a, 1, a) do
+        [] ->
+          {0, ""}
+
+        [extra] ->
+          # a_meta is not broken down, so we have to parse stuff like (X, Y)
+          {extra |> String.split(",") |> Enum.count(), extra}
+      end
```
paulo-ferraz-oliveira, line 227: To me this seems like a silly way to calculate the arity, but I couldn't find a better one.

lib/ex_doc/language/erlang.ex
```diff
@@ -215,6 +215,24 @@ defmodule ExDoc.Language.Erlang do
...
+
+    type = type_from_fragment(:otp, a_attrs[:href])
```
paulo-ferraz-oliveira, line 229: We split "type-" from the actual type. I couldn't find this explicitly done in the current code, but could find composing into "type-<Type>".

lib/ex_doc/language/erlang.ex
starbelly: You can calculate the arity but you'll have to reduce, here's an example of what a non zero arity would look like : 

```
     {code,[],
        [<<"{that_odd_form, ">>,
         {a,
          [{href,
            <<"_build/default/lib/stdlib/doc/queue.html#type-queue">>}],
          [<<"queue:queue">>]},
         <<"(">>,
         {a,
          [{href,
            <<"_build/default/lib/stdlib/doc/uri_string.html#type-uri_string">>}],
          [<<"uri_string:uri_string()">>]},
         <<")}">>]}
```         

Or with a built-in 

```
      {code,[],
        [<<"{that_odd_form, ">>,
         {a,
          [{href,
            <<"_build/default/lib/stdlib/doc/queue.html#type-queue">>}],
          [<<"queue:queue">>]},
         <<"(any())}">>]}
```

Whether it's worth it I suppose is another question :)

josevalim:
I wouldn’t try to do anything smart. Can we just leave the HTML as is, making sure it does not crash?

paulo-ferraz-oliveira:
Wdym? Not putting `code` inside the `a`? Or the `type-` thing?

paulo-ferraz-oliveira:
Or simply avoiding the crash but not taking `{@type ...}` into account either (since it's deprecated)?

paulo-ferraz-oliveira:
I'll do a new commit to simply avoid the crash and nothing else.

paulo-ferraz-oliveira:
This is what's visible with the updated code (01b60652)

image

Links are not usable, because generated HTML (in doc) points to e.g. `#type-myList`, while anchor is identified (above) as `#t:myList/0`.

paulo-ferraz-oliveira:
Using all `@type` examples [from EDoc](https://www.erlang.org/doc/apps/edoc/chapter.html) we get

image

and no crashes 👍 (except in the tests - if I try to create some for the examples, but I'm not sure if you want me to "fix" that too)

paulo-ferraz-oliveira:
I added this to the tests (locally - `autolink_doc` - where `search` and `arity` are new function arguments)

```elixir
        {:docs_v1, _, _, "application/erlang+html", %{}, _, extra} ->
          [{_, _, _, %{"en" => doc}, _}] =
            Enum.filter(extra, fn
              {{:type, ^search, ^arity}, _, _, _, _} -> true
              _ -> false
            end)
```

as well as more tests, but maybe it's not worth it (?) I can push if you wanna.

josevalim (approved):
If @wojtekmach is happy, I am happy!

wojtekmach:
Thank you! I'd strip out the `<a>` and just leave `<code>` but it definitely doesn't hurt to keep it this way for now.

PR #1798, opened 2023-10-29
Restore warning on missing references and `.html` link (for `.md`)

paulo-ferraz-oliveira:
While 9579ce8 (from #1617) added features to `ex_doc`, it also stopped using hardcoded language `Elixir`, which seems to have had the unintended consequence of removing other features, like:

* the ability to report on invalid links (e.g. `warning: documentation references file "examples" but it does not exist`), and
* auto-linking from `.md` to `.html` (links aren't "transformed"), e.g. as per using `rebar3_ex_doc`.

I found the commit by `git bisect`'ing, and it helped me understand where the origin of the issue was.

This pr is just an initial step as I understand you might not want the Erlang handler calling the Elixir handler, but:

1. I don't know where to put the "generic" code,
3. I don't know if it's acceptable to simply copy the code over.

I did copy the tests over from the Elixir counterpart, though, as that seemed to make sense (same requirements, I guess).

### DIFF
--- a/lib/ex_doc/language/elixir.ex
+++ b/lib/ex_doc/language/elixir.ex
@@ -4,7 +4,6 @@ defmodule ExDoc.Language.Elixir do
   @behaviour ExDoc.Language
 
   alias ExDoc.Autolink
-  alias ExDoc.Refs
   alias ExDoc.Language.Erlang
 
   @impl true
@@ -392,10 +391,6 @@ defmodule ExDoc.Language.Elixir do
   defp process_type_ast({:"::", _, [d | _]}, :opaque), do: d
   defp process_type_ast(ast, _), do: ast
 
-  ## Autolinking
-
-  @autoimported_modules [Kernel, Kernel.SpecialForms]
-
   defp walk_doc(list, config) when is_list(list) do
     Enum.map(list, &walk_doc(&1, config))
   end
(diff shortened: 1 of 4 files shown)


lib/ex_doc/language/elixir.ex
```diff
@@ -440,7 +440,7 @@ defmodule ExDoc.Language.Elixir do
 
   @ref_regex ~r/^`(.+)`$/
 
-  defp custom_link(attrs, config) do
+  def custom_link(attrs, config) do
```
paulo-ferraz-oliveira, line 443: This is the function I mention in the description. Where to put it, and the code it depends on?

lib/ex_doc/language/erlang.ex
```diff
@@ -303,7 +303,13 @@ defmodule ExDoc.Language.Erlang do
         inner
 
       _ ->
-        ast
+        case ExDoc.Language.Elixir.custom_link(attrs, config) do
+          nil ->
```
paulo-ferraz-oliveira, line 307: I also understand this is rather ugly, but maybe having the function moved can help me understand better how to code this bit...

lib/ex_doc/language/elixir.ex
josevalim: You can put it in the ExDoc.Autolink module :)

paulo-ferraz-oliveira: Got it, thanks. I'll move it, and the code it depends on, to that one.

paulo-ferraz-oliveira: ce25ca50

josevalim (approved):

wojtekmach:
Thank you!

Issue #1799, opened 2023-11-02
Visual readability of  sidebar tabs


sodapopcan:
Hello ExDoc team!

I've always had a bit of trouble with the visual readability of the sidebar tabs.  It actually took me an embarrassingly long time to figure out that they were even there (I've known about them for a while now 😅).  My issue is with how they are visually grouped with the project info as opposed to with the content they are link to.  There have several posts on Elixir Forums lately that seem to suggest newbies don't realize that the docs also contain guides.  It's my hypothesis that this may have something to do with it.

Now, of course I am a sample size of one don't have much of a clue if anyone else feels this way.  I was going to post on Elixir Forum but wanted to post here first as I actually have a quick little prototype to show and feel like sharing a visual design in a larger public forum can be asking for a headache 😅  I figured I'd share here first and if you don't immediately veto, I could ask for further feedback if necessary.  Otherwise, I'm happy to incorporate feedback, polish, and make a PR.

So here's my prototype:

<img width="1441" alt="image" src="https://github.com/elixir-lang/ex_doc/assets/3752792/a03a75db-f439-4dab-a29a-5367dd05396c">

Just classic tabs!  I did this very quickly and have a slightly smoother version in mind, but I'm ultimately going for "utilitarian while still visually pleasing enough."  I feel this does a better of saying, "Hey, there is more interesting stuff over here!" as it ties the tabs to their content. 

No worries if you're not a fan but I wanted to share.

✌️

josevalim:
I will leave it open for feedback but I recommend trying on an expanded sidebar so there is a scrollbar appearing. IIRC, one of the main reasons for having the spacing around the title-box/sidebar-header was so the scrollbar "disappears". You can see this happening if I remove the margin from the current layout:

<img width="730" alt="Screenshot 2023-11-02 at 22 56 32" src="https://github.com/elixir-lang/ex_doc/assets/9582/cc0fd90b-c9ba-4dae-b6eb-b897740049dd">

It now feels the scrollbar is encroaching. :)

sodapopcan:
It would have probably been a good idea to post a screenshot with the scrollbar, eh?  I did check it and actually _made_ it touch the top like that as I thought it was nicer 😅  Though I have no strong feelings there.  I'm one of those "hide scrollbars when no in use"-type people.

Here's one with scrollbars set to "always show" and some padding at the top (this is scrolled all the way up):

[...]

PR #1800, opened 2023-11-04
Redesign sidebar tabs

sodapopcan:
This is the result of discussion in #1799.

Make section tabs more tab-like, connecting the tabs to their content.  Ideally this will draw more attention to the different sections as it seems new users sometimes have trouble finding the guides.

Here's a screencap:

image


josevalim:

DavidOliver:
Nice!

PR #1801, opened 2023-11-04
Prevent search bar from overlapping settings cog

sodapopcan:
This PR fixes an issue where the settings cog gets partially hidden under the search bar at certain sizes.

I've just added a `row-gap` between the search-bar and settings cog which seems to do the trick.  I feel there is some unnecessary CSS (which I will annotate) which doesn't seem to do anything, at least not once there is the gap.  I could be related to other browsers, of course.

I also removed some padding which seems to line up the cog with the code icon below a bit better, but I will undo that if requested (it's very subtle).

### Before

https://github.com/elixir-lang/ex_doc/assets/3752792/ce3c3142-9a3e-4294-b441-191821b5c07f

### After

https://github.com/elixir-lang/ex_doc/assets/3752792/2e81abaf-e0ca-43c3-913c-829e53d03edf

### DIFF
--- a/assets/css/search-bar.css
+++ b/assets/css/search-bar.css
@@ -4,6 +4,7 @@
 
 .search-settings {
   display: flex;
+  column-gap: 12px;
   align-items: center;
 }
 
@@ -13,7 +14,6 @@
   height: 48px;
   position: relative;
   flex-grow: 1;
-  width: 96%;
 }
 
 .top-search .search-bar .search-input {
@@ -95,7 +95,6 @@
 
 .top-search .search-settings .icon-settings {
   font-size: 20px;
-  width: 4%;
   float: right;
   color: var(--iconAction);
   text-decoration: none;
@@ -103,7 +102,7 @@
   transition: color .3s ease-in-out;
   background-color: transparent;
   cursor: pointer;
-  padding: 1px 0px 1px 5px;
+  padding: 0;
 }
 
 .top-search .search-settings .icon-settings:hover {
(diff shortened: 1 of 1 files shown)


sodapopcan:
Oh right, I can't comment on lines that aren't part of a changed hunk.

[These lines](https://github.com/elixir-lang/ex_doc/pull/1801/files#diff-6b3b7e77e186f9552354c48e01abf92fdb9e35513a112691e0cb42fa12751c73R123-R125) and [these line](https://github.com/elixir-lang/ex_doc/pull/1801/files#diff-6b3b7e77e186f9552354c48e01abf92fdb9e35513a112691e0cb42fa12751c73R134-R136) seem unnecessary, at least after adding the gap.  I could be wrong, of course.

josevalim:
Can we keep the gap consistent? Why does it reduce as we slide? Also feel free to remove those lines. :)

sodapopcan:
> Can we keep the gap consistent?

Yes!  It just means deleting some more code and I felt bad about doing that 😅  Well, also nervous about doing it as I only have a Mac to test on.  It's always a crapshoot with CSS as it's _so_ easy to leave unnecessary styles behind.

I'll see what I can do.

sodapopcan:
How's this?

https://github.com/elixir-lang/ex_doc/assets/3752792/d3b38421-8226-42d2-a0b7-7ff24c09e372

josevalim:
Perfect, thank you!

sodapopcan:
*ahem* I'm only working on these to get the hearts...

Naw, I'm kidding!  No prob!

PR #1802, opened 2023-11-06
Add Erlang Markdown doc support

garazdawi:
This PR introduces a lot of changes in order to make it possible to use ExDoc for generating html+epub documentation for Erlang/OTP projects using Markdown as the input source. The aim is to make it possible for Erlang projects (including the Erlang/OTP project) to use ExDoc + Markdown for documentation.

The major additions are:

* Place language specific part sof the auto-linking process in the language callbacks. This includes both links to built-in/prefined functions/types and parsing of MFA links.
* Make it possible to visually group types, just as functions and extra pages are allowed to be grouped today.
* Add support for using `e:` for linking to extra pages in dependencies (no checking of the extra page actually exists is done)
* Add support for using `m:` for explicitly linking to a module in a regular_link. Also added an option to make this mandatory which is set to true for Erlang and false for Elixir. So in Erlang you would write either `` `m:foo` `` or `` [`foo`](`foo`) ``  which would produce the same thing, `<a href="lists.html"><code ...>foo</code></a>`.

Please see each individual commit for details about the changes and some minor changes/fixes that I encountered along the way.

Feel free to change/update/cherry-pick anything you like in here.


test/ex_doc/language/erlang_test.exs
```diff
@@ -180,7 +180,193 @@ defmodule ExDoc.Language.ErlangTest do
     end
   end
 
-  describe "autolink_doc/2 for extras" do
+  describe "autolink_doc/2 for markdown" do
+    test "module in module code", c do
+      assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
+               assert autolink_markdown("`erlang_bar`", c) ==
+                        ~s|<code class="inline">erlang_bar</code>|
+             end) == ""
```
wojtekmach, line 188: there's no need to assert nothing was printed. While we don't have an automatic way to fail the build if warning was emitted, it's fine for now to rely on ourselves to spot these when running tests locally. Btw don't worry about it, I'll fix these up after this PR lands.

```suggestion
    assert autolink_markdown("`erlang_bar`", c) ==
             ~s|<code class="inline">erlang_bar</code>|
```

wojtekmach (commented):
@garazdawi this is amazing. I left a few minor comments but I'm definitel happy with the code as is. Thank you.

Do you have an OTP branch that uses ExDoc that I could try locally? I'd love to do some local testing which will for sure answer some of the questions I left.

lib/ex_doc/language/erlang.ex
```diff
@@ -215,116 +260,140 @@ defmodule ExDoc.Language.Erlang do
...
         end
 
       "https://erlang.org/doc/link/seemfa" ->
-        {kind, url} =
+        {prefix, url} =
           case String.split(attrs[:href], "Module:") do
-            [url] -> {:function, url}
-            [left, right] -> {:callback, left <> right}
+            [_] ->
+              case String.split(attrs[:href], "Mod:") do
```
wojtekmach, line 310: We can also do `String.split(..., ["Module:", "Mod:"])`. As you mentioned in the commit message perhaps we don't need it after all? I'd rather fix up the docs upstream. I have an old PR for diameter docs, https://github.com/erlang/otp/pull/5160, and happy to create more for stranglers if any.

lib/ex_doc/autolink.ex
```diff
@@ -370,7 +370,7 @@ defmodule ExDoc.Autolink do
   def parse_function("%"), do: {:function, :%}
 
   def parse_function(string) do
-    case Code.string_to_quoted("& #{string}/0") do
+    case Code.string_to_quoted("& #{string}/0", warnings: false) do
```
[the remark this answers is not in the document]
garazdawi: It warned for `` `...` `` and similar.

lib/ex_doc/language/erlang.ex
```diff
@@ -265,7 +265,17 @@ defmodule ExDoc.Language.Erlang do
   end
 
   defp walk_doc({:code, attrs, [code], meta} = ast, config) when is_binary(code) do
-    case Autolink.url(code, :regular_link, config) do
+    ## Translate foo() to foo/0
```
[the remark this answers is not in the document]
garazdawi: The feature is already there for types in Erlang ExDoc, I only expanded it to functions and callbacks for consistency.

That being said, I'm not commited to this functionality, tought I do think it makes some docs looks nicer to have `()` instead of `/0`.

lib/ex_doc/language/erlang.ex
garazdawi: We can drop this commit, but the `Module` part needs to remain as edoc still uses the convention.

test/ex_doc/language/erlang_test.exs
garazdawi: You are braver than I am :smile: I've stopped trusting myself to notice things manually

garazdawi:
You can use this branch to test things: https://github.com/garazdawi/otp/commits/lukas/convert-docs-to-markdown/

```
ERL_TOP=`pwd`
./otp_build setup -a
make docs                    ## You can access the docs in doc/index.html
cd lib/stdlib/ && make docs  ## or lib/*/doc/html/index.html for specific applications
```

You need the ex_doc in your path for the docs to build.

lib/ex_doc/autolink.ex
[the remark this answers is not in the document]
garazdawi: Or maybe it was `` `mod:.../2` ``. Something with `...` in the name.

lib/ex_doc/language/erlang.ex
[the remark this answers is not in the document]
josevalim: @garazdawi types are the only place where we traverse the AST to generate links, so I would argue this is a bit different. I would indeed go with `foo/0` for consistency. :)

test/ex_doc/language/erlang_test.exs
josevalim: We also support concurrent captures, which means that `== ""` may fail if we end-up capturing something from another test. :)

lib/ex_doc/language/erlang.ex
josevalim: So what uses "Mod:" and what uses "Module:"? We are very happy to support both but at the same time it feels like an opportunity to be consistent? Or is "Module:" edoc and "Mod:" the erlang docs?

lib/ex_doc/doc_ast.ex
```diff
@@ -170,8 +170,21 @@ defmodule ExDoc.DocAST do
...
+
+      {lang, lexer, opts} ->
+        try do
+          render_code(pre_attr, lang, lexer, opts, code, outer_opts)
+        rescue
+          MatchError ->
+            IO.puts(:stderr, [
+              IO.ANSI.format([:yellow, "warning: ", :reset]),
+              "Crashed while formatting #{full_block}\n"
+            ])
```
josevalim, line 184: Is this for debugging or is this necessary permanently because "invalid" code blocks are being given?

lib/ex_doc/language/elixir.ex
```diff
@@ -183,9 +183,170 @@ defmodule ExDoc.Language.Elixir do
...
+  def parse_module(":" <> _ = string, :custom_link) do
+    do_parse_module(string)
+  end
+
+  def parse_module(_, _) do
+    :error
+  end
+
+  defp do_parse_module(string) do
+    case Code.string_to_quoted(string, warn_on_unnecessary_quotes: false) do
```
josevalim, line 325:
```suggestion
    case Code.string_to_quoted(string, warn_on_unnecessary_quotes: false, emit_warnings: false) do
```

josevalim: I'm adding emit_warnings as a flag to emit all warnings.

lib/ex_doc/autolink.ex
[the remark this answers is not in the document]
josevalim: I will make it public on Elixir v1.16+ as:

```suggestion
    case Code.string_to_quoted("& #{string}/0", emit_warnings: false) do
```

lib/ex_doc/language/erlang.ex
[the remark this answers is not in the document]
garazdawi: I'm not sure I follow the connection to traversing the AST? These are links from extra pages to types that take the form `` `t:sofs:type()` ``.

lib/ex_doc/doc_ast.ex
garazdawi: I would prefer to always have it as then we can put the style and github will render it with the correct syntax highlighting without having to fix the bug in the specific makeup flavor that crashes. I found that there are quite a few of our C code snippets that fail in `makeup_c`.

mix.exs
```diff
@@ -24,7 +24,7 @@ defmodule ExDoc.Mixfile do
 
   def application do
     [
-      extra_applications: [:eex, :crypto] ++ extra_applications(Mix.env()),
+      extra_applications: [:eex] ++ extra_applications(Mix.env()),
```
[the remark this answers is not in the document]
garazdawi: It was unclear to me why crypto was a dependency and I wanted to be able to build the docs even on systems where crypto was not available. So I removed it to trigger a discussion :)

lib/ex_doc/language/erlang.ex
garazdawi: `Module` is edoc and most of Erlang/OTP. `Mod:` is only diameter as far as I know and that should be fixed.

I will remove the `Mod` parts from this commit.

test/ex_doc/formatter/html/erlang_test.exs
```diff
@@ -54,7 +53,11 @@ defmodule ExDoc.Formatter.HTML.ErlangTest do
       extras: []
     ]
 
-    ExDoc.generate_docs(config[:project], config[:version], config)
+    ## Silence warnings generated by docs
```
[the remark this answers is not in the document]
garazdawi: I'm not sure how to do that... I've pushed a version that removed the silence but I think I need a bit of help figuring out how to selectively silence only what we want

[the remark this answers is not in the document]
wojtekmach: @garazdawi I don't we need to worry about it too much. Something I wanted to do for a long time was for these modules to _return_ warnings instead of printing them right away so the problem will go away. I'm happy to work on that after this lands.

lib/ex_doc/language/erlang.ex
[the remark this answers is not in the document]
josevalim: Oh, I see. I misunderstood then. Shouldn't we update those and be consistent so we have `t:sofs:type/0`?

lib/ex_doc/retriever.ex
```diff
@@ -357,9 +375,18 @@ defmodule ExDoc.Retriever do
...
     annotations = if type_data.type == :opaque, do: ["opaque" | annotations], else: annotations
     doc_ast = doc_ast(content_type, doc, file: source.path)
+    metadata = Map.put(metadata, :__doc__, :type)
+    group = GroupMatcher.match_function(groups_for_docs, metadata)
+
+    id =
+      if name == nil do
+        "t:nil/#{arity}"
+      else
+        "t:#{name}/#{arity}"
```
josevalim, line 385: Perhaps worth creating a tiny function called `name_or_nil` that we can reuse across this module?

lib/ex_doc/doc_ast.ex
josevalim, line 184:
```suggestion
          e ->
            IO.puts(:stderr, [
              IO.ANSI.format([:yellow, "warning: ", :reset]),
              "crashed while highlighting #{lang} snippet:\n\n",
              full_block,
              "\nFull error message shown next:\n\n",
              Exception.format(:error, e, __STACKTRACE__)
            ])
```

lib/ex_doc/formatter/html.ex
```diff
@@ -72,22 +72,24 @@ defmodule ExDoc.Formatter.HTML do
       extras: extra_paths(config),
       skip_undefined_reference_warnings_on: config.skip_undefined_reference_warnings_on,
       skip_code_autolink_to: config.skip_code_autolink_to,
+      force_module_prefix: config.force_module_prefix,
```
josevalim, line 75: What is the use case for this one? :)

lib/ex_doc/language/erlang.ex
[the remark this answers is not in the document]
garazdawi: I don't mind doing that, you can always write `` [`sofs:type()`](`t:sofs:type/0`) `` if you want the brackets.

Just be aware that doing that would be a backwards-incompatible change, albeit an undocumented behaviour, but there are tests verifying that it works.

lib/ex_doc/autolink.ex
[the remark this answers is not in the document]
garazdawi: Then I will leave it as is for now. I'll see if I can add a testcase triggering the warning.

test/ex_doc/language/erlang_test.exs
```diff
@@ -223,6 +223,24 @@ defmodule ExDoc.Language.ErlangTest do
              end) == ""
     end
 
+    test "escaped function in module", c do
+      assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
+               assert autolink_markdown(~S"`\c:ls/0`", c, deps: [stdlib: "stdlib/"]) ==
```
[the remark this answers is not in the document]
garazdawi: I was thinking about the same thing, but decided against it since all other prefixes are single letter.

I'm fine with changing it if you want to.

lib/ex_doc/formatter/html.ex
garazdawi: Either to allow elixir fork to forced their projects to use the `m:` prefix (from the other threads I gathered that possibly @wojtekmach wanted this?). Also, it is for people that are used to writing docs in Elixir and want to keep the same ergonomics in Erlang docs.

It was easy to add, so I added it. I'm fine with removing it if you think it makes things more complicated than they need to be.

wojtekmach: do you mean that Elixir folks could configure their project to require writing `` `m:String` `` because otherwise it won't autolink? Yeah let's skip this for now.

wojtekmach: btw, In Elixir project I noticed `` `m:EarmarkParser` `` creates `<a href=...><code>m:EarmarkParser</code></a>`, that is, it does _not_ remove the `m:` prefix in the renderer text. I also noticed `m::lists` does _not_ autolink. That being said, I think you don't explicitly document `m:` working for Elixir so I believe it's fine to be rough around the edges for now.

garazdawi: I'll remove the config and fix the `m:` prefix removal in elixir.

Regaring `m::lists`, it should work but it only works as the link target (i.e. a `:custom_link`). `:lists` has as far as I can tell never worked as a ":regular_link".

garazdawi: as I hit enter I realized that the `m:` should not be used in `:custom_links`. So I'll leave that to you to figure out if you want to allow `m::lists` in regular links when `:lists` is not allowed.

lib/ex_doc/retriever.ex
garazdawi: done

lib/ex_doc/autolink.ex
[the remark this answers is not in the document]
garazdawi: Added testcase. You needed `` `...a/0` `` to trigger it.

lib/ex_doc/formatter/html.ex
wojtekmach: agreed, there's no need for `m:` in custom links. (Come to think of it, it's less about whether it's regular vs custom link it's all about whether we're doing relaxed vs strict reference parsing. Should better express that in code one of those days.)

I think `m::lists` would be a nice addition for Elixir projects but it's not high priority at the moment IMHO. WDYT @josevalim?

lib/ex_doc/language/erlang.ex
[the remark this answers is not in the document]
josevalim: @garazdawi I would say it is _your_ decision. I am assuming this will be used as the foundation for future docs, so now would be the best time to start enforcing more consistent rules. I am fine with supporting your choice. :)

lib/ex_doc/formatter/html.ex
josevalim: Regarding this option, let's remove it. I think it is fine for Erlang to be more explicit here (`m:...`) while Elixir allows both, especially since modules in Elixir have a different casing than regular atoms. :) If someone reaaaaaaally wants it, they can send a PR later :D

garazdawi: `` `m:Enum` `` works as it should now in Elixir. But `` `m::lists` `` does not create a link. I'll leave things like this and you can update the elixir parts if you want to have `` `m::lists` ``.

I've removed the option to enforce `m:`.

josevalim:
@garazdawi one last question on the `m:mod` topic. Do you warn in Erlang if you do `m:lists` and the module does not exist? Because it should, right? It is explicit.

josevalim:
There are only two open questions, once they are answered, this is good to merge to me! 🚀

josevalim:
Cool, I created an issue then to make sure `m::lists` work in Elixir and also warn :)

garazdawi:
I decided to only allow arity for now in markdown. I've continued to allow () for types in `application/erlang+html` so that edoc continues to work without updates.

I've squashed and polished the commits so if you are happy, feel free to merge.

josevalim (approved):

josevalim:
Btw, please run `mix format` one last time. :)

garazdawi:
`mix format` has been run.

Any reason why you limit CI to run on `main` only? Would have cought this in my repo if CI ran on any branch.

wojtekmach:
Thank you!

Regarding CI, I believe it is to not run _two_ builds for PRs, use less compute.

test/ex_doc/formatter/html/erlang_test.exs
```diff
@@ -15,13 +15,13 @@ defmodule ExDoc.Formatter.HTML.ErlangTest do
 
     %% @doc
     %% f/0 function.
-    -spec foo(atom()) -> atom().
+    -spec foo(t:atom()) -> t:atom().
```
wojtekmach, line 18: @garazdawi this change prints warnings:

```
warning: documentation references type "t:atom/0" but it is undefined or private
```

and they were not caught manually by myself. 😅 Just making sure, we need to use `t:` in _markdown_ docs but we don't need to do so in -type and -spec expressions. Is it OK if I revert this change? Or you meant to test atom() type on a t module?

wojtekmach: ah, I see. the purpose of the test was to ensure we link to erlang.org/doc for these. I'll investigate other breakages. Thanks!

wojtekmach: oh, we did _not_ link `atom()` and friends to erlang.org but now we _do_ and that's why tests broke. I'll update accordingly!

wojtekmach:
> Any reason why you limit CI to run on main only? Would have cought this in my repo if CI ran on any branch.

@garazdawi thanks again for calling this out, I had a wrong understand how it all works. It is:

1. `on: push` trigger runs the exact commit on the branch.
2. `on: pull_request` runs the merge against branch.

(https://twitter.com/jarredsumner/status/1724299219358642550)

and it is useful to have both. I'll be changing our projects accordingly.

PR #1809, opened 2023-11-10
fixing test error

paulo-valim:


wojtekmach:
Can you explain this change? When does it break without this patch and why? It doesn't hurt to merge this, just curious.

josevalim:
It doesn't happen on CI, so I am wondering if it is related to concurrency (i.e. some tests have ansi enabled and others do not).

wojtekmach:
I'm gonna close this for now but if you run into it again please ping me @paulo-valim and include the seed from your test run.

PR #1812, opened 2023-11-10
Build precompiled ExDoc scripts on every release

### LINKED ISSUE #1810: Build precompiled ExDoc scripts on every release using GitHub Actions
This can be done by calling `mix escript.build`. We should probably support the last three Erlang/OTP versions.

Pinging @garazdawi, @gilacost, @wingyplus, as you were helpful in building Elixir and Erlang GitHub Actions. If any of you would like to help us with this, it would be very welcome!

gilacost:
~Just cheking if I am on the right path here for [this one](https://github.com/elixir-lang/ex_doc/issues/1810) @josevalim.~

1. **`release.yml`**:
     - Checks for the existence of a release corresponding to the pushed tag.
     - Creates a new release if it does not already exist.
     - Deploys pre-built binaries for different OTP versions.
     - Supports OTP versions 24, 25, and 26, with placeholders for future versions.

2. **`release_pre_built/action.yml`**:
     - Configurable for different major and minor OTP versions.
     - Utilizes `erlef/setup-beam@v1.16.0` for setting up the Elixir environment.
     - Generates SHA1 and SHA256 checksums for the built binaries.
   

closes #1810

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,56 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - v*
+
+permissions:
+  contents: write
+
+jobs:
+  create_release:
+    continue-on-error: true
+    runs-on: ubuntu-22.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Create release
+        run: |
+          echo "Creating release..."
+          gh release create \
+            --repo ${{ github.repository }} \
+            --title ${{ github.ref_name }} \
+            ${{ github.ref_name }}
+
+  release_pre_built:
+    needs: create_release
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - otp: 24
+            otp_version: "24.0"
+          - otp: 25
+            otp_version: "25.0.4"
+          - otp: 26
+            otp_version: "26.0.2"
+          # - otp: 27
+          #   otp_version: "27.0"
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 50
+      - uses: ./.github/workflows/release_pre_built
+        with:
+          otp_version: ${{ matrix.otp_version }}
+          otp: ${{ matrix.otp }}
+
+      - name: Upload Pre-built
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload --clobber "${{ github.ref_name }}" \
+            ex_doc_otp_${{ matrix.otp }} \
+            ex-doc-otp-${{ matrix.otp }}.sha{1,256}sum \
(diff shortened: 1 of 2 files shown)


.github/workflows/release_pre_built/action.yml
```diff
@@ -0,0 +1,25 @@
...
+        # version-type: strict
+    - name: Build ex_doc
+      shell: bash
+      run: |
+        mix deps.get
+        mix escript.build
+        zip -9 -r Precompiled.zip ex_doc
+        mv Precompiled.zip ex-doc-otp-${{ inputs.otp }}.zip
+        shasum -a 1 ex-doc-otp-${{ inputs.otp }}.zip > ex-doc-otp-${{ inputs.otp }}.zip.sha1sum
+        shasum -a 256 ex-doc-otp-${{ inputs.otp }}.zip > ex-doc-otp-${{ inputs.otp }}.zip.sha256sum
```
josevalim, line 24: I think they are tiny, we can probably skip the zipping.

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+permissions:
+  contents: write
+
+jobs:
+  create_draft_release:
+    runs-on: ubuntu-22.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Create draft release
```
josevalim, line 17: I think we can create the release if it doesn't exist yet, no need for draft. :)

josevalim:
Looks beautiful :heart:

gilacost:
How does [this](https://github.com/gilacost/ex_doc/releases/tag/build-precompiled-ex-docs-scripts-on-every-release) look?

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+          echo "RELEASE_EXISTS=$RELEASE_EXISTS" >> $GITHUB_ENV
+
+      - name: Create release
+        if: env.RELEASE_EXISTS == 'release not found'
+        run: |
+          echo "Creating release..."
+          gh release create \
+            --repo ${{ github.repository }} \
+            --title ${{ github.ref_name }} \
+            ${{ github.ref_name }}
```
josevalim, line 33: Quick question: could we always run release create? I assume if we try to create a release for an existing ref, it is going to fail and we can just ignore it? If that's the case, then we can always try to create the release within each job. WDYT?

gilacost: I think it makes sense, then you avoid any intervention in order to remove the release 🚀.
I have updated it so it continues-on-error when creating the release.

gilacost: Example of run that continues-on-error https://github.com/gilacost/ex_doc/actions/runs/6829969263

josevalim:
Absolutely beautiful, thank you! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

gilacost:
> Absolutely beautiful, thank you! 💚 💙 💜 💛 ❤️

At your service 🫡🫶🏽

PR #1813, opened 2023-11-10
Add noindex meta tag to 404 and Search pages

florish:
@josevalim As discussed in https://github.com/hexpm/hexdocs/issues/36#issuecomment-1806407623 

Couldn't find any related test coverage I could hook into, please let me know if tests are needed!

### DIFF
--- a/lib/ex_doc/formatter/html/templates/head_template.eex
+++ b/lib/ex_doc/formatter/html/templates/head_template.eex
@@ -9,7 +9,7 @@
     <%= if config.authors do %>
       <meta name="author" content="<%= Enum.join(config.authors, ", ") %>">
     <% end %>
-    <%= if page.title in ["404", "Search"] do %>
+    <%= if page.noindex do %>
       <meta name="robots" content="noindex">
     <% end %>
     <title><%= page.title %> — <%= config.project %> v<%= config.version %></title>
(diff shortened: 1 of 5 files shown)


lib/ex_doc/formatter/html/templates/head_template.eex
```diff
@@ -9,6 +9,9 @@
     <%= if config.authors do %>
       <meta name="author" content="<%= Enum.join(config.authors, ", ") %>">
     <% end %>
+    <%= if page.title in ["404", "Search"] do %>
```
josevalim, line 12: Instead of checking the title, when you call head_template(...) from these templates, pass `noindex: true`, then you can do:

```suggestion
    <%= if page.noindex do %>
```

:)

florish: Changed! Should I add a default `noindex: false` somewhere to prevent `(KeyError) key :noindex not found` exceptions on other pages?

florish: I now have `mix test` up and running locally and the answer is a clear "yes". I'll add `noindex: false` to the other places where `head_template/2` is called.

florish: Should be fixed now.

josevalim:

PR #1816, opened 2023-11-14
Add borders dividing table rows in rendered content

sevenseacat:
It looks like they were attempted to be added, but weren't rendering due to missing `border-collapse: collapse` on the table.

So now there are borders for both light and dark mode:

**After**:

Screenshot 2023-11-14 at 11 16 16 am

Screenshot 2023-11-14 at 11 18 23 am


josevalim:

PR #1807, opened 2023-11-10
More accurate autolink warnings line

wojtekmach:
Demo:

image

PR #1818, opened 2023-11-15
Fix Erlang Scanning & Warnings 

garazdawi:
This PR does 3 things:

1. Change Erlang+Elixir autolinking to parse function links using specialized code instead of sharing. There are too many incompatible overlaps between the languages to make it possible to use the same parser for both.
2. Use correct line numbers for regular links, just as #1807 did for Elixir
3. Make Erlang docs warn when an unkown type or callback is used in regular links.

I'm a bit unsure about the last one, why is it that ExDoc Elixir does not warn for these?

```
`t:Bad:type/0`
```

Seems like there is very little chance of a false positive, and if there is one you can always silence it.

### DIFF
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -413,7 +413,7 @@ defmodule ExDoc.Autolink do
       {:type, _visibility} ->
         case config.language.try_builtin_type(name, arity, mode, config, original_text) do
           nil ->
-            if mode == :custom_link do
+            if mode == :custom_link or config.language == ExDoc.Language.Erlang do
               maybe_warn(config, ref, visibility, %{original_text: original_text})
             end
 
@@ -482,7 +482,9 @@ defmodule ExDoc.Autolink do
 
         nil
 
-      {:regular_link, _module_visibility, :undefined} when not same_module? ->
+      {:regular_link, _module_visibility, :undefined}
+      when not same_module? and
+             (config.language != ExDoc.Language.Erlang or kind == :function) ->
         nil
 
       {_mode, _module_visibility, visibility} ->
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -371,29 +371,6 @@ defmodule ExDoc.Autolink do
     end
   end
 
-  # There are special forms that are forbidden by the tokenizer
-  def parse_function("__aliases__"), do: {:function, :__aliases__}
-  def parse_function("__block__"), do: {:function, :__block__}
-  def parse_function("%"), do: {:function, :%}
-
-  def parse_function(string) do
-    case Code.string_to_quoted("& #{string}/0", warnings: false) do
-      {:ok, {:&, _, [{:/, _, [{:__aliases__, _, [function]}, 0]}]}} when is_atom(function) ->
-        ## When function starts with capital letter
-        {:function, function}
-
-      ## When function is 'nil'
-      {:ok, {:&, _, [{:/, _, [nil, 0]}]}} ->
-        {:function, nil}
-
-      {:ok, {:&, _, [{:/, _, [{function, _, _}, 0]}]}} when is_atom(function) ->
-        {:function, function}
-
-      _ ->
-        :error
-    end
-  end
-
   def kind("c:" <> rest), do: {:callback, rest}
   def kind("t:" <> rest), do: {:type, rest}
   ## \\ does not work for :custom_url as Earmark strips the \...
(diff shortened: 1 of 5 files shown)


josevalim:
Elixir should warn for types I think. @wojtekmach, maybe it is worth implementing that when you implement `m:`. All of `m:`, `t:`, and `c:` should warn if not found.

lib/ex_doc/autolink.ex
```diff
@@ -432,7 +413,7 @@ defmodule ExDoc.Autolink do
       {:type, _visibility} ->
         case config.language.try_builtin_type(name, arity, mode, config, original_text) do
           nil ->
-            if mode == :custom_link do
+            if mode == :custom_link or config.language == ExDoc.Language.Erlang do
```
josevalim, line 416: Is this check here to avoid warning in Elixir? If so, you can go ahead and warn. :)

lib/ex_doc/autolink.ex
```diff
@@ -501,7 +482,9 @@ defmodule ExDoc.Autolink do
 
         nil
 
-      {:regular_link, _module_visibility, :undefined} when not same_module? ->
+      {:regular_link, _module_visibility, :undefined}
+      when not same_module? and
+             (config.language != ExDoc.Language.Erlang or kind == :function) ->
```
josevalim, line 487: Same here, feel free to let it warn on Elixir, assuming it is one of `m:`, `t:` or `c:`.

lib/ex_doc/autolink.ex
garazdawi: Yes it is, but if I remove it a lot (~20) of tests break (or not break, but they start to emit warnings to stdout).

josevalim: Ok, @wojtekmach please look into it afterwards then. :)

Ideally we want to avoid hardcoding the language here, an option would be better, but since this is temporary, it is fine by me. :)

josevalim (approved):

garazdawi:
Turns out that it was an issue in `:elixir_errors` [that has already been fixed](https://github.com/elixir-lang/elixir/commit/665a459a5938608fa970c6e5c45eb8e81fdc83c8).

Feel free to merge whenever you feel like.

wojtekmach:
Thank you!

> Same here, feel free to let it warn on Elixir, assuming it is one of m:, t: or c:.

Good call, will do!

wojtekmach:
@garazdawi I think after this PR, your OTP branch with ExDoc+markdown is now crashing and I was able to narrow it down to ExDoc emitted warnings for docs defined in .hrl files. ExDoc gets the file as .erl and tries to render a line in it that doesn't exist, it should look into .hrl file. Unfortunately I couldn't reproduce this in a test yet. Our markdown doc tests are going through the legacy edoc machinery in test_helper and I think ideally we'd have another code path where we extract things from `-doc`. I'd appreciate any advice how to go about that!

garazdawi:
Yes, I'm aware. I've fixed it in a local branch together with some other fixes. The problem happens when we have doc entries with different file anno than the moduledoc and then when ExDoc tries to warn it takes the snippet from the wrong file. 

The branch with my current fixes is: https://github.com/garazdawi/ex_doc/commits/lukas/fix-erlang-doc-support/

and the fix for this particular issue is: https://github.com/garazdawi/ex_doc/commit/0cfc40b5221d58131f31551d48c7feafbb3bfe9b

josevalim:
FWIW EEP 48 expects the Anno node to point to the source and it has an optional metadata that points to the edit_url. I am not sure we use them both in ExDoc but we will be glad to accommodate!

garazdawi:
> FWIW EEP 48 expects the Anno node to point to the source and it has an optional metadata that points to the edit_url. I am not sure we use them both in ExDoc but we will be glad to accommodate!

From what I've gathered, for ExDoc to work, Anno should point to the first line of the documentation. This is used when printing warnings by ExDoc etc. For `application/html+erlang`, the Anno points to the source code (or should anyway, seems like there are bugs there).

Then ExDoc outputs a "view source" button that points to the source code, this is currently taken from the anno in the AST. (This only works for types in Erlang right now, yet another thing to fix :) )

Maybe it would make sense to have an "edit documentation" button that takes you to a page where you can update the docs?

wojtekmach:
@garazdawi I created an issue to discuss Edit Documentation link: https://github.com/elixir-lang/ex_doc/issues/1825.

wojtekmach:
@garazdawi I cherry-picked all commits except using makeup_erlang fork (as we can't release to Hex with that) from https://github.com/garazdawi/ex_doc/commits/lukas/fix-erlang-doc-support/ on main.

If you could send a PR to makeup_erlang we can release it too. If there's any other ExDoc things you'd like to land before the release, let us know too!

garazdawi:
Great! I've meant to open a PR for weeks now, but other things have always gotten in the way. I will send a pr to makeup_erlang.

I don't have anything right now lined up for ExDoc, but I don't doubt that as we start converting more and more of our docs and other people start to look at it there will be issues and inconsistencies that come up.

PR #1826, opened 2023-11-24
Do not parse math expressions as markdown

### LINKED ISSUE #1571: Preprocess .livemd documents
Unfortunately Markdown documents try to escape backslashes. Which means that a Katex formula like this:

```
$$
\begin{bmatrix}
  1 & 2 \\
  3 & 4
\end{bmatrix}
$$
```

Will appear in the HTML as:

```
$$
\begin{bmatrix}
  1 & 2 \
  3 & 4
\end{bmatrix}
$$
```

Which will render the formula without the newline. However, Livebook handles this case correctly. This means that rendering Livebooks as Markdown documents differ within Livebook and using ExDoc.

@jonatanklosko, how hard do you think it would be to add some pre-processing here? Although I think the big challenge will be really to preprocess `$`? 🤔 

The only other option I can think about is to change the notation from `$$` to something that includes `` ` `` in it, such as `` $`..`$ `` or `` `$...$` ``.

jonatanklosko:
Closes #1571. Depends on https://github.com/RobertDober/earmark_parser/pull/142.

With this change we parse math expressions in Markdown and then rewrite them back to `$` and `$$`. Rendering the expressions is still up to the user, but this way we don't parse the expression contents as Markdown.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -34,7 +34,8 @@ defmodule ExDoc.Mixfile do
 
   defp deps do
     [
-      {:earmark_parser, "~> 1.4.38"},
+      # {:earmark_parser, "~> 1.4.38"},
+      {:earmark_parser, github: "jonatanklosko/earmark_parser", branch: "jk-math"},
       {:makeup_elixir, "~> 0.14"},
       {:makeup_erlang, "~> 0.1"},
       # Add other makeup lexers as optional for the executable
--- a/mix.exs
+++ b/mix.exs
@@ -34,8 +34,7 @@ defmodule ExDoc.Mixfile do
 
   defp deps do
     [
-      # {:earmark_parser, "~> 1.4.38"},
-      {:earmark_parser, github: "jonatanklosko/earmark_parser", branch: "jk-math"},
+      {:earmark_parser, "~> 1.4.39"},
       {:makeup_elixir, "~> 0.14"},
       {:makeup_erlang, "~> 0.1"},
       # Add other makeup lexers as optional for the executable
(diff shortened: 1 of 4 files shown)


mix.exs
```diff
@@ -34,7 +34,8 @@ defmodule ExDoc.Mixfile do
 
   defp deps do
     [
-      {:earmark_parser, "~> 1.4.38"},
+      # {:earmark_parser, "~> 1.4.38"},
+      {:earmark_parser, github: "jonatanklosko/earmark_parser", branch: "jk-math"},
```
josevalim, line 38:
```suggestion
      {:earmark_parser, "~> 1.4.39"},
```

mix.lock
```diff
@@ -1,5 +1,5 @@
 %{
-  "earmark_parser": {:hex, :earmark_parser, "1.4.38", "b42252eddf63bda05554ba8be93a1262dc0920c721f1aaf989f5de0f73a2e367", [:mix], [], "hexpm", "2cd0907795aaef0c7e8442e376633c5b3bd6edc8dbbdc539b22f095501c1cdb6"},
+  "earmark_parser": {:git, "https://github.com/jonatanklosko/earmark_parser.git", "b73e6591983320f4eced3292b6ed12f4fcbdbe2b", [branch: "jk-math"]},
```
josevalim, line 2:
```suggestion
  "earmark_parser": {:hex, :earmark_parser, "1.4.39", "424642f8335b05bb9eb611aa1564c148a8ee35c9c8a8bba6e129d51a3e3c6769", [:mix], [], "hexpm", "06553a88d1f1846da9ef066b87b57c6f605552cfbe40d20bd8d59cc6bde41944"},
```

josevalim:

PR #1827, opened 2023-11-24
Improve search bar navigation

jonatanklosko:
Allows navigation with <kbd>ctrl</kbd>+<kbd>n/p</kbd> on macOS and automatically scrolls the list so that the focused item is always visible.


josevalim:

PR #1833, opened 2023-12-21
Fix module anchors

garazdawi:
Since I had not written proper tests for module anchors for Erlang, @wojtekmach's [fix for module anchors in Elixir](https://github.com/elixir-lang/ex_doc/commit/d2cbfb45299a436c818c76c50ab4a523a52be669) broke the support in Erlang.

So here is a PR with proper tests and working Erlang anchors.


wojtekmach:
Sorry about that, thank you for the fix!

PR #1834, opened 2023-12-21
More work to make Erlang/OTP use ExDoc

garazdawi:
Hopefully the final PR to make ExDoc work with generating docs for Erlang/OTP.

This PR does the following: 

* Fixes so that source and doc file path and line numbers are correct for complex Erlang files such as erl_parse.yrl.
* `source_url_relative_path` can now be configured so that the cwd when running ex_doc does not need to be the root of the source tree.
* `source` can be configured for extra files so point to another file than the .md file. This is useful for when we generate markdown files and want to point to the original source.
* Issue (suppressable) warnings when a `-spec` refers to a type that is not in the documentation.
* Expand the fields of records when included in a spec/type/callback.

There are also some small other fixes, see the commits for more details.

The fix to make source and doc file paths work properly turned out to have to touch a lot of different places in the code, so please especially have a look at that and see if I have broken any invariants that Elixir code relies on.

There is no rush to review of merge this. I will be away over the holidays and back next year.


lib/ex_doc/language.ex
```diff
@@ -102,6 +115,8 @@ defmodule ExDoc.Language do
 
     * `:line` - the line where the code is located
 
+    * `:source` - the source file where the code is located
```
josevalim, line 118: Should we make these consistent? In some cases we have line+source, in others we have source_line+source_file? Or are they different on purpose?

lib/mix/tasks/docs.ex
```diff
@@ -103,8 +103,8 @@ defmodule Mix.Tasks.Docs do
     * `:extras` - List of paths to additional Markdown (`.md` extension), Live Markdown
       (`.livemd` extension), Cheatsheets (`.cheatmd` extension) and plain text pages to
       add to the documentation. You can also specify keyword pairs to customize the
-      generated filename and title of each extra page; default: `[]`. Example:
-      `["README.md", "LICENSE", "CONTRIBUTING.md": [filename: "contributing", title: "Contributing"]]`
+      generated filename, title and source file of each extra page; default: `[]`. Example:
```
josevalim, line 106: Isn't the source file in this case the `key` here? So the example above would be written as:

```suggestion
  "CONTRIBUTING.md": [filename: "contributing", title: "Contributing", source: "CONTRIBUTING.mdx"]
```

Or are you telling me that you have .mdx files, which are converted to markdown, which are then converted to HTML?

If the second, maybe we should include custom preprocessors in ExDoc, so ExDoc is the one converting .mdx to .md, and you don't need a separate pipeline for it? I assume other projects could benefit from .mdx too?

lib/mix/tasks/docs.ex
```diff
@@ -184,6 +184,10 @@ defmodule Mix.Tasks.Docs do
       https://mydomain.org/user_or_team/repo_name/src/main/%{path}#cl-%{line}
       ```
 
+    * `:source_url_relative_path` - The absolute path pointing to where ExDoc should
+      consider the base of the repo. That is where the `%{path}` in the `:source_url_pattern`
+      starts. The default is the current working directory.
```
josevalim, line 189: Instead of adding a new option, what if we make `source_url_pattern` accept a function/MFA, that receives the path and the line, and then you can do whatever you want with it? Would that work for your use cases?

lib/ex_doc/language/source.ex
```diff
@@ -0,0 +1,261 @@
...
+
+  The basedir is the cwd of the Elixir/Erlang compiler when compiling the module.
+  All `-file` attributes in the module is relative to this directory.
+  """
+  def get_basedir(abst_code, module) do
+    ## We look for the first -file attribute to see what the source file that
+    ## was compiled is called.
+    filename =
+      Enum.find_value(abst_code, fn
+        {:attribute, _anno, :file, {filename, _line}} ->
```
josevalim, line 105: Don't we have to guarantee this comes _before_ the module attribute? Otherwise we may find the filename of a function. Or do both abstract code guarantee there will be a `-file` at the top? I don't remember. :)

lib/ex_doc/language/source.ex
```diff
@@ -0,0 +1,261 @@
...
+      ## We use `module_info(:compile)[:source]` to get an absolute path
+      ## to the source file and calculate the basedir from that
+
+      module.module_info(:compile)[:source]
+      |> String.Chars.to_string()
+      |> Path.absname()
+      |> Path.split()
+      |> Enum.reverse()
+      |> Enum.split(Path.split(filename) |> Enum.count())
+      |> elem(1)
```
josevalim, line 127: `drop` or `take` will give you the sides of `split`. I think here you want drop (please double check)

```suggestion
      |> Enum.drop(Path.split(filename) |> Enum.count())
```

lib/ex_doc/language/source.ex
josevalim: Actually, if you can add a small note here, it would be very helpful. Does Elixir and Erlang include it at the top or not? Do they include in relative or absolute format or both? This way if we ever want to untangle both languages for whatever reason, we can at least have some idea of which branch is needed for each. :)

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -74,10 +74,14 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
-      [all] -> all
-    end
+    doc =
+      case :binary.split(doc, "</p>") do
+        [left, _] -> String.trim_trailing(left, ":") <> "</p>"
+        [all] -> all
+      end
+
+    ## Remove any anchors found in synopsis.
+    Regex.replace(~r|(<[^>]*) id="[^"]*"([^>]*>)|, doc, ~S"\1\2", [])
```
josevalim, line 84: Anchors in synopsis should be fine, no? If they are shown buggy, then we fix that instead?

lib/ex_doc/language/source.ex
garazdawi: If there is no `-file`, then the code falls back to using the `[:compile][:source]` attribute.

lib/ex_doc/formatter/html/templates.ex
garazdawi: Since the synopsis is present in two places, the anchor will be duplicated. I think that when setting an anchor in the docs the user want it to be in the main body of the documention and not at the listing of all functions.

lib/ex_doc/language.ex
garazdawi: yes. I'll fix that.

lib/mix/tasks/docs.ex
garazdawi: hmm, yes that might work. I will give it a go and see if any issues pop up.

lib/ex_doc/formatter/html/templates.ex
josevalim: Oh, I get it now. Follow up question: what are you doing that you end-up with anchors in synopsis? :D

garazdawi: We have (lots of) old function docs where an anchor has been put at the beginning of the docs in order to link to it using a name. This was from before the time when we added the possibility to link directly to `M:F/A` in the xml docs. So when converting those old docs to markdown, we endup with anchors in the synopsis.

Ideally I would like to remove them from our docs completely, but there are enough of them that it is a lot of work to fix so I did this change instead :)

lib/ex_doc/language/source.ex
josevalim: I am not worried about when there is no `-file`, I am worried about file used to denote the location of a specific function, rather than the whole module.

You could do this:

```
-file {"foo.eex", 1}.
function_generated_by_template(...) -> ....
```

In this case, the `-file` applies only to that function, and this style is much more common in Elixir. I _think_ Erlang requires the `-file` for the whole module to be before the module attribute. If that's the intent here, we should stop looking for the file as soon as we find we see the `-module` attribute (but it has been ages since I looked into this).

lib/ex_doc/language.ex
josevalim: The last work is to keep it as `line`+`source` everywhere, but it is your call!

lib/ex_doc/retriever.ex
```diff
@@ -235,18 +239,25 @@ defmodule ExDoc.Retriever do
        ) do
     {:docs_v1, _, _, content_type, _, module_metadata, _} = module_data.docs
     {{type, name, arity}, anno, signature, source_doc, metadata} = doc_element
+    doc_file = anno_file(anno, source)
     doc_line = anno_line(anno)
-    source = anno_file(anno, source)
+
+    ## When do we not have a source_line?
```
josevalim, line 245: We should always have one, no? If we are not sure, remove it and we will know if it is ever an issue. :D

lib/ex_doc/language/source.ex
garazdawi: Do you have a full example in Elixir that I can look at?

lib/ex_doc/retriever.ex
garazdawi: The original seemed to think that it was possible for us to not have a line, so I assumed that there was a reason for that. I'm happy to remove it and see what breaks :)

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -74,10 +74,14 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
-      [left, _] -> String.trim_trailing(left, ":") <> "</p>"
-      [all] -> all
-    end
+    doc =
+      case :binary.split(doc, "</p>") do
+        [left, _] -> String.trim_trailing(left, ":") <> "</p>"
+        [all] -> all
+      end
+
+    ## Remove any anchors found in synopsis.
```
josevalim, line 83:
```suggestion
    # Remove any anchors found in synopsis.
    # Old Erlang docs placed anchors at the top of the documentation
    # for links. Ideally they would have been removed but meanwhile
    # it is simpler to guarantee they won't be duplicated in docs.
```

lib/ex_doc/formatter/html/templates.ex
josevalim: Cool, I pushed a code suggestion.

lib/ex_doc/retriever.ex
josevalim: I guess the bigger question is: which one should win? We have both source_line and source_file and anno_file and anno_line. Shouldn't they be the same? When they are not the same? And we use doc_file and doc_line further down below as well. Shouldn't we use the same throughout?

The `Anno` in EEP 48 is meant to be the Anno of the source code, so shouldn't it always match the source_file+source_line?

lib/ex_doc/language/source.ex
josevalim: Go to `ExDoc.Markdown` and add this before the `to_ast` function:

```
  @doc """
  Converts the given markdown document to HTML AST.
  """
  @file "gotcha.ex"
  def to_ast(text, opts \\ []) when is_binary(text) do
```

The docs generated will be the same.

Now add `raise "oops"` to the first line of the `to_ast` function and run `mix build`, you will get this:

```
Generated ex_doc app
Generating docs...
** (RuntimeError) oops
    (ex_doc 0.31.1-dev) gotcha.ex:37: ExDoc.Markdown.to_ast/2
```

See how the file information was used in the stacktrace?

josevalim: And, fwiw, I don't think Elixir is adding the file information to the AST, rather somewhere in the Erlang compiler the `-file` attribute is used for that.

josevalim: The abstract code for this example is:

```
{:ok,
 {ExDoc.Markdown,
  [
    abstract_code: {:raw_abstract_v1,
     [
       {:attribute, 1, :file, {~c"lib/ex_doc/markdown.ex", 1}},
       {:attribute, 1, :module, ExDoc.Markdown},
       {:attribute, 1, :compile, [:no_auto_import]},
       {:attribute, 19, :callback,
        {{:to_ast, 2},
         [
           {:type, {19, 45}, :fun,
            [
              {:type, {19, 45}, :product,
               [
                 {:remote_type, {19, 26},
                  [{:atom, 0, String}, {:atom, 0, :t}, []]},
                 {:remote_type, {19, 39},
                  [{:atom, 0, Keyword}, {:atom, 0, :t}, []]}
               ]},
              {:type, {19, 48}, :term, []}
            ]}
         ]}},
       {:attribute, 24, :callback,
        {{:available?, 0},
         [
           {:type, {24, 26}, :fun,
            [{:type, {24, 26}, :product, []}, {:type, {24, 29}, :boolean, []}]}
         ]}},
       {:attribute, 1, :export,
        [
          __info__: 1,
          get_markdown_processor: 0,
          put_markdown_processor: 1,
          to_ast: 1,
          to_ast: 2
        ]},
       {:attribute, 1, :spec,
        {{:__info__, 1},
         [
           {:type, 1, :fun,
            [
              {:type, 1, :product,
               [
                 {:type, 1, :union,
                  [
                    {:atom, 1, :attributes},
                    {:atom, 1, :compile},
                    {:atom, 1, :functions},
                    {:atom, 1, :macros},
                    {:atom, 1, :md5},
                    {:atom, 1, :exports_md5},
                    {:atom, 1, :module},
                    {:atom, 1, :deprecated},
                    {:atom, 1, :struct}
                  ]}
               ]},
              {:type, 1, :any, []}
            ]}
         ]}},
       {:function, 0, :__info__, 1,
        [
          {:clause, 0, [{:atom, 0, :module}], [], [{:atom, 0, ExDoc.Markdown}]},
          {:clause, 0, [{:atom, 0, :functions}], [],
           [
             {:cons, 0,
              {:tuple, 0,
               [{:atom, 0, :get_markdown_processor}, {:integer, 0, 0}]},
              {:cons, 0,
               {:tuple, 0,
                [{:atom, 0, :put_markdown_processor}, {:integer, 0, 1}]},
               {:cons, 0, {:tuple, 0, [{:atom, 0, :to_ast}, {:integer, 0, 1}]},
                {:cons, 0,
                 {:tuple, 0, [{:atom, 0, :to_ast}, {:integer, 0, ...}]},
                 {nil, 0}}}}}
           ]},
          {:clause, 0, [{:atom, 0, :macros}], [], [nil: 0]},
          {:clause, 0, [{:atom, 0, :struct}], [], [{:atom, 0, nil}]},
          {:clause, 0, [{:atom, 0, :exports_md5}], [],
           [
             {:bin, 0,
              [
                {:bin_element, 0,
                 {:string, 0, [98, 13, 150, 29, 223, 104, 220, 12, 166, ...]},
                 :default, :default}
              ]}
           ]},
          {:clause, 0, [{:match, 0, {:var, 0, :Key}, {:atom, 0, :attributes}}],
           [],
           [
             {:call, 0,
              {:remote, 0, {:atom, 0, :erlang}, {:atom, 0, :get_module_info}},
              [{:atom, 0, ExDoc.Markdown}, {:var, 0, :Key}]}
           ]},
          {:clause, 0, [{:match, 0, {:var, 0, :Key}, {:atom, 0, :compile}}], [],
           [
             {:call, 0,
              {:remote, 0, {:atom, 0, :erlang}, {:atom, 0, :get_module_info}},
              [{:atom, 0, ExDoc.Markdown}, {:var, 0, :Key}]}
           ]},
          {:clause, 0, [{:match, 0, {:var, 0, :Key}, {:atom, 0, :md5}}], [],
           [
             {:call, 0,
              {:remote, 0, {:atom, 0, :erlang}, {:atom, 0, :get_module_info}},
              [{:atom, 0, ExDoc.Markdown}, {:var, 0, :Key}]}
           ]},
          {:clause, 0, [{:atom, 0, :deprecated}], [], [nil: 0]}
        ]},
       {:function, 68, :find_markdown_processor, 0,
        [
          {:clause, 68, [], [],
           [
             {:call, 69, {:remote, 69, {:atom, 69, Enum}, {:atom, 69, :find}},
              [
                {:cons, 69, {:atom, 69, ExDoc.Markdown.Earmark}, {nil, 69}},
                {:fun, 69,
                 {:clauses,
                  [
                    {:clause, 69, [{:var, 69, :_module@1}], [],
                     [{:case, 70, {:call, ...}, [...]}]}
                  ]}}
              ]}
           ]}
        ]},
       {:function, 45, :get_markdown_processor, 0,
        [
          {:clause, 45, [], [],
           [
             {:case, 46,
              {:call, 46,
               {:remote, 46, {:atom, 46, Application}, {:atom, 46, :fetch_env}},
               [{:atom, 46, :ex_doc}, {:atom, 46, :markdown_processor}]},
              [
                {:clause, 47,
                 [
                   {:tuple, 47,
                    [
                      {:atom, 47, :ok},
                      {:tuple, 47, [{:var, 47, :_processor@1}, {:var, 47, ...}]}
                    ]}
                 ], [],
                 [
                   {:tuple, 47,
                    [{:var, 48, :_processor@1}, {:var, 48, :_options@1}]}
                 ]},
                {:clause, 50, [{:atom, 50, :error}], [],
                 [
                   {:match, 51, {:var, 51, :_processor@2},
                    {:case, 51, {:call, 51, ...}, [{...}, ...]}},
                   {:call, 52, {:atom, 52, :put_markdown_processor},
                    [{:tuple, 52, [...]}]},
                   {:tuple, 0, [{:var, 53, :_processor@2}, {nil, 0}]}
                 ]}
              ]}
           ]}
        ]},
       {:function, 60, :put_markdown_processor, 1,
        [
          {:clause, 60, [{:var, 60, :_processor@1}],
           [
             [
               {:call, 60,
                {:remote, 60, {:atom, 60, :erlang}, {:atom, 60, :is_atom}},
                [{:var, 60, :_processor@1}]}
             ]
           ],
           [
             {:call, 61, {:atom, 61, :put_markdown_processor},
              [{:tuple, 61, [{:var, 61, :_processor@1}, {nil, 61}]}]}
           ]},
          {:clause, 64,
           [{:tuple, 64, [{:var, 64, :_processor@1}, {:var, 64, :_options@1}]}],
           [],
           [
             {:call, 65,
              {:remote, 65, {:atom, 65, Application}, {:atom, 65, :put_env}},
              [
                {:atom, 65, :ex_doc},
                {:atom, 65, :markdown_processor},
                {:tuple, 65,
                 [{:var, 65, :_processor@1}, {:var, 65, :_options@1}]}
              ]}
           ]}
        ]},
       {:function, 74, :raise_no_markdown_processor, 0,
        [
          {:clause, 74, [], [],
           [
             {:call, 75,
              {:remote, 75, {:atom, 75, :erlang}, {:atom, 75, :error}},
              [
                {:call, 75,
                 {:remote, 75, {:atom, 75, RuntimeError},
                  {:atom, 75, :exception}},
                 [
                   {:bin, 75,
                    [{:bin_element, 75, {:string, 75, ...}, :default, :default}]}
                 ]},
                {:atom, 75, :none},
                {:cons, 75,
                 {:tuple, 75, [{:atom, 75, :error_info}, {:map, 75, [{...}]}]},
                 {nil, 75}}
              ]}
           ]}
        ]},
       {:attribute, 36, :file, {~c"gotcha.ex", 123}},
       {:function, 36, :to_ast, 1,
        [
          {:clause, 36, [{:var, [generated: true, location: 0], :_@1}], [],
           [
             {:call, 36, {:atom, 36, :to_ast},
              [{:var, [generated: true, location: 0], :_@1}, {nil, 36}]}
           ]}
        ]},
       {:attribute, 36, :file, {~c"gotcha.ex", 123}},
       {:function, 36, :to_ast, 2,
        [
          {:clause, 36, [{:var, 36, :_text@1}, {:var, 36, :_opts@1}],
           [
             [
               {:call, 36,
                {:remote, 36, {:atom, 36, :erlang}, {:atom, 36, :is_binary}},
                [{:var, 36, :_text@1}]}
             ]
           ],
           [
             {:call, 37,
              {:remote, 37, {:atom, 37, :erlang}, {:atom, 37, :error}},
              [
                {:call, 37,
                 {:remote, 37, {:atom, 37, RuntimeError},
                  {:atom, 37, :exception}},
                 [{:bin, 37, [{:bin_element, ...}]}]},
                {:atom, 37, :none},
                {:cons, 37, {:tuple, 37, [{:atom, ...}, {...}]}, {nil, 37}}
              ]},
             {:match, 38,
              {:tuple, 38, [{:var, 38, :_processor@1}, {:var, 38, :_options@1}]},
              {:call, 38, {:atom, 38, :get_markdown_processor}, []}},
             {:call, 39,
              {:remote, 39, {:var, 39, :_processor@1}, {:atom, 39, :to_ast}},
              [
                {:var, 39, :_text@1},
                {:call, 39, {:remote, 39, {:atom, ...}, {...}},
                 [{:var, 39, ...}, {:var, ...}]}
              ]}
           ]}
        ]}
     ]}
  ]}}
```

So answer my earlier question, Elixir always guarantees the file at the top. :)

garazdawi: So from what I can tell, Elixir places all functions and macros with `@file` attributes at the end of the AST, so that works as it should. I suppose you don't have anything similar to `-include` in elixir?

I will add a comment explaining the way parsing the ast should work.

josevalim: There is no `-include` yeah. I guess it makes sense for us to place `-file` at the end, because it is not tied to a function but rather it is processed in order?

lib/ex_doc/retriever.ex
garazdawi: > The Anno in EEP 48 is meant to be the Anno of the source code, so shouldn't it always match the source_file+source_line?

EEP-48 Anno is the location of the documentation, not the source code. It is what we use to generate warnings for documentation issues, such as broken links etc.

We get the source code location from the AST.

Most of the time EEP-48 Anno does not contain any file, and then we assume that the docs are in the same file as the source code.

lib/ex_doc/language/source.ex
josevalim: Anyway, we always place one at the top, so we are good to go from the Elixir side either way, not sure about Erlang. :)

garazdawi: >  because it is not tied to a function but rather it is processed in order?

yes indeed.

> Anyway, we always place one at the top, so we are good to go from the Elixir side either way, not sure about Erlang. :)

We also do that.

lib/ex_doc/retriever.ex
josevalim: Ok, so this all makes sense then. And I actually know when we won't have a source line. It is for Elixir's `Kernel.SpecialForm`, where we have stub definitions of functions/macros. However, this only applies to functions/macros. We will always have a line for modules, types, callbacks, and so on.

Given this is a Elixir specific issue, what if we solve this on the Elixir side things, and fallback to `anno_line` there, and in here we assume `source_line` always exists?

lib/ex_doc/language/source.ex
```diff
@@ -0,0 +1,261 @@
...
+
+  @doc """
+  Get the basedir of a module
+
+  The basedir is the cwd of the Elixir/Erlang compiler when compiling the module.
+  All `-file` attributes in the module is relative to this directory.
+  """
+  def get_basedir(abst_code, module) do
+    ## We look for the first -file attribute to see what the source file that
+    ## was compiled is called.
```
josevalim, line 102:
```suggestion
    ## was compiled is called. Both Erlang and Elixir place one at the top.
```

lib/ex_doc/retriever.ex
garazdawi: yes, that make sense.

lib/ex_doc/language.ex
garazdawi: just realised that it is already called source_line and source_file everywhere, it was just the callback specs that were not updated. Will fix the callback definitions.

garazdawi:
Pushed fixes for everything except `source_url_pattern`. Will attempt a fix to that now.

lib/mix/tasks/docs.ex
garazdawi: Turns out that in #1529 someone else had the same idea as you and implemented it :) It just wasn't documented so I did not find that functionality. So I added documentation and removed this code.

josevalim: Beautiful!

garazdawi:
All issues should be fixed now. Feel free to merge if you think it is ready. Otherwise I will be back next year to look at it again.

josevalim (approved):

lib/ex_doc/retriever.ex
```diff
@@ -436,25 +449,40 @@ defmodule ExDoc.Retriever do
...
+      source.path
+    end
+    |> then(fn path ->
+      if function_exported?(Path, :relative_to, 3) do
+        Path.relative_to(path, File.cwd!(), force: true)
+      else
+        Path.relative_to(path, File.cwd!())
+      end
+    end)
+  end
```
wojtekmach, line 475:
```suggestion
    |> path_relative_to_cwd(force: true)
  end

  # TODO: Remove when we require Elixir 1.16
  if function_exported?(Path, :relative_to_cwd, 2) do
    defp path_relative_to_cwd(path, options), do: Path.relative_to_cwd(path, options)
  else
    defp path_relative_to_cwd(path, _options), do: Path.relative_to_cwd(path)
  end
```

test/ex_doc/language/erlang_test.exs
```diff
@@ -9,45 +9,45 @@ defmodule ExDoc.Language.ErlangTest do
 
   describe "autolink_doc/2 for edoc" do
     test "module", c do
-      assert autolink_edoc("{@link erlang_bar}", c) ==
+      assert autolink_edoc("{@link erlang_bar}", c) =~
```
wojtekmach: @garazdawi I don't think we need this change and can keep the stricter assertion here and in similar spots, i.e. ensure the exact html output. I'd use `=~` only on error messages. I went ahead and push this change to your branch but lmk if I was missing something!

wojtekmach (approved):

test/ex_doc/language/erlang_test.exs
garazdawi: Without that change, the tests break for OTP-27 as all edoc renderings return `<p>....</p>` instead of `...`.

wojtekmach: Got it!

PR #1848, opened 2024-01-21
Fix SVG logo without width/height attrs not displaying

### LINKED ISSUE #1847: SVG logo is not shown
`img` tag is 0x0 and logo is invisible. Works fine with PNG.

It used to work in v0.30.9. Doesn't work starting from v0.31.0

DavidOliver:
Sets width and (auto) height of logo image.

Also allows for 'jpeg' filename extension to be used and updates docs.

Fix #1847

### DIFF
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -53,8 +53,8 @@
 }
 
 .sidebar .sidebar-projectImage img {
-  width: 48px;
-  height: auto;
+  display: block;
+  max-width: 48px;
   max-height: 48px;
 }
 
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -53,7 +53,8 @@
 }
 
 .sidebar .sidebar-projectImage img {
-  max-width: 48px;
+  width: 48px;
+  height: auto;
   max-height: 48px;
 }
 
--- a/lib/ex_doc/cli.ex
+++ b/lib/ex_doc/cli.ex
@@ -174,9 +174,8 @@ defmodule ExDoc.CLI do
           --homepage-url  URL to link to for the site name
           --language      Identify the primary language of the documents, its value must be
                           a valid [BCP 47](https://tools.ietf.org/html/bcp47) language tag, default: "en"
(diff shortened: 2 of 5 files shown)


josevalim:
I am afraid forcing a width will be a breaking change. :( since I now know we don’t guarantee SVG support, I think we should continue saying SVG is not supported and remove the CSS changes. Thoughts?

DavidOliver:
As I don't imagine authors have been relying on logos displaying at any width other than the max-width of 48px, I don't think this change is likely to pose an issue. The docs previously said the image would be shown at 64x64, so the expectation of a square image has already been stated.

SVGs are great, and it looks like it's been possible to use them since 0.19.2, so I think it would be nice to formalise use of them.

josevalim:
I tried the CSS rules on Nx and that's how it will render:

Screenshot 2024-01-21 at 10 58 44

So I think forcing a width will definitely break expectations. :(

josevalim:
Or we document that SVG must have a width/height attached to them?

DavidOliver:
Oh - thanks. I intended the max-height to result in proportional scaling-down with tall logos.

Shall I see if I can get proportional/same-as-before scaling working? There's probably a way, and I should be able to limit setting of width to SVG images if necessary.

josevalim:
@DavidOliver it is your call :) You can try applying whatever changes you do to both https://hexdocs.pm/elixir/main/ and https://hexdocs.pm/nx, directly on the browser inspector, and see if they still look good. :)

DavidOliver:
Will do. Thanks. 😺

DavidOliver:
I thought I was getting close with use of flexbox and object-fit, but matching the previous resizing and alignment in all cases of a) different image formats, b) different image aspect ratios, and c) different browsers has eluded me, at least for now and without getting into SVG-specific styling and possibly JS, which I don't think is desirable given that we can instead choose to specify some basic requirements for SVGs.

I've pushed another commit that goes back to the previous CSS (with one tweak - see commit description) and adds documentation of the requirement of including relevant attrs in the SVG.

josevalim:

PR #1850, opened 2024-01-24
feat(epub): add support for group_for_modules in toc

milmazz:
Initial implementation in the EPUB formatter to support the `group_for_modules` option in the _table of contents_

The following is the current status:

Screenshot 2024-01-23 at 9 19 15 PM

This is after the initial implementation of this PR:

Screenshot 2024-01-23 at 9 20 46 PM


lib/ex_doc/formatter/epub/templates/nav_grouped_modules_template.eex
```diff
@@ -0,0 +1,19 @@
+<% nodes = Map.new(nodes, fn %{id: id} = node -> {id, node} end) %>
+<li><span><%= name %></span>
+  <ol>
+  <%= for {title, modules} <- config.groups_for_modules, is_list(modules) do %>
```
milmazz, line 4: There are no guarantees that all the available modules are listed here. So, I guess we better put those modules at the end in another loop.

lib/ex_doc/formatter/epub/templates/nav_template.eex
```diff
@@ -16,7 +16,11 @@
           <% end %>
         <% end %>
 
-        <%= nav_item_template "Modules", nodes.modules %>
+        <%= if Enum.empty?(config.groups_for_modules) do %>
+          <%= nav_item_template "Modules", nodes.modules %>
+        <% else %>
+          <%= nav_grouped_modules_template "Modules", nodes.modules, config %>
```
milmazz, line 22: It seems we can get rid of the first argument (`"Modules") here. So far, this template is only for modules.

lib/ex_doc/formatter/epub/templates/nav_grouped_modules_template.eex
milmazz: I just noticed that based on the docs from `lib/mix/tasks/docs.ex`, `groups_for_modules` can also hold 

> A regex or the string name of the module is also supported.

So, I should adapt this template to support those cases. Or add an intermediate function that adapts the output and then delegates that to `nav_grouped_modules_template` to make it cleaner.

josevalim: The group for a module is already in the module struct and I believe they are already sorted according to groups. So I think a `Enum.chunk_by(modules, & &1.group)` may be enough to get groups and the correct order. Otherwise we can likely use the same algorithm/logic we use for extras.

josevalim: I realized we could also refactor extras with my suggestion above, so here it goes: 3fbe0bc344b7d30edb87460db518f786fb7354bf :)

milmazz: @josevalim Thanks for the context and advice, now the logic for extras and modules is the same at least in the table of content generation, here is the result:

Screenshot 2024-01-24 at 7 39 07 p m
Screenshot 2024-01-24 at 7 38 56 p m

josevalim (approved):

PR #1853, opened 2024-01-25
Use blue links for Erlang

josevalim:
Examples below. Please use thumbs up (you prefer blue) or down (you prefer the current) to state your preference.

Screenshot 2024-01-25 at 21 51 31

Screenshot 2024-01-25 at 21 51 43

### DIFF
--- a/assets/css/custom-props/theme-light.css
+++ b/assets/css/custom-props/theme-light.css
@@ -9,8 +9,6 @@
 
   --links:                       var(--black);
   --linksVisited:                var(--black);
-  --linksNoUnderline:            var(--mainDark);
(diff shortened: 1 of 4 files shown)


assets/css/custom-props/theme-light.css
```diff
@@ -9,8 +9,6 @@
 
   --links:                       var(--black);
   --linksVisited:                var(--black);
-  --linksNoUnderline:            var(--mainDark);
-  --linksNoUnderlineVisited:     var(--mainDarkest);
```
josevalim, line 13: We just need to extract these default colors into per-language configuration.

josevalim:

PR #1855, opened 2024-01-30
Add before_closing_footer_tag and adjust annotation metadata

garazdawi:
This PR does two things:

1. Adds `before_closing_footer_tag` config option for html formatter to insert a custom footer. We use it to insert a copyright notice at the end of each page.
2. Add kmfa to the metadata before calling `annotations_for_docs` so that the fun doing the processing knows which entry it is currently working with. We use this to classify functions/types into different categories, such as built-in/predefined.

and as a bonus it also adds a warning for when an `extra` that is not in deps is referenced in a link.

### DIFF
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -304,7 +304,16 @@ defmodule ExDoc.Autolink do
           end
 
         config.deps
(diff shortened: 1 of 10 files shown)


josevalim:

PR #1858, opened 2024-02-03
Autolink .cheatmd files

sabiwara:
Fixes https://github.com/elixir-lang/elixir/issues/13315

Screenshot 2024-02-03 at 15 38 35


wojtekmach:
Thank you!

PR #1859, opened 2024-02-04
Warn if there is a newer version available

agonzalezro:
Check hex.pm to see if there is a newer published version of ExDoc available and warn if that's the case.

The GET request was a timeout of 1s and will not fail in case the response from hex.pm is different from 200.

Jason isn't use only for testing anymore and HTTPoison was added as a dependency as well.

This idea was discussed between @zachdaniel and @josevalim  in an X's thread: https://twitter.com/josevalim/status/1745910892854681818


wojtekmach:
Thank you for the PR. We already have #1843 so let’s continue the discussion there.

agonzalezro:
Sorry about that, I didn't know it was already open.

PR #1861, opened 2024-02-08
Correctly link otp apps not in deps to erlang.org

garazdawi:


lib/ex_doc/autolink.ex
```diff
@@ -121,52 +122,52 @@ defmodule ExDoc.Autolink do
...
+
+      maybe_path ->
+        case :application.get_application(module) do
+          {:ok, app} ->
+            app
+
+          _ ->
+            with true <- is_list(maybe_path),
+                 [_, "ebin", app, "lib" | _] <- maybe_path |> Path.split() |> Enum.reverse() do
+              String.split(app, "-") |> Enum.at(0) |> String.to_atom()
```
josevalim, line 137:
```suggestion
              String.split(app, "-") |> hd() |> String.to_atom()
```

lib/ex_doc/autolink.ex
```diff
@@ -121,52 +122,52 @@ defmodule ExDoc.Autolink do
...
 
-        {otp?, app}
+      true ->
+        :ex_doc
     end
   end
 
+  defp is_app_otp(app) do
+    maybe_lib_dir_path = :code.lib_dir(app)
+    is_list(maybe_lib_dir_path) && List.starts_with?(maybe_lib_dir_path, :code.root_dir())
```
josevalim, line 168:
```suggestion
    is_list(maybe_lib_dir_path) and List.starts_with?(maybe_lib_dir_path, :code.root_dir())
```

lib/ex_doc/autolink.ex
```diff
@@ -302,16 +303,32 @@ defmodule ExDoc.Autolink do
...
+          case tool do
+            :otp ->
+              @otpappdocs
+
+            :ex_doc ->
+              @hexdocs
+          end
+
+        # TODO: In ex_doc_app_url we check if `app in config.apps` and
+        # then don't # link to hexdocs. Should we do the same here?
```
josevalim, line 318: Apps are only used by umbrella projects when generating docs for several projects at once. So in theory that could happen here if they have multiple Erlang sources. Could we use `ex_doc_app_url` here? Or would that be too complicated?

lib/ex_doc/autolink.ex
```diff
@@ -121,52 +122,52 @@ defmodule ExDoc.Autolink do
...
-            _ ->
-              with true <- is_list(maybe_path),
-                   [_, "ebin", app, "lib" | _] <- maybe_path |> Path.split() |> Enum.reverse() do
-                String.split(app, "-") |> Enum.at(0) |> String.to_atom()
-              else
-                _ -> nil
-              end
-          end
+    cond do
+      is_app_otp(app) and app not in apps ->
```
josevalim, line 158:
```suggestion
      is_app_otp(app) and app not in config.apps  ->
```

I can't think of when an otp app would be in deps. :) This way we can get rid of the `apps` variable.

lib/ex_doc/autolink.ex
josevalim: Yeah, so maybe we don't use `tool_app` at all and instead write:

```
cond do
  app in config.apps -> # no prefix
  is_app_otp(app) -> @otpappdocs
  true -> @hexdocs
end
```

?

garazdawi: Ah, so then maybe `e:` link prefix should not work for umbrella apps and they should use relative local links for extra pages?

josevalim: I think they should work because umbrella projects allow you generating docs both at the root (altogether) and individually, so linking should likely support both depending on config.apps?

josevalim: We also can write references to ourselves, right? So if `config.apps == [:stdlib]`, then `e:stdlib` should not have any prefix I think.

lib/ex_doc/autolink.ex
garazdawi: When I build the docs for Erlang/OTP, otp apps are in deps :)

josevalim: @garazdawi then wouldn't the tool accuse to be ExDoc and generate links for HexDocs? Or maybe it does not matter because deps will use their own URLs anyway, so we can apply this change anyway?

lib/ex_doc/autolink.ex
garazdawi: Ah, ok. I've got no idea how umbrellas work so didn't know it would be that easy. I'll apply your suggest above after lunch and add a test or two for that case.

garazdawi: So, just to clarify what we want:

```elixir
def prefix(app, config) do
  cond do
    app in config.apps ->
      ""
    Keyword.has_key?(app,config.deps) ->
     "#{Keyword.fetch!(app,config.deps)}"
    is_otp_app(app) ->
     @otpappdocs
    true ->
     @hexdocs
  end
end
```

How would I go about to warn for the case when an application that is not part of the dependencies is linked to? That is the code that currently does this:

```elixir
config.deps
|> Keyword.get_lazy(String.to_atom(app), fn ->
  maybe_warn(
    config,
    "documentation references \"e:#{string}\" but #{app} cannot be found in deps.",
    nil,
    %{}
    )
     @hexdocs <> "#{app}"
end)
```

I want to be able to catch misspellings for application names.

josevalim: Maybe this:

```elixir
def prefix(app, config) do
  cond do
    app in config.apps ->
      ""
    Keyword.has_key?(app,config.deps) ->
     "#{Keyword.fetch!(app,config.deps)}"
    is_otp_app(app) ->
     @otpappdocs
    is_elixir_app(app)
    true ->
     warn("omgomgomgomg")
     @hexdocs
  end
end
```

garazdawi: I pushed an update that now works I think :) Basically the code became this:

```elixir
def prefix(app, config) do
  cond do
    app in config.apps ->
      Atom.to_string(app)
    Keyword.has_key?(app,config.deps) ->
     "#{Keyword.fetch!(app,config.deps)}"
    is_otp_app(app) ->
     @otpappdocs <> Atom.to_string(app)
    true ->
     if Application.load(app) != :ok do
        warn(....)
     end
     @hexdocs <> Atom.to_string(app)
  end
end
```

which is quite close to what you proposed.

lib/ex_doc/autolink.ex
```diff
@@ -302,16 +298,34 @@ defmodule ExDoc.Autolink do
...
+            {:error, {:already_loaded, ^app}} ->
+              nil
+
+            _ ->
+              maybe_warn(
+                config,
+                "documentation references \"e:#{string}\" but #{app} cannot be found.",
+                nil,
+                %{}
+              )
```
josevalim, line 318:
```suggestion
          if Application.ensure_loaded(app) != :ok do
            maybe_warn(
              config,
              "documentation references \"e:#{string}\" but #{app} cannot be found.",
              nil,
              %{}
            )
```

josevalim:

PR #1863, opened 2024-02-12
Hide search bar when printing

rhcarvalho:
Noticed this when trying to print a cheatsheet, though it should apply to printing any documentation page as we cannot use it to search on paper / PDF :)


josevalim:

PR #1868, opened 2024-02-20
Fix sidebar toggle flickering on page load

simonprev:
This prevent the animation from triggering after every page load. The margin/padding have also been adjusted to prevent flickering on toggle click.

You can see the before/after here where I showcase a toggle and a page refresh:

| Before  | After |
| ------------- | ------------- |
| <video src="https://github.com/elixir-lang/ex_doc/assets/464900/9eb759f0-2961-4564-a016-d6e75b6ae87e"/>  | <video src="https://github.com/elixir-lang/ex_doc/assets/464900/cae52fb1-75d8-489d-ba42-b23bfd17dcfd" />  |

The small viewport view looks the same:
image


josevalim:

Issue #1860, opened 2024-02-06
UX issues related to search bar location


zachallaun:
While I'm generally quite positive on the redesigned search bar, there are a handful of UX issues I've noticed that are a result of its new location at the top of the content portion of the page as opposed to the sidebar (which is sticky).

1. Least important, but worth mentioning: It is less discoverable, especially if you navigate directly to an anchor link that is farther down the page.
2. The `/` search focus keyboard shortcut instantly scrolls you to the top of the page, which is jarring and causes you to lose context. (I am often searching for something related to what I'm looking at.)
3. Related to 2, the jump to page top causes a loss of scroll history. Consider: you're halfway down a page, press `/` to search, navigate to a search result, then press the back button. Previously, your browser would restore the scroll position of the page you were originally on, but that is no longer possible as the `/` caused a scroll to the top of the page.

The easiest way to experiment with this is using the Elixir docs:

* [Elixir 1.16.1](https://hexdocs.pm/elixir/1.16.1/Kernel.html), which uses the new search
* [Elixir 1.15.7](https://hexdocs.pm/elixir/1.15.7/Kernel.html), which uses the old search

This issue is related to #1835, which could be used as inspiration for a solution.

josevalim:
I think we could always make it show up on scroll up, regardless if on mobile/small screens. `/ ` would just bring it up immediately. WDYT?

zachallaun:
Sounds great to me!

zachallaun:
@sodapopcan That would be perfectly acceptable to me.

josevalim:
Please do send a PR for `/` then! Feel free to add super k as well.

sodapopcan:
Will do!  It's already in the works and I should have a PR soon.

PR #1870, opened 2024-02-21
Add cmd-k/ctrl-k shortcut to focus searchbar

sodapopcan:
Hello ex_doc team!

This PR adds `cmd-k` (Mac) and `ctrl-k` (Win, Linux) to focus the search input.

This is the first of two features I agreed to take on discussed in #1860.  The second part is prevent the entire page from scrolling to the top when focusing the search bar and instead making it slide down over the content.  I'm happy to roll that into this PR but I wanted to get a review on this now since I don't have easy access to Windows of Linux machines so I would appreciate if someone could test the `ctrl-k` version.

I tried to stay very low-ceremony with this PR doing the simplest thing I could think of.  I've added a simple `hasModifier` option for shortcuts which will require that the system-appropriate shortcut key be used.  While of course `/` still works, I've made the placeholder hint show the `K` versions instead since I feel that is far more common these days.  Because it's system dependent, this had to be done with JavaScript.

Thanks, and the resolution to #1860 is coming soon!


assets/js/search-bar.js
```diff
@@ -102,6 +103,16 @@ function addEventListeners () {
...
 
+function setInputPlaceholder () {
+  const searchInput = qs(SEARCH_INPUT_SELECTOR)
+
+  if (isMacOS()) {
+    searchInput.setAttribute('placeholder', 'Search - ⌘K')
+  } else {
+    searchInput.setAttribute('placeholder', 'Search - ctrl K')
+  }
+}
```
josevalim, line 114: We should not change the placeholder. Let's keep it as it is right now. I think `/` is still the most convenient for most users (and also what GitHub uses).

garazdawi: Just as one datapoint to the opposite, `/` is (IMO) very inconvenient on a Swedish keyboard. Looking forward to using ctrl-k in ex_doc!

josevalim: Oh, good point. You can also press `s`. Maybe we should change the placeholder to say press `s` instead? 🤔

sodapopcan: No problem-o!  I forgot about github as for some reason it's one of the only sites where I don't use its keyboard shortcuts.  As a Vim user I do find `/` an odd choice and since it's used for search _within_ a document whereas `k` is used to search the docs, but github is obviously good precedent!  In any event I'm happy to bring back the placeholder.  Let me know if you'd rather I replace it with `s`.

josevalim: Let's only remove the placeholder changes for now, we can revisit in another PR.

sodapopcan: Done!

josevalim:

PR #1872, opened 2024-02-24
Bring search bar into view on desktop

sodapopcan:
This commit changes how the search bar is brought into view on desktop as brought up in #1860: when using one of the keyboard shortcuts to focus the input, the entire page jumps to the top making you lose your scroll position.  On mobile, technically smaller screens, this is remedied by causing the input to slide into view on scroll up, but I don't believe this is a desirable solution for desktop.

The following changes are introduced:

  * Using one of the keyboard shortcuts will focus the search input causing it to stick to the top.
  * It will continue to stick to the top so long as it is focused allowing you to scroll with it open.
  * The slide-on-scroll effect has been changed to only fire on touch-enabled devices as opposed to just smaller screens.  This is allows us to make the hexdocs window very small and still use keyboard shortcuts--Useful on laptops.

## PR notes

This turned out to be a little hairier than I imagined due to thinking there was more that needed to be done than there was.  Turns out it's a fairly small PR!  I did have to make the following concessions but overall I feel the UX is pretty solid.

- When the search bar is brought into view, it needed a little bit of padding.  Rather than get into branching code that checks if the search bar is visible in the viewport or not, I've just permanently added the padding (only on desktop).  It's very slight.
- I haven't added a slide-in/out effect as that is a bit more ceremony for potentially not amazing gains.  I can certainly add it if you like!  It's not that bad or anything, but needs to know the static search bar height (which probably isn't a problem).
- Closing the search bar has been hooked into closing autocomplete.  I realized this JustWorks after quite a while of futzing with a conditional blur event as well as contemplating bring in a click-away event.  I haven't done any function renaming as it sort of gets weird between touch and desktop.  Again I'm happy to put some time in a refactor there!
- If you are going to be testing using devtools to switch between desktop and a touch device be aware things get a little wonky when you switch so you need to reload.  I put no time into fixing this since devices don't switch between being touch enabled and not in real scenarios.

### DIFF
--- a/assets/css/layout.css
+++ b/assets/css/layout.css
@@ -122,7 +122,28 @@ body.sidebar-closed .content {
   left: 0;
 }
 
-@media screen and (max-width: 768px) {
+@media screen and (hover: hover) and (max-width: 768px) {
+  body.sidebar-opening .content {
+    left: 0;
+    width: 100%;
+  }
+
+  body.sidebar-closed .sidebar-button {
+    position: absolute;
+    top: 14px;
+  }
+
+  body.sidebar-closed .sidebar-button.fixed {
+    position: fixed;
+    padding: 16px 12px 18px 19px;
+  }
+
+  body.sidebar-closed .sidebar-button.fixed-top {
+    position: fixed;
+  }
+}
+
+@media screen and (hover: none) {
   .content,
   body.sidebar-opening .content {
     left: 0;
@@ -137,6 +158,7 @@ body.sidebar-closed .content {
 
   body.sidebar-closed .sidebar-button {
     position: absolute;
+    top: 14px;
   }
 
   .sm-fixed {
--- a/assets/js/helpers.js
+++ b/assets/js/helpers.js
@@ -161,10 +161,21 @@ export function getProjectNameAndVersion () {
 }
 
 /**
- * Return `true` if the client's OS is MacOS
+ * Return `true` if the client's OS is MacOS.
  *
  * @return {Boolean}
  */
 export function isMacOS () {
   return /(Mac|iPhone|iPod|iPad)/i.test(navigator.platform)
+}
+
+/**
+ * Return `true` if the client's device is touch-enabled.
+ *
+ * @return {Boolean}
+ */
+export function isTouchDevice () {
+  return (('ontouchstart' in window) ||
+    (navigator.maxTouchPoints > 0) ||
+    (navigator.msMaxTouchPoints > 0))
 }
--- a/assets/js/search-bar.js
+++ b/assets/js/search-bar.js
@@ -7,10 +7,11 @@ import {
   AUTOCOMPLETE_CONTAINER_SELECTOR,
   AUTOCOMPLETE_SUGGESTION_SELECTOR
 } from './autocomplete/autocomplete-list'
-import { isMacOS, qs } from './helpers'
+import { isMacOS, isTouchDevice, qs } from './helpers'
 
 const SEARCH_INPUT_SELECTOR = 'form.search-bar input'
 const SEARCH_CLOSE_BUTTON_SELECTOR = 'form.search-bar .search-close-button'
+const TOP_SEARCH_SELECTOR = '.top-search'
 
 /**
  * Initializes the sidebar search box.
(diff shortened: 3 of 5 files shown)


assets/js/helpers.js
```diff
@@ -161,10 +161,21 @@ export function getProjectNameAndVersion () {
 }
 
 /**
- * Return `true` if the client's OS is MacOS
+ * Return `true` if the client's OS is MacOS.
```
sodapopcan, line 164: Just added a missing period from my last PR :)

sodapopcan:
I noticed a problem with this PR I thought I had taken care of so I've converted it to a draft until I fix it.  It looks like I may need to bring in click-away since click the clear button or gear causes the whole search bar to close.

josevalim:
> I haven't added a slide-in/out effect as that is a bit more ceremony for potentially not amazing gains. I can certainly add it if you like! It's not that bad or anything, but needs to know the static search bar height (which probably isn't a problem).

Could we just naturally make it so scrolling down makes it disappear again?

If we scroll up, it continues at the top, if we scroll down, it goes away. The idea is that it behaves like in mobile, except that scrolling up on large screens does not bring it down, only pressing "/".

sodapopcan:
> 
> Could we just naturally make it so scrolling down makes it disappear again?
> 

Absolutely!  It certainly crossed my mind in the several ways I realized this could work so I'll go with that as it will also solve my other problem.

sodapopcan:
Just an update: I'm very sorry I have not gotten back to this.  I will in the next day or so!  I just had some stuff happen that is making me not want to write code and be social, even though for today that has manifested in _talking_ about code a whole lot of Elixir Forum 🙃

assets/css/layout.css
```diff
@@ -122,7 +122,28 @@ body.sidebar-closed .content {
...
+    position: absolute;
+    top: 14px;
+  }
+
+  body.sidebar-closed .sidebar-button.fixed {
+    position: fixed;
+    padding: 16px 12px 18px 19px;
+  }
+
+  body.sidebar-closed .sidebar-button.fixed-top {
```
sodapopcan, line 141: The burger menu has to be moved downward to be properly aligned with the searchbox when we are scrolled down the page.  When we are at the top of the page, the `top-search` div is padded differently so we have to compensate in this situation.

assets/js/search-bar.js
```diff
@@ -34,6 +35,16 @@ export function setSearchInputValue (value) {
  */
 export function focusSearchInput () {
   const searchInput = qs(SEARCH_INPUT_SELECTOR)
+
+  if (!isTouchDevice()) {
+    qs(TOP_SEARCH_SELECTOR).classList.add('sticky')
+    if (window.scrollY === 0) {
```
sodapopcan, line 41: See comment above.

assets/js/search-bar.js
```diff
@@ -139,38 +150,47 @@ function hideAutocomplete () {
...
+      sidebarMenu.classList.add('sm-hidden')
+      backgroundLayer.classList.add('sm-hidden')
+    } else {
+      // Scrolling up or at the top of the page
+      topSearch.classList.remove('sm-hidden')
+      sidebarMenu.classList.remove('sm-hidden')
+      backgroundLayer.classList.remove('sm-hidden')
+    }
+  } else if (currentScroll !== lastScrollTop) {
+    topSearch.classList.remove('sticky')
```
sodapopcan, line 190: I was going to extract these into a `closeSearchbar` function but it's only called once and it would require passing in the DOM elements as dependencies otherwise we'd running three DOM queries in a scroll callback which would be worse than this run-on sentence :)

sodapopcan:
This is ready for review!

This was honestly more of a slog than I imagined it to be 😰  Certainly not the worst, but there are more states to consider than I was thinking.  I only mention that as I'm a bit burned out on it so I never brought in the animation.  That said, if for some reason you want to reject this don't worry, I won't take it poorly (I know you're probably used to this I always just like to reassure for open source contributions).  I added some review comments and otherwise I think it's best to just try it out as opposed to me trying to describe it, but here is a summary anyway:

*On Desktop*

- `s` & friends focus the search bar as normal when `scrollY` is `0`.
- In any other scroll position it will bring down the searchbar in an overlay at the top.
- Scrolling in either direction will hide it
- With the sidebar closed and the viewport small enough, the burger menu will come down with the searchbar

*On Mobile*

- Nothing has changed other than, as mentioned, I've switched it to only fire on touch-enabled devices as opposed to screen size.

As per that last point, all behaviour related to showing the searchbar is dependent on being on a touch-enabled device or not and nothing to do with viewport size.

josevalim:

PR #1873, opened 2024-03-04
Add `equiv` handling for types and callbacks for Erlang

garazdawi:
This PR adds `equiv` handling for types and callbacks to the Erlang mode by adding `doc_fallback` support to the retriever for types and callbacks. 

Also sneaking in a doc fix to describe keys available in metadata in `groups_for_*` functions.

### DIFF
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -342,8 +342,8 @@ defmodule ExDoc.Retriever do
         callback_data.extra_annotations ++ annotations_from_metadata(metadata, module_metadata)
 
     doc_ast =
-      (source_doc && doc_ast(content_type, source_doc, file: doc_file, line: doc_line + 1)) ||
-        (callback_data[:doc_fallback] && callback_data.doc_fallback.())
+      doc_ast(content_type, source_doc, file: doc_file, line: doc_line + 1) ||
+        doc_fallback(callback_data)
 
     group =
       GroupMatcher.match_function(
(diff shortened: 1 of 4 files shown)


lib/ex_doc/retriever.ex
```diff
@@ -341,7 +341,9 @@ defmodule ExDoc.Retriever do
       annotations_for_docs.(metadata) ++
         callback_data.extra_annotations ++ annotations_from_metadata(metadata, module_metadata)
 
-    doc_ast = doc_ast(content_type, source_doc, file: doc_file, line: doc_line + 1)
+    doc_ast =
+      (source_doc && doc_ast(content_type, source_doc, file: doc_file, line: doc_line + 1)) ||
+        (callback_data[:doc_fallback] && callback_data.doc_fallback.())
```
josevalim, line 346: Apologies in case I am missing something obvious, but can `source_doc` ever be `nil`? However, even if `nil`, `doc_ast` would return `nil` anyway, so I think we can skip the `source_doc &&` bit. I would also add a small function, `doc_fallback(callback_data)` for the second part, so we end-up wit this:

```suggestion
      doc_ast(content_type, source_doc, file: doc_file, line: doc_line + 1) ||
        doc_fallback(callback_data)
```

WDYT?

garazdawi: > Apologies in case I am missing something obvious, but can source_doc ever be nil? 

I was just doing the same thing as for function_data, might very well be that source_doc cannot be `nil`.

Refactor pushed.

josevalim:

Issue #1866, opened 2024-02-18
Admonitions not styled properly within cheat sheets


davidsulc:
[I'm adding this issue on the assumption that admonitions should work/display in cheat sheets just as they do in normal docs.]

Admonition blocks can't be properly used in cheatsheets as their CSS rules are being overwritten by `.page-cheatmd` rules.

Actual:
![actual](https://github.com/elixir-lang/ex_doc/assets/356709/934779bc-93b0-4766-9d66-f44da5769944)

Expected:
![expected](https://github.com/elixir-lang/ex_doc/assets/356709/2f389728-97da-4eef-b09c-601542229f29)

DavidOliver:
I can take a look at this soon.

josevalim:
Thank you @DavidOliver. However, if deemed too complicated, I'd say it is fine to not supported admonitions in `.cheatmd`. They are also not supported `.livemd` as well (as in: Livebook does not render them properly).

josevalim:
Hi @DavidOliver, just checking: is this worth investigating/addressing?

DavidOliver:
@josevalim Sorry for the delay. I've now refreshed my memory of the cheatsheet styles, and they do seem to deliberately take a simple approach and assume a limited set of elements. I get the sense that the limited support of features in cheatsheets makes sense given their purpose. (Semi-related, tabset content isn't included in the output HTML in cheatsheets.)

Perhaps we could update the intro paragraph of the Cheatsheet User Guide to:

> This document is a cheatsheet on how to use cheatsheets. Cheatsheets are Markdown files with the .cheatmd extension which support a limited range of the ExDoc features. Refer to the source of this document.

josevalim:
@DavidOliver whatever you decide is fine by me. Could you please submit a PR with your proposed changes? :)

DavidOliver:
@josevalim, will do. I think I can improve on the wording I gave above.

davidsulc:
Thanks for taking the time to look into it! :+1:

PR #1876, opened 2024-03-06
Update Cheatsheet guide wording

DavidOliver:
Prompted by #1866 

I've tried to hint that not all features of ExDoc are intended to be used in cheatsheets without listing unsupported features with "which support a limited range of elements and formatting".

I've also shortened and tweaked some other wording in an attempt to simplify the experience of reading through and increase accuracy/relevance.


josevalim:

PR #1878, opened 2024-03-10
Make a function acceptable for :skip_code_autolink_to option

irisTa56:
Currently, the `:skip_code_autolink_to` option only accepts a list.
This approach becomes impractical when dealing with a large number of terms.

For example, I'm working on a project where some hidden modules are automatically defined, and some documentation tries to link to those modules (to be precise, types defined in those modules).
My problem there is that the number of those terms is too many to list in `:skip_code_autolink_to` option.

So it is useful if `:skip_code_autolink_to` option accepts regular expressions or, more flexibly, a function.
I choose the function approach as it aligns with the discussion in https://github.com/elixir-lang/ex_doc/pull/1759#discussion_r1314751777.
(I'm pleased to implement the same feature for `:skip_undefined_reference_warnings_on` option if this PR is approved.)

As a caveat, because `ExDoc.Autolink` is internal (`@moduledoc false`), I just removed code that expects a list for the `skip_code_autolink_to` field.
Should I keep backward compatibility at `ExDoc.Autolink` level too?

### DIFF
--- a/lib/ex_doc/autolink.ex
+++ b/lib/ex_doc/autolink.ex
@@ -26,7 +26,8 @@ defmodule ExDoc.Autolink do
   #
   # * `:skip_undefined_reference_warnings_on` - list of modules to skip the warning on
   #
-  # * `:skip_code_autolink_to` - list of terms that will be skipped when autolinking (e.g: "PrivateModule")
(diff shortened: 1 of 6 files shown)


josevalim:
Adding similar to `skip_undefined_reference_warnings_on` will be welcome!

PR #1883, opened 2024-04-03
Implement search preview

zachdaniel:
Hey all!

I've discussed the ability to preview pages in autocomplete with @josevalim, and this is the groundwork for that feature. There is more to be done, and I hate to be one to throw things over the fence, but with Ash's 3.0 release coming soon, multiple trainings ahead, and various other things on my plate, I just haven't had time to wrap this up and I don't know when I will have time to get back to it. Additionally, this isn't my wheelhouse so it is possible that there are folks who are more familiar w/ making good UIs that could be more effective. Here is a video showing the current behavior:

https://github.com/elixir-lang/ex_doc/assets/5722339/5213b86c-0acc-4e56-a58d-b6d42ce4be95

And here is what remains to be done:

1. The button that is displayed currently doesn't do anything on click.
2. There is an issue I can't nail down with the iframe's height. Ideally, it should adapt to the height of its context.
3. the right arrow should expand, and the left arrow should close, whereas currently they both just toggle.

Again, sorry I haven't had the time lately to get this over the line, but I think it is close enough to open a draft PR, hopefully someone has some cycles to wrap it up! If not, I can come back to it in  maybe 4-5 weeks


assets/js/autocomplete/autocomplete-list.js
```diff
@@ -89,12 +90,63 @@ export function moveAutocompleteSelection (offset) {
...
+  const suggestionList = qs(AUTOCOMPLETE_SUGGESTION_LIST_SELECTOR)
+
+  if(state.previewOpen && elementToSelect) {
+    suggestionList.classList.add('previewing')
+    const newContainer = document.createElement('div')
+    newContainer.classList.add('autocomplete-preview')
+    const iframe = document.createElement('iframe')
+    const previewHref = elementToSelect.href.replace('.html', '.html?preview=true')
+    // The minimum permissions necessary for the iframe to run JavaScript and communicate with the parent window.
+    iframe.setAttribute('sandbox', 'allow-scripts allow-same-origin allow-popups allow-forms')
```
josevalim, line 137: Quick question: is there a reason why you add allow-popups and allow-forms?

zachdaniel: IIRC it was necessary for the iframe to render properly, but it was something I set early so we should explore removing it. I think `allow-popups` was to allow the hover info to work inside of previews.

zachdaniel:
@josevalim that makes sense! I wasn't actually sure why that route was chosen, but I understand now. That will also likely solve the iframe height issues, if we just have it push all of its contents and then create an element, it won't have the iframe sizing problems :)

josevalim:
I did a follow up with some refactorings on #1884. Mostly JS left now :D

PR #1885, opened 2024-04-04
WIP: search preview

ThaddeusJiang:
Sorry, It is still work in progress.

- [x]  Fix the left and right arrow navigation. The right arrow should expand, and the left arrow should close, whereas currently they both just toggle

CleanShot 2024-04-04 at 23 04 50

- [x]  The buttons displayed doesn't do anything on click

CleanShot 2024-04-05 at 14 14 36

- [ ]  Use postMessage. The iframe should start as hidden and we should only show the iframe once it uses postMessage. The postMessage should include the content height, so we can render the maximum iframe height accordingly

PR #1886, opened 2024-04-05
Autocomplete preview: handle tab and arrow keys, fix lost focus on iframe load

arathunku:
I started working on this before https://github.com/elixir-lang/ex_doc/pull/1885 was open but decided to submit it anyway because it has a slightly different approach plus some additional things. 

I'm using mostly keyboard to navigate and wanted to get the UX just right :smile: 

MR addresses the problem with clicking the preview buttons and keeps the focus inside the input all the time. See the video:

https://github.com/elixir-lang/ex_doc/assets/749393/af85330b-12ac-4f10-a2d5-1378e0e54a2a

I'm happy to revert/drop changes from my MR as https://github.com/elixir-lang/ex_doc/pull/1885 progresses because this branch has one more thing... The "Close button" and "Open button" didn't match rest of the UI so I've added `eye` and `eye-open` to the bundle, and it feels slightly better. 

One more thing that was missing from me was being able to click "Open" without touching arrow keys, with the icon as button instead of preview I think the icon could be visible all the time. With previous buttons I tried to do that but the UI looked very "busy". What do you think?


josevalim:
I have merged this locally and I will push soon!

josevalim:

PR #1887, opened 2024-04-08
fix: click open / close preview button

ThaddeusJiang:
before: 
CleanShot 2024-04-08 at 17 47 56

after:
CleanShot 2024-04-08 at 17 48 33


arathunku:
:wave: @ThaddeusJiang I quickly tested the change in Firefox and I cannot close the preview with left arrow, it moves the cursor inside the input field. Another issues is that now autocomplete is no longer closed when we click outside the container. I hope this is helpful, there are so many edge cases in getting the preview and keyboard navigation right!

assets/js/handlebars/templates/autocomplete-suggestions.handlebars
```diff
@@ -6,7 +6,7 @@
         Autocompletion results for <span class="bold">"{{term}}"</span>
       </span>
       <span class="press-return">
-        Press <span class="bold">RETURN</span> for full-text search, <span class="bold">TAB</span> for previews
+        Press <span class="bold">RETURN</span> for full-text search, <span class="bold">→</span> for previews
```
josevalim, line 9: I am sorry but I think we should go with TAB, as someone can still use the arrows to change the input itself. Can we revert to tab?

ThaddeusJiang: @josevalim ok, understood

ThaddeusJiang: @josevalim Q: how to close preview by keyboard? Is it good idea using  `shift + Tab` ?

josevalim: Generally speaking, I think the current navigation looks good, so unless you find a bug (code bug or UI bug), I'd be happy to keep it as is. :)

josevalim: @ThaddeusJiang it should just be pressing Tab again. It should work right now as well!

ThaddeusJiang: @josevalim understood

As my test, the TAB key was broken.

before:
CleanShot 2024-04-08 at 15 47 54

So, I create this PR.

after: 

CleanShot 2024-04-08 at 15 49 34

josevalim: @ThaddeusJiang interesting. which browser are you using?

ThaddeusJiang: @josevalim Arc, I am using https://arc.net/

josevalim: @ThaddeusJiang this is weird. I have tested on both Arc and Safari  and they both work just fine. Have you run `mix build` before testing it, just in case?

josevalim: I have also noticed that your preview shows the search bar and it does not show up for me. :(

josevalim: That's it, great find. It should be "open doc/index.html". Maybe we should document it somewhere?

josevalim: I tried to improve the README instructions for clarity. Btw, do you think we need something from this PR now? Thanks!

ThaddeusJiang: @josevalim I have fixed the click issue in this PR.
Could you review it?
 https://github.com/elixir-lang/ex_doc/pull/1887/files

josevalim:

PR #1891, opened 2024-04-10
Fix search input focus text alignement and overscroll-behavior for results

simonprev:
This fixes the 1px offset on focus:

https://github.com/elixir-lang/ex_doc/assets/464900/199b6f89-deeb-4e3a-b1c5-8b3da2f2e476

And also added `overscroll-behavior` to prevent "body" scroll when we scroll to the end of the results. See the attached video, if I continue scrolling in the results, it scrolls the body underneath. `overscroll-behavior: contain` prevent this from happening, it’s like a "scroll trap" 😄 

https://github.com/elixir-lang/ex_doc/assets/464900/dc7cfd21-d3ee-45dc-938e-7c3ec9b66d47


josevalim:

PR #1892, opened 2024-04-10
style: fixup search bar's focus border radius

dbernheisel:
there is a gap between the nested borders on the search bar, this corrects that gap. The parent element has an 8px border-radius, and this element has a 1px border width, so 8-1 = 7px for perfected nested radii.

before:
Screenshot 2024-04-10 at 9 59 39 AM

after:
Screenshot 2024-04-10 at 9 59 35 AM


zachdaniel (approved):

josevalim:

PR #1894, opened 2024-04-16
Only remove -spec/-type from first line

garazdawi:
Before this change a spec looking like this:

```
-spec foo(integer()) -> ok;
         (atom()) -> ok.
```

would be transformed to:

```
-spec foo(integer()) -> ok;
    (atom()) -> ok.
```

that is the second clause would be indented incorrectly.

### DIFF
--- a/lib/ex_doc/language/erlang.ex
+++ b/lib/ex_doc/language/erlang.ex
@@ -709,7 +709,11 @@ defmodule ExDoc.Language.Erlang do
     offset = byte_size(Atom.to_string(type)) + 2
 
     options = [linewidth: 98 + offset]
-    :erl_pp.attribute(ast, options) |> IO.iodata_to_binary() |> trim_offset(offset)
+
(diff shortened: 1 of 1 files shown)


lib/ex_doc/language/erlang.ex
```diff
@@ -728,10 +728,6 @@ defmodule ExDoc.Language.Erlang do
   defp trim_offset(binary, offset) do
     binary
     |> String.trim()
-    |> String.split("\n")
-    |> Enum.map(fn line ->
-      binary_part(line, offset, byte_size(line) - offset)
-    end)
-    |> Enum.join("\n")
+    |> then(&binary_part(&1, offset, byte_size(&1) - offset))
```
josevalim, line 731: I think we can remove this function then and call `String.trim_leading(string, "-type ")` instead?

garazdawi: Fixed I think

josevalim:

PR #1897, opened 2024-05-01
Allow the modal to close when we click outside the modal content

ashkan117:
With many modals when you click outside the content the modal should close. This is helpful in the scenario where you press the 'g' keyboard shortcut and want to close it. The only way we can do so right now is by clicking the 'X' on the modal. I think this makes things more consistent with standard modal UI experiences and is a nice utility to have.

### DIFF
--- a/assets/js/modal.js
+++ b/assets/js/modal.js
@@ -34,6 +34,15 @@ function renderModal () {
   qs(MODAL_CLOSE_BUTTON_SELECTOR).addEventListener('click', event => {
     closeModal()
   })
+
+  qs(MODAL_SELECTOR).addEventListener('click', event => {
+    const classList = event.target.classList
+    // if we clicked on the modal overlay/parent but not the modal content
+    if (classList.contains('modal') && classList.contains('shown') && classList.length === 2) {
+      closeModal()
+    }
+  })
+
 }
 
 /**
--- a/assets/js/modal.js
+++ b/assets/js/modal.js
@@ -42,7 +42,6 @@ function renderModal () {
       closeModal()
     }
   })
-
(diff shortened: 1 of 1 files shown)


assets/js/modal.js
```diff
@@ -34,6 +34,15 @@ function renderModal () {
...
   })
+
+  qs(MODAL_SELECTOR).addEventListener('click', event => {
+    const classList = event.target.classList
+    // if we clicked on the modal overlay/parent but not the modal content
+    if (classList.contains('modal') && classList.contains('shown') && classList.length === 2) {
+      closeModal()
+    }
+  })
+
```
josevalim, line 45:
```suggestion
```

assets/js/modal.js
```diff
@@ -34,6 +34,14 @@ function renderModal () {
   qs(MODAL_CLOSE_BUTTON_SELECTOR).addEventListener('click', event => {
     closeModal()
   })
+
+  qs(MODAL_SELECTOR).addEventListener('click', event => {
+    const classList = event.target.classList
+    // if we clicked on the modal overlay/parent but not the modal content
+    if (classList.contains('modal') && classList.contains('shown') && classList.length === 2) {
```
josevalim, line 41: Should we check if the ID is "modal" instead of hardcoding classes?

ashkan117: Sorry for the delayed response! Yeah that would be a better way to go about things. I can refactor the code and make a new PR for that using the id instead

ashkan117: Ah nevermind I see that it's been done already. Awesome!

PR #1898, opened 2024-05-01
fix: mix task regex didn't consider tasks with underscores

zachdaniel:
`mix ash_phoenix.gen.live` was not auto-linked, but it is now


josevalim:

PR #1905, opened 2024-05-16
Make light theme sidebar light

### LINKED ISSUE #1900: Truly light theme
I was referred here from Erlang OTP where I [raised this problem](https://github.com/erlang/otp/issues/8447). I am raising it, because I tried using the new documentation in Erlang OTP 27 RC3 which uses `ex_doc`. I prefer to use light themes for my reading. But for documentation it means completely light, that is every part should have dark text on light background, even the sidebar.

I noted the dark sidebar in light theme is intentional as mentioned in [other issue](https://github.com/elixir-lang/ex_doc/issues/1774).

For me, this is a big usability problem. The contrast between the almost black sidebar and the white main area makes it hard for me to read anything in the sidebar. I have to consciously mentally switch just to be able to navigate elsewhere or merely scan the list of available types/functions/etc. and then switch back to read the text.

When comparing the new OTP documentation with the old one side by side, I initially thought the new one is somehow worse. But then I compared details and realized the new one is actually slightly improved in contents and even has some nice bells and whistles. It is just the difference in theming between the sidebar and the detail contents that makes it considerably less usable to me.

If you want to keep the current Light theme as is, could you please add a third, truly light theme with light sidebar? I would much appreciate it, since it will soon be thrust upon me with the new OTP release.

(I noticed there are three themes: System, Dark and Light. I initially suggested to change the Light theme, since it seems identical to the System theme to me. I didn't realize the System theme actually automatically changes between Dark and Light depending on browser settings.)

fififionek:
As proposed in [Truly light theme](https://github.com/elixir-lang/ex_doc/issues/1900) issue, I updated the light theme stylesheets so the sidebar coloring is dark text on light background.

I used only already existing color variables.
I had to change `sidebar.css`, because inactive navigation tabs were styled with non-themed variables for hover.

resolves #1900

### DIFF
--- a/assets/css/custom-props/theme-light.css
+++ b/assets/css/custom-props/theme-light.css
@@ -76,18 +76,18 @@
                                                  var(--white) 20%,
                                                  var(--white-opacity-50) 70%,
                                                  var(--white-opacity-0) 100%);
-  --sidebarAccentMain:           var(--gray50);
-  --sidebarBackground:           var(--gray800);
-  --sidebarHeader:               var(--gray700);
-  --sidebarMuted:                var(--gray300);
-  --sidebarHover:                var(--white);
-  --sidebarScrollbarThumb:       var(--coldGray);
+  --sidebarAccentMain:           var(--gray800);
+  --sidebarBackground:           var(--gray50);
+  --sidebarHeader:               var(--gray100);
+  --sidebarMuted:                var(--gray600);
+  --sidebarHover:                var(--black);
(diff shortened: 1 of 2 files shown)


fififionek:
resolves #1900

DavidOliver:
I just tried this out, and it's looking good to me.

josevalim:
Merged manually, thank you. We will give it a try on main for a while and collect feedback.

josevalim:

PR #1910, opened 2024-06-01
Give example when showing deprecation warning for binary assets config

zachallaun:
I encountered this warning when upgrading and it wasn't immediately clear to me what I should do to fix it. This change updates the printed warning to show the exact valid config you could copy/paste into your `mix.exs`.


josevalim:

PR #1914, opened 2024-06-06
Avoid unwanted showing/hiding of search bar on mobile when scrolling

### LINKED ISSUE #1913: Mobile: Search bar is too sensitive to page scroll
Currently, on small screens, the search bar appears when you drag up and disappears when you drag down. However, it appears to do so with only one pixel of movement. I will often "flick" to scroll quickly and, which I reach the desired section, press and drag slowly to align the content I'm interested in. I noticed, however, then when I lift my finger, it will often cause an extremely small amount of movement in the opposite direction, which causes the search bar to either disappear or reappear. Similarly, which trying to make it appear by dragging up a bit, it will often disappear when I lift my finger because of the small counter movement. 

To fix this, some small threshold of movement could be introduced within which the search bar does not toggle. 

To reproduce: visit https://hexdocs.pm/elixir/main/Kernel.html on your phone, scroll to middle of the page, observe the search bar behavior when dragging up/down.

DavidOliver:
Fixes #1913

Please note the removal of the code commented with "Add 'scroll-sticky' class when not at the top". This removal stops the search bar from appearing when a fragment identifier link is loaded, preventing the obscuring of content of linked-to sections, such as a heading. Is there any reason this code should stay?

### DIFF
--- a/assets/js/search-bar.js
+++ b/assets/js/search-bar.js
@@ -172,29 +172,23 @@ function hideAutocomplete () {
   hideAutocompleteList()
 }
 
-let lastScrollTop = window.scrollY
-const scrollThreshold = 70 // Set a threshold for scroll, adjust as needed
+let lastScroll = window.scrollY
+const scrollThreshold = 70 // Threshold beyond which search bar may be hidden
+const scrollConsideredIntentional = -2 // Pixels (negative sign) considered to indicate intent
 
 window.addEventListener('scroll', function () {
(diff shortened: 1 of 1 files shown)


josevalim:
@DavidOliver yes, we don't know to show it on anchors. As long as the sidebar shows up when scrolling up, it is fine to remove it.

DavidOliver:
@josevalim, thanks. The search bar is still showing when scrolling up.

PR #1915, opened 2024-06-06
Consolidate Apple OS detection

DavidOliver:
There were previously two functions that checked for Apple OS.

Converts previous OS module to Styling and modernises the function in Helpers.


josevalim:

PR #1916, opened 2024-06-10
Adds a `v` shortcut to open/focus the version select

andersonmcook:
When searching for docs in a search engine, the results don't always include the latest version as the first choice. This adds a `v` shortcut that will open the version select (if the user's browser supports it, otherwise it will focus it) to make selecting versions easier.

### DIFF
--- a/assets/js/keyboard-shortcuts.js
+++ b/assets/js/keyboard-shortcuts.js
@@ -1,5 +1,6 @@
 import { isAppleOS, qs } from './helpers'
-import { toggleSidebar } from './sidebar/sidebar-drawer'
+import { isSidebarOpened, openSidebar, toggleSidebar } from './sidebar/sidebar-drawer'
+import { openVersionSelect } from './sidebar/sidebar-version-select'
 import { focusSearchInput } from './search-bar'
 import { cycleTheme } from './theme'
 import { openQuickSwitchModal } from './quick-switch'
@@ -33,6 +34,11 @@ export const keyboardShortcuts = [
     key: 'k',
     hasModifier: true,
     action: searchKeyAction
+  },
+  {
+    key: 'v',
+    description: 'Open/focus version select',
+    action: versionKeyAction
   },
   {
     key: 'g',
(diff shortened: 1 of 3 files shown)


assets/js/sidebar/sidebar-drawer.js
```diff
@@ -136,13 +143,13 @@ export function openSidebar () {
   sessionStorage.setItem('sidebar_state', 'opened')
   qs(SIDEBAR_TOGGLE_SELECTOR).setAttribute('aria-expanded', 'true')
 
-  requestAnimationFrame(() => {
```
andersonmcook, line 139: `openSidebar` and `closeSidebar` did not actually return `Promise`s so I made a small update so that pressing `v` with a closed sidebar would pop open the sidebar and _then_ open/focus the version select.

assets/js/keyboard-shortcuts.js
```diff
@@ -34,6 +35,11 @@ export const keyboardShortcuts = [
     hasModifier: true,
     action: searchKeyAction
   },
+  {
+    key: 'v',
+    description: 'Open/focus version select',
```
andersonmcook, line 40: We could use `'showPicker' in HTMLSelectElement.prototype` to determine the language.

assets/js/sidebar/sidebar-version-select.js
```diff
@@ -70,3 +70,29 @@ function handleVersionSelected (event) {
...
+ *   - the browser's HTMLSelectElement lacks the showPicker method
+ *   - there has been transient user interaction
+ */
+export function openVersionSelect () {
+  const select = qs(VERSIONS_DROPDOWN_SELECTOR)
+
+  if (select) {
+    select.focus()
+
+    // Prevent subsequent 'v' press from submitting form
```
andersonmcook, line 86: This confused me, but the `v` submits the form. You can test this on the [main hexdocs page for Elixir](https://hexdocs.pm/elixir/Kernel.html).

Click the version select without selecting a version, press `escape` (the select is still focused), press `v`, and it will take you to the version below your current selection.

Issue #1917, opened 2024-06-10
Warn about docs for out-of-date versions


pdavies:
Hi, as a relative newcomer to Elixir, I've been caught out several times by looking at outdated documentation without realising that it was for an old version of the language or a library. This is particularly easy to do because of Google's tendency to return old versions at the top of results:

<img width="675" alt="image" src="https://github.com/elixir-lang/ex_doc/assets/1106188/b50a984d-dd70-4924-a67d-60157197a7a6">

<img width="656" alt="image" src="https://github.com/elixir-lang/ex_doc/assets/1106188/280448c8-5a2a-443e-9d80-324a78820fca">

HTML docs built with exdoc could warn when viewing old versions, but they do not, and this seems like a shortcoming worth fixing. There is plenty of precedence for warnings like this in versioned HTML docs:

It is built into docs.rs:

<img width="412" alt="image" src="https://github.com/elixir-lang/ex_doc/assets/1106188/3fecfefc-4269-469c-be11-66bba47fea7c">

And is a fairly common sight elsewhere:

<img width="483" alt="image" src="https://github.com/elixir-lang/ex_doc/assets/1106188/d01b317a-3079-47dc-bef0-bbe1c2ef1f2f">

I'll raise a PR! 🙂

josevalim:
Closing in favor of the PR!

PR #1918, opened 2024-06-10
Add links to jump to latest version of docs

pdavies:
Hello! 👋 

This PR addresses #1917 . It adds support for specifying which version is the latest via docs_config.js:

```javascript
var versionNodes = [
    { version: 'v1.0.0', url: '/foo'},
    { version: 'v2.0.0', url: "/bar", latest: true },
    { version: 'v0.34.0-dev', url: "/baz" }
];
```

 which will trigger a warning hyperlink to the latest version when viewing other versions:

image

It is not required to have a latest version in docs_config.js. Existing docs_config.js files will continue to produce the existing behaviour with no warning links.

Rather than try to determine which version is the latest through javascript logic, it seemed more appropriate to delegate to the author of docs_config.js, for example hexdocs. This avoids the risk of behaviour that differs from the documentation host's opinion on which version is the latest. This does mean that the PR doesn't do anything interesting without a small enhancement to (in practice) hexdocs.

This PR has the drawback that docs published prior to this PR will not get warnings. But due to the way exdocs works, I don't see how any attempt at fixing #1917 could get around this.

### DIFF
--- a/assets/js/handlebars/templates/versions-dropdown.handlebars
+++ b/assets/js/handlebars/templates/versions-dropdown.handlebars
@@ -10,7 +10,8 @@
       {{/each}}
     </select>
     {{#if jumpToLatestUrl }}
-      <span class="sidebar-projectVersionsLatestLink"><a href="{{jumpToLatestUrl}}">⚠️ Go to latest</a></span>
+      <br />
+      <span class="sidebar-projectVersionsLatestLink"><a href="{{jumpToLatestUrl}}">Go to latest</a></span>
     {{/if}}
   </label>
 </form>
--- a/assets/js/handlebars/templates/versions-dropdown.handlebars
+++ b/assets/js/handlebars/templates/versions-dropdown.handlebars
@@ -9,5 +9,8 @@
         </option>
       {{/each}}
     </select>
+    {{#if jumpToLatestUrl }}
+      <span class="sidebar-projectVersionsLatestLink"><a href="{{jumpToLatestUrl}}">⚠️ Go to latest</a></span>
+    {{/if}}
   </label>
 </form>
--- a/assets/js/handlebars/templates/versions-dropdown.handlebars
+++ b/assets/js/handlebars/templates/versions-dropdown.handlebars
@@ -9,9 +9,10 @@
         </option>
       {{/each}}
     </select>
-    {{#if jumpToLatestUrl }}
+    {{#if latestUrl }}
       <br />
-      <span class="sidebar-projectVersionsLatestLink"><a href="{{jumpToLatestUrl}}">Go to latest</a></span>
+      <span class="sidebar-projectVersionsLatestLink"><a href="{{latestUrl}}">View latest
+          ({{latestVersion}})</a></span>
     {{/if}}
   </label>
 </form>
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -110,8 +110,15 @@
 }
 
 .sidebar .sidebar-projectVersionsLatestLink {
-  background-color: red;
-  color: white;
+  border-left: var(--navTabBorderWidth) solid var(--sidebarLanguageAccentBar);
+  padding: 0 10px;
+  background-color: var(--sidebarAccentMain);
+}
+
+.sidebar .sidebar-projectVersionsLatestLink a {
+  color: var(--sidebarBackground);
+  font-weight: bold;
+  ;
 }
 
 .sidebar .sidebar-listNav {
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -109,6 +109,11 @@
   display: none;
 }
 
+.sidebar .sidebar-projectVersionsLatestLink {
+  background-color: red;
+  color: white;
+}
+
 .sidebar .sidebar-listNav {
   display: flex;
   margin: 0;
--- a/assets/js/sidebar/sidebar-version-select.js
+++ b/assets/js/sidebar/sidebar-version-select.js
@@ -14,16 +14,19 @@ export function initialize () {
     const versionsContainer = qs(VERSIONS_CONTAINER_SELECTOR)
     // Initially the container contains only text with the current version
     const currentVersion = versionsContainer.textContent.trim()
+    const latestVersionNode = versionNodes.find(node => node.latest)
+    const jumpToLatestUrl = latestVersionNode?.version !== currentVersion ? latestVersionNode?.url : null
+
     const nodes = decorateVersionNodes(versionNodes, currentVersion)
 
-    renderVersionsDropdown({ nodes })
+    renderVersionPicker(nodes, jumpToLatestUrl)
   }
 }
 
-function renderVersionsDropdown ({ nodes }) {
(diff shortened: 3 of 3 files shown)


josevalim:
This looks good to me but the indicator is calling too much attention. What if we keep only the warning sign? And then on mouse over it says this is not the latest version and you can click it to update?

halostatue:
@josevalim I think that it would be better to have the words than the warning sign for mobile users, since there is no "mouse over" capability there. I think the colours could definitely be muted, but from an accessibility point of view, words are better.

josevalim:
On mobile, we usually render the project name and version on every page, below the title, because the sidebar is not open. Shall we have the go to latest there instead (for mobile)?

halostatue:
I think so, but it might be something worth rethinking the render, because having that as a parenthetical next to a parenthetical is odd.

mobile-ish rendering of Oban docs

pdavies:
Thanks for the feedback!

Visuals: I agree the white on red is unpleasant and should be toned down. It also did not play nicely with light/dark theming. However there is a balance to strike - if it isn't eyecatching, it isn't useful. I've revised the appearance to be theme-aware and less painful on the eyes. It still isn't _beautiful_ but I'd need to defer to someone with better frontend design skills to improve it any further!

Mobile view: I hadn't given any thought to this. My preference would be to tackle that separately - more honestly for _someone else_ to tackle that separately. This is mainly because I'm not sure how I would get the necessary information to module_template.eex in a nice way.

image

image

assets/js/handlebars/templates/versions-dropdown.handlebars
```diff
@@ -9,5 +9,9 @@
         </option>
       {{/each}}
     </select>
+    {{#if jumpToLatestUrl }}
+      <br />
+      <span class="sidebar-projectVersionsLatestLink"><a href="{{jumpToLatestUrl}}">Go to latest</a></span>
```
halostatue, line 14: The Rust docs say "go to latest version", not "go to latest"; I realize we have limited space because this is in the sidebar, but I wonder if it might instead be possible to say something like `View latest (v{{latestVersion}})`?

pdavies:
Hello there, this PR slipped off my radar but I was reminded of it by this tweet: https://x.com/jskalc/status/1816920900660236370

I've pushed a change to address @halostatue's suggestion. Anything more you need from me here? Keen to get it in if possible!

MurdoqZodoq (approved):

josevalim:
Hi @pdavies, I believe the only blocker is making this fit visually. Perhaps just an alert icon after the version selector is enough? On mouse over we could say "Go to latest version", which will be done on click? Unsure.

josevalim:
Someone dropped a suggestion on the forum which I quite like:

https://elixirforum.com/t/search-engines-return-old-library-versions/58873/10

image

image

What do you think? It is quite subtle but probably enough to portray the meaning. Then the person can click on the select to go to latest. So no additional buttons needed.

zachdaniel:
Great idea 👌👌👌

wojtekmach:
Can we go regular color for latest, otherwise orange, i.e. no green? This would be subjectively less busy visually. The downside would be for docs generated with prior ExDoc versions there would be no difference but less of an issue in the long term.

halostatue:
@josevalim The problem with the colour approach is red-green colour blindness. I think that some sort of iconography or plain text is going to be best and most accessible.

josevalim:
@halostatue what if we just show yellow/orange if outdated? No need to show it on green if latest.

I am also open to using icons, if there is a good suggestion. But if we don't have one such proposal for now, my suggestion is to ship this and we continue refining it in the future (adding icons, notifications, etc).

josevalim:
@pdavies, are you ok with someone else carrying this over the finish line or do you want some time? It is your call, it was pending on the ExDoc team decision for sometime anyway. Thanks!

halostatue:
Simulated deuteranopia

CleanShot 2025-01-07 at 21 36 04@2x

Simulated Deuteranomaly

CleanShot 2025-01-07 at 21 36 56@2x

Simullated Protanopia

CleanShot 2025-01-07 at 21 37 40@2x

Simulated Protanomaly

CleanShot 2025-01-07 at 21 38 14@2x

Simulated Tritanopia

CleanShot 2025-01-07 at 21 39 00@2x

Simulated Tritanopoly

CleanShot 2025-01-07 at 21 39 34@2x

Simulated Monochromacy

CleanShot 2025-01-07 at 21 40 02@2x

Simulated Partial Monochromacy

CleanShot 2025-01-07 at 21 40 43@2x

All of this is with SimDaltonism: https://github.com/michelf/sim-daltonism for macOS.

(I am not colour blind myself, but I have a friend who is.)

josevalim:
Thank you, those are helpful. Given we only plan to highlight it when it is outdated, we can compare the color of the version with the “Elixir” title in the first example, and it seems it is noticeable in all images. So we can move forward.

although I also I think the color is too subtle overall, even in regular cases without the simulation, so we should definitely have additional mechanisms or icons. Suggestions and mockups are definitely welcome, so we iterate until we find something nice looking and useful.

halostatue:
> Thank you, those are helpful. Given we only plan to highlight it when it is outdated, we can compare the color of the version with the “Elixir” title in the first example, and it seems it is noticeable in all images. So we can move forward.

I think something like this might work:

⏷ v1.17.3
⏷ v1.17.2 ⚠️

is likely to be a better choice — if the warning sign is a [popover](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API/Using) so that the *meaning* of that sign can be tapped/clicked (in either mobile or desktop) to see the meaning (something like `This is not the current version of the documentation.`).

I *do* worry that it's going to be too subtle, and _also_ think that it would be good to show the version on desktop like we do on mobile *if* the sidebar is hidden. We could then *also* use the same hooks that the dropdown is using to append the ⚠️ popover activator to present the same message without showing the sidebar.

CleanShot 2025-01-07 at 23 37 38@2x

halostatue:
The warning sign looks like it's going to be pretty stable in terms of display size and style across operating systems and devices, per [emojipedia](https://emojipedia.org/warning#designs) (scrolling required to see the options).

josevalim:
We should not use the emoji IMO because it is too dissonant. We don’t use them anywhere else in ExDoc UI. We do use Remix icons though, and perhaps the warning one can be used. I believe it is already in the build as we use it for admonition.

halostatue:
Looking at ones that suggest a warning, I think that `spam`, `alarm-warning`, and `thunderstorms` are the best choices — and they may not be used, but I don't think that's a negative here because this should be *seen*.

CleanShot 2025-01-08 at 00 35 27@2x

josevalim:
Or "spam" or the one we already use:

Screenshot 2025-01-08 at 08 05 51

josevalim:
We already have alert-line as part of the bundle, so we can probably just use it, and it looks like the emoji:

Screenshot 2025-01-08 at 08 08 03

josevalim:
Thank you. I think we can make the version number and the remix icon colored in orange/yellow when outdated. At least have that as a starting point. @alisinabh, would you like to pick it up?

alisinabh:
@josevalim yes sure. Would be happy to do it.

PR #1929, opened 2024-07-02
Add top margin to nested module prefix in sidebar

zachallaun:
I was browsing some docs earlier that used the `:nest_modules_by_prefix` config and found that, with a very long module list, it was very difficult to see the prefixes and almost impossible to quickly scan for (e.g. while scrolling down to find the next section).

This PR just adds 10px of margin above those prefixes, which is sufficient to visually set them apart. (10px is chosen because it's the same amount used when you expand a module subheader in the sidebar.)

### Before

image

### After

image

### DIFF
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -245,6 +245,7 @@
   font-size: 0.9em;
   line-height: 1.8em;
   color: var(--sidebarSubheadings);
+  margin-top: 10px;
   padding-left: 15px;
 }
 


assets/css/sidebar.css
```diff
@@ -245,6 +245,7 @@
   font-size: 0.9em;
   line-height: 1.8em;
   color: var(--sidebarSubheadings);
+  margin-top: 10px;
```
josevalim, line 248: The grouping functionality uses 1.5em. Should we use the same?

```suggestion
  margin-top: 1.5em;
```

zachallaun: I'm find with that if you'd prefer. I personally think it errs on the side of too much margin, but I don't feel strongly. Here's what that looks like:

image

zachallaun: Looking at the grouping functionality: that font size is `.8em` whereas the module nesting uses `.9em`, so `1.5em` margin ends up being a bit larger for module nesting. To make them consistent, we could update module nesting to use `.8em` and `1.5em` margin.

zachallaun: (Sorry for the repeated messages.) Looking at it now, bumping the nested modules down to `.8em` seems a bit too small. The other grouping uses ALL CAPS, so it's more readable at that smaller size.

I'll ultimately defer to you.

josevalim:

PR #1937, opened 2024-08-10
Switch anchor `title` to `aria-label`

### LINKED ISSUE #1935: Screen reader text obscures search
There are a few spans that use the `.sr-only` class to hide text that's still readable by screen readers. When searching, it seems that browsers still find these.

A simple reproduction: Go to https://hexdocs.pm/ex_doc/ExDoc.Markdown.html, Ctrl/Cmd-F, type "link". The word "link" doesn't appear in the page except for the five links to the two callbacks and three functions in that module.

This isn't a huge deal, and if it helps with screen readers, it's probably worth keeping it as is. I think there's a way to basically hide the text from what the browser search sees though, I just can't remember how right now.

harrisi:
Fixes #1935.

I thought it would be more work, but this seems to be all that's required.

### DIFF
--- a/lib/ex_doc/formatter/html/templates/detail_template.eex
+++ b/lib/ex_doc/formatter/html/templates/detail_template.eex
@@ -3,15 +3,13 @@
     <span id="<%=enc "#{default_name}/#{default_arity}" %>"></span>
   <% end %>
   <div class="detail-header">
-    <a href="#<%=enc node.id %>" class="detail-link" data-no-tooltip title="Link to this <%= pretty_type(node) %>">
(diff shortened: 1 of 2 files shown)


halostatue:
I am *not* an accessibility expert, but I think that in this case it may be desirable to have both the `title` and the `aria-label` with the same value because different screen readers will do different things.

harrisi:
My reasoning for not having both is due to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns) and, actually, the [HTML specification itself](https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute). As a fellow non-accessibility expert, I was kind of surprised but this, but it seems like the global `title` attribute is discouraged in general.

josevalim:

PR #1938, opened 2024-08-10
Convert admonition blockquotes to sections for screen reader users

### LINKED ISSUE #1939: Admonition blockquotes should be converted to divs
For accessibility reasons, admonition blockquotes should be converted to divs.

voughtdq:
Some screen readers will announce that an element is a blockquote. It could be potentially surprising to hear something announced as a blockquote despite the inner contents not being a quote.

References:
 - https://adrianroselli.com/2023/07/blockquotes-in-screen-readers.html#Example01
 - https://v5.chriskrycho.com/journal/psa-on-blockquotes/

Closes #1939

### DIFF
--- a/assets/css/content/admonition.css
+++ b/assets/css/content/admonition.css
@@ -23,14 +23,14 @@
   background-color: var(--tipBackground);
 }
 
-.content-inner section.admonition > :first-child {
+.content-inner section.admonition > .admonition-title {
   color: var(--contrast);
   margin: 0 -1.2rem;
   padding: .7rem 1.2rem .7rem 3.3rem;
   font-weight: 700;
   font-style: normal;
 }
-.content-inner section.admonition > :first-child::before {
+.content-inner section.admonition > .admonition-title::before {
   color: var(--contrast);
   position: absolute;
   left: 1rem;
@@ -41,52 +41,52 @@
   -moz-osx-font-smoothing: grayscale;
 }
 
-.content-inner section.admonition > :first-child.warning {
+.content-inner section.admonition > .admonition-title.warning {
   background-color: var(--warningHeadingBackground);
   color: var(--warningHeading);
 }
-.content-inner section.admonition > :first-child.warning::before {
+.content-inner section.admonition > .admonition-title.warning::before {
   content: var(--icon-error-warning);
   color: var(--warningHeading);
 }
 
-.content-inner section.admonition > :first-child.error {
+.content-inner section.admonition > .admonition-title.error {
   background-color: var(--errorHeadingBackground);
   color: var(--errorHeading);
 }
-.content-inner section.admonition > :first-child.error::before {
+.content-inner section.admonition > .admonition-title.error::before {
   content: var(--icon-error-warning);
   color: var(--errorHeading);
 }
 
-.content-inner section.admonition > :first-child.info {
+.content-inner section.admonition > .admonition-title.info {
   background-color: var(--infoHeadingBackground);
   color: var(--infoHeading);
 }
-.content-inner section.admonition > :first-child.info::before {
+.content-inner section.admonition > .admonition-title.info::before {
   content: var(--icon-information);
   color: var(--infoHeading);
 }
 
-.content-inner section.admonition > :first-child.neutral {
+.content-inner section.admonition > .admonition-title.neutral {
   background-color: var(--neutralHeadingBackground);
   color: var(--neutralHeading);
 }
-.content-inner section.admonition > :first-child.neutral::before {
+.content-inner section.admonition > .admonition-title.neutral::before {
   content: var(--icon-double-quotes-l);
   color: var(--neutralHeading);
 }
 
-.content-inner section.admonition > :first-child.tip {
+.content-inner section.admonition > .admonition-title.tip {
   background-color: var(--tipHeadingBackground);
   color: var(--tipHeading);
 }
-.content-inner section.admonition > :first-child.tip::before {
+.content-inner section.admonition > .admonition-title.tip::before {
   content: var(--icon-information);
   color: var(--tipHeading);
 }
 
-.content-inner section.admonition > :first-child code {
+.content-inner section.admonition > .admonition-title code {
   margin: 0 0.5ch;
 }
 
@@ -102,7 +102,7 @@
   color: var(--admCodeColor);
 }
 
-.content-inner section.admonition > :first-child :is(a, a:visited) {
+.content-inner section.admonition > .admonition-title :is(a, a:visited) {
   color: inherit;
   text-decoration-color: currentColor;
 }
@@ -116,7 +116,7 @@
     border-radius: 0;
   }
 
-  .content-inner section.admonition > :first-child {
+  .content-inner section.admonition > .admonition-title {
     margin: 0 calc(-1 * var(--content-gutter));
   }
 }
--- a/assets/css/content/admonition.css
+++ b/assets/css/content/admonition.css
@@ -1,36 +1,36 @@
-.content-inner div.admonition:is(.warning, .error, .info, .neutral, .tip) {
+.content-inner section.admonition:is(.warning, .error, .info, .neutral, .tip) {
   border-radius: 10px;
   border-left: 0;
 }
 
-.content-inner div.admonition.warning {
+.content-inner section.admonition.warning {
   background-color: var(--warningBackground);
 }
 
-.content-inner div.admonition.error {
+.content-inner section.admonition.error {
   background-color: var(--errorBackground);
 }
 
-.content-inner div.admonition.info {
+.content-inner section.admonition.info {
   background-color: var(--infoBackground);
 }
 
-.content-inner div.admonition.neutral {
+.content-inner section.admonition.neutral {
   background-color: var(--neutralBackground);
 }
 
-.content-inner div.admonition.tip {
+.content-inner section.admonition.tip {
   background-color: var(--tipBackground);
 }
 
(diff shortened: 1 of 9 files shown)


lib/ex_doc/markdown/earmark.ex
```diff
@@ -74,6 +76,44 @@ defmodule ExDoc.Markdown.Earmark do
...
+      |> then(fn
+        "" ->
+          ""
+
+        {"class", classes} ->
+          classes
+          |> String.split(" ")
+          |> Enum.filter(&(&1 in @admonition_classes))
+          |> Enum.join(" ")
+      end)
```
voughtdq, line 97: I tried to keep everything self-contained instead of adding helper functions, so that's why this is a little messy. The "coercion" to `""` is there since the call to `Enum.join/2` will return an empty string for an empty list.

test/ex_doc/markdown/earmark_test.exs
```diff
@@ -75,6 +75,43 @@ defmodule ExDoc.Markdown.EarmarkTest do
...
+        ], %{}}
+      ]
+
+      warning_error = """
+      > ### Warning! Error! {: .warning .error}
+      > A warning and an error.
+      """
+
+      assert Markdown.to_ast(warning_error, []) == [
+        {:div, [class: "error warning", role: "note"], [
```
voughtdq, line 108: I realized that Earmark doesn't reverse the attrs after they get processed. Probably need to create an issue in the Earmark repo about this.

lib/ex_doc/markdown/earmark.ex
```diff
@@ -74,6 +76,44 @@ defmodule ExDoc.Markdown.Earmark do
...
+
+    if h_admonition_classes != "" do
+      blockquote_attrs =
+        case Enum.split_with(blockquote_attrs, &match?({"class", _}, &1)) do
+          {[], attrs} ->
+            [{"class", h_admonition_classes}, {"role", "note"} | attrs]
```
voughtdq, line 103: I'm a little iffy on setting the `role` to `note` here. It would be nice if someone more familiar with aria roles could advise.

voughtdq:
Might not need [this javascript](https://github.com/elixir-lang/ex_doc/blob/v0.34.2/assets/js/content.js#L37-L45) anymore.

whatyouhide:
Would it make sense to turn this into a `<section>` instead? That's what the first blog post you linked to does.

voughtdq:
I don't see why not! I'll push some changes.

josevalim:
/cc @DavidOliver, who, iirc, worked on this feature. 

/cc @wojtekmach, who, iirc, worked on fixup.

@voughtdq can you please take a look at tabset.css and print.css? They also reference blockquotes and I think they should be addressed as well. There may be others.

assets/css/content/admonition.css
```diff
@@ -1,36 +1,36 @@
-.content-inner blockquote:is(.warning, .error, .info, .neutral, .tip) {
+.content-inner section.admonition:is(.warning, .error, .info, .neutral, .tip) {
```
josevalim, line 1: We can now simplify most of these rules. We no longer need `:is(...)`, since the section.admonition is enough to identify it.

josevalim:
> Might not need [this javascript](https://github.com/elixir-lang/ex_doc/blob/v0.34.2/assets/js/content.js#L37-L45) anymore.

That looks like the case as well. We should give it a try. Thank you!

assets/css/content/admonition.css
voughtdq: I don't have any objections to that, but it seems like the `:is` actually simplifies most of the selectors here. That is, removing `:is` for `.content-inner section.admonition:is(.warning, .error, .info, .neutral, .tip)` would result in

```css
.content-inner section.admonition.warning,
.content-inner section.admonition.error,
.content-inner section.admonition.info ...
```

Do you still want me to move forward with "de-`:is`ing"?

DavidOliver: @voughtdq, I think that for the rules that select all of the types, we can simplify to `.content-inner .admonition`?

voughtdq: Fixed that. I'm going to look closer at the selectors that use `:is` and see if some of them can be simplified.

DavidOliver: Okay. To clarify, whereas before we needed to include all the types of admonitions because there could be other types of blockquotes on the page, now that there is a section with a class of `admonition` I think selecting using that single class should be possible. Does that make sense?

josevalim: Yeah, to be clear, this should simply be: `content-inner section.admonition {`

josevalim: I would keep the section to avoid accidentally applying styles. Sorry for the confusion.

voughtdq:
I think we're good to go. @DavidOliver can you take a look?

assets/css/print.css
```diff
@@ -52,11 +52,11 @@
     display: none;
   }
 
-  .content-inner blockquote:is(.warning, .error, .info, .neutral, .tip) {
+  .content-inner section.admonition {
     border: 2px solid var(--gray400);
   }
 
-  .content-inner blockquote :is(h3, h4):is(.warning, .error, .info, .neutral, .tip) {
+  .content-inner section.admonition :is(h3, h4):is(.warning, .error, .info, .neutral, .tip) {
```
DavidOliver, line 59: `.content-inner section.admonition :is(h3, h4)` ?

DavidOliver:
@voughtdq, so far I've only looked through the latest CSS changes - I haven't tried out your branch yet - but I've only found one thing remaining, which I've commented on directly.

I would usually tend not to rely on `:first-child` and stick to `is:(h3, h4)`, to avoid being dependent on markup order, but I appreciate the headings will almost certainly be the first children and that the markup is controlled.

Thanks.

voughtdq:
(Saw your other comment about the selector I missed, pushed those changes - sorry about that).

I agree that using `:first-child` is not ideal. 

The problem I'm running into with `:is(h3, h4)` is that it will affect all child heading elements. So something like this:

```markdown
> ### Warning {: .warning}
> Don't feed your dog too much peanut butter.
>
> #### What about my cats?
> They don't really like peanut butter.
```

will end up having the second `h4` given the styles applied to `:is(h3, h4)`. It can be discriminated further with `> :is(h3, h4):first-child`, but at that point, since we already have control of the markup, `> :first-child` is basically equivalent.

In this case, I think it's pragmatic to use `> :first-child` to avoid applying styles to subsequent heading elements. There's a real possibility that someone might want additional headings in their admonition. 

That said, I never have strong opinions on these things, so I'm open to alternatives.

The alternative I can think of is adding an additional class (*e.g.* `.admonition`) to the heading element that gets an admonition class and then select on `section.admonition .admonition`.

DavidOliver:
@voughtdq, I see. Thanks!

josevalim:
@voughtdq given that we control the markup, we can adjust the markup by adding a .header or title to the h3/h4. As you prefer. Whatever leads to the best combination of code + CSS.

However, I am remembering that we explicitly chose to not go down this path. We didn't want to change the generated Markdown because that would further couple us to the underlying markdown engine and we didn't want to have a subset/superset of Markdown. So it was very important that our rules were expressed with only CSS. Perhaps it is fine to deviate at this point, but I wanted to write down why we didn't go with this approach early on.

voughtdq:
I took some time to consult how other markdown/documentation implementations do it. There is a consensus of setting a class for the admonition title, but not on what elements/classes to use. I see a lot of implementations that use `.admonition-title` with a few others using less common names. I've decided to go ahead and add the `.admonition-title` class.

There is *a lot* of fragmentation in markdown admonition syntax and then further fragmentation in what HTML is generated for it. I anticipate that any new engines added, markdown or otherwise, will require transformation of the admonition markup simply because there's no true consensus or standard.

lib/ex_doc/markdown/earmark.ex
```diff
@@ -74,6 +76,46 @@ defmodule ExDoc.Markdown.Earmark do
     "$$\n#{content}\n$$"
   end
 
+  # Convert admonition blockquotes to sections for screen reader accessibility
+  defp fixup(
+         {"blockquote", blockquote_attrs, [{tag, h_attrs, h_content, h_meta} | rest] = ast,
+          blockquote_meta}
+       )
+       when tag in ["h3", "h4"] do
+    {h_classes, h_attrs} = Enum.split_with(h_attrs, &match?({"class", _}, &1))
```
josevalim, line 85: You can also use List.keytake and similar. :) Similar below.

josevalim:
@voughtdq I dropped one minor nitpick. Also, can you please go ahead and delete the JavaScript function you linked to? https://github.com/elixir-lang/ex_doc/blob/v0.34.2/assets/js/content.js#L37-L45 Then I will merge it! Thank you!

lib/ex_doc/markdown/earmark.ex
voughtdq: Ok, I'm glad to hear that 😅
I think I can make it look better with keytake/keystore

josevalim: If you take, you probably don't need keystore, you can just prepend it: `[{"class", val} | attrs]`

voughtdq:
Working on that when I wake up. 

I have a clearer implementation with List.keyfind/List.keystore, but it also means different ordering of attrs in some cases. Is it ok to sort the attrs in the tests? I will go back to keytake/append if not ok.

josevalim:
Yes, it is ok!

josevalim:

PR #1940, opened 2024-08-12
Warn when trying to reference types in hidden modules

### LINKED ISSUE #1932: More explicit warning when type not visible because module is not visible
Recently I was working on one of my libraries (https://github.com/axelson/password-validator commit 4bb73) and I received the following warning:
```
> mix docs
Generating docs...
    warning: documentation references type "PasswordValidator.Validators.CharacterSetValidator.Config.t()" but it is hidden or private
    │
 79 │     def validate_character_set(
    │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    │
    └─ (password_validator 0.5.1) lib/password_validator/validators/character_set_validator.ex:79: PasswordValidator.Validators.CharacterSetValidator.validate_character_set/3
```

I was initially surprised because IEx the type exists and is public (i.e. `@type` instead of `@typep`) and IEx finds the type just fine:
```elixir
iex(2)> t PasswordValidator.Validators.CharacterSetValidator.Config.t
@type t() :: %PasswordValidator.Validators.CharacterSetValidator.Config{
        allowed_special_characters: :all | String.t() | nil,
        custom_messages: any(),
        lower_case: [:infinity | non_neg_integer()],
        numbers: [:infinity | non_neg_integer()],
        special: [:infinity | non_neg_integer()],
        upper_case: [:infinity | non_neg_integer()]
      }
```

After a good bit of head scratching I finally realized that the issue was that the _module_ (not the type) was hidden via `@moduledoc false`

Coming back to the error message "documentation references type "PasswordValidator.Validators.CharacterSetValidator.Config.t()" but it is hidden or private" I think I thought that the `it` in the sentence was referring to the type, but it is really referring to the module, or maybe the message isn't explicit/obvious that the type isn't visible because the module isn't visible.

My suggested updated warning message would be "warning: documentation references type "PasswordValidator.Validators.CharacterSetValidator.Config.t()" but the module `PasswordValidator.Validators.CharacterSetValidator.Config` is hidden or private".

voughtdq:
Closes #1932
When a module is hidden, its types will also be hidden. We check for two conditions and warn the user if both are met:
 - The module defining the type is hidden
 - The type itself is hidden If the type is undefined, the standard warning is displayed instead

It does not seem possible to do this check for Erlang modules.

### DIFF
--- a/test/ex_doc/language/elixir_test.exs
+++ b/test/ex_doc/language/elixir_test.exs
@@ -500,7 +500,9 @@ defmodule ExDoc.Language.ElixirTest do
     ExDoc.Refs.insert([
       {{:module, AutolinkTest.Foo}, :public},
       {{:function, AutolinkTest.Foo, :bar, 1}, :hidden},
-      {{:type, AutolinkTest.Foo, :bad, 0}, :hidden}
+      {{:type, AutolinkTest.Foo, :bad, 0}, :hidden},
+      {{:module, AutoLinkTest.Hidden}, :hidden},
+      {{:type, AutoLinkTest.Hidden, :my_type, 0}, :hidden}
     ])
 
     opts = [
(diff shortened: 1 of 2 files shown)


josevalim (approved):
This looks good to me but I have a feeling this was tried before and we reverted it for some reason. I will let @wojtekmach merge it accordingly.

wojtekmach:
@josevalim we were already warning under these circumstances. This PR improves the error message!

@voughtdq thank you!

test/ex_doc/language/elixir_test.exs
```diff
@@ -500,7 +500,9 @@ defmodule ExDoc.Language.ElixirTest do
     ExDoc.Refs.insert([
       {{:module, AutolinkTest.Foo}, :public},
       {{:function, AutolinkTest.Foo, :bar, 1}, :hidden},
-      {{:type, AutolinkTest.Foo, :bad, 0}, :hidden}
+      {{:type, AutolinkTest.Foo, :bad, 0}, :hidden},
+      {{:module, AutoLinkTest.Hidden}, :hidden},
+      {{:type, AutoLinkTest.Hidden, :my_type, 0}, :hidden}
```
wojtekmach, line 505: note to self: The testing setup is not fantastic. I don't know off the top of my head whether _here_ :hidden is supposed to reflect the type is hidden or the type is (so to speak) public but because it is in hidden module it ends up hidden. This is an implementation detail.

A more robust way to test this would be more end-to-end:

image

PR #1947, opened 2024-08-30
Support modules with encrypted debug info

voughtdq:
Addresses #1928.

- Add configuration options to support decrypting debug info
- Update documentation with new options, examples, and advice
- Update test helper so that debug info options are forwarded to the compiler
- Add basic tests
- Update changelog

### DIFF
--- a/lib/mix/tasks/docs.ex
+++ b/lib/mix/tasks/docs.ex
@@ -213,64 +213,10 @@ defmodule Mix.Tasks.Docs do
   ## Encrypted debug info
 
   If a module is compiled with [encrypted debug info](`:compile.file/2`), ExDoc will not be able to
-  extract its documentation without first setting a decryption function or utilizing
-  `.erlang.crypt` as prescribed by `m::beam_lib#module-encrypted-debug-information`. Two
-  convenience options (see below) are provided to avoid having to call `:beam_lib.crypto_key_fun/1`
-  out-of-band and/or to avoid using `.erlang.crypt`.
-
-  If you prefer to set the key out-of-band, follow the instructions provided in the
-  `m::beam_lib#module-encrypted-debug-information` module documentation.
-
-  > ### Key exposure {: .warning}
-  >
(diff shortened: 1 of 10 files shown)


lib/mix/tasks/docs.ex
```diff
@@ -200,6 +210,68 @@ defmodule Mix.Tasks.Docs do
       where path is either an relative path from the cwd, or an absolute path. The function
       must return the full URI as it should be placed in the documentation.
 
+  ## Encrypted debug info
+
+  If a module is compiled with [encrypted debug info](`:compile.file/2`), ExDoc will not be able to
+  extract its documentation without first setting a decryption function or utilizing
+  `.erlang.crypt` as prescribed by `m::beam_lib#module-encrypted-debug-information`. Two
+  convenience options (see below) are provided to avoid having to call `:beam_lib.crypto_key_fun/1`
+  out-of-band and/or to avoid using `.erlang.crypt`.
```
josevalim, line 219: This looks great, but let's support only the `.erlang.crypt` style for now. The reason for this is because erlang.config files cannot define functions (afaik), so it is not a feature they would be able to use anyway. Let's start with the minimum necessary and expand on that later. WDYT?

josevalim (commented):
Sorry for the delay, I was busy during ElixirConf. I dropped a comment, can you please take a look? Thank you!

lib/mix/tasks/docs.ex
voughtdq: To clarify, you want documentation about `.erlang.crypt`, but all the other changes should be reverted? Sounds good to me. 

Do you want me to add tests or is that not necessary? Adding a test for this could be invasive - either putting a `.erlang.crypt` file in the root project directory or messing with the `XDG_CONFIG_HOME` environment variable. I'm leaning toward no, what do you think?

voughtdq:
@josevalim hope you had fun 😁

I'll work on scaling this back and should have something for you within the next 48 hours.

voughtdq:
whoa ew
going to just do a new pull request with the changes

PR #1949, opened 2024-09-11
Document using .erlang.crypt with ExDoc

### LINKED ISSUE #1928: No support for debug_info keys?

voughtdq:
Closes #1928.

- Update documentation
- Update test helper so that debug info options are forwarded to the compiler
- Add basic tests
- Update changelog
- Use "CI" environment variable to skip the encrypted debug info test


test/ex_doc/retriever/erlang_test.exs
```diff
@@ -354,6 +354,61 @@ defmodule ExDoc.Retriever.ErlangTest do
              |> Erlang.autolink_spec(current_module: :mod, current_kfa: {:type, :type, 0}) ==
                "type() :: #a{a :: <a href=\"https://www.erlang.org/doc/apps/erts/erlang.html#t:pos_integer/0\">pos_integer</a>(), b :: <a href=\"https://www.erlang.org/doc/apps/erts/erlang.html#t:non_neg_integer/0\">non_neg_integer</a>(), c :: <a href=\"https://www.erlang.org/doc/apps/erts/erlang.html#t:atom/0\">atom</a>(), d :: <a href=\"https://www.erlang.org/doc/apps/erts/erlang.html#t:term/0\">term</a>(), e :: <a href=\"https://www.erlang.org/doc/apps/erts/erlang.html#t:term/0\">term</a>()}."
     end
+
+    @tag :ci
```
josevalim, line 358: We don’t want to tag it as CI. We probably want it to be called encrypted_info or similar and check if the encrypted info is enabled or not in the test helper. CI just makes sure to put the file in place. WDYT?

josevalim:
Ping :)

josevalim:
Closing this. Please let us know if there is still an interest in delivering this!

Issue #1951, opened 2024-09-17
Looking for a way to preserve old links


zachdaniel:
Hey folks! I'm looking for a way to preserve old links even as documentation changes. As our project evolves, many things change, pages move, modules are renamed, etc. For now, I've added a "this page has moved" extra for one of our extras that has moved that was popular before. However, this sort of thing can also happen for modules, mix tasks, etc.

Two proposals for how this might be resolved:

## Explicit `moved_pages` config

We add a configuration in `ex_doc` called `moved_pages`, which would create pages that are not displayed in the sidebar (and probably not in search?) that simply show a message that a given page has been moved, like so:

```elixir
moved_pages: %{
  "upgrade.html" => "upgrading-to-3.0.html"
}
```

Since this uses the html route, it can work for modules/mix tasks/anything else.

## Add the ability to hide pages in the sidebar

This would only work for extras, since we don't want to leave dead code in a project just for the sake of documentation purposes, but with extras that would be fine. So we could do something like:

```elixir
extras: [{"Extra", sidebar: false}]
```

Then if something moves, you can just leave the old page there but hidden in the sidebar.

josevalim:
We would love a redirects configuration that generates html pages with redirects. +1!

josevalim:
Probably without the extension, so we automatically generate it.

zachdaniel:
I will look into it :) I hadn't considered redirects TBH because I was thinking static sites didn't have a good way to do this, but I was wrong and there is metadata you can set to trigger a refresh 👍

zachdaniel:
Looks like we also already have the setup for redirects for `index.html` 🥳 this might be easy.

josevalim:
yes, we can probably even have `"index" => main` just be an entry that is automatically merged into redirect.

zachdaniel:
👍 perfect. PR incoming shortly.

PR #1952, opened 2024-09-17
improvement: add `redirects` config, and documentation explaining it

zachdaniel:
Proposed solution for #1951

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,18 +464,31 @@ You can improve the developer experience on everything but function names changi
 by using the `redirects` configuration. For example, if you changed the module `MyApp.MyModule`
 to `MyApp.My.Module` and the extra `get-started.md` to `quickstart.md`, you can
 setup the following redirects:
+<!-- tabs-open -->
+
+### Elixir
+
+For this example, we've changed the module `MyApp.MyModule` to `MyApp.My.Module`, and the extra `get-started.md` to `quickstart.md`
 
 ```elixir
-defp docs do
-  [
-    ...,
-    redirects: %{
-      MyApp.MyModule => MyApp.My.Module,
-      "get-started" => "quickstart"
-    }
-  ]
-end
+redirects: %{
+  "MyApp.MyModule" => "MyApp.My.Module",
+  "get-started" => "quickstart"
+}
 ```
+
+### Erlang
+
+For this example, we've changed the module `:my_module` to `:my_module2`, and the extra `get-started.md` to `quickstart.md`
+
+```erlang
+{redirects, [
+  {"my_module", "my_module2"},
+  {"get-started", "quickstart"}
+]}.
+```
+
+<!-- tabs-close -->
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -457,8 +457,8 @@ As your project grows, your documentation may very likely change, even structura
 - Links to *functions* are actually links to modules with anchor links. If you change the function name, the link does
   not break but will leave users at the top of the module's documentation.
 
-Because these docs are static files, when a user gets to a page that is not found, they will see a generic 404 page.
-They will not be redirected to your package's home page. This can potentially be jarring for users.
+Because these docs are static files, the behavior of a missing page will depend on where they are hosted.
+In particular, [hexdocs.pm](https://hexdocs.pm) will show a 404 page.
 
 With this in mind, it is a good idea to preserve links to your old documentation. We do this with the `redirects` configuration.
 This can solve for everything but the function names changing.
--- a/README.md
+++ b/README.md
@@ -448,6 +448,35 @@ Similarly to the example above, if your Markdown includes Mermaid graph specific
 
 For more details and configuration options, see the [Mermaid usage docs](https://mermaid-js.github.io/mermaid/#/usage).
 
+## Changing documentation over time
+
+As your project grows, your documentation may very likely change, even structurally. There are a few important things to consider in this regard:
+
+- Links to your *extras* will break if you change or move file names.
+- Links to your *modules, and mix tasks* will change if you change your code.
+- Links to *functions* are actually links to modules with anchor links. If you change the function name, the link does
+  not break but will leave users at the top of the module's documentation.
+
+Because these docs are static files, when a user gets to a page that is not found, they will see a generic 404 page.
+They will not be redirected to your packages home page. This can potentially be jarring for users.
+
+With this in mind, it is a good idea to preserve links to your old documentation. We do this with the `redirects` configuration.
+This can solve for everything but the function names changing.
+
+For this example, we've changed the module `MyApp.MyModule` to `MyApp.My.Module`, and the extra `get-started.md` to `quickstart.md`
+
+```elixir
+defp docs do
+  [
+    ...,
+    redirects: %{
+      MyApp.MyModule => MyApp.My.Module,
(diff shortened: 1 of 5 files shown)


lib/ex_doc/formatter/html.ex
```diff
@@ -390,6 +385,35 @@ defmodule ExDoc.Formatter.HTML do
...
+  end
+
+  defp stringify_redirect_item(item) when is_binary(item) do
+    item
+  end
+
+  defp stringify_redirect_item(item) when is_atom(item) do
+    inspected = inspect(item)
+
+    case to_string(item) do
```
zachdaniel, line 407: Is there a better way to inspect an atom that will leave off both `:` and `Elixir.`?

zachdaniel: Or should we just treat any atoms as module names? and do `inspect(item)`. I was just getting ahead of folks writing:

```elixir
redirects: %{
  :upgrade => :quickstart,
  Old.Module => New.Module
}
```

README.md
```diff
@@ -448,6 +448,35 @@ Similarly to the example above, if your Markdown includes Mermaid graph specific
...
+## Changing documentation over time
+
+As your project grows, your documentation may very likely change, even structurally. There are a few important things to consider in this regard:
+
+- Links to your *extras* will break if you change or move file names.
+- Links to your *modules, and mix tasks* will change if you change their name.
+- Links to *functions* are actually links to modules with anchor links. If you change the function name, the link does
+  not break but will leave users at the top of the module's documentation.
+
+Because these docs are static files, when a user gets to a page that is not found, they will see a generic 404 page.
```
zachdaniel, line 460: Can we fix this on `hexdocs.pm`? On a not found page for a package that actually exists, could we render a nicer message, or at least a link to the page?

Example not found page: https://hexdocs.pm/ash/something-random

CleanShot 2024-09-17 at 08 49 56

josevalim: You would have to ask them.

README.md
```diff
@@ -448,6 +448,35 @@ Similarly to the example above, if your Markdown includes Mermaid graph specific
...
+They will not be redirected to your package's home page. This can potentially be jarring for users.
```
josevalim, line 461:
```suggestion
Because these docs are static files, the behavior of a missing page will depend on where they are hosted.
In particular, [hexdocs.pm](https://hexdocs.pm) will show a 404 page.
```

README.md
```diff
@@ -448,6 +448,35 @@ Similarly to the example above, if your Markdown includes Mermaid graph specific
...
+
+With this in mind, it is a good idea to preserve links to your old documentation. We do this with the `redirects` configuration.
+This can solve for everything but the function names changing.
+
+For this example, we've changed the module `MyApp.MyModule` to `MyApp.My.Module`, and the extra `get-started.md` to `quickstart.md`
```
josevalim, line 466:
```suggestion
You can improve the developer experience on everything but function names changing
by using the `redirects` configuration. For example, if you changed the module `MyApp.MyModule`
to `MyApp.My.Module` and the extra `get-started.md` to `quickstart.md`, you can
setup the following redirects:
```

lib/ex_doc/formatter/html.ex
josevalim: There is likely already a function in this project that does that, but it may be specific for each supported language. In that case, our options are:

1. Do not support atoms
2. Support atoms but simply call `to_string` on it and yank the "Elixir." prefix

lib/ex_doc/formatter/html.ex
```diff
@@ -390,6 +384,36 @@ defmodule ExDoc.Formatter.HTML do
...
+      "Elixir." <> ^inspected -> inspected
+      other -> other
+    end
+  end
+
+  defp stringify_redirect_item(item) do
+    raise ArgumentError,
+          "Redirect source and destination must be a string or an atom, got: #{inspect(item)}"
```
josevalim, line 414: Error messages typically start in lowercase:

```suggestion
          "redirect source and destination must be a string or an atom, got: #{inspect(item)}"
```

mix.exs
```diff
@@ -107,6 +108,15 @@ defmodule ExDoc.Mixfile do
...
 
+  defp redirects(:dev) do
+    %{
+      "old-admonition" => "admonition",
+      Exdoc.OldMarkdown => ExDoc.Markdown
+    }
+  end
+
+  defp redirects(_), do: %{}
+
```
josevalim, line 119: We should not add those here and instead have a proper test. :)

lib/ex_doc/formatter/html.ex
zachdaniel: 👍 that was the other alternative I considered, but thought there could be pages that start with `Elixir.`. So perhaps just only supporting strings is the better answer?

README.md
```diff
@@ -448,6 +448,35 @@ Similarly to the example above, if your Markdown includes Mermaid graph specific
...
+
+Because these docs are static files, the behavior of a missing page will depend on where they are hosted.
+In particular, [hexdocs.pm](https://hexdocs.pm) will show a 404 page.
+
+You can improve the developer experience on everything but function names changing
+by using the `redirects` configuration. For example, if you changed the module `MyApp.MyModule`
+to `MyApp.My.Module` and the extra `get-started.md` to `quickstart.md`, you can
+setup the following redirects:
+
+```elixir
```
josevalim, line 468: The docs in the README typically have examples for both Erlang/Elixir, so we would need to do that. Alternatively, you can move this the `mix docs` docs.

mix.exs
zachdaniel: Oops 🤦

README.md
zachdaniel: Made it support lists of tuples in addition to maps, and used that syntax in the Erlang example.

mix.exs
zachdaniel: Test added, removed from here.

lib/ex_doc/formatter/html.ex
zachdaniel: It no longer supports atoms.

lib/ex_doc/formatter/html.ex
```diff
@@ -390,6 +384,24 @@ defmodule ExDoc.Formatter.HTML do
     |> Enum.sort_by(fn extra -> GroupMatcher.group_index(groups, extra.group) end)
   end
 
+  def generate_redirects(config, ext) do
+    config.redirects
+    |> Map.put_new("index", config.main)
```
josevalim, line 389: This will fail if we give it a list, so we need to add a test, replace `Enum.map` by `Map.new`, and call it at the end.

zachdaniel: Ah, sorry about that, fixed  and tested :)

josevalim:
Excellent turn around here, thank you!

garazdawi:
Excellent feature! I'm a bit late to the party, but would it not be good to add this option to the list of options in https://github.com/elixir-lang/ex_doc/blob/main/lib/mix/tasks/docs.ex? So that we have a common list for all options available?

zachdaniel:
Will add when I'm at a computer 👍👍

zachdaniel:
PR here: https://github.com/elixir-lang/ex_doc/pull/1953/files

PR #1960, opened 2024-10-31
allow searching for atoms inside backticks

SteffenDE:
In the LV repo you were not able to search for atoms inside backticks, for example when searching for "validate_attrs" the corresponding option of the `Phoenix.Component.slot/3` macro could not be found:

image

This commit fixes this by stripping the backticks from tokens and also removing trailing colons. I'm not familiar with the ex_doc codebase at all, so this may very well NOT be the way to properly fix, so please double check.

After this change:

image

cc @josevalim

### DIFF
--- a/assets/js/search-page.js
+++ b/assets/js/search-page.js
@@ -196,7 +196,7 @@ function docTokenFunction (token) {
     tokens.push(token.clone().update(() => toSplitWords))
   } else if (toSplitWords.startsWith(':')) {
     // allow searching for atoms without `:`
-    toSplitWords = toSplitWords.slice(1)
+    toSplitWords = toSplitWords.substring(1)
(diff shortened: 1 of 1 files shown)


assets/js/search-page.js
```diff
@@ -188,6 +194,10 @@ function docTokenFunction (token) {
     // also make it searchable as foo_bar
     toSplitWords = toSplitWords.substring(1)
     tokens.push(token.clone().update(() => toSplitWords))
+  } else if (toSplitWords.startsWith(':')) {
+    // allow searching for atoms without `:`
+    toSplitWords = toSplitWords.slice(1)
```
josevalim, line 199:
```suggestion
    toSplitWords = toSplitWords.substring(1)
```

josevalim:

PR #1963, opened 2024-11-19
Add support for nominal types

lucioleKi:
This PR adds support for `-nominal` to be used in the same way as `-type` in documentation. Without this change, OTP's CI can't build documentation for nominal types.

Related PR in OTP: [PR-9079](https://github.com/erlang/otp/pull/9079).


josevalim:
It was my bad, I was too quick on merging. We just need to tag the tests, I will push a fix soon :)

josevalim:
Pushed!

PR #1965, opened 2024-11-20
Fix for missing nominal case in autolink_spec

lucioleKi:
Small fix for missing nominal case in autolink_spec.


josevalim:
Thank you! Is there a test we could add here?

lucioleKi:
What didn't match was something like this. It crashed when I build docs for OTP. I have difficulty writing a test that works at this moment, perhaps due to my limited knowledge about the syntax. I'm not sure how to build a term like this in the test. 
`{:attribute, [file: "/otp/lib/stdlib/src/erl_anno.erl", location: {122, 2}], :nominal, {:line, {:type, {122, 20}, :non_neg_integer, []}, []}}` 

After the nominal PR is merged, the following test could be added at line 578 `test/language/erlang_test.exs` to test for integration:

```    
test "OTP nominal", c do
      assert autolink_extra("`t:erl_anno:location/0`", c) ==
               ~s|<a href="https://www.erlang.org/doc/apps/stdlib/erl_anno.html#t:location/0"><code class="inline">erl_anno:location/0</code></a>|
```

josevalim:
You can add the test today and wrap it in a conditional:

```
if System.otp_release() >= "28" do
  test "OTP nominal", c do
      assert autolink_extra("`t:erl_anno:location/0`", c) ==
               ~s|<a href="https://www.erlang.org/doc/apps/stdlib/erl_anno.html#t:location/0"><code class="inline">erl_anno:location/0</code></a>|
  end
end
```

WDYT?

garazdawi:
From what I understand this PR fixes when creating docs for code like this:

```erlang
-doc "abd".
-nominal foo() :: file:name().
```

So the testcase needs to compile that kind of erlang code for it to work. I would look how the tests for spec works, like this one: 

```elixir
  test "type", c do
      assert autolink_spec("-type foo() :: t().", c) ==
               ~s|foo() :: <a href="#t:t/0">t</a>().|
    end
```

https://github.com/elixir-lang/ex_doc/blob/v0.35.0/test/ex_doc/language/erlang_test.exs#L716C5-L719C8

lucioleKi:
It turns out that existing test cases are sufficient. I can test the new changes locally with OTP 28. Once the CI here is updated to test for 28, it will also show up here.

josevalim:

PR #1968, opened 2024-11-21
Edoc still includes private functions so need to remove them

garazdawi:
In cf49545 the pruning of edoc function was removed which causes more functions than there should be to be part of the generated docs.

This commit restores the old behaviour and adds a testcase.


josevalim:

josevalim:
v0.35.1 is out.

PR #1971, opened 2024-11-24
Make headings selectable

### LINKED ISSUE #1969: Unable to select heading text
There's a bug in the HTML docs which makes it is impossible to select the contents of a heading.

Here's a screen-cast that shows the issue:

https://github.com/user-attachments/assets/16cb488e-f190-49c9-9bdd-b747db116002

ExDoc version: latest

josevalim:
Closes #1969.

### DIFF
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -186,14 +186,6 @@
   opacity: 1;
 }
 
-/* Allow section link to be hovered and used “through” text */
-.content-inner .section-heading .text {
-  pointer-events: none;
-}
-.content-inner .section-heading .text a {
-  pointer-events: all;
-}
-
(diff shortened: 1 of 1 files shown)


josevalim:
@DavidOliver, this was added in https://github.com/elixir-lang/ex_doc/commit/3189db663d2e6ef07f8b2870fd96fe4f7073492d, so I am pinging you to see if you are happy with this revert. :)

josevalim:
Great catch!

> Shall we aim for the hover link to show when either the heading or the hover link is hovered, but to leave the heading selectable, requiring pointing to the hover link icon itself to use the fragment link? I can have a look at doing that if there is agreement.

Yes, please!

PR #1972, opened 2024-11-26
Improve content headings

### LINKED ISSUE #1969: Unable to select heading text

DavidOliver:
https://github.com/user-attachments/assets/dedb197e-fb52-44b9-9fa3-3e5626f314fe

> [!IMPORTANT]
> @josevalim, if you'd prefer the hover links to remain invisible until the heading is hovered/focussed, let me know and I'll revert that change.

Allows selection of heading text. Also makes hover link always (slightly) visible for discoverability.

Also improves:

- hover link colour
- hover link icon spacing
- heading spacing

Fixes #1969

### DIFF
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -39,7 +39,8 @@
 
 .content-inner h2 {
   font-size: 1.6em;
-  margin: 1em 0 0.5em;
+  padding-top: 1em;
+  margin-bottom: 0.5em;
   font-weight: 700;
 }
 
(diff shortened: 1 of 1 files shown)


josevalim:
Please let's keep it visible only on hover please.

And if we are going to change the color, can you please make sure it works with `mix docs --proglang erlang`? Generally speaking, I prefer to avoid accent, exactly because we need to make sure it works for both Erlang and Elixir, both on light and dark mode. So it is a bit too much. :)

PR #1973, opened 2024-11-26
Set smooth scroll behaviour

DavidOliver:
Just an idea. Merge/close as desired, of course. :)

https://github.com/user-attachments/assets/77034762-31fa-4bb9-9286-fdf26a84f86e


josevalim:
Thank you! We actually had this in Livebook and we received several comments saying it was jarring, and we reverted it there. So it is probably best to not enable it here. :)

PR #1976, opened 2024-12-08
add markdown formatter / exporter

mayel:
This is a quick proof of concept of https://elixirforum.com/t/generate-docs-markdown-similar-to-epub/67946

### DIFF
--- a/lib/ex_doc/cli.ex
+++ b/lib/ex_doc/cli.ex
@@ -103,7 +103,7 @@ defmodule ExDoc.CLI do
   defp normalize_formatters(opts) do
     formatters =
       case Keyword.get_values(opts, :formatter) do
-        [] -> opts[:formatters] || ["html", "epub"]
+        [] -> opts[:formatters] || ["html", "epub", "markdown"]
         values -> values
       end
 
--- a/lib/ex_doc/formatter/markdown.ex
+++ b/lib/ex_doc/formatter/markdown.ex
@@ -1,0 +1,126 @@
+defmodule ExDoc.Formatter.Markdown do
+  @moduledoc false
+
+  @mimetype "text/markdown"
+  @assets_dir "MD/assets"
+  alias __MODULE__.{Assets, Templates}
+  alias ExDoc.Formatter.HTML
+  alias ExDoc.Utils
+
+  @doc """
+  Generates Markdown documentation for the given modules.
+  """
+  @spec run([ExDoc.ModuleNode.t()], [ExDoc.ModuleNode.t()], ExDoc.Config.t()) :: String.t()
+  def run(project_nodes, filtered_modules, config) when is_map(config) do
+    Utils.unset_warned()
+
+    config = normalize_config(config)
+    File.rm_rf!(config.output)
+    File.mkdir_p!(Path.join(config.output, "MD"))
+
+    project_nodes =
+      HTML.render_all(project_nodes, filtered_modules, ".md", config, highlight_tag: "samp")
+
+    nodes_map = %{
+      modules: HTML.filter_list(:module, project_nodes),
+      tasks: HTML.filter_list(:task, project_nodes)
+    }
+
+    extras =
+      config
+      |> HTML.build_extras(".xhtml")
+      |> Enum.chunk_by(& &1.group)
+      |> Enum.map(&{hd(&1).group, &1})
+
+    config = %{config | extras: extras}
+
+    static_files = HTML.generate_assets("MD", default_assets(config), config)
+    HTML.generate_logo(@assets_dir, config)
+    HTML.generate_cover(@assets_dir, config)
+
+    # generate_nav(config, nodes_map)
+    generate_extras(config)
+    generate_list(config, nodes_map.modules)
+    generate_list(config, nodes_map.tasks)
+
+    {:ok, epub} = generate_zip(config.output)
+    File.rm_rf!(config.output)
+    Path.relative_to_cwd(epub)
+  end
+
+  defp normalize_config(config) do
+    output =
+      config.output
+      |> Path.expand()
+      |> Path.join("#{config.project}")
+
+    %{config | output: output}
+  end
+
+  defp generate_extras(config) do
+    for {_title, extras} <- config.extras do
+      Enum.each(extras, fn %{id: id, title: title, title_content: _title_content, source: content} ->
+        output = "#{config.output}/MD/#{id}.md"
+        content = """
+        # #{title}
+
+        #{content}
+        """
+
+        if File.regular?(output) do
+          Utils.warn("file #{Path.relative_to_cwd(output)} already exists", [])
+        end
+
+        File.write!(output, content)
+      end)
+    end
+  end
+
+
+
+
+  defp generate_list(config, nodes) do
+    nodes
+    |> Task.async_stream(&generate_module_page(&1, config), timeout: :infinity)
+    |> Enum.map(&elem(&1, 1))
+  end
+
+  defp generate_zip(output) do
+    :zip.create(
+      String.to_charlist("#{output}-markdown.zip"),
+      files_to_add(output),
+      compress: [
+        ~c".md",
+        ~c".jpg",
+        ~c".png"
+      ]
+    )
+  end
+
+  ## Helpers
+
+  defp default_assets(config) do
+    [
+      {Assets.dist(config.proglang), "MD/dist"},
+      {Assets.metainfo(), "META-INF"}
+    ]
+  end
+
+  defp files_to_add(path) do
+    Enum.reduce(Path.wildcard(Path.join(path, "**/*")), [], fn file, acc ->
+      case File.read(file) do
+        {:ok, bin} ->
+          [{file |> Path.relative_to(path) |> String.to_charlist(), bin} | acc]
+
+        {:error, _} ->
+          acc
+      end
+    end)
+  end
+
+  defp generate_module_page(module_node, config) do
+    content = Templates.module_page(config, module_node)
+    File.write("#{config.output}/MD/#{module_node.id}.md", content)
+  end
+
+end
--- a/lib/ex_doc/formatter/html/templates.ex
+++ b/lib/ex_doc/formatter/html/templates.ex
@@ -93,7 +93,7 @@ defmodule ExDoc.Formatter.HTML.Templates do
   defp presence([]), do: nil
   defp presence(other), do: other
 
-  defp enc(binary), do: URI.encode(binary)
+  def enc(binary), do: URI.encode(binary)
 
   @doc """
   Create a JS object which holds all the items displayed in the sidebar area
--- a/lib/ex_doc/formatter/markdown/assets.ex
+++ b/lib/ex_doc/formatter/markdown/assets.ex
@@ -1,0 +1,17 @@
+defmodule ExDoc.Formatter.Markdown.Assets do
+  @moduledoc false
+
+  defmacrop embed_pattern(pattern) do
+    ["formatters/markdown", pattern]
+    |> Path.join()
+    |> Path.wildcard()
+    |> Enum.map(fn path ->
+      Module.put_attribute(__CALLER__.module, :external_resource, path)
+      {Path.basename(path), File.read!(path)}
+    end)
+  end
+
+  def dist(_proglang), do: []
+
+  def metainfo, do: embed_pattern("metainfo/*")
+end
--- a/lib/ex_doc/formatter/markdown/templates.ex
+++ b/lib/ex_doc/formatter/markdown/templates.ex
@@ -1,0 +1,197 @@
+defmodule ExDoc.Formatter.Markdown.Templates do
+  @moduledoc false
+
+  require EEx
+
+  import ExDoc.Utils,
+    only: [before_closing_body_tag: 2, before_closing_head_tag: 2, h: 1, text_to_id: 1]
+
+  alias ExDoc.Formatter.HTML.Templates, as: H
+
+  @doc """
+  Generate content from the module template for a given `node`
+  """
+  def module_page(config, module_node) do
+    summary = H.module_summary(module_node)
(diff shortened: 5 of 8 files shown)


lib/ex_doc/cli.ex
```diff
@@ -103,7 +103,7 @@ defmodule ExDoc.CLI do
   defp normalize_formatters(opts) do
     formatters =
       case Keyword.get_values(opts, :formatter) do
-        [] -> opts[:formatters] || ["html", "epub"]
+        [] -> opts[:formatters] || ["html", "epub", "markdown"]
```
josevalim, line 106: I don't think we shuold enable it by default :)

mayel: Ah I misread and thought that was the allow-list of supported formatters

lib/ex_doc/formatter/markdown.ex
```diff
@@ -0,0 +1,126 @@
...
+
+
+
+  defp generate_list(config, nodes) do
+    nodes
+    |> Task.async_stream(&generate_module_page(&1, config), timeout: :infinity)
+    |> Enum.map(&elem(&1, 1))
+  end
+
+  defp generate_zip(output) do
```
josevalim, line 88: We should skip the zip as well.  The simplest we can genereate the best :)

josevalim:
Thank you @mayel, I think the general direction is good and I think we can continue exploring it. The testing structure will be very important too.

Just a heads up, we will be slow with reviews on our side, since we are focused on Elixir v1.18 and launching Livebook Teams.

mjrusso:
This is awesome @mayel! And thanks for letting me know about this PR :)

Having spent some time thinking about this, a few requirements suggestions for discussion/debate. My ideal Markdown export would generate:

- an exact mirror of existing HTML page structure, but in Markdown format (with .md extension for each file, and working hyperlinks between all .md files)
- at least one "single file" export with all documentation included in a single Markdown file
  - (there might be opportunities to generate a few different types of "single file" exports; [here's what hex2txt is currently doing](https://github.com/mjrusso/hex2txt/blob/2efe6d0582f593ba7d9d81a67f08369c82752723/lib/hex2txt/generator.ex#L23-L34), but it would probably make sense for exdoc to generate a version of the single-file export with all "extras" included)
- a top-level Markdown file that simply links to every generated Markdown file (like a sitemap, i.e. the intention behind the [llms.txt proposal](https://llmstxt.org))

I would also recommend generating all of this by default, so tooling can start to rely on these files existing :)

mayel:
@mjrusso Thanks for the feedback! 

The structure of the should already match that of the html docs.

In terms of a single file, that was the intention of the ZIP archive containing all the md docs, so it can easily be downloaded from hexdocs and devs can choose which files/modules they want to add as context rather than always including everything, but now I'm thinking we could add a cli flag to generate either a single file or a ZIP with seperate files, and leave the discussion of what the default should be for later?...

I've pushed some WIP I hadn't staged which includes generating an `index.md` with structured links to all the .md docs.

mjrusso:
> The structure of the files and markdown contents should already match that of the html docs.

Perfect :) I was mostly trying to enumerate my ideal requirements independent of what was already written, just for ease of debate.

> add a cli flag to generate either a single file or a ZIP with seperate files, and leave the discussion of what the default should be for later?...

In general my preference would be to make Markdown generation (in whatever form we decide) the default, with no other configuration options (other than disabling it if you really don't want it) so tools can rely on a common approach.

On the topic of the zip, single file generation, etc.:

> In terms of a single file, that was the intention of the ZIP archive containing all the md docs, so it can easily be downloaded from hexdocs and devs can choose which files/modules they want to add as context rather than always including everything, but now I'm thinking we could add a cli flag to generate either a single file or a ZIP with seperate files, and leave the discussion of what the default should be for later?...

Since we can already download a tarball of all docs from hex.pm (the md files, if generated, would be included by default there as well, correct?), I think we can forego the zip archive. Easy enough to get the markdown files from there. (And would these be fetched by default with `mix hex.docs`?)

Thinking through this a bit more, I think we could forego the single-file generation (at least for now). Realistically for AI tooling integration that works we are going to need a server in between that can manage pulling the right chunks of documentation for any given task. The individual md files being produced here provide the right building blocks.

(Also, instead of "Download Markdown version", perhaps "View Markdown version", which just links to the _index.md_ file.)

mayel:
Ah the downloadable docs from hex.pm had completely slipped me by! It may be useful to also include that link in the doc footers next to the ePub. 

And yeah makes sense to me, hoping I find some time to work on it a bit more soon :)

mayel:
> we can already download a tarball of all docs from hex.pm (the md files, if generated, would be included by default there as well, correct?)

The epub is included in that ZIP so I'm guessing so

mayel:
Alright I've experimented with updating the footer so it includes:

- Hex Package (if a hex package is set)
- View Code: 
  - Source Repo (if a source url is known)
  - Hex Preview (if a hex package is set)
- View Markdown version (if markdown formatter enabled)
- Download docs archive (from hex, should include html, markdown and epub) 
- Search HexDocs

Screenshot 2024-12-27 at 15 11 32

mayel:
OK I'm starting to feel pretty good with the generated output (tested with a bunch of projects) and the test suite, probably missing some things but could use some feedback on the implementation and test coverage :)

lib/ex_doc/doc_ast.ex
```diff
@@ -64,6 +64,70 @@ defmodule ExDoc.DocAST do
     Enum.map(attrs, fn {key, val} -> " #{key}=\"#{val}\"" end)
   end
 
+  @doc """
+  Transform AST into a markdown string.
+  """
+  def to_markdown_string(ast, fun \\ fn _ast, string -> string end)
```
josevalim, line 70: Why not render the original content instead? 🤔

josevalim:
Thanks everyone for the work so far. I believe this is a great direction and, at the same time, it shows we need to some clean up before moving forward:

1. We will want to refactor some of the template handling because there is too much shared logic between HTML, EPUB, Markdown. I think some of it should be moved to the retriever.

2. I believe it should be multiple files indeed and no zips, since we can also use it to generate man pages and either. See #1992, which is now a duplicate? /cc @eksperimental

lib/ex_doc/doc_ast.ex
mayel: We do when the original content is markdown. This was needed for cases where an AST node is created manually, like for type specs.

josevalim: I see. In this case maybe we should push the functionality to the retriever, so it adds specs both in text and html format.

josevalim: Or maybe we use a separate function that knows how to render the specs for a given node with the given format.

mayel: What about the autolink functions? They depend on the parsed AST of the extra markdown docs and transform it.

mayel:
>     1. We will want to refactor some of the template handling because there is too much shared logic between HTML, EPUB, Markdown. I think some of it should be moved to the retriever.

Is this about DRY and not having duplicate logic, or about having more separation between the code for each format? Some guidance would be helpful here. 

>     2. I believe it should be multiple files indeed and no zips, since we can also use it to generate man pages and either. 

Yeah it's generating separate files now, following the same structure/naming as the html ones.

> See [WIP: Markdown formatter #1992](https://github.com/elixir-lang/ex_doc/pull/1992), which is now a duplicate? /cc @eksperimental

Ah yeah seems so, I can look at that PR to see if there's any approach or piece of code (thinking especially of the templates) that looks better and port them to this one?

lib/ex_doc/doc_ast.ex
josevalim: Do we want the markdown links in this case? Would the markdown links be useful for man pages? cc @eksperimental

mayel: Dunno. And on the LLM side not sure if links would make a difference, any idea @mjrusso?

eksperimental:
Hi everyone. I woud like to discuss about this more in detail, I think we could open up an issue so we don't divert the conversation from this PR? Doing the formatter I noticed the duplication but also the limitations of the current approach.

mayel:
> I also think that specs/types/callbacks should be inside `` ```erlang/elixir `` blocks. That means that we need to remove the links, but as links are not working anyway it does not matter.

how are links not working? and yeah it's either having links or formatting there, not sure which is preferable...

mayel:
> Hi everyone. I woud like to discuss about this more in detail, I think we could open up an issue so we don't divert the conversation from this PR? Doing the formatter I noticed the duplication but also the limitations of the current approach.

where you going to open an issue @eksperimental? otherwise not sure how to proceed here @josevalim?

josevalim:
I should also say that we are adding search over HexDocs, which would allow you to search only certain packages for a given term, and submit the filtered results to a LLM. Would that be better than giving the whole docs of a bunch of deps? Which I assume would consume too many tokens?

mjrusso:
> I should also say that we are adding search over HexDocs, which would allow you to search only certain packages for a given term, and submit the filtered results to a LLM. Would that be better than giving the whole docs of a bunch of deps? Which I assume would consume too many tokens?

Yes, there is downstream work required to effectively use the documentation (at least for LLM consumption), which also happens to look a lot like a search problem. 

[This Livebook](https://livebook.dev/run/?url=https%3A%2F%2Fgithub.com%2Fmjrusso%2Fhex2txt%2Fblob%2Fmain%2Fnotebooks%2Fhex2context.livemd) is a simple prototype; there's tons of opportunities for improvement but it does work and provide reasonable results. (This happens to use hex2txt to get the docs, but the nice part is that the approach is general and could work with any Markdown as input. I want a standalone app like this that I can run locally that exposes as a Model Content Protocol server, but that's getting off topic :)

garazdawi:
> how are links not working?

The links in specs are working, but the autolinks in the markdown documentation does not (that is `t:String.t/0` is not resolved to anything).

> yeah it's either having links or formatting there, not sure which is preferable...

I'm going to guess that this depends on what it will be used for. For the usecase that @zuiderkwast wants (that is converting to man pages), the formatting is to prefer as there are no links in man pages anyway. Either way it is easy enough for some postprocessing tool to strip links and re-format the specs.

eksperimental:
> were you going to open an issue @eksperimental? otherwise not sure how to proceed here @josevalim?

Sorry for the delay. I will write something in an new issue later today.

zuiderkwast:
Is this PR close to being merged? @josevalim

It would be nice to have this before OTP 28 is released so we can use it to generate OTP man pages and include them in OTP 28.

josevalim:
No, unfortunately we need time to review all of the building blocks here and I don't think it will be before Erlang/OTP 28. Sorry :(

josevalim:
Apologies folks for the delay. There has been a lot of changes around this area, both in ExDoc and in AI, and I have been postponing this.

Perhaps more interesting, the following has happened:

1. We have improved some of ExDoc so there is less duplication across formatters
2. Other documentation tools are now including "copy markdown" or "view markdown" functionality on their pages, as seen here: https://modelcontextprotocol.io/introduction
3. Some LLM tools are now including a llms.txt that is basically an index and points to other files, as generating a single llms.txt may blow up the context

Therefore, I propose the following:

1. Extract the building of extras to happen outside of the formatter and instead pass it as arguments to the formatter, more likely by introducing a `ExDoc.ExtraNode` construct
2. Change the HTML formatter to generate markdown files by default (which can reuse good chunks of this PR)
3. Add a "copy markdown" button to the top of all pages
4. Generate a llms.txt as an index by default as part of HTML formatter

Is everyone happy with this direction? Any takers?

zachdaniel:
@josevalim for clarity, the generation of markdown files would essentially allow going to any given piece of docs and saying `<path_to_docs>.md` to get markdown for those docs? 


A separate thought: agents set a user-agent header indicating what they are. What if we tailored that experience more heavily? 

For example:

- automatically redirecting to markdown
- showing summaries with just headers/functions and letting it pass query parameters to view specific items?

josevalim:
> for clarity, the generation of markdown files would essentially allow going to any given piece of docs and saying <path_to_docs>.md to get markdown for those docs?

Yes. The copy markdown is then a matter of fetching the .md on the side and copying to clipboard.

> A separate thought: agents set a user-agent header indicating what they are. What if we tailored that experience more heavily?

That's outside of the ExDoc concern. We don't say how the pages are served. Plus I am not sure there will be a unified way of checking for agents, each likely pass a different user agent.

garazdawi:
I think this approach would work from our perspective.

mjrusso:
@josevalim +1 for this proposal. Just one minor additional suggestion, re:

> Some LLM tools are now including a llms.txt that is basically an index and points to other files, as generating a single llms.txt may blow up the context
>
> ...
>
> Generate a llms.txt as an index by default as part of HTML formatter

It would be nice to generate `llms.txt` (as an index, as you proposed) but to _also_ generate `llms-full.txt`, which includes everything in-line. You can see https://llmstxt.site for examples of this in-the-wild (e.g., Anthropic has both https://docs.anthropic.com/llms.txt and https://docs.anthropic.com/llms-full.txt).

mjrusso:
(Although, curiously, Anthropic's `llms-full.txt` includes content in a seemingly-random order; it would be more intelligible if it were ordered in the same way as the index.)

yordis:
Hey @josevalim any resolutions on the topic?

josevalim:
Here is my proposal: https://github.com/elixir-lang/ex_doc/pull/1976#issuecomment-2963511394

Anyone is welcome to go ahead with the implementation or rebase this one.

yordis:
@mayel I am gonna wait to see if you would like to continue the work, otherwise, hopefully I do not forget and pick it up from here.

yordis:
@josevalim I can continue the work at https://github.com/elixir-lang/ex_doc/pull/2146

Right now the CSS and JS isn't properly working, but I can focus on that last if that is Ok with you

PR #1980, opened 2024-12-15
Fix typespec with (...) -> any(), closes #1974

### LINKED ISSUE #1974: A certain type definition with multiple any() crashes ex_doc
The below type definition causes a crash on main (commit d8146c397fde7df2a09497b379a801519f3c2e66)
```
-type foo() :: fun((...) -> any()) | [any()].
```
crash:
```
       (ex_doc 0.35.1) lib/ex_doc/language/erlang.ex:699: ExDoc.Language.Erlang.pop/0
       (ex_doc 0.35.1) lib/ex_doc/language/erlang.ex:617: anonymous fn/2 in ExDoc.Language.Erlang.replace/3
       (elixir 1.14.0) lib/string.ex:1620: String.do_replace/4
       (elixir 1.14.0) lib/string.ex:1623: String.do_replace/4
       (elixir 1.14.0) lib/string.ex:1577: String.replace_guarded/4
       (ex_doc 0.35.1) lib/ex_doc/language/erlang.ex:601: ExDoc.Language.Erlang.autolink_spec/4
```
I couldn't shrink it further what is special about that contrived type definition, but the below type definitions work fine:
```
-type foo() :: fun() | [any()].
-type foo() :: fun((...) -> integer()) | [any()].
-type foo() :: fun((...) -> any()) | [integer()].
```

josevalim:
@gomoripeti this is fixed in main but it will require Elixir v1.18+.

@garazdawi in the long term, we probably want the code that converts a typespec into a string with links to be part of Erlang/OTP? Perhaps by adding some sort of fallback to the existing pretty printing function? Or perhaps, instead of converting it to Elixir AST and traversing it, we'd traverse the Erlang AST typespec directly.

### DIFF
--- a/lib/ex_doc/language/erlang.ex
+++ b/lib/ex_doc/language/erlang.ex
@@ -546,9 +546,14 @@ defmodule ExDoc.Language.Erlang do
             {op, _, [int]}, acc when is_integer(int) and op in [:+, :-] ->
               {nil, acc}
 
-            # fun() (spec_to_quoted expands it to (... -> any())
-            {:->, _, [[{name, _, _}], {:any, _, _}]}, acc when name == :... ->
-              {nil, acc}
+            # fun() (spec_to_quoted expands it to (... -> any() in Elixir v1.17 and earlier)
+            # TODO: Remove me when we require Elixir v1.18+
+            {:->, _, [[{name, _, _}], {:any, _, _}]} = node, acc when name == :... ->
+              if Version.match?(System.version(), ">= 1.18.0-rc") do
+                {node, acc}
+              else
+                {nil, acc}
+              end
 
             # record{type :: remote:type/arity}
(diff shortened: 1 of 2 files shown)


josevalim:

garazdawi:
> Or perhaps, instead of converting it to Elixir AST and traversing it, we'd traverse the Erlang AST typespec directly.

I think this would be a better option. IIRC it should be possible to use erl_syntax_lib:fold to get something relatively close to what we have today.

garazdawi:
I found a regression with this change. If you do:

```
-spec foo(fun()) -> ok.
```

It will try to create a link for a local type called `fun/0`, while `fun()` is actually syntax (as opposed to `function()` which is more like a type()). Any ideas on how to solve this would be appreciated. I will work around it by replacing it with `function()` for now and will start digging to see if I can come up with a solution.

PR #1983, opened 2024-12-20
Use view transitions to improve ux

### LINKED ISSUE #1982: Use something like pjax for hosted docs
Today we perform full page navigation but we can use something like pjax for navigating across pages as long as they have relative paths. However, we need to be careful to not use pjax when running from the local file system. Note we don't need to use pjax per se, just something similar, to keep the navigation smoother.

josevalim:
Closes #1982


assets/js/content.js
```diff
@@ -16,6 +16,30 @@ export function initialize (isPreview) {
...
+      const href = link.getAttribute('href')
+      if (!href) return
+
+      // Only relative links
+      if (href.startsWith('http') || href.startsWith('/')) return
+      e.preventDefault()
+
+      document.startViewTransition(() => {
+        window.location.href = href
+      })
```
jonatanklosko, line 40: Unfortunately `startViewTransition` doesn't work across document navigations, it's for animating DOM state on the same page.

Issue #1962, opened 2024-11-13
Output man pages


zuiderkwast:
When OTP switched to using `ex_doc` in OTP 27, we lost the man pages for the OTP modules.

Would you accept a PR to produce man pages, or is there a way to access the intermediate markdown format and use an external tool (e.g. pandoc) to generate man pages from it?

Example of an OTP man page from OTP 26:

![image](https://github.com/user-attachments/assets/8fb927a6-e252-43e0-8b0e-11f325846fb2)

josevalim:
It probably makes the most sense for us to have a markdown output. Then you can use something like `pandoc` to convert it to man. :) PRs exploring this would be welcome!

eksperimental:
I remember some years ago I proposed having a Markdown formatter. I think having the documentation in this format opens up the possibility to a lot of things.

Now that I have some free time I can look into an implementation. What would be the basic requirements?

josevalim:
I can't think of many requirements, actually. I guess the only question is if it is one markdown file per module or one per function. If it is one per module, then it is equivalent to our other formatters, but probably considerably simpler.

zuiderkwast:
One per module matches the old OTP man pages, so for this use case, it's perfect.

eksperimental:
I would go with one file per module per file. This way the documentation could be readable over a terminal with `less` or markdown reader.

garazdawi:
We have an escript that converts a subset of markdown to man pages: https://github.com/erlang/otp/blob/master/make/markdown_to_man.escript. So if ex_doc could generate markdown documents for each module, that escript could be used to create man pages for it.

Note that the escript is on intended for general use, only of the subset of markdown used by the Erlang/OTP docs.

eksperimental:
I'm already working on it, and I have a draft working version. I will create soon create an issue so we can discuss about anything related to this feature.

eksperimental:
Here's the PR https://github.com/elixir-lang/ex_doc/pull/1992

josevalim:
Erlang is adding their own man page handling on top of Markdown, which probably makes more sense since they can handle it directly from the terminal without generating the man pages upfront.

If someone wants to do a man pages version for ExDoc, we will gladly accept it once the Markdown work is done. Thanks!

zuiderkwast:
> Erlang is adding their own man page handling on top of Markdown, which probably makes more sense since they can handle it directly from the terminal without generating the man pages upfront.

[...]

PR #1992, opened 2024-12-31
WIP: Markdown formatter

eksperimental:
Original implementation as discussed in https://github.com/elixir-lang/ex_doc/issues/1962

The generated docs can be seen in this repo: https://github.com/eksperimental/ex_doc_markdown_formatter_docs

This is a WIP. There are many details to fix such as some characters that need to be properly escaped/unescaped.

I would like to get some feedback before moving forward.

My approach was to keep things simple and add things if they are needed.

One challenge was to rewrite the headings levels within docs, since that would break the hierarchical structure of the document.

Screenshots

**API REFERENCE**
image

**KERNEL**
image

**KERNEL**
image


mix.exs
```diff
@@ -44,9 +44,10 @@ defmodule ExDoc.Mixfile do
       # Add other makeup lexers as optional for the executable
       {:makeup_c, ">= 0.1.0", optional: true},
       {:makeup_html, ">= 0.1.0", optional: true},
-      {:jason, "~> 1.2", only: :test},
+      {:jason, "~> 1.4"},
       {:floki, "~> 0.0", only: :test},
-      {:easyhtml, "~> 0.0", only: :test}
+      {:easyhtml, "~> 0.0", only: :test},
+      {:mdex, "~> 0.3"}
```
leandrocp, line 50: Hey mdex author here. I'm happy to help, make changes upstream, or answer any question to support this PR. Thanks!

lib/ex_doc/formatter/markdown.ex
```diff
@@ -0,0 +1,539 @@
...
+
+      %MDEx.Heading{level: level}, lowest_level when level < lowest_level ->
+        {:cont, level}
+
+      _, lowest_level ->
+        {:cont, lowest_level}
+    end)
+  end
+
+  defp bump_levels(%MDEx.Document{nodes: nodes} = document, levels_to_bump) do
```
leandrocp, line 161: Hi @eksperimental you could leverage the Access protocol if you prefer. Here's an example: https://github.com/leandrocp/mdex/commit/d44bdd69e3ab6fa506a93a46de102b4c1fffbdfb

mix.exs
eksperimental: Thank you @leandrocp for your help.  I will reach out if I need anything.

lib/ex_doc/formatter/markdown.ex
eksperimental: Thank you. Applied in 805b8ba

yordis:
@eksperimental do you need any help with this one?

eksperimental:
> @eksperimental do you need any help with this one?

Hi @yordis. The PR is a proof of concept. As it is now it's OK. There's another PR going on in parallel. I wouldn't invest much more energy on this until we see what's going to happen with the other PR and the direction on this one. I think we need to wait for Jose's input on this. Only then if it's decided to continue with this I think I may use your help as there are plenty of details to be ironed out.

Thanks for the help. I will let know when if the time comes.

Issue #1987, opened 2024-12-21
Accessibility issue in ExDoc headers


josevalim:
The experience could be improved further if those links to "view source" and "link to this function" were semantically outside of the heading. Currently if you navigate to e.g. "API Reference", the announcement says "Navigated to view source API reference". It would require a bit of a CSS rewrite.

<img width="890" alt="SCR-20241221-qasn" src="https://github.com/user-attachments/assets/b7a5015e-ba8d-45e4-b34c-32b1b1fd200c" />

_Originally posted by @angelikatyborska in https://github.com/elixir-lang/ex_doc/issues/1986#issuecomment-2558181320_

josevalim:
@DavidOliver, if you have some availability, could you take a look? I think you have experience changing this part of the app and would probably be 10x more productive than me. If you cannot look at it, I understand. :) Thank you! :heart:

josevalim:
I think we should have something like this or similar:

```
<div id="modules" class="section-heading">
  <a href="#modules" class="hover-link">
    <i class="ri-link-m" aria-hidden="true"></i>
  </a>
  <h2 class="text">Modules</h2>
</div>
```

DavidOliver:
@josevalim, sure, I'll take a look.

DavidOliver:
Now in progress.

josevalim:
Closing in favor of the PR :)

PR #1993, opened 2025-01-01
Move action links out from heading tags

DavidOliver:
@angelikatyborska, @josevalim,

This draft PR only updates the headings which contain action links:

- top-of-page headings;
- callback and function headings.

Headings which contain hover links are not changed here; the hover links do not include text so I believe "Navigated to ..." announcements are okay, and changing these headings would require some more styling updates. If it needs to be done I can look at doing so, but I thought I'd get you to take a look before spending more time on it.

@angelikatyborska, if you could take another look, that would be great.

Still to do:

- Checking callback/function annotations (e.g., 'macro'), which may be affected, though I'm hoping they will still be correctly aligned.
- Test updates.

I'll wait to hear back.

---

Avoids action link text being read out as part of "navigated to..." announcements.

#1987

### DIFF
--- a/assets/css/content/functions.css
+++ b/assets/css/content/functions.css
@@ -58,7 +58,7 @@
 }
 
 .content-inner .detail-header .icon-action {
-  margin-top: 4px;
+  margin-top: 3px;
 }
 
 .content-inner .specs pre {
--- a/assets/css/content/functions.css
+++ b/assets/css/content/functions.css
@@ -16,8 +16,8 @@
 }
 
 .content-inner .detail-header {
-  margin: 2 0 1em;
-  padding: 0.5em 1em;
+  margin: 1em 0;
+  padding: 0.5em 0.85em 0.5em 1em;
   background-color: var(--textDetailBackground);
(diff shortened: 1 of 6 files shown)


josevalim:
LGTM! :) Thank you!

DavidOliver:
Annotation-related ordering and alignment needs bringing into line. I'll update that and the tests soon.

DavidOliver:
Function annotations now re-aligned and tests updated.

Headings now use Flexbox rather than floats, and respond better to narrow viewports:

image

josevalim:
Closing and reopening for CI.

assets/css/content/functions.css
```diff
@@ -62,6 +57,10 @@
   }
 }
 
+.content-inner .detail-header .icon-action {
+  margin-top: 4px;
```
josevalim, line 61:
```suggestion
  margin-top: 3px;
```

This makes the alignment the same as main. Apologies for pixel shaving. :D

josevalim:

PR #1995, opened 2025-01-02
Page load improvements

liamcmitchell:
Main improvement is reducing the layout shift caused by the sidebar popping in on page load.

* Set sidebar state & width in `inline_html.js`
* Reduce sidebar state & animation classes from 6 to 2
* Remove redundant body selectors
* Add max width to drawer
* Refactor `sidebar-drawer.js` to reduce indirection and simplify logic
* Register sidebar size observer on mousedown to avoid recording non-user changes and triggering DOM updates
* Simplify sidebar animation sequence reducing DOM updates and callbacks
* Refactor `theme.js`
* Remove toast on browser theme change (it made the code cleaner and I thought it was OK to go)
* Change script tags to [`defer`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#async_and_defer) - this removes the need for `onDocumentReady` helper but I thought I would stop my refactoring here

My machine is too old to capture accurate video of page load so instead I'm showing page load recordings from Chrome devtools performance tab.

## Before

https://github.com/user-attachments/assets/8a4af095-bf24-4b78-9726-ca3ed11b484a

## After

https://github.com/user-attachments/assets/ff67b618-a82c-4d40-aad1-fd76c54aca8d

Biggest layout shift is now `0.4` instead of `0.9`, page loads a little faster, assets are a little smaller.

### DIFF
--- a/assets/css/layout.css
+++ b/assets/css/layout.css
@@ -104,7 +104,7 @@ body {
     max-width: 90vw;
   }
 
-  .sidebar-button {
+  body:not(.sidebar-open) .sidebar-button {
     position: absolute;
   }
 }
--- a/assets/css/layout.css
+++ b/assets/css/layout.css
@@ -27,15 +27,17 @@ body {
   height: 100%;
 }
 
+/* Sidebar is closed by default and opened with body.sidebar-opened. */
 .sidebar {
-  display: flex;
+  display: none;
   flex-direction: column;
   width: var(--sidebarWidth);
   min-width: var(--sidebarMinWidth);
+  max-width: 50vw;
   height: 100%;
   position: fixed;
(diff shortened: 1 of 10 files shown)


josevalim:
This looks great @liamcmitchell! I found only one regression. On small screens, the hamburger for opening and closing the sidebar is moving with the page, instead of remaining static. You can see it it is out of place in the screenshot below:

Screenshot 2025-01-02 at 09 45 47

If you scroll further, then it fully disappears!

liamcmitchell:
Sorry, didn't check other stylesheets for the classes I removed.

Fixed this and a few other issues.

josevalim:

PR #2006, opened 2025-01-04
Page load improvements

liamcmitchell:
Minor performance improvements and refactoring.

* remove `mix compile --force` from asset build watch script
* styling
  * set `apple-os` class in `inline_html.js`
  * remove `non-apple-os` class, replace with `:not(.apple-os)` selector
  * update `isAppleOs()` helper to query this class for source of truth
* content
  * replace `fixLinks()` & `.no-underline` with CSS-only `:has(code, img)` selector
  * remove `fixSpacebar()`, there is no need to focus content to enable keybinds
  * inline and refactor `setLivebookBadgeUrl()` - the only remaining code
* copy-button
  * optimise button HTML, remove unneeded attrs, replace `span.sr-only` with appropriate attrs on svg
  * replace `.insertAdjacentHTML()` with `.cloneNode()` (only parses HTML once)
  * implement array filters in query selectors
* makeup
  * inline many things
  * implement mouseenter/leave callbacks in single function, no need for closures
  * only search parent for matching delimiters
* search-bar
  * remove reading of `window.scrollY` on init, this forces layout and is not required
  * refactor logic
  * remove `scrollThreshold` of 70 px, looking at git history and PR comments, this is no longer needed
* tooltips
  * inline HTML, no need for handlebars
  * inline a few functions
  * move filter logic into query selectors
  * remove need for closures in event handlers

Performance improvement is so minor, I can't record it reliably. JS is 0.8kb smaller.

### DIFF
--- a/assets/README.md
+++ b/assets/README.md
@@ -22,7 +22,7 @@ $ npm run --prefix assets build
 
 Build a complete production bundle, including JavaScript and CSS.
 
-(Note that this is not required to be manually run when generating docs: if you run `mix build` at the `ExDoc` root after changing your assets, the assets will be recompiled, `mix compile --force` will be invoked, and fresh docs with your changes will be generated.)
+(Note that this is not required to be manually run when generating docs: if you run `mix build` at the `ExDoc` root after changing your assets, the assets will be recompiled and fresh docs with your changes will be generated.)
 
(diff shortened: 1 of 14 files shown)


josevalim:

PR #2007, opened 2025-01-07
Page load improvements

liamcmitchell:
* support importing plain `.html` templates, bypassing handlebars completely
* sidebar-list:
  * remove redundant ids & classes, prefer using aria attrs
  * small adjustment to expand button position
  * initialize on sidebar open - no need to do anything until visible
  * init once only, no need to re-render on nav - **NOTE: expand/collapse states will now persist across SWUP nav**
  * convert from handlebars to vanilla JS
  * move tab and tabset render from elixir to JS - keep it all together, less HTML
  * remove nodes_map from many functions
  * much inlined
  * refactored callbacks
* copy-button:
  * move button to template
  * create template lazily
  * inline `addCopyButtons`
* modal:
  * convert to html template
  * move state to plain vars intead of nested properties, no need for object
  * remove init, render lazily
  * inlined more things
* quick-switch:
  * convert body to html template
  * build result el inline
* sidebar-drawer:
  * register update callback
  * call sidebar-list init on open
* tabsets: get heading text without requiring layout
* toast: init lazily

## Before

before

```
../formatters/html/dist/html-OD6LFIR3.js                             144.0kb
../formatters/html/dist/html-elixir-FAAZMXJW.css                      65.4kb
```

## After

after

```
../formatters/html/dist/html-HTRMLDRB.js                             130.2kb
../formatters/html/dist/html-elixir-EGPMA5T5.css                      64.4kb
```

-14kb JS, -1kb CSS, no more forced layout in init, sidebar-list renders about 4x faster


josevalim:
It looks so smooth now, thank you! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

josevalim:

josevalim:
@liamcmitchell I found a bug when running it under swup.js. To reproduce:

```
$ mix build
$ python3 -m http.server 8080
$ open http://localhost:8080/doc/
```

If you click on the sidebar to navigate across pages, for example within "Pages", it doesn't close the current open item. Can you please take a look?

I am wondering if we still need swup.js but I would say the answer is yes to avoid flashing pages on slow connections?

assets/js/modal.js
```diff
@@ -1,45 +1,38 @@
...
-}
+// State
+
+/** @type {HTMLDivElement | null} */
+let modal = null
+/** @type {HTMLElement | null} */
+let prevFocus = null
+/** @type {HTMLElement | null} */
+let lastFocus = null
+let ignoreFocusChanges = false
```
jonatanklosko, line 14: FWIW I consider a single state object good for maintainability/readability, because it makes it clear when we modify the global state rather than deal with local variables :)

assets/js/handlebars/helpers.js
```diff
@@ -1,45 +1,5 @@
...
-    if (context.lastModuleSeenInGroup !== node.nested_context) {
-      return options.fn(this)
-    }
-  } else {
-    // track the most recently seen module
-    // prevents emitting a duplicate entry for nesting when
-    // the nesting prefix matches an existing module
-    context.lastModuleSeenInGroup = node.title
-  }
-})
```
jonatanklosko, line 29: This is great! FTR if someone is motivated enough we could drop handlebars altogether. We could build the HTML by hand `` x.innerHTML = `...` `` or perhaps use something lightweight like lit-html or similar (which relies on tagged string literals such as `` html`<div>${foo}</div>` ``).

assets/js/modal.js
liamcmitchell: I assumed it was done for that reason (I've also done something similar in the past) but I had my code-golfing hat on.

I don't think the global/local distinction is very useful here. To me, global is JS global (`window`/`global`), everything else is local. I don't see why module state needs to be treated differently to function state. The `let` keyword should signal that the var is state and it should be no harder to trace back. Maybe rename it to `let stateModal` or similar if it needs to be special? But I don't think it's special.

Anyway, reasons to avoid the object wrapper:
* one less object
* property accesses have a runtime cost
* i don't think it can be minified as well
  * object: `let o={modal:null};o.modal`
  * let: `let m=null;m`

Super minor I know but I didn't think the object wrapper was very useful.

assets/js/handlebars/helpers.js
liamcmitchell: We'll see how far I get. I've been replacing most of it with vanilla JS and an `el()` helper:
https://github.com/elixir-lang/ex_doc/blob/22103e7d6f7bdee1924faa770350020198393ddb/assets/js/helpers.js#L203-L222

jonatanklosko: `el()` is totally fine also! Unless the code ends up overly nested/convoluted for the slightly bigger handlebars templates (probably manageable though).

assets/js/modal.js
jonatanklosko: > I don't see why module state needs to be treated differently to function state

Function state is short lived and usually it's easy to see all variables the function defines. So the rationale would be that reading a function 100 lines into the file, seeing `state.modal = ...` immediately gives the idea that it alters a long-lived state (like a state store), while `model = ...` looks just like a function-local assignment and you need to look around to rule that out.

I wouldn't expect the runtime cost of key access to be relevant in practice, but minification is a fair argument.

That said, it's not a strong opinion and I don't mind the change. Just mentioning for completeness :)

assets/js/handlebars/helpers.js
liamcmitchell: Yeah, there is definitely value in templates but the handlebars runtime alone is 50kb.

I'm looking at the remaining logic in handlebars and thinking about how it would look as a mix of plain HTML templates and vanilla JS.
The most complex templates are:
* [autocomplete-suggestions.handlebars](https://github.com/elixir-lang/ex_doc/blob/main/assets/js/handlebars/templates/autocomplete-suggestions.handlebars)
* [search-results.handlebars](https://github.com/elixir-lang/ex_doc/blob/main/assets/js/handlebars/templates/search-results.handlebars)
* [tooltip-body.handlebars](https://github.com/elixir-lang/ex_doc/blob/main/assets/js/handlebars/templates/tooltip-body.handlebars)

I think it's doable but prob also worth investigating [Svelte](https://svelte.dev/), which has a very similar [template syntax](https://svelte.dev/docs/svelte/each) but much smaller runtime cost.

I think I'll try converting one of them both ways and compare.

josevalim:
> I mentioned this in the PR desc, maybe I should have bolded more :)

It was my bad then. :) Perhaps we can have "breaking changes" listed separately?

> I thought it was more a feature than a bug. I thought if a user has opened a menu, then closing it without explicit intent (clicking collapse button) might be disruptive. I can also see that auto-closing menus keeps the menu clean. I'll make a PR to change it.

I thought about this but I'd prefer for now to keep a consistent experience across them.

> Menu could be updated immediately, no need to wait for content load.

Originally, I was think we should update the menu immediately and clear the content area, but when I looked at other places, like Docusaurus, it behaves like today. What are your general thoughts?

Other than that, given we still see the sidebar flashing without swup.js, we should probably keep it in?

liamcmitchell:
> Originally, I was think we should update the menu immediately and clear the content area, but when I looked at other places, like Docusaurus, it behaves like today. What are your general thoughts?

I think it's good as-is. Doing all changes together keeps the nav consistent with visible content and means less layout shift.

> Other than that, given we still see the sidebar flashing without swup.js, we should probably keep it in?

I think the experience is worth the 30kb. I would keep it.

assets/js/handlebars/helpers.js
jonatanklosko: Exactly, handlebars runtime is a good reason to remove .

I mentioned lit-html, because it is much smaller than something like Svelte, and for rendering small chunks of html it should be enough.

jonatanklosko: I actually didn't hear about it either, but I searched around for `` html`...` `` literal, since I had a vague idea that it either exists or should exist as an optimised version of `` x.innerHTML = `...` `` :D There is also lighterhtml (and probably others),  but based on popularity and activity lit-html is definitely a safer bet. The reason I would pick it over Svelte here is that the `html` literal feels almost exactly like a regular interpolated string, so there is no abstraction overhead on contributors. So for our purpose, I would lean towards that, but if you have more thoughts, happy to hear them!

josevalim:
Btw, this pull request breaks the sidebar rendering on iOS. No icons show up. I am trying to understand why it happens, but nothing is logged to the console. :D

josevalim:
Ok, it happens on any small screen. So if you reduce the browser width, you can reproduce it. I found the exact line of code, I will drop a comment.

assets/js/sidebar/sidebar-drawer.js
```diff
@@ -45,6 +48,7 @@ export function initialize () {
 export function update () {
   const pref = sessionStorage.getItem(SIDEBAR_STATE_KEY)
   const open = pref !== SIDEBAR_PREF_CLOSED && !isScreenSmall()
+  if (open) initializeList()
```
josevalim, line 51: This is the problematic line. It doesn't ever initialize the list on small screens. :) I will revert it on main but feel free to push a better fix if this should be maintained somehow.

PR #2011, opened 2025-01-08
feat: allow configuring `custom_autocompletions` list

zachdaniel:
Doing the non-hacky version of this work here: https://github.com/elixir-lang/ex_doc/pull/2009

The plan is to try and set it up as incremental improvements that can be used in conjunction to do what I want. I believe this will be the first of three changes.

This one allows users to add custom sidebar-items (but not actually in the sidebar) that will show up in autocomplete.
CleanShot 2025-01-08 at 10 07 57@2x

no. 2 would be configuring extras to collapse headers into dropdown lists the way that modules are split into `Types` and `Callbacks` and `Functions` in the sidebar

no. 3 would be configuring extras to not put their headers into the autocomplete list at all, because I will be providing custom ones (so a `hidden: true` key being set on sidebar-items.js)

How does this plan sound?


lib/ex_doc/formatter/html.ex
```diff
@@ -183,7 +183,9 @@ defmodule ExDoc.Formatter.HTML do
   end
 
   defp generate_sidebar_items(nodes_map, extras, config) do
-    content = Templates.create_sidebar_items(nodes_map, extras)
+    content =
+      Templates.create_sidebar_items(nodes_map, extras, config.custom_autocompletions || [])
```
josevalim, line 187:
```suggestion
      Templates.create_sidebar_items(nodes_map, extras, config.custom_autocompletions)
```

josevalim (approved):
This PR looks good to me. My suggestion is to add docs and ship it.

zachdaniel:
Alright @josevalim small tweaks for consistency (in one place it was `auto_completion` and in another place it was `autocompletion` and added to the readme. More PRs to follow, but probably next week :)

README.md
```diff
@@ -492,9 +492,46 @@ For this example, we've changed the module `:my_module` to `:my_module2`, and th
   {"get-started", "quickstart"}
 ]}.
 ```
-
 <!-- tabs-close -->
 
+## Customizing Autocomplete
```
josevalim, line 497: Let's move this to the `mix docs` task. That's where we generally put options and configuration. WDYT?

zachdaniel: That’s fine with me but should I do that for the redirects section above? I put it in the resume because that’s where that was (Also something I added so i could just be wrong x 2 😅)

josevalim: Yes, for sure. I should have caught the redirect one the first time around as well. Please move both of them there and follow the style there!

zachdaniel: Done :)

The `redirects` option was already documented in the task, but I moved its accompanying long form docs there as well.

josevalim:

PR #2012, opened 2025-01-08
Add warning for stale version of documentation

alisinabh:
Continuation of #1918 *(could not push to the same branch since I'm not a maintainer)*

---

Based on feedback in #1918 here is the revised version of warning color and icon added when an outdated version is detected.

@halostatue I tried yellow but I could not make it work on Light mode. It would end-up look green-ish when lightness was reduced.

### Screenshots

Screenshot 2025-01-08 at 2 30 25 PM
Screenshot 2025-01-08 at 2 30 37 PM
Screenshot 2025-01-08 at 2 30 45 PM

**NOTE**: Initially I wanted to put the icon after the version, but the `<select>` HTML element width is set to the largest possible item. This will create an unwanted padding that looks ugly in projects which have versions like `1.2.3-RC0`.

### DIFF
--- a/assets/css/custom-props/theme-dark.css
+++ b/assets/css/custom-props/theme-dark.css
@@ -80,6 +80,7 @@ body.dark {
   --sidebarHover:                var(--white);
   --sidebarScrollbarThumb:       var(--coldGray);
   --sidebarScrollbarTrack:       var(--sidebarBackground);
+  --sidebarStaleVersion:         var(--orangeDark);
   --sidebarSubheadings:          var(--gray400);
   --sidebarItem:                 var(--gray200);
   --sidebarInactiveItemBorder:   var(--gray400);
--- a/assets/css/custom-props/theme-dark.css
+++ b/assets/css/custom-props/theme-dark.css
@@ -80,7 +80,7 @@ body.dark {
   --sidebarHover:                var(--white);
   --sidebarScrollbarThumb:       var(--coldGray);
   --sidebarScrollbarTrack:       var(--sidebarBackground);
-  --sidebarStaleVersion:         var(--orangeDark);
+  --sidebarStaleVersion:         var(--orangeLight);
   --sidebarSubheadings:          var(--gray400);
   --sidebarItem:                 var(--gray200);
   --sidebarInactiveItemBorder:   var(--gray400);
--- a/assets/css/custom-props/theme-light.css
+++ b/assets/css/custom-props/theme-light.css
@@ -80,6 +80,7 @@
   --sidebarHover:                var(--black);
   --sidebarScrollbarThumb:       var(--coldGrayLight);
   --sidebarScrollbarTrack:       var(--sidebarBackground);
+  --sidebarStaleVersion:         var(--orangeLight);
   --sidebarSubheadings:          var(--black);
   --sidebarItem:                 var(--black);
   --sidebarInactiveItemBorder:   var(--gray500);
--- a/assets/css/custom-props/theme-light.css
+++ b/assets/css/custom-props/theme-light.css
@@ -80,7 +80,7 @@
   --sidebarHover:                var(--black);
   --sidebarScrollbarThumb:       var(--coldGrayLight);
   --sidebarScrollbarTrack:       var(--sidebarBackground);
-  --sidebarStaleVersion:         var(--orangeLight);
+  --sidebarStaleVersion:         var(--orangeDark);
   --sidebarSubheadings:          var(--black);
   --sidebarItem:                 var(--black);
   --sidebarInactiveItemBorder:   var(--gray500);
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -109,6 +109,22 @@
   display: none;
 }
 
+.sidebar .sidebar-staleVersion {
+    display: inline;
+}
+
+.sidebar .sidebar-staleVersion > a {
+    color: var(--sidebarStaleVersion);
+    font-weight: 400;
+}
+
+.sidebar .sidebar-staleIcon {
+    font-size: 18px;
+    position: relative;
+    top: 3px;
+    line-height: 0;
+}
+
 .sidebar .sidebar-list-nav {
   display: flex;
   margin: 0;
--- a/assets/js/handlebars/templates/versions-dropdown.handlebars
+++ b/assets/js/handlebars/templates/versions-dropdown.handlebars
@@ -10,4 +10,12 @@
       {{/each}}
     </select>
   </label>
+  {{#if latestVersion}}
+    <div class="sidebar-staleVersion">
+      <a href="{{latestVersion}}" title="This version of package is not the latest version">
+        <i class="ri-alert-line sidebar-staleIcon" aria-hidden="true"></i>
+        <span>Go to latest</span>
+      </a>
+    </div>
+  {{/if}}
 </form>
--- a/assets/js/sidebar/sidebar-version-select.js
+++ b/assets/js/sidebar/sidebar-version-select.js
@@ -26,10 +26,30 @@ if (!isEmbedded) {
       isCurrentVersion: node.version === currentVersion
     }))
 
-    versionsContainer.innerHTML = versionsDropdownTemplate({ nodes })
+    const latestVersionNode = versionNodes.find(node => node.latest)
+    const latestVersion = latestVersionNode?.version !== currentVersion ? latestVersionNode?.url : null
 
-    qs(VERSIONS_DROPDOWN_SELECTOR).addEventListener('change', handleVersionSelected)
+    versionsContainer.innerHTML = versionsDropdownTemplate({ nodes, latestVersion})
+
+    const select = qs(VERSIONS_DROPDOWN_SELECTOR)
+    select.addEventListener('change', handleVersionSelected)
+    adjustWidth(select)
   }
+}
+
+// Function to adjust the width of the select element
+function adjustWidth (select) {
+  // Create a temporary element to measure the width
+  const temp = document.createElement('span')
+  temp.style.visibility = 'hidden'
+  temp.style.position = 'absolute'
+  temp.style.whiteSpace = 'nowrap'
+  temp.style.font = window.getComputedStyle(select).font
+  temp.textContent = select.options[select.selectedIndex].text
+
+  document.body.appendChild(temp)
+  select.style.width = `${temp.offsetWidth + 20}px`
+  document.body.removeChild(temp)
 }
 
 function handleVersionSelected (event) {
--- a/assets/css/custom-props/common.css
+++ b/assets/css/custom-props/common.css
@@ -52,6 +52,8 @@
   --black: hsl(0, 0%, 0%);
   --black-opacity-10: hsla(0, 0%, 0%, 10%);
   --black-opacity-50: hsla(0, 0%, 0%, 50%);
+  --orangeDark: hsl(30, 90%, 40%);
+  --orangeLight: hsl(30, 80%, 50%);
 }
 
 @media screen and (max-width: 768px) {


halostatue:
Screenshot 2025-01-08 at 14 49 22
Screenshot 2025-01-08 at 14 49 28
Screenshot 2025-01-08 at 14 49 38
Screenshot 2025-01-08 at 14 49 44
Screenshot 2025-01-08 at 14 49 51

assets/js/handlebars/templates/versions-dropdown.handlebars
```diff
@@ -2,7 +2,10 @@
   <label>
     <span class="sidebar-projectVersionsDropdownCaret" aria-hidden="true">&#x25bc;</span>
     <span class="sr-only">Project version</span>
-    <select class="sidebar-projectVersionsDropdown">
+    {{#if latestVersion}}
+      <i class="ri-alert-line sidebar-staleIcon" aria-hidden="true"></i>
+    {{/if}}
+    <select class="sidebar-projectVersionsDropdown{{#if latestVersion}} sidebar-staleVersion{{/if}}"{{#if latestVersion}} title="This version is not the latest version of the package."{{/if}}>
```
halostatue, line 8: I think that this should be both more SR friendly and mobile friendly with the use of a [popover](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API/Using). I am thinking:

```suggestion
{{#if latestVersion}}
  <i class="ri-alert-line sidebar-staleIcon" aria-hidden=true popovertarget="staleVersion"></i>
  <span class="sr-only">This is not the latest version of this package.</span>
  <div id="staleVersion" popover>This is not the latest version of this package.</div>
{{/if}}
<select class="sidebar-projectVersionsDropdown{{# if latestVersion}} sidebar-staleVersion{{/if}}">
```

I’m not sure if this will work, though.

halostatue (changes requested):

assets/js/sidebar/sidebar-version-select.js
```diff
@@ -15,15 +15,19 @@ export function initialize () {
     const versionsContainer = qs(VERSIONS_CONTAINER_SELECTOR)
     // Initially the container contains only text with the current version
     const currentVersion = versionsContainer.textContent.trim()
+
+    const latestVersionNode = versionNodes.find(node => node.latest)
+    const latestVersion = latestVersionNode?.version !== currentVersion ? latestVersionNode?.version : null
```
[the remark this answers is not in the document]
alisinabh: Good point. Addressed in 8ef47f02

assets/js/handlebars/templates/versions-dropdown.handlebars
alisinabh: Since popover works with clicks and not hover, this won't work as clicking on the icon will open the version selector.

alisinabh: SR text addressed in da05c1d9

josevalim:
@alisinabh this looks great to me. A couple things:

1. Shall we have the icon after the version? I feel it calls more attention, otherwise it is kind of stuck between the caret and the v. Here is a mockup:

401281824-ecfb3120-b23b-4fc9-9189-f90c5f498208

2. I think the dark orange works better for light mode, because it is darker and has more contrast. Maybe we have a single variable `--orange` with the current value of `--orangeDark`?

josevalim:
Sorry, a couple additional thoughts (we are going in circles here and I apologize):

1. The most likely thing the user will want to do is to jump to latest

2. Both --orange and --orangeDark today in light mode today has low AA rating on Firefox

So my suggestion is:

1. have --orangeDark be hsl(30, 90%, 40%) and use it on light mode
2. have --orangeLight be hsl(30, 80%, 50%) and use it on dark mode
3. Add a link to jump to latest

Here is a quick HTML I came up with:

```
<div style="display: inline;padding-left: 8px;color: hsl(30, 90%, 40%);">
  <i aria-hidden="true" class="ri-alert-line"></i> <span>Go to latest</span>
</div>
```

And here is how it looks like:

PNG image


WDYT?

halostatue:
> Sorry, a couple additional thoughts (we are going in circles here and I apologize):
> 
> 1. The most likely thing the user will want to do is to jump to latest
> 2. Both --orange and --orangeDark today in light mode today have low AA accessibility rating on Firefox
>
> So my suggestion is:
> 
> 1. have --orangeDark be hsl(30, 90%, 40%) and use it on light mode
> 2. have --orangeLight be hsl(30, 80%, 50%) and use it on dark mode
> 3. Add a link to jump to latest

You mean something like this?

```html
<div class="sidebar-projectVersion" translate="no">
  <form autocomplete="off">
    <label>
      <span class="sidebar-projectVersionsDropdownCaret" aria-hidden="true">▼</span>
      <span class="sr-only">Project version</span>
      <select class="sidebar-projectVersionsDropdown">
          <option translate="no" value="https://hexdocs.pm/elixir/1.18.1">v1.18.1</option>
          <option translate="no" value="https://hexdocs.pm/elixir/1.18.0">v1.18.0</option>
          <option translate="no" value="https://hexdocs.pm/elixir/1.18.0-rc.0">v1.18.0-rc.0</option>
          <option translate="no" value="https://hexdocs.pm/elixir/1.17.3">v1.17.3</option>
      </select>
    </label>
  </form>
  <div style="display: inline;padding-left: 8px;color: hsl(30, 90%, 40%);">
    <i aria-hidden="true" class="ri-alert-line"></i> <span>Go to latest</span>
  </div>
</div>
```

halostatue:
FWIW, I can't seem to get it to work well on Safari with anything *except* beneath the `select` or `form` (so the `inline` can go away).

Screenshot 2025-01-09 at 10 38 45

josevalim:
Unfortunately it needs to after the version dropdown, otherwise will break the logo positioning in several designs.

josevalim:
I understand the issue the width of the `select` is dictated by its largest option. We would need to use some JavaScript to adjust it. Here is what ChatGPT proposed:

```

    const adjustWidth = () => {
      // Create a temporary element to measure the width
      const temp = document.createElement('span');
      temp.style.visibility = 'hidden';
      temp.style.position = 'absolute';
      temp.style.whiteSpace = 'nowrap';
      temp.style.font = window.getComputedStyle(select).font;
      temp.textContent = select.options[select.selectedIndex].text;

      document.body.appendChild(temp);
      select.style.width = `${temp.offsetWidth + 20}px`; // Add some padding
      document.body.removeChild(temp);
    };
```

But I didn't try it. We could run this before we inject the select.

assets/js/sidebar/sidebar-version-select.js
```diff
@@ -17,14 +18,38 @@ export function initialize () {
...
     renderVersionsDropdown({ nodes, latestVersion })
+    initializeVersionSelect()
   }
 }
 
+function initializeVersionSelect () {
+  const select = qs(VERSIONS_DROPDOWN_SELECTOR)
+  const text = qs(VERSIONS_TEXT_SELECTOR)
+
+  text.textContent = select.options[select.selectedIndex].text
```
josevalim, line 34: Instead of doing this here, it would probably be better to do it in .handlebars itself, by passing `currentVersion` when we render?

assets/js/sidebar/sidebar-version-select.js
```diff
@@ -17,14 +18,38 @@ export function initialize () {
...
+
+  text.addEventListener('click', (event) => {
```
josevalim, line 36: Did you find this simpler than setting up the width? My biggest concern is if this will work from mobile, I will check soon and let you know.

josevalim:
I tested it on iOS and it worked well, so we are good with going this route. However I have found two bugs:

1. The `select` has font-weight of 400 but the text currently does not have
2. Clicking on the caret does not open up the select, perhaps we should move the caret inside the text?

alisinabh:
@josevalim Sorry I was testing my solution and realized it does not work on Safari (needs two clicks since safari does not allow showPicker in some conditions). Tried what you posted and it worked better also the logic is more simple.

Can you please test again?

assets/css/sidebar.css
```diff
@@ -109,6 +109,22 @@
   display: none;
 }
 
+.sidebar .sidebar-staleVersion {
+    display: inline;
+    padding-left: 4px;
```
josevalim, line 114:
```suggestion
```

josevalim: No longer necessary as the width padding on select adds enough space.

assets/css/sidebar.css
```diff
@@ -109,6 +109,21 @@
...
+
+.sidebar .sidebar-staleVersion > a {
+    color: var(--sidebarStaleVersion);
+    font-weight: 400;
+}
+
+.sidebar .sidebar-staleIcon {
+    font-size: 18px;
+    position: relative;
+    top: 3px;
```
josevalim, line 124:
```suggestion
    top: 3px;
    line-height: 0;
```

This is required so the caret on the left-side preserves its alignment. :)

assets/css/custom-props/theme-dark.css
```diff
@@ -80,6 +80,7 @@ body.dark {
   --sidebarHover:                var(--white);
   --sidebarScrollbarThumb:       var(--coldGray);
   --sidebarScrollbarTrack:       var(--sidebarBackground);
+  --sidebarStaleVersion:         var(--orangeDark);
```
josevalim, line 83:
```suggestion
  --sidebarStaleVersion:         var(--orangeLight);
```

The brighter orange goes in the darker barkground. :)

assets/css/custom-props/theme-light.css
```diff
@@ -80,6 +80,7 @@
   --sidebarHover:                var(--black);
   --sidebarScrollbarThumb:       var(--coldGrayLight);
   --sidebarScrollbarTrack:       var(--sidebarBackground);
+  --sidebarStaleVersion:         var(--orangeLight);
```
josevalim, line 83:
```suggestion
  --sidebarStaleVersion:         var(--orangeDark);
```

josevalim:
@alisinabh I have dropped two last comments and this is good to go! Please rebase the PR and submit it so it doesn't contain changes the `formatter` directory. Thank you so much! :heart:

assets/css/custom-props/theme-light.css
alisinabh: I though it looks off 😅 
Addressed in 5dd3ff6f

josevalim:

assets/js/sidebar/sidebar-version-select.js
```diff
@@ -26,12 +26,32 @@ if (!isEmbedded) {
...
+  temp.style.whiteSpace = 'nowrap'
+  temp.style.font = window.getComputedStyle(select).font
+  temp.textContent = select.options[select.selectedIndex].text
+
+  document.body.appendChild(temp)
+  select.style.width = `${temp.offsetWidth + 20}px`
+  document.body.removeChild(temp)
+}
+
 function handleVersionSelected (event) {
```
wojtekmach, line 55: `handleVersionSelected` makes sure that when we are on, say, https://hexdocs.pm/elixir/1.18.0/Task.html and from version selector we choose v1.18.1, we'll end up at https://hexdocs.pm/elixir/1.18.1/Task.html.

I think it'd be nice for "Go to latest" to work similarly, preserve relative path.

I'm not very well versed in JS but this is a PoC (copy pasting the implementation with minor changes) for this functionality:

```diff
diff --git a/assets/js/sidebar/sidebar-version-select.js b/assets/js/sidebar/sidebar-version-select.js
index 48968c88..de80813f 100644
--- a/assets/js/sidebar/sidebar-version-select.js
+++ b/assets/js/sidebar/sidebar-version-select.js
@@ -34,6 +34,9 @@ if (!isEmbedded) {
     const select = qs(VERSIONS_DROPDOWN_SELECTOR)
     select.addEventListener('change', handleVersionSelected)
     adjustWidth(select)
+
+    const versionsGoToLatest = qs(".sidebar-staleVersion a")
+    versionsGoToLatest.addEventListener('click', handleGoToLatestClicked)
   }
 }

@@ -67,6 +70,22 @@ function handleVersionSelected (event) {
     })
 }

+function handleGoToLatestClicked (event) {
+  const url = this.href
+  const pathSuffix = window.location.pathname.split('/').pop() + window.location.hash
+  const otherVersionWithPath = `${url}/${pathSuffix}`
+  event.preventDefault();
+
+  checkUrlExists(otherVersionWithPath)
+    .then(exists => {
+      if (exists) {
+        window.location.href = otherVersionWithPath
+      } else {
+        window.location.href = url
+      }
+    })
+}
```

@alisinabh if this piques your interest perhaps you could send another PR?

Issue #2013, opened 2025-01-09
Introduce latest and related packages search


josevalim:
Now that we added package search to Hexdocs, we can start exposing it in ExDoc.
This will be done as two distinct features, latest search and related packages search.
We are also adding search feature to Hexdocs landing page and you should familiarize
yourself with [the new screens there before moving forward](https://github.com/hexpm/hexdocs/issues/49#issuecomment-2558537425).

## Overall description

We will add new features to ExDoc that allows you to perform both search and
autocompletion on the latest or related versions of a package. We will use the
existing `docs_config.js` to enable this functionality. Once enabled, a dropdown
arrow will appear embedded at the end of the search allowing you to choose the
type of search you want to perform, similar to this:

<img width="563" alt="Screenshot 2025-01-09 at 10 30 27" src="https://github.com/user-attachments/assets/e5b9f66b-5025-45f8-9b1d-e86f12d341ae" />

The text in the search bar will indicate which mode is active:

  * `related` - Type / to search this and related packages
  * `latest` - Type / to search the latest version
  * `current` - Type / to search this version

You can click the dropdown to choose the mode, these options will show up as:

  * `:related` - Search the latest version of this package and related packages
  * `:latest` - Search the latest version of this package
  * `:current - Search the current version of this package (works offline)

To enable these modes, a `searchNodes` variable must be injected into `docs_config.js`.
This variable is an array of `{name: "package-name", version: "package-version"}` entries,
and it must have at least one entry, which is the current package itself and its latest
version. The modes should be available in this order and according to these conditions:

  * `:related` - If searchNodes is available and it contains more than one entry
  * `:latest` - If searchNodes is available and it contains more than zero entries
  * `:current` - Always

Furthermore, we also want to improve the search results page to be closer to the one in
HexDocs and provide a more consistent experience overall. This means support for toggling
previews too.

## Implementation

  * [ ] Update the current search results to leverage previews and stay closer to HexDocs mockups

  * [x] Change HexDocs to include the `searchNodes` in all public packages today.
    We should also simplify the `versionNodes` one, there is no need to include "latest" on every entry
    (cc @wojtekmach)
  
  * [ ] Add support for `searchNodes` in ExDoc's autocompletion and search,
    with the dropdown for configuration, using the new TypeSense API. We can use Elixir itself to test
    this feature, as its docs_config.js is generated by hand rather than Hex.pm

[...]

PR #2014, opened 2025-01-09
Revert "feat: allow configuring `custom_autocompletions` list"

josevalim:
We realized this won't work with #2013, since the search comes from Typesense
which uses search nodes. We will explore other approaches.

Reverts elixir-lang/ex_doc#2011


josevalim:

PR #2015, opened 2025-01-09
Refactor module init

liamcmitchell:
Simplify `html.js` by moving all init into modules.

* run preview & hint first, they remove DOM, reducing work for other modules
* register `swup:page:view` callbacks from within modules
* move `isEmbedded`, `isPreview` and `isHint` into globals
* run module init from within modules
* adjust `descriptionElementFromHash` to also move comment nodes - preserving tabset comments
* inline many things in preview
* fix scroll caused by preview loading (on chrome/edge at least)
* fix quick switch link after swup nav (init wasn't called after nav)
* inline many things in sidebar version select
* move swup init into own module, it doesn't need to know about other modules now
* inline many things in hint-page

One thing I noticed when refactoring the init logic:
```js
  if (isPreview && isEmbedded) {
    initPreview()
  } if (isHint && isEmbedded) {
    initHintsPage()
  } else {
  //  init swup, versions, keyboard shortcuts, qs, sidebar...
```
Notice that the second `if` is not an `else if` even though it's formatted like one. I assume that it was intended to be an `else if` and have refactored it that way.

By shuffling the module init and preventing a lot of unnecessary work, previews load almost 2x faster for me locally.

## Before
image

## After
image

-0.4kb JS


josevalim:

PR #2016, opened 2025-01-09
Fix sidebar open/close on mobile

liamcmitchell:
Fixes https://github.com/elixir-lang/ex_doc/pull/2007#issuecomment-2580965575

Sorry, I put the init in the wrong update fn and didn't test.

I moved the init to the right fn, added a comment and renamed the wrong update fn to it's prev name before I changed it 🤦🏼‍♂️

Found another bug while testing. The close on content click didn't work after SWUP nav, the content element it was registered to is replaced on nav. Fixed by listening to body clicks.


josevalim:

PR #2017, opened 2025-01-11
feat: customize `search_data` for an extra

zachdaniel:
This allows configuring the "search data" for any given extra. When you do this, you take over the way that the extra appears both in autocomplete *and* in search_data. The current system assumes that you give this information in markdown, which I think is a safe assumption, but we could do something like allow for an `html_body` or `markdown_body` to be given, etc.

This is likely not going to be useful for 99% of people, and as such I think its okay that its not super ergonomic. Specifically, you have to completely manage the search content etc. The way this will be used, however, is in `Spark` where we generate markdown files for DSL documentation. With this change, we can add a function called `Spark.Dsl.search_data_for(Your.Dsl)` which will improve the autocomplete experience and the search experience in various ways (by adding more things to autocomplete than just headers, and by customizing the `type` of the search data nodes

@josevalim I know we had discussed a sort of "unification" of search-data and sidebar-items, but I think this does let me kick the can down the road a little bit. 

With this PR  `ex_doc` provides a way to customize the search data for extras (which I need) that encompasses all of the information that currently drives both search-data *and* sidebar-items. This means that we are free to refactor those two files internally, and my code that customizes searchable content for extras won't be affected. Additionally, since you configure the `search_data` for an extra *completely* (it replaces the autogenerated search data), we don't double index the same material but differently.

If this approach is amenable then I can work up some documentation and add some additional tests for how it modifies search data. Only didn't do that because I think there will be some discussion/back and forth on naming, implementation, etc. (or even if this is something that ex_doc will accept 😄 )


zachdaniel:
To get an idea of how it works, see this test:

https://github.com/elixir-lang/ex_doc/pull/2017/files#diff-f2a0f2316eaaba625bd96b239eaff839eacb84d3df2c5a619d10d9f6a6c5378aR600

lib/ex_doc/formatter/html.ex
```diff
@@ -445,6 +451,32 @@ defmodule ExDoc.Formatter.HTML do
...
+
+      %{
+        link: "#{id}.html##{search_data.anchor}",
+        title: search_data.title,
+        body: search_data.body,
+        description: search_data.description,
+        type: search_data.type
+      }
+    end)
+  end
```
josevalim, line 473: I'm very happy with this contract!

test/ex_doc/formatter/html_test.exs
```diff
@@ -592,6 +592,41 @@ defmodule ExDoc.Formatter.HTMLTest do
...
 
+    test "custom search data is added to the sidebar", %{tmp_dir: tmp_dir} = context do
+      generate_docs(
+        doc_config(context,
+          source_beam: "unknown",
+          extras: [
+            {"test/fixtures/README.md",
+             search_data: [
+               %{
+                 anchor: "heading-without-content",
```
josevalim, line 603: There is one particular annoyance here. The top item of the search data is the "module body" of the page, and it doesn't have an anchor. So we need to allow anchor to be an empty string and add a test that we emit the proper code (i.e. no trailing #).

josevalim: Alternatively, assume that the top-level one is always generated. This will impact other places, please see comment above.

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -90,6 +100,13 @@ defmodule ExDoc.Formatter.HTML.Templates do
         group: to_string(group),
         headers: extract_headers(content)
       }
+      |> then(fn item ->
+        if is_nil(extra[:search_data]) do
+          item
+        else
+          Map.put(item, :hide_in_autocomplete, true)
+        end
+      end)
```
josevalim, line 109: I'd like to suggest one alternative approach here. Instead of introducing `custom` as its own separate thing, make this be:

    Map.put(item, :custom, [%{id: ..., anchor: ...}, ...])

Where the entry in the list is as closest to the header as possible, but feel free to add additional fields if you find necessary. I think this will require fewer changes to the autocomplete code and be similar in nature.

The only caveat is the empty anchor. You would either have to remove it here, since we would still automatically autocomplete the top/parent node, or we take the decision that search_data is only about the children data. The top/parent node is always automatically included, both in autocomplete, and in search.

zachdaniel: Sorry, I’m not following 😓

josevalim: To recap: instead of storing a separate `nodes.custom`, we will have an optional `.search` field inside each extra. Then the autocomplete code will do this: if the element has a `.search` field, you will search for entries inside the .search field, otherwise business as usual. This means that the results will still appear under the "extras" category instead of adding a new one. Also, the entries in the .search field should mirror the ones in `headers` as much as possible for simplicity. And then we can also get rid of `hide_in_autocomplete`.

zachdaniel:
@josevalim I believe I've addressed your comments. Instead of using `custom` I've used the key `searchData` since it aligns more with how it is configured and will be used in general.

Also added a test in the js spec.

assets/js/autocomplete/suggestions.js
```diff
@@ -18,7 +18,8 @@ const SUGGESTION_CATEGORY = {
   moduleChild: 'module-child',
   mixTask: 'mix-task',
   extra: 'extra',
-  section: 'section'
+  section: 'section',
+  custom: 'custom'
```
josevalim, line 22: I think the custom stuff should now be gone?

assets/js/autocomplete/suggestions.js
```diff
@@ -131,14 +144,30 @@ function childNodeSuggestion (childNode, parentId, query, category, label) {
  * Builds a suggestion for a node section.
  */
 function nodeSectionSuggestion (node, section, query, category, label) {
-  if (!matchesAny(section.id, query)) { return null }
+  if (!matchesAny(section.title || section.id, query)) { return null }
```
josevalim, line 147: I am not the biggest fan of the added conditionals. It makes the code hard to follow. Can't we normalize them to be as close as possible when generating the search data from Elixir side?

assets/js/autocomplete/suggestions.js
```diff
@@ -211,7 +240,8 @@ function categoryPriority (category) {
     case SUGGESTION_CATEGORY.module: return 1
     case SUGGESTION_CATEGORY.moduleChild: return 2
     case SUGGESTION_CATEGORY.mixTask: return 3
-    default: return 4
+    case SUGGESTION_CATEGORY.custom: return 4
+    default: return 5
```
josevalim, line 244: Revert?

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -84,12 +84,32 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
+
       %{
         id: to_string(id),
         title: to_string(title),
         group: to_string(group),
         headers: extract_headers(content)
       }
+      |> then(fn item ->
+        if search_data do
+          Map.put(item, :searchData, search_data)
```
josevalim, line 108: Sorry, something go lost in translation. We should not send the search data as is, but the transformed version of the search data. Make it look like `headers` as much as possible. I'd rather transform it on Elixir code than JS code. :D

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -84,12 +84,32 @@ defmodule ExDoc.Formatter.HTML.Templates do
     for extra <- extras do
       %{id: id, title: title, group: group, content: content} = extra
 
+      search_data =
+        case extra do
```
josevalim, line 88: Move this clause perhaps to the conditional below? We don't need to have two conditionals :)

lib/ex_doc/formatter/html/templates.ex
zachdaniel: Right 🤦‍♂️

zachdaniel: This came from an intermediate design when I though I had to support both the child node and section branches with a single object and IIRC they had different fields, but the conditional for "if searchData" solves that.

assets/js/autocomplete/suggestions.js
```diff
@@ -133,12 +145,20 @@ function childNodeSuggestion (childNode, parentId, query, category, label) {
 function nodeSectionSuggestion (node, section, query, category, label) {
   if (!matchesAny(section.id, query)) { return null }
 
+  let link
+
+  if (section.anchor === '' || !section.anchor) {
```
zachdaniel, line 150: This is required because when using `searchData` there may be nodes with no anchor link.

assets/js/autocomplete/suggestions.js
```diff
@@ -133,12 +145,20 @@ function childNodeSuggestion (childNode, parentId, query, category, label) {
...
+  } else {
+    link = `${node.id}.html#${section.anchor}`
+  }
+
   return {
-    link: `${node.id}.html#${section.anchor}`,
+    link,
     title: highlightMatches(section.id, query),
-    description: node.title,
+    description: section.description || node.title,
```
zachdaniel, line 159: This conditional is necessary because search items may customize the text below the title, but currently extra sections use their document's name (`node.title`).

assets/js/autocomplete/suggestions.js
```diff
@@ -133,12 +145,20 @@ function childNodeSuggestion (childNode, parentId, query, category, label) {
...
     matchQuality: matchQuality(section.id, query),
-    labels: [label, 'section'],
+    labels: section.labels || [label, 'section'],
```
zachdaniel, line 161: This conditional is necessary because we determine the labels from the `type` provided by the search_data item, and it wouldn't make sense for them to have `['extra', 'section']` as labels.

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -90,6 +90,24 @@ defmodule ExDoc.Formatter.HTML.Templates do
         group: to_string(group),
         headers: extract_headers(content)
       }
+      |> then(fn item ->
+        case extra do
+          %{search_data: search_data} when is_list(search_data) ->
+            search_data =
+              Enum.map(search_data, fn search_item ->
+                search_item
+                |> Map.put(:id, search_item.title)
```
zachdaniel, line 99: This mapping is admittedly a bit strange, but we massage data primarily designed for the search page into the format expected by sidebar items for autocomplete here (as requested).

zachdaniel:
Okay @josevalim I think it should align better with what you are looking for now (which is better than what I had). There is still some conditional logic changes required in the autocomplete code, but I've explained each in a comment.

assets/js/autocomplete/suggestions.js
```diff
@@ -133,12 +145,20 @@ function childNodeSuggestion (childNode, parentId, query, category, label) {
 function nodeSectionSuggestion (node, section, query, category, label) {
   if (!matchesAny(section.id, query)) { return null }
 
+  let link
```
josevalim, line 148: ✂️

lib/ex_doc/formatter/html.ex
```diff
@@ -183,7 +184,9 @@ defmodule ExDoc.Formatter.HTML do
   end
 
   defp generate_sidebar_items(nodes_map, extras, config) do
-    content = Templates.create_sidebar_items(nodes_map, extras)
+    content =
+      Templates.create_sidebar_items(nodes_map, extras)
```
josevalim, line 188:
```suggestion
    content = Templates.create_sidebar_items(nodes_map, extras)
```

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -84,12 +84,29 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
+        }
+
+      case extra do
+        %{search_data: search_data} when is_list(search_data) ->
+          search_data =
+            Enum.map(search_data, fn search_item ->
+              search_item
+              |> Map.put(:id, search_item.title)
+              |> Map.put(:labels, [search_item.type])
+              |> Map.take([:id, :labels, :description, :anchor])
```
josevalim, line 102:
```suggestion
              |> Map.take([:description, :anchor])
              |> Map.put(:id, search_item.title)
              |> Map.put(:labels, [search_item.type])
```

lib/mix/tasks/docs.ex
```diff
@@ -339,14 +341,12 @@ defmodule Mix.Tasks.Docs do
...
-  As your project grows, your documentation may very likely change, even structurally.
-  There are a few important things to consider in this regard:
+  As your project grows, your documentation may very likely change, even structurally. There are a few important things to consider in this regard:
 
-    * Links to your *extras* will break if you change or move file names.
-    * Links to your *modules, and mix tasks* will change if you change their name.
-    * Links to *functions* are actually links to modules with anchor links.
-      If you change the function name, the link does not break but will leave users
-      at the top of the module's documentation.
+  - Links to your *extras* will break if you change or move file names.
```
josevalim, line 346: ExDoc uses `*` for lists with 2 space indentation (as seen above).

lib/mix/tasks/docs.ex
```diff
@@ -361,6 +361,27 @@ defmodule Mix.Tasks.Docs do
         "get-started" => "quickstart"
       }
 
+  ## Customizing Extras
+
+  - `:title` - The title of the extra page. If not provided, the title will be inferred from the filename.
+  - `:filename` - The name of the generated file. If not provided, the filename will be inferred from the source file.
+  - `:source` - The source file of the extra page. This is useful if you want to customize the filename or title but keep the source file unchanged.
```
josevalim, line 368: Let's make sure to keep these lines shorter, wrapping around 98 (Elixir's length)

lib/mix/tasks/docs.ex
```diff
@@ -361,6 +361,27 @@ defmodule Mix.Tasks.Docs do
...
+  ### Customizing Search Data
+
+  It is possible to fully customize the way a given extra is indexed, both in autocomplete and in search.
+  In most cases, this makes sense for _generated_ documentation. If `search_data` is provided, it completely
+  overrides the built in logic for indexing your document based on the headers and content of the document.
+  The following fields can be provided in a list of maps for `search_data`.
+
+  - `:anchor` - The anchor link for the search result. Use `""` to point to the top of the page.
+  - `:title` - The title of the search result.
+  - `:description` - A brief description of the search result.
```
josevalim, line 381:
```suggestion
  - `:title` - The title of the result.
  - `:description` - A brief description of the result (used during autocompletion)
```

assets/js/autocomplete/suggestions.js
josevalim: Remember that our goal is to unify this with TypeSense search, as the TypeSense search will be the default in the future for Ash most likely. And this description field is not indexed in TypeSense at all, so this feature will be lost anyway. I suggest removing it then.

assets/js/autocomplete/suggestions.js
josevalim: Is the `.anchor` never not set? It seems before it assumed it to always be present.

josevalim (commented):
We are almost there! I have only dropped nitpicks now. :)

assets/js/autocomplete/suggestions.js
zachdaniel: Right, the second is unnecessary

assets/js/autocomplete/suggestions.js
zachdaniel: Hmm... regardless of how things change, however, the autocomplete would look "bad" without the description I'm providing. I.e it would either be the title of the document or nothing. For now, I'll remove because it is always easier to add to an API than remove, and I can revisit if needed. Thinking about it, the title of the extra being the description isn't the worst thing ever. 👍🏻

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -84,12 +84,30 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
+          group: to_string(group),
+          headers: extract_headers(content)
+        }
+
+      case extra do
+        %{search_data: search_data} when is_list(search_data) ->
+          search_data =
+            Enum.map(search_data, fn search_item ->
+              %{
+                anchor: search_item.anchor,
```
zachdaniel, line 100: With the removal of `:description`, `Map.take` and `Map.put` here didn't make sense anymore.

zachdaniel:
@josevalim ready for another round 🙇

lib/mix/tasks/docs.ex
```diff
@@ -339,14 +341,12 @@ defmodule Mix.Tasks.Docs do
...
+  As your project grows, your documentation may very likely change, even structurally. There are a few important things to consider in this regard:
 
-    * Links to your *extras* will break if you change or move file names.
-    * Links to your *modules, and mix tasks* will change if you change their name.
-    * Links to *functions* are actually links to modules with anchor links.
-      If you change the function name, the link does not break but will leave users
-      at the top of the module's documentation.
+  - Links to your *extras* will break if you change or move file names.
+  - Links to your *modules, and mix tasks* will change if you change their name.
+  - Links to *functions* are actually links to modules with anchor links. If you change the function name, the link does
```
josevalim, line 348:
```suggestion
    * Links to your *extras* will break if you change or move file names.
    * Links to your *modules, and mix tasks* will change if you change their name.
    * Links to *functions* are actually links to modules with anchor links.
      If you change the function name, the link does
```

test/ex_doc/formatter/html_test.exs
```diff
@@ -592,6 +592,42 @@ defmodule ExDoc.Formatter.HTMLTest do
              ] = Jason.decode!(content)["extras"]
     end
 
+    test "custom search data is added to the sidebar", %{tmp_dir: tmp_dir} = context do
```
josevalim, line 595:
```suggestion
    test "custom search data is added to the sidebar and search nodes", %{tmp_dir: tmp_dir} = context do
```

And we need to test the generated search_data.js as well. Also please add one item without anchor so we get proper coverage on those branches too.

zachdaniel:
lemme squash all these commits actuall

lib/ex_doc/language/elixir.ex
```diff
@@ -289,7 +289,7 @@ defmodule ExDoc.Language.Elixir do
         {:local, :..}
 
       ["//", "", ""] ->
-        {:local, :"..//"}
+        {:local, :..//}
```
josevalim, line 292:
```suggestion
        {:local, :"..//"}
```

test/ex_doc/formatter/html_test.exs
```diff
@@ -592,6 +592,56 @@ defmodule ExDoc.Formatter.HTMLTest do
...
+
+                 Here it is :)
+                 """
+               }
+             ]}
+          ]
+        )
+      )
+
+      "sidebarNodes=" <> content = read_wildcard!(tmp_dir <> "/html/dist/sidebar_items-*.js")
```
josevalim, line 626: We need to test search_data-*.js too, sorry for all the back and forth!

zachdaniel: Not a problem :)  It's not right until it's right!

zachdaniel:
@josevalim done

josevalim:

PR #2018, opened 2025-01-12
Replace makeup with prism?

liamcmitchell:
Not sure if this is considered an improvement so opening draft PR for feedback.

I was looking for more page load wins and found the syntax highlighting HTML produced by makeup.

For example, code in a markdown block like the following:
```elixir
def deps do
```
Is transformed into the following HTML:
```html
<span class="kd">def</span><span class="w"> </span><span class="nf">deps</span><span class="w"> </span><span class="k" data-group-id="7816317888-1">do</span><span class="w">
```

I wondered how much that affected HTML size and render. I tested ExDocs readme.html (23 code blocks) with and without the makeup highlighting. Turns out it doesn't hurt that much. It compresses well so gzip size is only 10% bigger and render is also only about 10% slower.

I then wondered how other projects do it and found many/most use client side JS ([highlight.js](https://highlightjs.org)/[Prism](https://prismjs.com/)). As I'm writing this PR I was checking something in git history and see that this project used highlight.js until it was [removed](https://github.com/elixir-lang/ex_doc/pull/1300) 5 years ago 😬.

I tested the same readme.html with deferred syntax highlighting using Prism (had issues without plugin autoload in highlight.js). Full page load is about 2x slower but the only visible difference is a change in text color. The page without syntax highlighting should still render 10% faster and the syntax highlighting comes later.

The changes in this draft are just a POC:
* makeup disabled, not removed completely
* Prism and plugins are loaded from CDN - only works offline if cached
* uses different themes
* no open/closing highlighting on hover ([plugin exists](https://prismjs.com/plugins/match-braces))
* leading `$` in bash is not unselectable ([plugin exists](https://prismjs.com/plugins/command-line/))

I see clear benefits:
* support for 297 languages compared to 9
* no need for config
* support for nested languages

image

image

How strong is the preference for in-house tools?
How important is syntax highlighting offline?
Is there a preference for self-contained deps (each doc bundles own Prisma) or shared (all docs use CDN)?
Is there a preference of CDN?


josevalim:
Hi @liamcmitchell, thanks for asking!

The benefit of using makeup is keeping everything “in house”. We can do bug fixes and improve syntax highlighting for Elixir itself quite straight-forward, as it is all written in Elixir. It is also easier to manage if we want to support different formats (say we want to add  coloring `man` pages or similar).

so in this particular case, I’d prefer to leave it all as is :)

liamcmitchell:
OK, thanks!

hichemfantar:
my input on this is that Prism isn't maintained but [Shiki](https://shiki.style/) is really cool, they have a [playground](https://textmate-grammars-themes.netlify.app/?theme=aurora-x&grammar=elixir).

[Vitepress](https://vitepress.dev/) uses shiki under the hood.

PR #2021, opened 2025-01-16
Add main scrollbar dark mode support and improve scrollbar styles

hichemfantar:
Introduce support for dark mode in the main scrollbar and enhance scrollbar styles to use native styles as much as possible for better contrast and more predictable scrollbars adhering more to user expectations.

image

image

### DIFF
--- a/assets/css/autocomplete.css
+++ b/assets/css/autocomplete.css
@@ -114,6 +114,7 @@
   white-space: normal;
   overflow-x: hidden;
   overscroll-behavior-y: contain;
+  scrollbar-width: thin;
 }
 
 .autocomplete-suggestions.previewing:has(.selected) {
@@ -232,26 +233,6 @@
 
 .autocomplete-suggestions a {
   text-decoration: none !important;
-}
-
-/* Style the scrollbar track (the area behind the thumb) */
-.autocomplete-suggestions::-webkit-scrollbar {
-  width: 5px;
-  /* Set the width of the scrollbar */
-  border-radius: 7px;
-  /* Add rounded corners to the thumb */
-  flex-shrink: 0;
-}
-
-/* Style the scrollbar thumb (the draggable part) */
-.autocomplete-suggestions::-webkit-scrollbar-thumb {
-  background-color: var(--autocompleteScrollbarThumb);
-  border-radius: 7px;
-}
-
-/* Style the scrollbar track on hover */
-.autocomplete-suggestions::-webkit-scrollbar-track {
-  background-color: var(--autocompleteScrollbarTrack);
 }
 
 @media screen and (hover: none) {
--- a/assets/css/content/cheatsheet.css
+++ b/assets/css/content/cheatsheet.css
@@ -114,11 +114,6 @@
   padding: var(--vertical-space) var(--horizontal-space);
 }
 
-.page-cheatmd .content-inner pre code::-webkit-scrollbar {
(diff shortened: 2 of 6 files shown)


josevalim:
Thank you @hichemfantar!

@liamcmitchell @DavidOliver @jonatanklosko, can you please let me know if you are happy with this direction?

hichemfantar:
@josevalim no problem! I opened this mainly to fix the main scrollbar not supporting dark mode but also noticed the inconsistent styles of the scrollbars so I decided to rely on native styles as much as possible as from my experience that reduces problems with cross browser support.

DavidOliver:
Thanks, @hichemfantar. I like the way the scrollbars appear in your screenshots.

Testing out in the two browsers I happen to use in the Pantheon desktop environment is giving mixed results.

## Vivaldi

main:
main-vivaldi

better-scrollbars:
better-scrollbars-vivaldi

Scrollbars for side menu and search results end up with less contrast, but all three page/in-page scrollbars are made consistent.

## Firefox

main:
main-firefox

better-scrollbars:
better-scrollbars-firefox

Scrollbars are even thinner than they were.

---

I don't have much experience with customising scrollbars, so I don't have anything much to offer other than the screenshots. (Don't you just love web dev? :wink: )

If your new scrollbar styling is simpler and works well for other browsers, maybe it's the right move anyway.

hichemfantar:
the thinning behavior is intentional but I guess it can be too thin for some people. I'll revert the change and make them normal width!

hichemfantar:
the aggressive firefox thin behavior is odd and it might related to your OS because it looks thin in the before and after.

hichemfantar:
Alight, the sidebar scrollbar width is now normal and I only kept the thin width in the autocomplete section.

jonatanklosko:
Change to the main scrollbar definitely looks good, it has no custom styles at the moment anyway.

For the secondary ones, whenever I use custom styles it's because the scrollbars are too chunky/disruptive. Perhaps `scrollbar-width: thin` fixes that, but it would be good to get some screenshots from Windows also to get a fuller picture.

liamcmitchell:
> can you please let me know if you are happy with this direction?

+1 for native scrollbar styling and use of `color-scheme: dark;`.

hichemfantar:
it looks the same as this on windows because chromium based browsers tend to be the most consistent across platforms

image

josevalim:
Thanks everyone for the feedback!

hichemfantar:
thanks for merging! any idea when the changes will go live? is it automated deployment?

DavidOliver:
> the aggressive firefox thin behavior is odd and it might related to your OS (perhaps it expands when you hover over it?) because it looks thin in the before and after.

I expect this is the case. The Pantheon desktop environment's scrollbars are somewhat thin by default, and I guess Firefox uses them, but set to (even) thinner. They do expand on hover, but they are still very thin, especially for the search results.

Anyway, I can't say it particularly bothers me. Thanks. :)

hichemfantar:
@DavidOliver can you share what it looks like when expanded?

DavidOliver:
Sure. Window sized to approximately 1920 x 1080p.

https://github.com/user-attachments/assets/6fc53463-3e02-456b-a9ee-18731ede0b15

I'm not sure we should spend too long worrying about this if the scrollbar widths in other DEs and browsers are deemed sufficiently wide.

PR #2023, opened 2025-01-17
Refactor border-radius styles to use CSS variable for consistency

hichemfantar:
I noticed border radius inconsistencies so i addressed them via a single variable.
Update border-radius styles across various components to utilize a CSS variable, ensuring consistency in design. Adjust the variable value for improved appearance.

image

image

image

### DIFF
--- a/assets/css/copy-button.css
+++ b/assets/css/copy-button.css
@@ -14,7 +14,9 @@ pre .copy-button:focus {
   top: 5px;
   right: 5px;
   padding: 4px;
-  background-color: var(--codeBackground);
+  background-color: transparent;
+  backdrop-filter: blur(2px);
+  border-radius: var(--borderRadius);
   border: none;
   cursor: pointer;
   transition: all 150ms;
--- a/assets/css/copy-button.css
+++ b/assets/css/copy-button.css
@@ -15,7 +15,7 @@ pre .copy-button:focus {
   right: 5px;
   padding: 4px;
   background-color: transparent;
-  backdrop-filter: blur(2px);
+  backdrop-filter: blur(4px);
   border-radius: var(--borderRadius);
   border: none;
   cursor: pointer;
--- a/assets/css/autocomplete.css
+++ b/assets/css/autocomplete.css
@@ -107,7 +107,7 @@
 
 .autocomplete-suggestions {
   background-color: var(--autocompleteBackground);
-  border-radius: 8px;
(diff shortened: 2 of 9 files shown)


hichemfantar:
oh my bad i understand now i'll make the change

assets/css/copy-button.css
```diff
@@ -14,7 +14,9 @@ pre .copy-button:focus {
   top: 5px;
   right: 5px;
   padding: 4px;
-  background-color: var(--codeBackground);
+  background-color: transparent;
+  backdrop-filter: blur(2px);
```
josevalim, line 18:
```suggestion
  background-color: var(--codeBackground);
```

I tried it in a couple places but I found the blurred colors confusing. I did like the border though, can we revert it partially?

josevalim:
To be clear, I meant the one for `mix help docs` should stay with 2px: example

assets/css/copy-button.css
hichemfantar: understandle, i increased the contrast. if you still find it confusing i'll revert the change

before
image

after
image

hichemfantar:
i find 4px to be well balanced

2px
image

4px
image

hichemfantar:
@josevalim another example zoomed out

2px
image

4px
image

josevalim:
Foo `bar` baz

DavidOliver:
@hichemfantar, @josevalim, did you mean for the tabs to be affected?

image

image

Although the tabs are now more visually distinguished from each other, the flatter look feels neater to me.

hichemfantar:
looks better with a gap, i'll push a fix

image

hichemfantar:
https://github.com/elixir-lang/ex_doc/pull/2027

DavidOliver:
Do you prefer the new corner style?

Separately to preferring the visual feel of how it was, a (minor) concern I have is that on narrow/mobile screens this approach is more likely to see the tabs taking up more space than is available. Would you be averse to reverting the tabs style?

But if you and @josevalim are happy, not to worry.

hichemfantar:
I prefer the one with spacing because it gives the elements room to breathe and is more aligned with the theme thanks to the rounded corners.
Also the original design was also broken despite the small radius because you could see the rounded corners clashing with each other due to the lack of spacing

hichemfantar:
Wrote a new a implementation that's pretty much better that all the ones we discussed 
Pretty clean right? 

image

PR #2025, opened 2025-01-17
Refactor transitions for consistency across styles

hichemfantar:
Standardize transition properties by introducing a common variable, ensuring uniformity in animation effects throughout the styles.

https://github.com/user-attachments/assets/dc3e4223-bd72-4934-9344-cdaf6b4dcd91

https://github.com/user-attachments/assets/c2a2f162-487c-43e2-8e41-2057f4d64060

### DIFF
--- a/assets/css/custom-props/common.css
+++ b/assets/css/custom-props/common.css
@@ -54,6 +54,7 @@
   --black-opacity-50: hsla(0, 0%, 0%, 50%);
   --orangeDark: hsl(30, 90%, 40%);
   --orangeLight: hsl(30, 80%, 50%);
+  --transition-all: all 150ms cubic-bezier(0.4, 0, 0.2, 1);
 }
 
 @media screen and (max-width: 768px) {
@@ -61,4 +62,8 @@
     --content-width: 100%;
     --content-gutter: 20px;
   }
+}
+
+.transition-all {
+  transition: var(--transition-all);
 }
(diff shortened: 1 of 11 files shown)


josevalim:
@hichemfantar if you mouse over the root items in the sidebar, there is a small jiggle now. Is that intentional? I am afraid part of it is because we are using `all` everywhere. I think I prefer to be explicit over what we want to animate. What if we keep the property explicit and use a variable exclusively for the timer and/or curve?

hichemfantar:
I fixed the bug you talked about and introduces multiple [presets](https://tailwindcss.com/docs/transition-property)

assets/css/custom-props/common.css
```diff
@@ -62,3 +78,27 @@
...
+  transition: var(--transition-shadow);
+}
+
+.transition-transform {
+  transition: var(--transition-transform);
+}
+
+.transition-none {
+  transition-property: none;
+}
```
josevalim, line 104: Do we need the classes? We don't use them anywhere for now, right? So we can skip them and add them later if necessary. :)

assets/css/custom-props/common.css
```diff
@@ -54,6 +54,22 @@
...
+  --transition-colors: color var(--transition-duration) var(--transition-timing),
+  background-color var(--transition-duration) var(--transition-timing),
+  border-color var(--transition-duration) var(--transition-timing),
+  text-decoration-color var(--transition-duration) var(--transition-timing),
+  fill var(--transition-duration) var(--transition-timing),
+  stroke var(--transition-duration) var(--transition-timing);
+
+  --transition-opacity: opacity var(--transition-duration) var(--transition-timing);
+  --transition-shadow: box-shadow var(--transition-duration) var(--transition-timing);
+  --transition-transform: transform var(--transition-duration) var(--transition-timing);
```
josevalim, line 72:
```suggestion
```

josevalim:

PR #2027, opened 2025-01-17
Add gap between tabset tabs for improved spacing

hichemfantar:
Introduce a gap between tabset tabs to enhance visual spacing and improve user experience.

image

### DIFF
--- a/assets/css/tabset.css
+++ b/assets/css/tabset.css
@@ -7,6 +7,7 @@
 .tabset-tablist {
   display: flex;
   overflow-x: auto;
+  gap: 10px;
 }
 
 .tabset-tab {


josevalim:
I prefer them without spacing, so let's ask others' opinions! cc @DavidOliver

hichemfantar:
I prefer the one with spacing because it gives the elements room to breathe and is more aligned with the theme thanks to the rounded corners.
Also the original design was also broken despite the small radius because you could see the rounded corners clashing with each other due to the lack of spacing

josevalim:
Oh, I see there is a discussion on #2023. Given both @DavidOliver and I prefer the old one, I will revert and use z-index to ensure the highlighted one is on top. Thanks @hichemfantar !

hichemfantar:
I wish i could change your mind on the matter because from a design standpoint, snapping together rounded corners (even tiny ones) is a no go.
Essentially no spacing works when the corners are sharp because that conveys continuity.
However with round corners, the principle is to go with spacing.
sharp corners work because they snap to each other.
@DavidOliver @josevalim @jonatanklosko @liamcmitchell

PR #2029, opened 2025-01-17
Add word-wrap to inline code to fix overflow

hichemfantar:
Improves the readability of inline code by adding word-wrap functionality.

before
image

after
image


josevalim:

PR #2030, opened 2025-01-17
Add border-radius to function styles for consistency

hichemfantar:
Enhance the visual appeal of function styles by incorporating a border-radius.

image

### DIFF
--- a/assets/css/content/functions.css
+++ b/assets/css/content/functions.css
@@ -23,6 +23,7 @@
   font-size: 1em;
   font-family: var(--monoFontFamily);
   position: relative;
+  border-radius: var(--borderRadius);
 }
 
 .content-inner .detail-header .signature {
--- a/assets/css/content/functions.css
+++ b/assets/css/content/functions.css
@@ -17,9 +17,8 @@
 
 .content-inner .detail-header {
   margin: 1em 0;
-  padding: 0.5em 0.85em 0.5em 1em;
+  padding: 1em;
   background-color: var(--textDetailBackground);
-  border-left: 3px solid var(--textDetailAccent);
   font-size: 1em;
   font-family: var(--monoFontFamily);
   position: relative;


josevalim:
Thanks. I prefer the current styles so let's see what are others preferences.

hichemfantar:
It's consistent with the rest of the theme. [example](https://docusaurus.io/docs/markdown-features/admonitions)

image

hichemfantar:
we can always get rid of the left border since i don't know what purpose it serves

hichemfantar:
i think it looks better now

image

DavidOliver:
Thanks. My opining, for what it's worth, as other opinions have been requested. :)

Border: I think the left-border helps clearly mark the callback and function headings. I don't believe a medium tone block on a dark background tone alone is sufficient in this regard.

Corners: I'm not averse to introducing rounded corners to the callback/function headings, but I would say that not everything in a design/theme needs to be the same as everything else. Contrast is an important aspect of design, and contrast can take the form of differing forms applied to different types of elements.

That being said, I'd probably go for 2/3px radius on the callback/function headings, which gives a more subtle feel. A smaller radius may also avoid the kind of curved lip on the inside of a border where it meets the corner, of which I'm not a fan.

hichemfantar:
@DavidOliver understandable however I don't understand the user of the left border?
Those kinds of borders are usually best left to admonitions (which are like notes that really need your attention) the one i referenced above.
the callback already has its own bg color and code icon.
spamming left border colors on many consecutive elements can be distracting.

hichemfantar:
admonition type styles should be kept to a minimum and only when needed, you can read about the matter here https://docusaurus.io/docs/markdown-features/admonitions

hichemfantar:
here as well https://github.com/orgs/community/discussions/16925

josevalim:
As we are starting to venture into larger changes, it is important to keep two things in mind:

1. ExDoc is used by more than one language. In this case, the left border gives a clear and early indicator that we are looking at an Elixir definition, especially if they usually keep the sidebar collapsed

2. Once we publish a new ExDoc version, anything published with the old version will remain the same. We have historically done changes to the sidebar and layout, but we have been more strict in relation to contents

This means that, if we change a very identifiable piece such as the function signature header, it will cause confusion, because now the user will suddenly find themselves navigating across 2 or 3 different styles.

For example, we specifically discussed not using left border for admonitions exactly to not confuse those components. So I agree @DavidOliver, consistency is important, but sometimes we should treat distinct elements as visually distinct as we can. And we need a strong reason to change an element that has stayed the same for at least 8 years.

Thank you!

DavidOliver:
The context here is a list of callback and functions, not continuous articles/guides, so I think the callbacks and functions headings should be the elements with the most visual emphasis so as to create a clear, visual structure. I feel that the slice of colour on the left gives the eye useful markers/anchors, helping it to navigate the content, and it's a nice way of injecting a small amount of the programming language's colour (purple or red) without making things gaudy.

From memory, I don't think callback/function documentation tends to contain many admonitions. However, where admonitions *do* appear within callback/function sections, if they, which are subordinate to headings in this hierarchy of information, have the same, or more, emphasis, than the headings, I think that creates discordance and requires a greater expenditure of effort by the reader. So I would tend to go in the other direction for these sections.

hichemfantar:
so the left border is mainly to indicate it's elixir code and not another language, right?

DavidOliver:
@hichemfantar, I think it does both:

1. help visually emphasise the headings and anchor the heading text;
2. reinforce the colour of the language.

PR #2031, opened 2025-01-18
Go to latest with path if exists

alisinabh:
Base on [this comment](https://github.com/elixir-lang/ex_doc/pull/2012#discussion_r1910302119) from @wojtekmach (thank you), this PR adds redirect to exact page user currently is visiting in case it exists in the latest version.

P.S sorry for the delay. I was unwell for the past week.


hichemfantar:
this code in now breaking for some reason, and it's breaking many other things along with it like the settings dialog
`versionsGoToLatest.addEventListener("click", handleGoToLatestClicked);`
```
Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
```

josevalim:
Ah, this needs to be done conditionally, as the version button won't be always there. I will push a fix shortly.

PR #2035, opened 2025-01-18
improvement: always find a preview to render

zachdaniel:
An unanticipated side effect of using the new `search_data` for customizing what appears in autocomplete and search is that, unless your search result points at a header (the entire thing we need to avoid), the preview option will show no content. 

https://github.com/user-attachments/assets/2c5fecff-8fbc-4d6b-95b2-e3ad6003d8ec

I've added a change that I'm hoping will be amenable that improves this. At worst case, it will help in the case of some unexpected structure of the page in normal usage. It only kicks in after the current preview finding logic, which goes to the header in question and grabs the following content.

What this does, is goes *up* until it finds a header, and uses that, or falls back to the top content. To me this seems like a reasonable solution because the preview is still always a header in the document, not something customizable in `search_data` for example. I will likely need to do additional reworking of the way that our DSL docs so that the first parent header is a sensible thing to preview for any given DSL option.

https://github.com/user-attachments/assets/433be8f6-2f70-41c5-9c7e-d1555f69da0f

If we *can't* do this or something like it, then we should revert the entire custom search data feature, because this points out that the only place safe to include in autocomplete into an extra is a header. I really really really hope we don't have to do that though 🙏


josevalim (commented):
LGTM!

zachdaniel:
In that case, I've marked it ready for review 🙏

josevalim:

PR #2036, opened 2025-01-19
Refactor sidebar styles for improved layout and consistency

hichemfantar:
Enhance the sidebar's layout and visual consistency by adjusting styles, padding, and alignment across various elements.
inspired by https://turbo.build/repo/docs

image

image


josevalim:
HI @hichemfantar, thank you for exploring this. I personally prefer the old version, but I know a lot of it boils down to taste. However, there is two objectives aspects that we can discuss:

1. This change removes some of the sidebar width available to contents. You can look at Elixir docs and see how titles are often pruned: https://hexdocs.pm/elixir/agents.html - we need to make sure the maximum amount of width possible is available for titles

2. Some projects are going to have only two or even a single tab at the top of the sidebar, so we need to make sure those look good. Here is an example with two: https://hexdocs.pm/livebook/ - i would also prefer if we keep a stronger distinction between the sidebar tabs and the contents. The current version loses some of it :)

jonatanklosko:
I remember an important point behind the current design was to have more indicators of nesting levels:

image

image

In the proposed design it is less clear (feels more flat):

image

image

Note that in the turbo.build sidebar, only one item is highlighted (i.e. parents of selected item are not highlighted). It also has the vertical bars, which I think are helpful to grasp the nesting.

hichemfantar:
@jonatanklosko i actually prefer that the parents are always highlighted, seems like the best choice overall in terms in clarity.
I’ll also look into the vertical bars

hichemfantar:
@josevalim 1 I feel like that’s always gonna be a problem with long titles but I’ll look into it
2 I think it looks good enough thanks to the existing tab indicators but I’ll look into it

hichemfantar:
@josevalim is there a way to pull and test other docs in my local setup?
Like the elixir or phoenix docs

hichemfantar:
@jonatanklosko good call with the left border
@josevalim what do you think?

image

josevalim:
I think having the padding and the left border is too cluttered. Personally speaking, I think the padding and background is not adding much to the sidebar. But I don't want to send you into a rabbit hole if we are not sure we are going to accept your changes.

> @josevalim is there a way to pull and test other docs in my local setup?

You can reproduce other layouts by changing ExDoc. You can comment this out in the mix.exs:
```
      extras:
        [
          "README.md",
          "Cheatsheet.cheatmd",
          "CHANGELOG.md"
        ] ++ test_dev_examples(Mix.env()),
```

You can also put `@moduledoc false` in `Mix.Tasks.Docs`, if you want to delete the "Mix Tasks" tab.

hichemfantar:
how about now, removed the bg and reduced the padding for extra width like you previously said
image

hichemfantar:
another iteration bringing back the bg color you mentioned

image

hichemfantar:
I believe this version addresses all of your concerns

image

hichemfantar:
blends in perfectly overall

image

hichemfantar:
i believe the new version saves even more in terms of horizontal space leaving more space for text

before
image

after
image

hichemfantar:
with this pr, we can conclude the redesign of the docs because the sidebar is only thing (if i haven't missed anything) that stuck out like a sore thumb

josevalim:
I am still worried the current tab style really won't work with a single element. You can even go ahead and delete them manually in the browser, and it looks like this:

Screenshot 2025-01-19 at 16 35 42

I'd also suggest aligning the expand/collapse icon with the hamburger menu. There are also other visual styles in the sidebar, such as nesting, not used by ExDoc, and I would need to verify them manually before.

If you want, we can also have a separate discussion to improve only the font style in the sidebar and spacing, without touching all of it. Smaller changes will be easier to accept in here, given how complex the sidebar can be across several different projects.

josevalim:
Oh, even in the font style, I remember some Linux OSes/browsers had strongly divergent behaviour on anti-aliasing, and sometimes the font would be either too big or too small, and it may be we have regressions here, which adds further reservations for doing even more changes to the sidebar.

hichemfantar:
letting you know that you're using an outdated version of the branch, make sure you pull the latest changes because they address many issues.

> Oh, even in the font style, I remember some Linux OSes/browsers had strongly divergent behaviour on anti-aliasing

i've tested on multiple OSs and everything looks fine, even tested on mobile

hichemfantar:
the code for this is very old and everything font wise seems stable now and even if some obscure linux distors behave erratically, i don't think that justifies making the reading experience worse for everyone.

hichemfantar:
@josevalim one more to thing to consider is accessibility. people spend a lot of time reading and it's important fonts are rendered predictably and consistently.

hichemfantar:
> I'd also suggest aligning the expand/collapse icon with the hamburger menu, the fact they are not aligned really stands out

are you sure about this because the hamburger icon position is calculated dynamically and it's gonna be practically impossible to have it align at all times.

even the old sidebar has this issue

hichemfantar:
I’ve done many tests across different browsers and OSes and I’m getting consistent results.
Will do some more but so far it’s really stable.
My changes are fairly standard to any browser to render consistently

hichemfantar:
@josevalim we can go ahead and change the text align to left (although i still prefer it to be center) in the tabs if that's your only concern because i believe i've addressed all your other concers.

hichemfantar:
i think i found the best implementation that looks good no matter how many elements there are

image

josevalim:
Hi @hichemfantar! As mentioned in a previous issue, such larges change to the design have a long term impact, because it makes the docs feel inconsistent, and in this case, I don't think the changes are worth the potential disruption and possibility of regressions. So we don't plan to go ahead with this one, thanks!

hichemfantar:
May I ask what exactly is disrupting about these changes? tested on multiple environments and functionality is kept exactly the same. Let me know if i missed anything 

before
image

after
image

josevalim:
Changing the user experience / ui is in itself a disruption. For example, GitHub has recently changed the issues tracker, and I am still getting used to it. However, in our case, we have an aggravating factor that all docs published in the last 6 years will have and retain the old style. When it comes to link styles and corners, those are minor changes, and tabs are used only from time to time, but changing several sidebar styles affects all docs and can be very disruptive.

hichemfantar:
I understand but this is not complete design and functionality overhaul, all the previous functionality is left unchanged and simply the readability and clarity.

here's the original tabs design preserved, now it's consistent with the one in the doc content.
Do you find this version agreeable in terms of tab design?

before
image

after
image

hichemfantar:
aligned the dropdown icon by default with the toast but it's impossible to have it align in other states. same issue exists in the old sidebar.

before
image

after
image

hichemfantar:
I'm willing to work with you on this to make the transition as painless as possible which i've handled pretty well thus far.
let me know

hichemfantar:
I understand, i'll make smaller scoped PRs that tackle individual issues

PR #2037, opened 2025-01-19
Make table responsive with block display and overflow properties

hichemfantar:
Adding block display and overflow properties to the table improves responsiveness and enhances the user experience on smaller screens.

bug
image

### DIFF
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -135,6 +135,8 @@
 .content-inner table {
   margin: 2em 0;
   border-collapse: collapse;
+  display: block;
+  overflow: auto;
 }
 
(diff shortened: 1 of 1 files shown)


assets/css/content/general.css
```diff
@@ -135,6 +135,8 @@
 .content-inner table {
   margin: 2em 0;
   border-collapse: collapse;
```
josevalim, line 137: Should we remove this one? I believe this only applies to `display: table`?

hichemfantar: those are valid styles regardless

josevalim:

PR #2038, opened 2025-01-19
Refactor blink-background animation keyframes for clarity

hichemfantar:
Now the animation transitions back to the initial state instead of stopping suddently


josevalim:

PR #2039, opened 2025-01-19
Remove uppercase transformation from table headers and add nowrap to content cells

hichemfantar:
Eliminate uppercase transformation for better readability and ensure content cells handle text without wrapping.

I couldn't find where the italic`em` is being applied so i can get rid of it

before
image

after
image

### DIFF
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -140,7 +140,6 @@
 .content-inner th {
   text-align: left;
   font-family: var(--sansFontFamily);
-  text-transform: uppercase;
   font-weight: 700;
   padding-bottom: 0.5em;
 }
@@ -167,6 +166,7 @@
   padding-left: 1em;
   line-height: 2em;
   vertical-align: top;
+  white-space: nowrap;
 }
 
 .content-inner .section-heading {


josevalim:

hichemfantar:
@josevalim any idea how to remove the italic styles? i couldn't find where to change the `em` element

josevalim:
Looking at the source, they choose to apply italics via markdown, we should not override their choice:

```

| _File Type_               | _File Name/Extension_ | _Documented in_                                     |
| ------------------------- | --------------------- | --------------------------------------------------- |
| Module                    | `.erl`                | [Erlang Reference Manual](`e:system:modules.md`)    |
| Include file              | `.hrl`                | [Erlang Reference Manual](`e:system:modules.md`)    |

```

hichemfantar:
Seems like a peculiar decision. I’ve worked with markdown rendering engines and I don’t think I’ve seen this before.
This is a separate package right? I’ll try and fix it upstream

josevalim:
It is not markdown specific, this was an author decision, and yeah, you could say it is not necessary, but I wouldn't sweat about it, as it seems intentional. The reason I agreed in removing the uppercase transformation is precisely because people could opt-in if they really want to, but if they do, that's their choice, not much we can do.

hichemfantar:
I’ll see if I can convince them. What can I say, accessibility is a passion

garazdawi:
> I’ll see if I can convince them. What can I say, accessibility is a passion

We are not very hard to convince :) any help with our docs is greatly appriciated! Thanks!

PR #2040, opened 2025-01-20
Reduce nav tab border width and adjust sidebar padding for improved layout

hichemfantar:
Decrease the navigation tab border width and remove unnecessary padding in the sidebar for a cleaner layout and more horizontal space.

before
image

after
image


josevalim:
@hichemfantar let's go ahead and reduce the left border of the active selection as well for consistency? I believe those two are meant to be in sync. :)

hichemfantar:
all good

josevalim:
Also, for a future somewhat related discussion, @DavidOliver @jonatanklosko @liamcmitchell @wojtekmach, which style of sidebar tabs do you prefer? The current ones or:

Screenshot 2025-01-20 at 18 18 14

Because if we want to change them, I'd do it as we thinner the borders.

hichemfantar:
> The left entries are no longer aligned:
> 
> Screenshot 2025-01-20 at 18 13 10
> We need to either bring everything a bit more to the left or the "Pages" a bit more to the right. Sorry for the pixel shaving but it feels this is the type of feedback you would want to give/receive. :)

fixed 🚀

josevalim:
Thank you, I am closing shop for today to focus on a new Elixir patch release, I should be back tomorrow. :)

hichemfantar:
alright so it's down to these 2

choice 1
image

choice 2
image

josevalim:
Let's go with v3. Do you already want to go ahead and change its border style in this PR or a future one? Please do not change the font style for now. :) Thank you!

DavidOliver:
Sorry for the delayed response. I agree the new style looks to work well, and I like that it matches the new style of content tabs. Title case instead of uppercase is also fine, in my opinion.

@hichemfantar, I expect you have focus (as well as hover) styles in place?

Thanks.

hichemfantar:
v3 applied

here's the focus state @DavidOliver 

image

PR #2041, opened 2025-01-20
Enhance sidebar styles for overflow and text display

hichemfantar:
Improve sidebar styles to handle overflow better and ensure text displays correctly without wrapping.

before
image

after
image

### DIFF
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -132,6 +132,8 @@
   display: flex;
   margin: 0;
   padding: 4px 4px 0 4px;
+  overflow: auto;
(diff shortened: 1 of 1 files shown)


josevalim:
Theoretically we don't have to worry about this at the moment, because you cannot have more than 3 elements anyway,  but I guess it doesn't hurt to merge this in. :)

hichemfantar:
🚀 always try to consider these use cases because you never know when it'll break and someones misses it because it looks good on their machine while it breaks on someone elses

josevalim:

PR #2042, opened 2025-01-20
Improve the font clarity of sidebar tab button styles for better accessibility

hichemfantar:
Removing the inherited font-weight from sidebar button styles improves the visual clarity and overall appearance of the sidebar.

this is an improvement in terms of accessibility as the font can be too thin to read on screens without good contrast levels and pixelating.

research:
https://accessibility.psu.edu/legibility/lightweight/

https://www.digitala11y.com/choosing-accessible-fonts-enhancing-readability-and-inclusivity/#:~:text=Font%20Weight,-Let's%20start%20with&text=But%20be%20cautious%3A%20overly%20bold,fonts%20can%20be%20problematic%20too.

image

image

image

image

fix applied to tabs
image

fix applied to everything
image

tldr: avoid light weight fonts as much as possible

### DIFF
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -3,7 +3,6 @@
   --sidebarLineHeight: 20px;
   font-family: var(--sansFontFamily);
   font-size: var(--sidebarFontSize);
-  font-weight: 300;
   line-height: var(--sidebarLineHeight);
   background-color: var(--sidebarBackground);
   color: var(--sidebarAccentMain);
@@ -12,10 +11,6 @@
   & .sidebar-tabpanel {
     scrollbar-width: thin;
   }
-}
-
-:root:not(.apple-os) .sidebar {
-  font-weight: 400; /* Non-Apple OSes render small light type too thinly */
 }
 
 .sidebar ul {
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -3,6 +3,7 @@
   --sidebarLineHeight: 20px;
   font-family: var(--sansFontFamily);
   font-size: var(--sidebarFontSize);
+  font-weight: 500;
   line-height: var(--sidebarLineHeight);
   background-color: var(--sidebarBackground);
   color: var(--sidebarAccentMain);
@@ -148,6 +149,7 @@
   -webkit-appearance: none;
   text-align: inherit;
   color: inherit;
+  font-weight: inherit;
   cursor: pointer;
   display: inline-block;
   line-height: 27px;
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -153,7 +153,6 @@
   -webkit-appearance: none;
   text-align: inherit;
   color: inherit;
-  font-weight: inherit;
   cursor: pointer;
   display: inline-block;
   line-height: 27px;


hichemfantar:
@josevalim i found the offending code https://github.com/hichemfantar/elixir-ex_doc-fork/blob/68429705e85cd39b579dafc68c1344b2f93f676f/assets/css/sidebar.css#L17-L20

it only looks very thin on macos because of this code.

DavidOliver:
I think the tabs look fine in the heavier weight you have.

For the second-level of links in the sidebar (Features, Usage, etc.), the text feels just a little too heavy to me in your fixed screenshot. Is there an in-between weight available for this level of links?

hichemfantar:
There's no in between weight from my testing. I believe we should add some extra vertical spacing and that will fix the issue because there's not enough space between the list items.

also what you're seeing here is actually what it already looks like in windows and linux because of this https://github.com/hichemfantar/elixir-ex_doc-fork/blob/68429705e85cd39b579dafc68c1344b2f93f676f/assets/css/sidebar.css#L17-L20

DavidOliver:
> There's no in between weight from my testing.

Ah, that's a shame. Let's see what others think about the weight.

> what you're seeing here is actually what it already looks like in windows and linux

Here's what the live ExDoc docs looks like for me in both Firefox and Vivaldi in Linux:

image

I believe it's usually the case that MacOS renders heavier than others (see, for example, [info on font dilation here](https://skip.house/blog/macos-font-rendering#font-dilation)), especially Windows. We previously attempted to counter this. It sounds like we're seeing different results for some reason.

I haven't looked into it for a while and I have to get back to other things for the next few hours, but I suppose it would be great if we could tell Macs to render like the others and then choose weights for all accordingly, if that didn't negatively affect Mac users.

DavidOliver:
Yes, I agree that things don't have to look the same across devices. I think in this case we felt either Mac or non-Mac users were getting a sufficiently poor weight, hence our attempt to close the gap.

Anyway, I should be able to test those browsers in Linux and browsers in Windows as necessary later if it's helpful, once others have confirmed which weight(s) they're happy to aim for.

hichemfantar:
awesome, i pushed the changes we talked so you could test what it looks like when the font weight is consistent across all OSes

josevalim:
I am using macos and the font-weight is definitely too heavy for me.

josevalim:
I am using macos and the font-weight is definitely too heavy for me. And that's definitely not something I am "used to".

Screenshot 2025-01-21 at 19 59 48

hichemfantar:
It's the opposite for me, but don't forget that the thin font (which is already uncomfortably thin imo) may become too thin on other apple devices.
this is an accessibility concern that's best left to the os and browser to handle natively

hichemfantar:
also it may feel a bit heavy for you because there may not be enough vertical spacing between elements

hichemfantar:
A slightly bold font is better than a light font because of the concerns and research mentioned in the original issue

hichemfantar:
even docusaurus uses the same font weight as vitepress which is 500

josevalim:
> Notice how vite and the new font weight are pretty much identical (further proof (besides the research i linked) it's the more correct weight to use because vitepress has a large team behind with lots of experience in building great web experiences)

Right, but we cannot look at things in a vacuum. Both use a lighter font color too, which helps balance things out, but at the same both Vitepress and Docusaurus gets only a AA check on their contrast.

josevalim:
Here is an example from Gmail. The Gmail font is a bit thicker than our current one, but it is definitely thinner than the one proposed here. And the Gmail team has definitely spent enough time and resources on UX. So I don't buy the argument that font-weight 500 is the recommended value, especially because there is variation between fonts and we cannot look at those things in isolation.

Screenshot 2025-01-21 at 21 36 58

Another option is to choose another font, that either has a slightly heavier Light mode or a slightly lighter Normal mode. It seems Lato is neither here nor there.

hichemfantar:
Funny you should mention gmail because i also find the gmail font weight to be a bit too thin haha

[Inter](https://fonts.google.com/specimen/Inter?query=inter) is a really well battle tested font or we can simply rely on native system fonts (that's what docusaurus does)?
native systems font are best because you don't have to wait for the font to load, they're pretty reliable, and the user is used to them.

josevalim:
I tried a couple founds and Nunito Sans seems to be the least disruptive. It is quite similar but thicker. Here are them side by side:

Screenshot 2025-01-21 at 22 23 58

But that's keeping the current font weight. Increase the weight, once again, make it too heavy for me.

hichemfantar:
Let me try with system fonts because that may just solve all of our problems

josevalim:
We moved most of it to system fonts in the past but we couldn't find a reasonable looking one for the sidebar. Let us know if you have better luck.

hichemfantar:
https://github.com/elixir-lang/ex_doc/pull/2054 my attempt at system fonts which looks pretty consistent imo

josevalim:
Closing in favor of #2053.

PR #2048, opened 2025-01-21
Update content width for improved accessibility

hichemfantar:
Adjust the content width variable to enhance the responsiveness of the layout.

w3 recommends `70em` which is a nice balance between showing the max amount of content (reduce scrolling) without compromising readability

https://www.w3.org/WAI/tutorials/page-structure/styling/#line-length

before
image

after
image

### DIFF
--- a/assets/css/custom-props/common.css
+++ b/assets/css/custom-props/common.css
@@ -1,6 +1,6 @@
 :root {
   /* Layout & Whitespace */
-  --content-width: 949px;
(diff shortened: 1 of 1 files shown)


josevalim:
The reasoning for 70em according to linked resource is to show 80 characters. Today we are showing 108, which is close to Elixir's formatter line length of 98, but this change extends it with ~20 more chars. I don't believe we should go ahead with it. Thank you.

PR #2049, opened 2025-01-21
Refactor table wrap for improved text handling

hichemfantar:
Only allow td wrapping to avoid headers overflowing because of long td cells and having to horizontally scroll to see the rest of the headers.

before
image

after
image


josevalim:

PR #2051, opened 2025-01-21
Center align project image and adjust project version layout

hichemfantar:
Center aligns the project image in the sidebar and modifies the layout of the project version form for improved aesthetics.

before
image

after
image

### DIFF
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -74,7 +74,6 @@
   word-wrap: break-word;
   display: block;
   width: calc(100% - 12px);
-  margin-bottom: 2px;
 }
 
 .sidebar .sidebar-projectVersion {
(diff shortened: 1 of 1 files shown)


hichemfantar:
perhaps set the logo top align?

image

josevalim:
I completely understand why you would want to do that, and that's something that I'd love to do too, but when we did so in the past, it broke the positioning of logos from several projects. They have published logos tailored to that precise space and positioning and changing that unfortunately breaks a ton of things. :(

josevalim:
I think the Phoenix one is actually an example of where it is worse with this change, while now it is neatly center aligned:

Screenshot 2025-01-21 at 20 10 42

josevalim:
I will go ahead and close this, unfortunately, because it is hard to do these changes without breaking projects. Thank you!

PR #2053, opened 2025-01-21
Use Nunito Sans instead of Lato

josevalim:
This increases readability in the sidebar,
decreases payload sizes for being variable.

image

cc @DavidOliver @hichemfantar


josevalim:
The goal is for this to replace #2042. We will still need to address #2043 using another technique.

josevalim:
You can download the zip above to try it out on your machine.

josevalim:
Please give me your opinions @hichemfantar and @DavidOliver if this is worse than keeping the current one. I need to do some padding adjustments to the sidebar tabs (above the purple border and below it), but other than that, this should feel pretty close.

hichemfantar:
It looks off compared to the body font, I think both should be the same font.

Look at how [tailwind](https://tailwindcss.com/docs/installation) perfectly uses Inter Variable on everything

josevalim:
@hichemfantar please let me know if you prefer this version or not. This is the only change I am considering for now.

DavidOliver:
Comparing Nunito Sans with Lato in the sidebar and for headings, I think I do prefer Nunito Sans. It loses a little character, if that makes sense, being more plain, but it has a nice chunkiness about it. And since the main text was changed to use the browser's set font a while back, it probably makes for a better pairing given that the main text will be sans serif for most users? Very subjective, of course.

In Linux here...

main:
image

jv-nunito-sans
image

josevalim:
I will avoid rocking the boat too much here. I moved it all to Lato 400 and enabled anti-aliasing for macOS.

josevalim:
@DavidOliver btw, I know you tried antialiasing in the past and it didn’t work for me on macOS. But for some miraculous reason it did work this time and it did allow us to simplify things! So thank you for teaching me the trick :)

DavidOliver:
@josevalim, oh, I did? This has all dissipated from my memory and I'd have to look up the details again. So those settings thin the type on Mac?

Glad you found something that worked.

josevalim:
@DavidOliver no need to dig back, they were very long convos. :)

PR #2054, opened 2025-01-21
Use system font for consistency, improved readability, and faster loading times

hichemfantar:
Standardize font weights and sizes across components to enhance readability and maintain consistency in typography.

[Docusaurus](https://docusaurus.io/) relies on system fonts and it looks pertty good and loads very quickly

looks great and consistent with all systems i've tested on (macos, window, ubuntu) tbh

if we get rid of these third party fonts then this saves 70.2kB from Lato

and 34.3kB from Inconsolata

look at how much time is wasted fetching these separate resources

image

image

### DIFF
--- a/assets/css/_html.css
+++ b/assets/css/_html.css
@@ -1,3 +1,6 @@
+@import "@fontsource-variable/inter";
+@import "@fontsource-variable/vazirmatn";
+
 @import "custom-props/common.css";
 @import "custom-props/theme-light.css";
 @import "custom-props/theme-dark.css";
--- a/assets/css/_html.css
+++ b/assets/css/_html.css
@@ -1,5 +1,6 @@
 @import "@fontsource-variable/inter";
 @import "@fontsource-variable/vazirmatn";
+@import "@fontsource-variable/open-sans";
 
 @import "custom-props/common.css";
 @import "custom-props/theme-light.css";
--- a/assets/css/_html.css
+++ b/assets/css/_html.css
@@ -1,8 +1,3 @@
-@import "@fontsource/lato/300.css";
-@import "@fontsource/lato/400.css";
-@import "@fontsource/lato/700.css";
-@import "@fontsource-variable/inconsolata";
-
 @import "custom-props/common.css";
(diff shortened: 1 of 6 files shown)


hichemfantar:
cc @josevalim

hichemfantar:
@DavidOliver i tested this on linux and it looked pretty good, feel free to test on your machine and report your experience

DavidOliver:
@hichemfantar, thanks. I'll try it out tomorrow.

hichemfantar:
Alright i removed the deps

hichemfantar:
you can see which font is being used by scrolling al the way down in the dev tools computed tab

image

hichemfantar:
added some extra vertical padding to separate them more clearly

josevalim:
Thank you. I am totally not convinced about the sidebar, even after playing with the font-weight and size a bit. It makes it feel there is no identity whatsoever and the contents get lost, as if it was text.

I am ok-ish with the monospace one but having different fonts means it is harder to format code snippets so they show up reasonably across all places. I think having a consistent experience for authors and readers is worth the 35kb.

hichemfantar:
I'm leaning towards Inter and Vazirmatn
Both are used by vitepress and both support variable fonts

https://fontsource.org/fonts/inter
https://fontsource.org/fonts/vazirmatn

hichemfantar:
pushed the new fonts so feel free to test them, lato seems to lack many weights btw

josevalim:
I tried Inter and I was not every happy with it. I will try https://fontsource.org/fonts/vazirmatn and let you know :)

hichemfantar:
pushed a new font
I think you're gonna love [Open Sans](https://fontsource.org/fonts/open-sans), it's an open variant of google product sans

josevalim:
I tried OpenSans, Raleway, MontSerrat and a whole bunch of them yesterday. I just tried vazirmatn now. I beleive the one that stays closest to the one we have while still improving the status quo is Nunito Sans in #2053, so let's move the discussion there. :) Thank you!

PR #2055, opened 2025-01-21
feat: Add "latest" and "related" search.

jimsynz:
Here's my first stab at adding latest and related search as per #2013 

I'm keen for early feedback, thus this draft PR.

I have a few questions:

1. Is there a standard function in the JS (I couldn't find one) to generate a link to hexdocs for a specific package name, version and ref?  I feel iffy about just having string interpolation in the code.
2. It looks like it might be difficult (but not impossible) to correctly generate the metadata (and thus the excerpts) from the response payload. Each hit contains a list of highlight snippets however it treats underscores as token separators so searching for `to_ast` is returns results for `to ast`.
3. It seems to return the same hits over and over again. Each returned hit has a different document ID so I think it's actually an index problem not a search problem. I _can_ dedup them by `ref` but I'd rather that the search index returned better results.

I haven't yet changed any of the markup as I wanted to get this step right before I go adding things like package names/versions to results, etc.


josevalim:
> Is there a standard function in the JS (I couldn't find one) to generate a link to hexdocs for a specific package name, version and ref? I feel iffy about just having string interpolation in the code.

There isn't because this is the first time we do something like this. However, since it comes from hexdocs, I believe it is fair to assume a naming scheme.

> Each hit contains a list of highlight snippets however it treats underscores as token separators so searching for to_ast is returns results for to ast.

@ruslandoga, is this something we can change in the indexer?

> It seems to return the same hits over and over again. Each returned hit has a different document ID so I think it's actually an index problem not a search problem. I can dedup them by ref but I'd rather that the search index returned better results.

The reason it returns the same results over and over again is because we break the documentation of a single function into multiple entries, one per h2 to be more precise. The benefit is that we can provide more precise links too. For now, I don't think we dedup them for regular results right? So I would not dedup them here, but we can dedup them later if we want to (we can even render it like Google results, where we show the main entry and below we refer to specific sections in the result).

---

This is fantastic progress. :)

garazdawi:
FYI, in the algolia for https://erlang.org/doc/search we disable `_` as punctuation just as  proposed here for the same reasons.

ruslandoga:
👋 

I couldn't find and options to make highlighting ignore the token separators. I think we might need to roll our own highlighter, similar to the one used in autocomplete: https://github.com/elixir-lang/ex_doc/blob/51cd42207f193d34172de79c8896985c787bfa38/assets/js/autocomplete/suggestions.js#L289-L311

assets/js/search-page.js
```diff
@@ -27,30 +28,85 @@ function initialize () {
...
     try {
-      // We cannot match on atoms :foo because that would be considered
-      // a filter. So we escape all colons not preceded by a word.
-      const fixedValue = value.replaceAll(/(\B|\\):/g, '\\:')
-      const results = searchResultsToDecoratedSearchItems(index.search(fixedValue))
+      let results = []
+      const searchNodes = getSearchNodes()
+
+      if (['related', 'latest'].includes(queryType) && searchNodes.length > 0) {
+        results = await remoteSearch(value, queryType, searchNodes)
```
ruslandoga, line 47: Can we can a race condition here? When the previous request returns after the current request and updates the items to stale results?

Might be helpful: https://github.com/ruslandoga/ex_doc/pull/1

assets/js/search-page.js
```diff
@@ -27,30 +28,85 @@ function initialize () {
...
+}
+
+async function remoteSearch (value, queryType, searchNodes) {
+  let filterNodes = searchNodes
+
+  if (queryType === 'latest') {
+    filterNodes = searchNodes.slice(0, 1)
+  }
+
+  const filters = filterNodes.map(node => `package:=${node.name}-${node.version}`).join(' || ')
```
ruslandoga, line 75: Unless we start using custom URLs like the one mentioned in https://github.com/hexpm/hexdocs/issues/49#issuecomment-2583622344, an array filter like this might help keep the URL shorter (no `package:=` repetition):

```js
if (nodes && nodes.length > 0) {
  const packages = nodes.map((node) => `${node.name}-${node.version}`);
  params.filter_by = `package:=[${packages.join(",")}]`;
}
```

Adapted from Typesense POC: https://gist.github.com/ruslandoga/d544addc4a17e1f4853d7e9ae97818a4

assets/js/search-page.js
jimsynz: Thanks. I'm sure you're right. As you can probably tell it's been almost a decade since I wrote any JavaScript so I'm still getting the hang of the new idioms.

jimsynz: So looking at the code more carefully, it appears that the `search` function is only called on page load, so should only be run once in the page's lifecycle.

jimsynz: Additionally, the search result handlebars template takes care of whether any results were actually returned.

jimsynz:
Okay so I've pushed some updates:

1. Uses the shorter package filter syntax in the API query.
2. Repurpose the existing highlighter code to highlight search terms in the API search results.

jimsynz:
I'll be moving on to adding the search type dropdown in the search bar next.

jimsynz:
search box

Okay, so I've added the search type to the UI, although you can probably tell that CSS is not my strong suit. It enables and disables the "latest" and "related" search based on the contents of `searchNodes` and `versionNodes`. It also completely disables the select box when the browser is not online.

josevalim:
Thank you @jimsynz. I have been a bit busy this week but I will try to add review and add polish to this feature this week.

jimsynz:
Thanks @josevalim 🙌

assets/js/search-page.js
```diff
@@ -26,30 +28,95 @@ function initialize () {
...
       renderResults({ value, errorMessage: error.message })
     }
   }
 }
 
+async function localSearch (value) {
+  const index = await getIndex()
+
+  // We cannot match on atoms :foo because that would be considered
+  // a filter. So we escape all colons not preceded by a word.
```
josevalim, line 65: Is this considered a filter by typesense?

lib/ex_doc/formatter/html/search_data.ex
```diff
@@ -2,22 +2,51 @@ defmodule ExDoc.Formatter.HTML.SearchData do
...
+          %{name: name, version: to_string(version)}
+
+        name when is_atom(name) ->
+          %{name: name, version: "*"}
+      end)
+
+    Enum.concat(search_nodes, related_packages)
   end
 
+  defp maybe_append_related_packages(search_nodes, _related_packages), do: search_nodes
```
josevalim, line 48: We should revert these changes for now, I believe the related packages will come from hex.pm itself. You can emulate it though by calling `mix docs` and then customizing the docs_config.js in doc yourself. :)

lib/ex_doc/formatter/html/templates/head_template.eex
```diff
@@ -21,7 +21,7 @@
       <link rel="canonical" href="<%= config.canonical %>" />
     <% end %>
     <script defer src="<%= Assets.rev config.output, "dist/sidebar_items-*.js" %>"></script>
-    <script defer src="docs_config.js"></script>
+    <script defer src="<%= Assets.rev config.output, "dist/search_data-*.js" %>"></script>
```
josevalim, line 24: This change should not be necessary either!

lib/ex_doc/formatter/html/search_data.ex
josevalim: Actually, we added our own for testing:

https://github.com/elixir-lang/ex_doc/blob/8a5217e9d1287d8b198f7b686dfdc83243f043d3/mix.exs#L127

So I think you can just change the `searchNodes` there, perhaps by adding Elixir itself or `earmark_parser` to searchNodes, so you can test it too.

assets/js/search-bar.js
```diff
@@ -137,6 +143,66 @@ function addEventListeners () {
...
+  return getSearchNodes().length > 1
+}
+
+function shouldEnableLatestSearch () {
+  const versionNodes = getVersionNodes()
+
+  if (versionNodes.length > 0) {
+    const latest = versionNodes[0]
+    const searchNodes = getSearchNodes()
+    const match = searchNodes.some(node => `v${node.version}` === latest.version)
```
josevalim, line 200: I don't think we need to check if latest or not. It is easier if the user doesn't have to think about it. Regardless if it we are in the latest package or not, we should allow latest to be chosen (unless there is no searchNodes). So basically, we will have this:

1. If searchNodes.length > 1, the related option is enabled, otherwise it is disabled
2. If searchNodes.length > 0, the latest option is enabled, otherwise it is disabled
3. Otherwise it is only current/local

If online, the default should be related > latest > current. If offline, it is the current one.

In other words, this should return `return versionNodes.length > 0`.

josevalim:
Thank you @jimsynz, I have dropped some comments. About the online vs offline, maybe instead of checking it upfront, we should try to run the search and, if the search fails for any reason, we fallback to regular search with an error message? I believe there is a toast we could use.

jimsynz:
Thanks for the review @josevalim - I'll get onto it.

assets/js/search-page.js
jimsynz: This came from the original implementation of local search. The filters passed off to the typesense API are correctly quoted due to the use of the browser's `URLSearchParams` API so unless I'm misunderstanding the question I'm 99% sure it's fine.

jimsynz:
I'm dreadfully sorry that it has taken this long for me to get back to this PR. I have rebased it and addressed most of the feedback. I haven't yet changed the offline-detection as suggested - but I wanted to get the PR in a good shape for when @zachdaniel runs out of patience with me and takes over :stuck_out_tongue_closed_eyes:

josevalim:
Hi @jimsynz, apologies for the delay on this as well, I decided to focus on the new hexdocs.pm homepage, as I knew it would influence the design here, and as consequence we even decided to go with a different approach that allows bringing custom search engines. Apologies for not going ahead with this after inviting you to contribute and thanks for your work nonetheless! :heart:

PR #2057, opened 2025-01-22
Add serve command for local documentation hosting

hichemfantar:
Introduce a new command in both `mix.exs` and `package.json` to facilitate local hosting of documentation using `http-server`.

i currently use https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer but i'd prefer if this was built-in like most dev tooling like vite does

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,9 +458,10 @@ The easiest way to test changes to ExDoc is to locally rebuild the app and its o
 
   1. Run `mix setup` to install all dependencies
   2. Run `mix build` to generate the docs and open up the generated `doc/index.html`
-  3. (optional) Run `npm run --prefix assets build:watch` if working on assets for automatic recompilation
-  4. Run `mix lint` to check linting and formatting (and `mix fix` to automatically fix it)
-  5. (important) Do not add the files in the `formatters/` directory to your commits, those will be handled by the maintainers
+  3. Run `mix serve` to serve the docs locally
+  4. (optional) Run `npm run --prefix assets build:watch` if working on assets for automatic recompilation
+  5. Run `mix lint` to check linting and formatting (and `mix fix` to automatically fix it)
+  6. (important) Do not add the files in the `formatters/` directory to your commits, those will be handled by the maintainers
 
 See the README in the `assets/` directory for more information on working on the assets.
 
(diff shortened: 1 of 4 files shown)


josevalim:
Do we need serve for the most common cases? The watcher works against the file system too, right?

josevalim:
Also, Elixir does have a way to serve them out of the box, so we should use that, instead of adding a new dependency. :)

hichemfantar:
> Do we need serve for the most common cases? The watcher works against the file system too, right?

yeah but you have to manually look for the index file, open it, and then manually refresh each time a change is made

> Also, Elixir does have a way to serve them out of the box, so we should use that, instead of adding a new dependency. :)

can you show an example?

i'm also thinking of creating a `dev` command that runs both the watcher auto compile and serves the docs at the same time

similar to what dev tooling like Vite does

hichemfantar:
[concurrently](https://www.npmjs.com/package/concurrently) would be perfect for this unless of a better way

josevalim:
@hichemfantar `:httpd.start(["doc/"])`

wojtekmach:
Its easy to support `mix build —-open` which would open browser with file://…/index.html or something along those lines.

I wonder what are pros/cons of using a web server for testing. Would there be things that work over http:// va file:// (so to speak) or vice versa? At the end of the day we need to support both because we have `mix hex.docs offline`.

If file watcher and auto rebuilding assets and those sort of things are what you are after then yes, something external to Elixir might be the most practical but if possible I would not add any new dependencies of any kind.

hichemfantar:
I added some changes and I'd like your feedback on the matter.

there is one odd thing, when i cancel the mix dev command, it seems the web server isn't killed which i think means `cmd` isn't properly sending the SIGTERM signal

hichemfantar:
Feel free to test the current setup and if it's to your liking then we can rewrite in elixir tooling ig

josevalim:
Thank you, I have added a TODO to look into adding this from the Elixir side itself. For now I will update the README steps.

hichemfantar:
awesome, a single `mix dev` command that both compiles the files on change and starts a local dev server would be great.

PR #2059, opened 2025-01-22
Use system monospace font

hichemfantar:
Eliminate the Inconsolata font dependency and adjust styles to use the system monospace font instead. This streamlines the font usage in the project and reduces page load times.

### DIFF
--- a/assets/css/custom-props/common.css
+++ b/assets/css/custom-props/common.css
@@ -9,7 +9,7 @@
   --defaultFontFamily: -apple-system, BlinkMacSystemFont, Helvetica, Arial, sans-serif,
     "Apple Color Emoji", "Segoe UI Emoji";
   --sansFontFamily: "Lato", sans-serif;
-  --monoFontFamily: "Inconsolata Variable", Menlo, Courier, monospace;
+  --monoFontFamily: monospace;
 
   /* Typography */
   --baseFontSize: 18px;
--- a/assets/css/_html.css
+++ b/assets/css/_html.css
@@ -1,6 +1,5 @@
 @import "@fontsource/lato/400.css";
 @import "@fontsource/lato/700.css";
-@import "@fontsource-variable/inconsolata";
 
 @import "custom-props/common.css";
 @import "custom-props/theme-light.css";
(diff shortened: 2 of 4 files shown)


assets/css/custom-props/common.css
```diff
@@ -9,7 +9,7 @@
   --defaultFontFamily: ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, Helvetica, Arial, sans-serif,
     "Apple Color Emoji", "Segoe UI Emoji";
   --sansFontFamily: "Lato", ui-sans-serif, system-ui, sans-serif;
-  --monoFontFamily: "Inconsolata Variable", Menlo, Courier, monospace;
+  --monoFontFamily: monospace;
```
josevalim, line 12:
```suggestion
  --monoFontFamily: ui-monospace, monospace;
```

josevalim:
We use monospaced in other places and they need to be double checked too. For example, this is definitely too big now:

Screenshot 2025-01-22 at 22 02 46

garazdawi:
Not sure if this has any bearing on anything, but I had to change monospace font [in this page](https://www.erlang.org/docs/28/apps/stdlib/terminal_interface.html) in order for the first code segment to render properly.

josevalim:
Change what, where and how? :)

josevalim:
Oh, I get it, Inconsolata may not have all Unicode codepoints. That's rough. That's a good reason to use the system one. Let's give it another try then.

garazdawi:
I added this IAL: https://github.com/erlang/otp/blob/master/lib/stdlib/doc/guides/terminal_interface.md?plain=1#L44 and then this in the config: https://github.com/erlang/otp/blob/master/make/ex_doc.exs#L231-L233.

And yes, as you discovered, the default font does not render all unicode things properly.

josevalim:
Ok, i have changed it to use system fonts and adjusted the font sizes, thank you @hichemfantar and @garazdawi!

@garazdawi, if you find further hiccups, please let me know!

garazdawi:
thanks!

> @garazdawi, if you find further hiccups, please let me know!

will do :)

hichemfantar:
awesome, relying on system fonts is always the safest choice

hichemfantar:
@garazdawi feel free to report any other bugs you encountered so we can fix them upstream

PR #2064, opened 2025-01-23
Increase padding in code block styles

hichemfantar:
Enhance the visual appearance of code blocks by increasing the padding.

before
image

after
image


josevalim:
I have moved the button in main. The system monospace font already increased the padding for me locally and I would prefer to keep it tighter. Thanks.

PR #2065, opened 2025-01-23
Refactor font sizes to use CSS variables for consistency

hichemfantar:
Consistently apply CSS variables for [font sizes](https://v3.tailwindcss.com/docs/font-size) across the application to enhance maintainability and ensure a unified design.

No size changes applied, custom values remain.

### DIFF
--- a/assets/css/autocomplete.css
+++ b/assets/css/autocomplete.css
@@ -172,7 +172,7 @@
   background-color: var(--autocompletePreview);
   cursor: pointer;
   padding: 4px 8px;
-  font-size: 14px;
+  font-size: var(--text-sm);
 }
 
 .autocomplete-preview-indicator button:hover {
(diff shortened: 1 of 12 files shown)


josevalim:
@hichemfantar thank you. We have places where we use `em` or `rem` fonts. Should we align those as well?

hichemfantar:
those places use custom uncommon sizes, it'd be difficult to create variables for those unless you wish to change them to use the [standard values](https://v3.tailwindcss.com/docs/font-size) i've applied

josevalim:
> those places use custom uncommon sizes

We use a lot of them for headers, which I agree we don't want to change. But we also use them in autocomplete, app-vsn and other places that could benefit from standardization. So maybe we try to standardize everything except monofonts, icons, and headers?

hichemfantar:
There's a a lot of variations present, it's best to tackle them individually and gradually

josevalim:
Ok, good point. We just need to remember to resort mostly to these new vars then.

josevalim:

PR #2066, opened 2025-01-23
Fix font cascading behavior and clean up CSS by removing unused variables and redundant declarations

hichemfantar:
Eliminate unused variables and redundant font family declarations to streamline the CSS files.

this pr introduces no visual changes but fixes font hierarchy because sansFontFamily which is meant for titles and the sidebar was being applied on the entire site and this fixes that so now we have proper font cascading behavior where the default font is actually the default

normalize already sets things globally https://github.com/sindresorhus/modern-normalize/blob/ba66c6fe35c7ae33954120c73e69d57de4249a0b/modern-normalize.css#L26

### DIFF
--- a/assets/css/layout.css
+++ b/assets/css/layout.css
@@ -12,7 +12,6 @@ body {
   background-color: var(--background);
   color: var(--textBody);
   font-size: var(--text-base);
-  font-family: var(--sansFontFamily);
   line-height: 1.6875em;
 }
 
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -1,5 +1,4 @@
 .content-inner {
-  font-family: var(--defaultFontFamily);
   font-size: 1em;
   line-height: 1.6875em;
   position: relative;
--- a/assets/css/content/summary.css
+++ b/assets/css/content/summary.css
@@ -21,7 +21,6 @@
 }
 
 .content-inner .summary .summary-row .summary-synopsis {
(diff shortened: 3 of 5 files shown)


assets/css/layout.css
```diff
@@ -12,7 +12,6 @@ body {
   background-color: var(--background);
   color: var(--textBody);
   font-size: var(--text-base);
-  font-family: var(--sansFontFamily);
```
josevalim, line 15: Isn’t it a good practice to declare something in the body?

hichemfantar: normalize already [does that](https://github.com/sindresorhus/modern-normalize/blob/ba66c6fe35c7ae33954120c73e69d57de4249a0b/modern-normalize.css#L26)

hichemfantar:
now we can get rid of defaultFontFamily var because we have proper font cascading behavior and normalizecss handles the default font for us in https://github.com/sindresorhus/modern-normalize/blob/ba66c6fe35c7ae33954120c73e69d57de4249a0b/modern-normalize.css#L26

hichemfantar:
and thanks to this fix we can avoid awkward  overrides like this [one](https://github.com/elixir-lang/ex_doc/pull/2066/files#diff-f84fe326a1636182dc0392140d35848057327b1310cda5c608fb79fe99c403bd)

hichemfantar:
again this pr introduces no visual changes but only fixes font cascading

josevalim:

PR #2067, opened 2025-01-24
Add 'favicon' config option

arcanemachine:
This pull request adds favicon support to the HTML formatter.

Much of the functionality was added by copying and renaming some of the 'logo' config functionality, so the code is very similar. (This pull request doesn't make any attempt to add favicons to the ePub formatter as I have no idea what the story is there.)

Favicon support seems to be a minefield when you start [digging into the issue](https://stackoverflow.com/questions/48956465/favicon-standard-2024-svg-ico-png-and-dimensions), but this is just a basic implemention. The output will look like this if the favicon is enabled (The file extension will change as needed, and supports the same filetypes as the `logo` option):

```
<link rel="icon" href="assets/favicon.png" />
```

Tested (manually) to work with JPG/PNG/SVG in Chrome and Firefox, and with JPG/PNG in Safari iOS 18.1 (Safari doesn't currently support SVG favicons).


josevalim:

PR #2069, opened 2025-01-24
Add exdoc:loaded event

### LINKED ISSUE #2068: Rendering Mermaid graph after navigation
Opening [Task.html](https://hexdocs.pm/elixir/Task.html) documentation and then navigating to the [Supervisor.html](https://hexdocs.pm/elixir/Supervisor.html) docs via the sidebar does not trigger mermaid and therefore the mermaid graph is not visible - refreshing the page will trigger mermaid and render the graph:

To display the graph when navigating from the sidebar mermaid would have to be triggered. Example code that achieves this:

```js
<script>
  function mermaidLoaded() {
    mermaid.initialize({
      startOnLoad: false,
      theme: document.body.className.includes("dark") ? "dark" : "default"
    });
    const runMermaid = () => {
      let id = 0;
      for (const codeEl of document.querySelectorAll("pre code.mermaid")) {
        const preEl = codeEl.parentElement;
        const graphDefinition = codeEl.textContent;
        const graphEl = document.createElement("div");
        const graphId = "mermaid-graph-" + id++;
        mermaid.render(graphId, graphDefinition).then(({svg, bindFunctions}) => {
          graphEl.innerHTML = svg;
          bindFunctions?.(graphEl);
          preEl.insertAdjacentElement("afterend", graphEl);
          preEl.remove();
        });
      }
    }
    const observer = new MutationObserver(runMermaid);

    // re run mermaid each time we navigate via sidebar
    observer.observe(document.querySelector("#sidebar"), {
      attributes: true,
      subtree: true,
      childList: true,
    });

    runMermaid();
  }
</script>
<script async src="https://cdn.jsdelivr.net/npm/mermaid@10.2.3/dist/mermaid.min.js" onload="mermaidLoaded();"></script>
```

josevalim:
Closes #2068.

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,7 +370,7 @@ docs: [
 ]
 ```
 
-On the JavaScript side, ExDoc emits the "exdoc:loaded" event. This event may be called multiple times, as you navigate across pages, so initialization that should happen only once must be conditional. We recommend external scripts to use `def`, not async, as shown in the examples below.
+On the JavaScript side, ExDoc emits the `"exdoc:loaded"` event. This event may be called multiple times, as you navigate across pages, so initialization that should happen only once must be conditional. We recommend external scripts to use `defer`, not `async`, as shown in the examples below.
 
 ### Rendering Math
 
--- a/README.md
+++ b/README.md
@@ -381,13 +381,17 @@ If you write TeX-style math in your Markdown, such as `$\sum_{i}^{N} x_i$`, it e
 <link href="https://cdn.jsdelivr.net/npm/katex-copytex@1.0.2/dist/katex-copytex.min.css" rel="stylesheet" type="text/css">
 <script defer src="https://cdn.jsdelivr.net/npm/katex-copytex@1.0.2/dist/katex-copytex.min.js" crossorigin="anonymous"></script>
 
-<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.4/dist/contrib/auto-render.min.js" integrity="sha384-+VBxd3r6XgURycqtZ117nYw44OOcIax56Z4dCRWbxyPt0Koah1uHoK0o4+/RRE05" crossorigin="anonymous"
-  onload="renderMathInElement(document.body, {
-    delimiters: [
-      {left: '$$', right: '$$', display: true},
-      {left: '$', right: '$', display: false},
-    ]
-  });"></script>
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.4/dist/contrib/auto-render.min.js" integrity="sha384-+VBxd3r6XgURycqtZ117nYw44OOcIax56Z4dCRWbxyPt0Koah1uHoK0o4+/RRE05" crossorigin="anonymous"></script>
+
+<script>
+  window.addEventListener("exdoc:loaded", () => {
+    renderMathInElement(document.body, {
+      delimiters: [
+        {left: '$$', right: '$$', display: true},
+        {left: '$', right: '$', display: false},
+      ]
+    })
+  })
 </script>
 ```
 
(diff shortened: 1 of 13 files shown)


assets/js/content.js
```diff
@@ -6,7 +6,7 @@ import { settingsStore } from './settings-store'
  * corresponding to the current documentation page.
  */
 
-window.addEventListener('swup:page:view', initialize)
+window.addEventListener('exdoc:loaded', initialize)
 initialize()
```
jonatanklosko, line 10: Since `exdoc:loaded` is also dispatched on `DOMContentLoaded`, it means we don't need the extra `initialize` call:

```suggestion
```

jonatanklosko:
Just one commend and looks good to me. @liamcmitchell please let us know if you have any concerns, thanks!

It may be worth documenting `exdoc:loaded` explicitly in the readme section, and say that: (a) it may run multiple times, so initialization that should happen once must be conditional; (b) for external scripts they should use `defer`, but not `async`.

assets/js/content.js
josevalim: This will delay all initialization but I guess we are fine? cc @liamcmitchell

README.md
```diff
@@ -370,6 +370,8 @@ docs: [
 ]
 ```
 
+On the JavaScript side, ExDoc emits the "exdoc:loaded" event. This event may be called multiple times, as you navigate across pages, so initialization that should happen only once must be conditional. We recommend external scripts to use `def`, not async, as shown in the examples below.
```
jonatanklosko, line 373:
```suggestion
On the JavaScript side, ExDoc emits the `"exdoc:loaded"` event. This event may be called multiple times, as you navigate across pages, so initialization that should happen only once must be conditional. We recommend external scripts to use `defer`, not `async`, as shown in the examples below.
```

josevalim:
@garazdawi you will have to update Erlang. I plan to release v0.37.0-rc.0 this weekend, which should also be compatible with Erlang/OTP 28.

assets/js/content.js
liamcmitchell: I don't see a problem with these calls being deferred 👍

assets/js/sidebar/sidebar-drawer.js
```diff
@@ -13,7 +13,7 @@ const smallScreenQuery = window.matchMedia(`screen and (max-width: ${SMALL_SCREE
 if (!isEmbedded) {
   setDefaultSidebarState()
```
liamcmitchell, line 14: This is now duplicated and could be removed

assets/js/sidebar/sidebar-list.js
```diff
@@ -89,7 +89,7 @@ export function initialize () {
   })
 
   window.addEventListener('hashchange', markCurrentHashInSidebar)
-  window.addEventListener('swup:page:view', markCurrentHashInSidebar)
+  window.addEventListener('exdoc:loaded', markCurrentHashInSidebar)
 
   markCurrentHashInSidebar()
```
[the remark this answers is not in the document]
josevalim: Thank you. I will push it soon. And to be clear, the next line does:

```
  // Triggers layout, defer.
  requestAnimationFrame(scrollNodeListToCurrentCategory)
```

That would still be fine, right?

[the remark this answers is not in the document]
josevalim: Ok, I can confirm this one breaks it and nothing is being marked on the sidebar. :)

assets/js/sidebar/sidebar-drawer.js
josevalim: I found no problems with this one. :)

assets/js/sidebar/sidebar-list.js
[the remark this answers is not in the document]
josevalim: I will ship a new version with this still using `swup:page:view`. If you have suggestions on how to unify it, it would be very welcome. :) Thank you :heart:

josevalim:

assets/js/sidebar/sidebar-list.js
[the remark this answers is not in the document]
liamcmitchell: Sorry, I missed that this init is lazy and will always come after the first `exdoc:loaded`. Pushed changes to unify and make more clear in https://github.com/elixir-lang/ex_doc/pull/2070

PR #2070, opened 2025-01-25
Remove remaining swup:page:view listeners

liamcmitchell:
Re: https://github.com/elixir-lang/ex_doc/pull/2069#discussion_r1929493618

The sidebar-list is lazy-rendered so will always be called after the first `exdoc:loaded` event.

I re-arranged a few lines and added some comments to make it more obvious that it's not the standard `exdoc:loaded` -> `init()` pattern.

### DIFF
--- a/assets/js/sidebar/sidebar-list.js
+++ b/assets/js/sidebar/sidebar-list.js
@@ -1,6 +1,8 @@
 import { el, getCurrentPageSidebarType, qs, qsAll } from '../helpers'
 import { getSidebarNodes } from '../globals'
 
+// Sidebar list is only rendered when needed.
+// Mobile users may never see the sidebar.
 let init = false
 export function initialize () {
   if (init) return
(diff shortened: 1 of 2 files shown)


assets/js/sidebar/sidebar-list.js
```diff
@@ -88,15 +90,15 @@ export function initialize () {
...
-  window.addEventListener('swup:page:view', markCurrentHashInSidebar)
+  // Update new sidebar list with current hash.
   markCurrentHashInSidebar()
 
   // Triggers layout, defer.
   requestAnimationFrame(scrollNodeListToCurrentCategory)
+
+  // Keep updated with future changes.
+  window.addEventListener('hashchange', markCurrentHashInSidebar)
+  window.addEventListener('exdoc:loaded', markCurrentHashInSidebar)
```
josevalim, line 101: Doesn't this mean `markCurrentHashInSidebar` will be invoked two times on boot? One here and another on DOMContentLoaded? Or is that fine?

liamcmitchell: Calling multiple times is fine but I don't think it should happen. 
On page load this would be called by the sidebar init fn which is now triggered by ex doc:loaded, meaning the event has already fired at this point and won't fire again until SWUP nav.
That's why it broke when you tried it. Adding the listener here is too late for the first render.

josevalim: Got it! Thanks for clarifying and sorry for being a slowpoke!

josevalim:

PR #2074, opened 2025-01-27
Bump Erlang and Elixir version used to create release escript

garazdawi:
The Elixir bump is needed in order for ExDoc to not use any pre-compiled regexps, and the Erlang bump is just because bumping is always good.. :)


josevalim:

garazdawi:
When you have the time, could we have an rc1 with this change?

PR #2075, opened 2025-01-27
Update bottom actions button colors for themes

hichemfantar:
Adjust colors for bottom action buttons to improve visibility and consistency across dark and light themes.

image

### DIFF
--- a/assets/css/content/bottom-actions.css
+++ b/assets/css/content/bottom-actions.css
@@ -30,7 +30,7 @@
 }
 
 .content-inner .bottom-actions .bottom-actions-button .title {
-  color: var(--mainLightest);
+  color: var(--bottomActionsBtnTitle);
 }
 
 .content-inner .bottom-actions .bottom-actions-button[rel="prev"] {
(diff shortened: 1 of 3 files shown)


hichemfantar:
I suppose the red border is the issue here

josevalim:
And here is the other option:

Screenshot 2025-01-27 at 16 05 52

I am thinking the current version strikes the best balance. The best way to think about it is that it is a button rather than a link.

hichemfantar:
i think it's good now

hichemfantar:
lgtm

DavidOliver:
LGTM, too, if we're not using red for Erlang's inner content links. I agree it's best to keep the red and blue away from each other in this case.

Thanks!

garazdawi:
I don’t have a preference, go with whichever you think works best.

josevalim:
@DavidOliver I know you said it looks good to you but do you _prefer_ this one?

DavidOliver:
@josevalim, I prefer it to combining a red outline and blue link text. Is that what you mean?

josevalim:
If you prefer the one currently in main or this one. :)

DavidOliver:
@josevalim, sorry. :smile: I prefer this one, with the page title in the language colour.

PR #2076, opened 2025-01-27
Ensure top heading action icon is to right when width allows

DavidOliver:
Only `top-heading` headings are affected, and only when content is wider than 600px.

Callback/function headings have not been changed; they continue to always wrap so as to prioritise text display.

A container has been registered in the layout styles, meaning that we can easily conditionally apply styles based on the content width irrespective of whether or not the sidebar is shown.

https://github.com/user-attachments/assets/59fd72b6-de59-472c-9a64-753d18ea8778

### DIFF
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -41,7 +41,7 @@
     gap: 32px;
     align-items: flex-start;
     .icon-action {
-      padding-top: 1.7rem; /* vertically aligned with first line of heading x-height */
+      padding-top: 1.7rem; /* vertically align with x-height of first line of heading */
     }
   }
 }
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -34,6 +34,17 @@
 .content-inner .heading-with-actions.top-heading .icon-action {
   font-size: 1.2rem;
 }
+/* When content is wide enough, ensure action icon is to the right of, not below, heading */
+@container content (width > 600px) {
+  .content-inner .heading-with-actions.top-heading {
+    flex-wrap: nowrap;
+    gap: 32px;
+    align-items: flex-start;
+    .icon-action {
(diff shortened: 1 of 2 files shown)


josevalim (approved):

assets/css/content/general.css
```diff
@@ -34,6 +34,17 @@
 .content-inner .heading-with-actions.top-heading .icon-action {
   font-size: 1.2rem;
 }
+/* When content is wide enough, ensure action icon is to the right of, not below, heading */
+@container content (width > 600px) {
+  .content-inner .heading-with-actions.top-heading {
+    flex-wrap: nowrap;
+    gap: 32px;
+    align-items: flex-start;
+    .icon-action {
```
josevalim, line 43: I think this is invalid, as we are not using a preprocessor, so I want ahead and remove it, as it seems not necessary. Please double check.

DavidOliver: Nested selectors are now supported in CSS. This ensures the icon isn't too close to the text.

DavidOliver: Correcting my edited comment: the padding-top is necessary for vertical alignment.

DavidOliver: (When the heading runs into two lines or more.)

josevalim: Oh, fantastic, I will add it back! esbuild is complaining, so we may need to either update it or move it to the top. Thanks!

josevalim: @DavidOliver this seems to do the trick: https://github.com/elixir-lang/ex_doc/commit/b0a2412936379aca6662dd9d270501c7dc1b2ad9

DavidOliver: Great - thanks!

PR #2078, opened 2025-01-27
Increase code font size

hichemfantar:
Adjust font sizes for better readability and refactor CSS to utilize variables for consistency across the styles.

The font is smaller that before on windows and is harder to read, bringing it up to 14px makes it easier to read

### DIFF
--- a/assets/css/content/functions.css
+++ b/assets/css/content/functions.css
@@ -59,7 +59,7 @@
 
 .content-inner .specs pre {
   font-family: var(--monoFontFamily);
-  font-size: var(--text-xs);
+  font-size: var(--text-sm);
   font-style: normal;
   line-height: 24px;
   white-space: pre-wrap;
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -241,7 +241,7 @@
   font-style: normal;
   line-height: 24px;
   font-weight: 400;
-  font-size: 13px;
+  font-size: 14px;
 }
 
 @media screen and (max-width: 768px) {
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -241,7 +241,7 @@
   font-style: normal;
   line-height: 24px;
   font-weight: 400;
-  font-size: 14px;
+  font-size: var(--text-sm);
 }
 
 @media screen and (max-width: 768px) {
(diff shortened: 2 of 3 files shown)


josevalim:
Thank you but the proposed sizes are definitely too big for macOS. If balancing system fonts size across OSes is going to be complicated, we should revert back to Inconsolata.

hichemfantar:
I don't know what settings you're working with but i'm on default and 14px looks similar to the regular text size.

can you share a screenshot?

josevalim:
Here is a possible solution:

```
@font-face {
  font-family: '...';
  size-adjust: 90%;
}
```

And we adjust the size for either the macOS or the Windows font.

josevalim:
Screenshot 2025-01-28 at 00 17 11

hichemfantar:
how is that too big? the left one looks the same as the rest of the text, the right one looks smaller than the rest of the text.

josevalim:
@hichemfantar the `@callback ...` one is the issue. Could it perhaps be made to be `13px` or would that be too small on Windows? It feels like it would. :(

hichemfantar:
the left screenshot is from the current pr, right?

josevalim:
> the right one looks smaller than the rest of the text.

That's the whole point of my comment though. As I said the previous time this was brought up, those tend to consume a lot of space from functions and arguments with long names.

And the callback one is smaller on purpose because they tend to be very long. Here is just one example: https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#stream/4

I will revert to Inconsolata tomorrow morning.

> the left screenshot is from the current pr, right?

Yes.

hichemfantar:
Please don't revert to a third party font, the whole reason we moved is because it doesn't support all characters properly and it was kinda heavy in size.

1px for improved readability is a much better solution than undoing all the improvements we made.

josevalim:
Do you think we can have  @font-face declaration that adjusts it on Windows then?

```
@font-face {
  font-family: '...';
  size-adjust: 90%;
}
```

I want to avoid rehashing discussions on the font size, it is not productive. So we either default to a single font or we figure out a way to make it work without having to adjust the sizes for every OS.

josevalim:
@hichemfantar I pushed some of those changes and other fixes related to font sizes, including increasing line length so tey get more breathing room. However, I kept `.specs pre` at 13px. If those are too small on Windows, please provide a font face config. Thanks!

josevalim:
I downloaded several monospace fonts and tried them one by one. Consolas is the only one too small, so I have adjusted its font-face accordingly.

hichemfantar:
@josevalim late but thanks for the update!

PR #2079, opened 2025-01-28
Fix so that `fun()` in Erlang does not try to generate any link

garazdawi:
Turns out that there actually was a test for this, but as the codepath was Elixir 1.18 only it was never tested in CI. So this PR adds Elixir 1.18 to the CI matrix and fixes the fun problem.

Fixes regression in #1980

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,8 +81,10 @@ jobs:
           # Test Erlang without -doc attribute support
           - elixir: "1.16"
             otp: "26"
-          # Test Erlang with -doc attribute support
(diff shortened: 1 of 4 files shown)


josevalim:

garazdawi:
@josevalim When you have the time it would be great with another rc with this change. I've now built ex_doc locally and everything works as it should.

josevalim:
RC2 is out. Once you are all happy with it let me know and I can ship final!

garazdawi:
Thanks! I will test it now.

garazdawi:
Testing takes a while, but now it is all done and working as it should. Thanks!

PR #2083, opened 2025-01-28
Increase margin-bottom of last textual element in tabset content

DavidOliver:
Increases margin of last element unless it's a `pre` (code box).

Left: before. Right: after.

image

### DIFF
--- a/assets/css/tabset.css
+++ b/assets/css/tabset.css
@@ -71,6 +71,14 @@
   }
 }
 
+/* tabpanel content: top margin of first and bottom margin of last */
+@media screen and (max-width: 768px) {
+  .tabset-panel > :is(:first-child) {
+    &:is(table) {
+      margin: .5em 0;
+    }
+  }
+}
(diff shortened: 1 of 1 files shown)


josevalim:
Should we check for `pre`? Or should we check for the last child being a `p`? For example, what if it is admonition? Or a table? Or an image?

DavidOliver:
I believe most other types of content should benefit from the 1.25em margin, but I'm setting up some tests now and will report back soon.

DavidOliver:
Good call - there were improvements to make. I've set margins for the first and last children of some different kinds. This leaves spacing between elements unaffected.

image

I'm going to look at margins here for narrow viewports now.

josevalim (approved):
It seems quite some work for small adjustments but if you are happy with it, ship it!

DavidOliver:
For the narrow viewport widths, I've just tightend up the first-child table margin-top and the last-child table margin-bottom.

PR #2089, opened 2025-02-05
Support for umbrella projects via CLI

### LINKED ISSUE #2071: Native support for "umbrella" Erlang projects
It is common for Erlang projects to be composed by several distinct applications. This is usually referred to as an [umbrella (or release) project](https://rebar3.org/docs/basic_usage/). Since `ex_doc` operates at application level, this means that the user has to figure out a way to produce an index for the project itself.

In the case of a `rebar3` project, this is partially solved by the [rebar3_ex_doc](https://github.com/jelly-beam/rebar3_ex_doc) plugin, which has [basic support for umbrella projects](https://github.com/jelly-beam/rebar3_ex_doc?tab=readme-ov-file#umbrella-support). For Erlang projects not using rebar3, this has to be solved case by case.

The most notorious example is [Erlang/OTP itself](https://www.erlang.org/doc/readme.html), where an index of all modules is generated [here](https://github.com/erlang/otp/blob/master/system/doc/top/src/otp_man_index.erl), and a fake _index_ application is [defined](https://github.com/erlang/otp/blob/debbc7b267a7b9799c9f955953d737e3cf919dc5/system/doc/top/docs.exs#L11) and [used](https://github.com/erlang/otp/blob/83882095f07c72844646572b0049dde544a2102f/make/ex_doc.exs#L32) to get the feeling of a single projects. This approach requires custom, duplicated logic for various projects.

Was it ever considered to add native support for umbrella projects to `ex_doc` itself? I believe it would simplify its usage outside of rebar3 and increase its adoption.

robertoaloi:
While the `ExDoc` API already supports _umbrella projects_, the functionality is not exposed through the `ex_doc` CLI, making it hard to create sites for projects that include more than one application.

Here we propose a simple change, allowing users to specify multiple `ebin` directories when invoking `ex_doc`. If multiple `ebin` directories are specified and no custom `groups_per_modules` option is specified, we automatically group modules by application, which seems to be a sensible default.

Fixes #2071

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,8 @@ It is also possible to specify multiple `ebin` directories in the case of _umbre
    $ ex_doc "PROJECT_NAME" "PROJECT_VERSION" _build/dev/lib/app1/ebin _build/dev/lib/app2/ebin -m "PROJECT_MODULE" -u "https://github.com/GITHUB_USER/GITHUB_REPO" -l path/to/logo.png
    ```
 
+If multiple `ebin` directories are specified, modules are grouped by application by default. It is possible to override this behaviour by providing a custom `groups_per_modules` option.
+
 You can specify a config file via the `--config` option, both Elixir and Erlang formats are supported. Invoke `ex_doc` without arguments to learn more.
 
 <!-- tabs-close -->
--- a/README.md
+++ b/README.md
@@ -114,6 +114,12 @@ You can use ExDoc via the command line.
    GITHUB_REPO     => ecto
    ```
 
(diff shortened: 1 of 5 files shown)


josevalim:
Thank you! I also think we need one test in ex_doc/config_test.exs for the group_for_modules config and we should be good to go!

robertoaloi:
Added some tests to verify all possible combinations. Let me know if that's what you had in mind as well.

josevalim:

PR #2092, opened 2025-02-18
Fix code highlighting for languages with non-alphanumeric characters

jonatanklosko:
Currently `` ```c++ `` is not highlighted, because `+` does not match `\w` in the regex. We could allow list `+`, but I relaxed the regex to pick up any class name.


josevalim:

PR #2097, opened 2025-03-17
Automatically close the sidebar when navigating sections on mobile

mcky:
I've been reading quite a few ExDoc sites from my mobile lately and I've found it slightly frustrating that clicking on some headings causes a full page change and the nav is closed, but within a module the sidebar stays open and it's unclear that you've even navigated

An example of the previous behaviour:

https://github.com/user-attachments/assets/b8e0090b-b1eb-4e5d-a101-4dd8bcb5c88a


DavidOliver:
@josevalim, is it okay with you to have the nav bar close on nav link click when mobile/narrow viewport?

josevalim:
@DavidOliver i am 80% sure this was the case in the past and it was changed because it would be undesired in many cases. It can double check later.

josevalim:
Given we always close when navigating away, I think we should remain consistent, so let's give it a try!

mcky:
Thanks folks! 🎉

PR #2103, opened 2025-03-31
Allow listing outside URLs in extras

### LINKED ISSUE #2084: Allow urls as extra
I would like to have the possibility to have URLs listed as extras in order to have links outside of the local app as part of the sidebar. For example:

```
extras: [
  "Wikipedia": [url: "https://wikipedia.com"]
]
```
Would create a sidebar item called `Wikipedia` that links directly to `https://wikipedia.com`. See #2081 for a bit of more context.

sorentwo:
This makes it possible to delcare URLs as extras and have them listed as links in the sidebar. For example, to set a "Wikipedia" url:

```elixir
extras: [
  "Wikipedia": [url: "https://wikipedia.com"]
]
```

This is currently light on some validations. I'll expand the tests and URI validation if the approach is improved. 

- [x] Handle urls in epub
- [x] ~Validate the `url` value is an actual URI~
- [x] ~Expand formatter tests~
- [x] Document usage of `url` options

Closes #2084


lib/ex_doc/formatter/html.ex
```diff
@@ -384,10 +386,31 @@ defmodule ExDoc.Formatter.HTML do
     end)
   end
 
+  defp normalize_extras(base) when is_binary(base), do: {base, %{}}
+  defp normalize_extras({base, opts}), do: {base, Map.new(opts)}
```
sorentwo, line 390: This is purely for pattern matching to flatten out `build_extra/5`. The alternative was a cond in `build_extra/5`, but this seemed a little clearer.

lib/ex_doc/formatter/html.ex
```diff
@@ -384,10 +386,31 @@ defmodule ExDoc.Formatter.HTML do
...
+    title = input_options[:title] || filename_to_title(input)
+    group = GroupMatcher.match_extra(groups, input)
+
+    # TODO: Can we make the content/source a link?
+
+    %{
+      group: group,
+      id: Utils.text_to_id(title),
+      source_path: input_options[:url],
+      source_url: input_options[:url],
```
sorentwo, line 407: These probably aren't necessary now.

lib/ex_doc/formatter/html/search_data.ex
```diff
@@ -18,36 +18,8 @@ defmodule ExDoc.Formatter.HTML.SearchData do
     ["searchData=" | ExDoc.Utils.to_json(data)]
   end
 
-  defp extra(map) do
-    if custom_search_data = map[:search_data] do
-      extra_search_data(map, custom_search_data)
-    else
```
sorentwo, line 24: Added a clause for this rather than a conditional in the function body.

lib/ex_doc/formatter/html.ex
```diff
@@ -602,7 +621,15 @@ defmodule ExDoc.Formatter.HTML do
 
       {path, opts} ->
         base = path |> to_string() |> Path.basename()
-        {base, opts[:filename] || Utils.text_to_id(Path.rootname(base))}
+
+        txid =
+          cond do
+            filename = opts[:filename] -> filename
+            url = opts[:url] -> url
```
josevalim, line 628: I am pretty sure you want to skip the `:url` in this case, because this is used to convert something like ``[foo](foo.md)`` into a link for `foo.html`, but there is nothing to link to in the URL case.

lib/ex_doc/formatter/html.ex
josevalim: Yes, we probably don't need `id`, `source_path`, `source_url` and `title_content`, as those are all about the rendered page.

josevalim:
For the `epub`, because there is no sidebar, should we even render them? Or should we just discard them upfront? Also, I wouldn't validate the `url`. It is the user docs, they can put whatever they want in there. :) Especially because even something like `foo.html` or `foo.pdf` would be valid ones.

sorentwo:
Suggestions applied, changes made , and residual TODO items addressed. This is all set!

lib/ex_doc/formatter/html/search_data.ex
```diff
@@ -59,6 +31,34 @@ defmodule ExDoc.Formatter.HTML.SearchData do
     end)
   end
 
+  defp extra(%{url: url} = map) do
+    [encode("#{map.id}", map.title, :extras, url)]
```
josevalim, line 35: Sorry, I should have caught it the first time around, but I am thinking this likely should not show up on search? I.e. we should skip URLs here too.

lib/mix/tasks/docs.ex
```diff
@@ -378,6 +379,7 @@ defmodule Mix.Tasks.Docs do
        title but keep the source file unchanged.
     * `:search_data` - A list of terms to be indexed for autocomplete and search. If not provided, the content
        of the extra page will be indexed for search. See the section below for more.
+    * `:url` - An external url to link to from the sidebar.
```
josevalim, line 382: I am thinking we could break it apart like this:

```
## Customizing Extras

There are two sources for extras, filenames and urls.

For filenames, the allowed configuration is:

  * ...

For urls:

  * title
  * url

```

lib/ex_doc/formatter/html/search_data.ex
josevalim: And let's add a test for whatever decision we pick here. :)

test/ex_doc/formatter/html_test.exs
```diff
@@ -112,6 +112,24 @@ defmodule ExDoc.Formatter.HTMLTest do
     assert LazyHTML.text(bar_content["h1"]) == "README bar"
   end
 
+  test "extras defined as external urls", %{tmp_dir: tmp_dir} = context do
+    config =
```
josevalim, line 116: I think we should add a test that shows grouping work. Imagine you want all URLs in a section in the sidebar called Important Links. The `Template.sidebar_extras` kinda implies URLs work with groups but we need to be sure it works "end to end".

assets/js/sidebar/sidebar-list.js
```diff
@@ -59,6 +59,7 @@ export function initialize () {
       const items = []
       const hasHeaders = Array.isArray(node.headers)
       const translate = hasHeaders ? undefined : 'no'
+      const href = node?.url || `${node.id}.html`
```
josevalim, line 62: Should we add an icon, such [as this one](https://remixicon.com/icon/external-link-line), for URLs?

If so, you can upload this bundle to remixicon.com, add external link, and get the new font back: https://github.com/elixir-lang/ex_doc/blob/main/assets/fonts/RemixIconCollection.remixicon

test/ex_doc/formatter/html_test.exs
josevalim: It seems groups expect the filename but there is none here, so we probably have to pass the filename or the url (and update its docs accordingly).

sorentwo: Grouping does work based on the input name, but it doesn't consider the URL. For example, with the extra `[Elixir: [url: "https://elixir-lang.org"]]` this will match on "Elixir", but it doesn't consider the url value at all.

```elixir
    input = to_string(input)
    title = input_options[:title] || input
    group = GroupMatcher.match_extra(groups, input)
```

That's a little different than how regular grouping works because the input lacks path information for easy grouping.

I've opted to expand `GroupMatcher.match_extra` to be url/path aware, so the pattern applies to either the title or url.

lib/ex_doc/formatter/html/search_data.ex
sorentwo: Done 👍

lib/mix/tasks/docs.ex
sorentwo: Good call. Done 👍

sorentwo:
@josevalim Suggestions and changes made, ready for another pass

assets/js/sidebar/sidebar-list.js
```diff
@@ -78,7 +79,10 @@ export function initialize () {
       }
 
       items.push(el('li', {}, [
-        el('a', {href: `${node.id}.html`, translate}, [node.nested_title || node.title]),
+        el('a', {href, translate}, [
+          node.url ? el('i', {class: 'external-link ri-external-link-line'}) : null,
+          node.nested_title || node.title
```
josevalim, line 84:
```suggestion
          node.nested_title || node.title,
          node.url ? el('i', {class: 'external-link ri-external-link-line'}) : null
```

lib/ex_doc/group_matcher.ex
```diff
@@ -48,11 +48,20 @@ defmodule ExDoc.GroupMatcher do
...
+        string when is_binary(string) -> path == string
+      end
+    end)
+  end
+
+  def match_extra(group_patterns, {:url, title, url}) do
+    match_group_patterns(group_patterns, fn pattern ->
+      case pattern do
+        %Regex{} = regex -> Regex.match?(regex, title) or Regex.match?(regex, url)
+        string when is_binary(string) -> title == string or title == url
```
josevalim, line 64: I am worried about matching on the title leading to false positives. I think we should match only on the URL, for consistency with filename. Then we can simplify the code. We should also document this under the `groups_for_extra` metadata (or whatever we call it).

josevalim (changes requested):
Sorry for the delay. I have only two tiny comments remaining and we can ship it!

sorentwo:
@josevalim Updated!

lib/ex_doc/group_matcher.ex
```diff
@@ -48,11 +48,11 @@ defmodule ExDoc.GroupMatcher do
   @doc """
   Finds a matching group for the given extra filename
```
josevalim, line 49:
```suggestion
  Finds a matching group for the given filename or url.
```

josevalim (approved):

josevalim:
Closing and reopening to kick CI back to life.

test/ex_doc/formatter/html_test.exs
```diff
@@ -112,6 +112,24 @@ defmodule ExDoc.Formatter.HTMLTest do
...
+      doc_config(context,
+        extras: [
+          "readme.md",
```
josevalim, line 119: Some OSes have case sensitive file systems, and we normalize everything to lowercase. So I lowercased everything here too, as this was not the concern of this test.

test/ex_doc/formatter/html_test.exs
```diff
@@ -112,6 +112,24 @@ defmodule ExDoc.Formatter.HTMLTest do
...
+        ]
+      )
+
+    File.write!("#{tmp_dir}/readme.md", "readme")
+
+    generate_docs(config)
+
+    content = File.read!(tmp_dir <> "/html/readme.html")
+
+    assert content =~ "https://elixir-lang.org"
```
josevalim, line 130:
```suggestion
          "#{tmp_dir}/readme.md",
          "Elixir": [url: "https://elixir-lang.org"]
        ]
      )

    File.write!("#{tmp_dir}/readme.md", "readme")
    generate_docs(config)

    content = File.read!(tmp_dir <> "/html/readme.html")
    assert content =~ "https://elixir-lang.org"
```

josevalim:

starbelly:
Hey @josevalim there was hard breaking change introduced here. That looks quite intentional. However, ex_doc does usually give a heads up in the form of a soft deprecation, although ex_doc is not obliged to per it still being 0.x. I would just like confirmation on this so I can make an informed decision about what to do in rebar3_ex_doc 😄

josevalim:
What was the breaking change? I would be glad to revert it to something less breaking.

starbelly:
> What was the breaking change? I would be glad to revert it to something less breaking.

I missed some context. I think this is just a bug that got introduced here : https://github.com/elixir-lang/ex_doc/pull/2103/files#diff-126f42fa83c5dbc502baeb9fce0945b4089753ce8fe10ec03df01f6ef761a3d5R614

Sorry, this was not a breaking change in ex_doc, rebar ex_doc plugin is at fault here. TL;DR we allow pasing in a map, and it used to be it just happened to work because there was no explicit usage of `Keyword`, however, extras has always been documented as being a keyword list. 

There's changes to make over there, not here 😄 Sorry for the noise.

josevalim:
No worries at all, if there is something we can do on our side, please let us know!

starbelly:
> No worries at all, if there is something we can do on our side, please let us know!

Short of opting to use maps instead of keyword lists, I don't think there is much to do on this side 😄

josevalim:
I believe we can accept both keyword lists and atoms and normalize them accordingly. My goal in the short term future is to convert them to structs instead of passing around raw maps.

Issue #2105, opened 2025-04-04
Replace EasyHTML by LazyHTML


josevalim:
Ref: https://github.com/dashbitco/lazy_html

DavidOliver:
A quick search indicates EasyHTML is only used a couple of times, in tests. Should I take this?

josevalim:
If you want to, go for it!!!

PR #2106, opened 2025-04-04
Replace EasyHTML with LazyHTML

DavidOliver:
Re. #2105

Is there a better way of asserting whether or not a node is matched by the CSS selectors in `ExDoc.Formatter.HTML.TemplatesTest` `outputs function groups` test than the utility function I've added? I wondered if it would be possible to pattern match against the LazyHTML result's number of nodes, but I think we don't have direct access and instead only see a reference? (`%LazyHTML{resource: reference()}`.)


test/ex_doc/formatter/html/templates_test.exs
```diff
@@ -497,13 +506,13 @@ defmodule ExDoc.Formatter.HTML.TemplatesTest do
...
-      refute doc["#legacy [id='example/2']"]
-      assert doc["#functions [id='example_1/0']"]
-      refute doc["#functions [id='example/2']"]
+      doc = LazyHTML.from_document(content)
+      assert num_nodes_matched(doc, "#example-functions a[href='#example-functions']") == 1
+      assert num_nodes_matched(doc, "#legacy a[href='#legacy']") == 1
+      assert num_nodes_matched(doc, "#example-functions [id='example/2']") == 1
+      assert num_nodes_matched(doc, "#legacy [id='example/2']") == 0
+      assert num_nodes_matched(doc, "#functions [id='example_1/0']") == 1
+      assert num_nodes_matched(doc, "#functions [id='example/2']") == 0
```
jonatanklosko, line 515: `LazyHTML` implements `Access` behaviour that simply does `query` and it implements `Enumerable`, so you can get the number of top-level nodes directly. So one way would be this:

```suggestion
      assert Enum.count(doc["#example-functions a[href='#example-functions']"]) == 1
      assert Enum.count(doc["#legacy a[href='#legacy']"]) == 1
      assert Enum.count(doc["#example-functions [id='example/2']"]) == 1
      assert Enum.count(doc["#legacy [id='example/2']"]) == 0
      assert Enum.count(doc["#functions [id='example_1/0']"]) == 1
      assert Enum.count(doc["#functions [id='example/2']"]) == 0
```

test/ex_doc/formatter/html_test.exs
```diff
@@ -85,6 +85,8 @@ defmodule ExDoc.Formatter.HTMLTest do
   end
 
   test "multiple extras with the same name", c do
+    import LazyHTML
```
jonatanklosko, line 88: We do just a few function calls, so I would say an import is not necessary. In general it's easier to understand code with qualified calls, unless it's well-established macros where `import` makes sense :)

test/ex_doc/formatter/html/templates_test.exs
DavidOliver: Ah - thanks! Examples are given in LazyHTML's docs, too, which I managed to not pick up on.

mix.exs
```diff
@@ -47,7 +47,7 @@ defmodule ExDoc.Mixfile do
       {:makeup_html, ">= 0.1.0", optional: true},
       {:jason, "~> 1.2", only: :test},
       {:floki, "~> 0.0", only: :test},
```
jonatanklosko, line 49: `easyhtml` depends on `floki`, it was likely used before, that's why both are listed. Looks like there is one direct usage in `erlang_test.exs`, which we should be able to replace with `lazy_html` as well :)

```suggestion
```

josevalim (approved):

jonatanklosko (approved):

PR #2109, opened 2025-04-06
Document the requirement to push releases

axelson:
Currently there isn't any documentation about the requirement to push releases/tags to the `source_url` to get links to specific versions of the code. So this PR adds this documentation.

I couldn't find a general spot where `source_url` is talked about so I added an admonition near where it is first introduced.

Background: recently (in https://github.com/PostHog/posthog-elixir/pull/14#issuecomment-2778825519) I was asked where the requirement for pushing git tags comes from. I wasn't able to find any official documentation (here or in https://hexdocs.pm/elixir/writing-documentation.html) so I have created this PR.

Here's what it looks like when I build the documentation locally:
Screenshot 2025-04-06 07-48-40@2x

### DIFF
--- a/lib/mix/tasks/docs.ex
+++ b/lib/mix/tasks/docs.ex
@@ -61,10 +61,9 @@ defmodule Mix.Tasks.Docs do
   > #### Tip {: .info}
   >
   > If you set `:source_url` to a GitHub/GitLab/BitBucket repo then whenever you
-  > publish a release you should publish the release to the repo (which will
-  > push a tag). If the release exists in the repo then ex_doc will generate
-  > links to the specific version of the code that the user is currently
-  > browsing.
+  > publish a new version of your package, you should run `git tag vVERSION`
+  > and push the tag. This way, ExDoc will generate links to the specific version
+  > the docs were generated for.
 
   ExDoc also allows configuration specific to the documentation to
   be set. The following options should be put under the `:docs` key
--- a/lib/mix/tasks/docs.ex
+++ b/lib/mix/tasks/docs.ex
@@ -58,13 +58,6 @@ defmodule Mix.Tasks.Docs do
         ]
       end
 
-  > #### Tip {: .info}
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/docs.ex
```diff
@@ -58,6 +58,14 @@ defmodule Mix.Tasks.Docs do
         ]
       end
 
+  > #### Tip {: .info}
+  >
+  > If you set `:source_url` to a GitHub/GitLab/BitBucket repo then whenever you
+  > publish a release you should publish the release to the repo (which will
+  > push a tag). If the release exists in the repo then ex_doc will generate
+  > links to the specific version of the code that the user is currently
+  > browsing.
```
josevalim, line 67:
```suggestion
  > publish a new version of your package, you should run `git tag vVERSION`
  > and push the tag. This way, ExDoc will generate links to the specific version
  > the docs were generated for.
```

However, I don't think this is automatic, you must also configure the `:source_ref` option.

axelson: Ah, yeah you're correct. I have that in so many projects (nearly every project I've forked into my `dev/forks` directory) that I forgot it was not automatic. Also it is something that works on 95% of libraries that I see on hexdocs. I also got thrown off by the explicit github/gitlab/bitbucket support on these lines (I didn't look too closely): https://github.com/elixir-lang/ex_doc/blob/8552c492fd35498c29983ac196988723bdc09e96/lib/ex_doc/config.ex#L284

I'll update the docs to add it. But it seems like it's outgrowing an admonition and will now need a new section specifically about setting up `source_url` and `source_ref`.

josevalim: A custom section looks great, yeah.

axelson:
@josevalim okay I've extracted this to a new section and it is now ready for another look!

josevalim:

PR #2110, opened 2025-04-06
Be slightly more careful to take the first paragraph

### LINKED ISSUE #2108: Bug: unbalanced <li> tag emitted in synopsis
While consuming ex_doc epub output and running through a strict linter, I discovered that ex_doc is emitting some bad HTML.  This task concerns the broken tags produced by the `ExDoc.Formatter.HTML.Templates.synopsis/1` function.

Given minimal .ex input as follows,
```elixir
defmodule ImbalanceTest do
  @doc """
  ## Options

  - one

  - two
  """
  def nonsense do
  end
end
```

The low-level method gets called like this,
```elixir
iex> ExDoc.Formatter.HTML.Templates.synopsis("<h2>Options</h2><ul><li><p>one</p></li><li><p>two</p></li></ul>")
"<h2>Options</h2><ul><li><p>one</p>"
```

The `<p>` between list items seems to be key to triggering this bug.

adamwight:
The previous algorithm was easily fooled into emitting unbalanced tags, so the logic is tightened up with a more specific regex.

Overall, this probably isn't doing the right thing.  Consider making the synopsis more robust by using eg. Floki in production.

Closes #2108 , sort of...

### DIFF
--- a/lib/ex_doc/formatter/html/templates.ex
+++ b/lib/ex_doc/formatter/html/templates.ex
@@ -53,9 +53,10 @@ defmodule ExDoc.Formatter.HTML.Templates do
 
   def synopsis(doc) when is_binary(doc) do
     doc =
-      case :binary.split(doc, "</p>") do
-        [left, _] -> String.trim_trailing(left, ":") <> "</p>"
-        [all] -> all
+      Regex.run(~r|<p>((?:(?!</p>).)*?):*</p>|, doc)
+      |> case do
+        nil -> doc
+        [_, first_paragraph_text] -> "<p>" <> first_paragraph_text <> "</p>"
       end
 
     # Remove any anchors found in synopsis.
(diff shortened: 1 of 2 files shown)


lib/ex_doc/formatter/html/templates.ex
```diff
@@ -53,9 +53,10 @@ defmodule ExDoc.Formatter.HTML.Templates do
...
   def synopsis(doc) when is_binary(doc) do
     doc =
-      case :binary.split(doc, "</p>") do
-        [left, _] -> String.trim_trailing(left, ":") <> "</p>"
-        [all] -> all
+      Regex.run(~r|<p>((?:(?!</p>).)*?):*</p>|, doc)
+      |> case do
+        nil -> doc
+        [_, first_paragraph_text] -> "<p>" <> first_paragraph_text <> "</p>"
       end
```
josevalim, line 60: We are kind of assuming that it starts with <p>, so perhaps I would enforce that instead?

```
case String.trim_leading(doc) do
  "<p>" <> _ = doc ->
    doc =
      case :binary.split(doc, "</p>") do
        [left, _] -> String.trim_trailing(left, ":") <> "</p>"
        [all] -> all
      end

    Regex.replace(~r|(<[^>]*) id="[^"]*"([^>]*>)|, doc, ~S"\1\2", [])

  _ ->
    nil
end
```

Alternatively we capture the synopsis over the AST instead of the rendered document. This would also make it easier to remove links and what not.

Issue #2104, opened 2025-03-31
Defining a text bloc for group_for_docs sections


lud:
Hello,

I would like to know if it is currently possible to inject some text in the generated docs below the section headers that are generated from `groups_for_docs`.

If not, is this something you'd consider adding?

Thank you.

josevalim:
Just to make sure we are on the same page, can you send a screenshot showing where you'd want to add the text?

josevalim:
In that case, a PR is definitely welcome.

lud:
Yes, exactly here.

![Image](https://github.com/user-attachments/assets/ed0ec363-fff7-40ff-8d0c-50edcbd2b2ec)
Below the "Schema Definition Utilities" header, that is a section header after the Summary section with signatures.

> In that case, a PR is definitely welcome.

I'll give it a try. Any direction or general rule to consider before starting?

josevalim:
> I'll give it a try. Any direction or general rule to consider before starting?

No, go for it!

garazdawi:
> Any direction or general rule to consider before starting?

As it is now possible to create groups via two mechanism, that is doc metadata and configuration file, IMO it should also be possible to add the "group description" with those two mechanisms. The syntax for writing the description should IMO be markdown (and not just text as deprections are), so that we can have links and other things in there.

josevalim:
I think you covered the options well but all of them come with trade-offs that I am not particularly excited about. Storing it as metadata in the moduledoc can be interesting, so we can keep all of the markup in one place, but you can also have reused groups (and perhaps reused descriptions).

lud:
If you want to reuse I guess instead of using the defaults, you provide `:description_for_module_group` like this:

```elixir
description_for_module_group: fn
  _module, :"Helpers" -> MyApp.Docs.description_for(:"Helpers")
  _, _ -> nil
end
```

I think it's possible to call app code from `mix.exs` right? As it is evaluated by a mix task.

lud:
Yes.

We could take shortcuts:

If `groups_for_docs` returns iodata then it's a match and the text is going to be used as the description. If it returns `true`, `false` or `nil` it behaves as usual (no description).

```elixir
defp groups_for_docs do
  [
    Helpers:
      &(&1[:section] == :helpers &&
          """
          Some docs here
          """),
    Deprecations: &(&1[:section] == :old_stuff)
  ]
end
```

In the catchall you can return a `{header, text}` tuple, or a text/`nil` as usual.

[...]

PR #2113, opened 2025-04-09
Display description text in docs groups

lud:
Related to https://github.com/elixir-lang/ex_doc/issues/2104

Still a work in progress. If we go in this direction we should add docs as well.

There are some limits with this solution:

* Since we want the description from `default_group_for_docs`, it means that each doc node can return a different description for a given group. I made the choice to ignore those descriptions based on the group title.
* As we want to keep the group order defined in moduledoc, the retriever will fetch those, possibly containing descriptions. I made the choice to not replace descriptions from moduledoc by descriptions returned by the config callback.
* Node groups will temporarily contain a title/description map, only to be replaced with just the title once the groups are built in the modulenode. This avoids a lot of "when is_binary() ... when is_map()" stuff and avoids to refactor many tests. This also allows to have each group rendered only once. A drawback is that in the `module_summary` function we are reconciliating the groups once again where each doc node could already bear the full group. Not a big deal imho.
* Since those groups are rendered, they must behave like most nodes. But they are not nodes for now. I guess we could add a struct for those, but it's not the real problem. What is more important is that I had to hardcode "text/markdown" and wrap the text in `%{"en" => text }`, as module and doc nodes are extracted from documentation. I could have written a specialized `doc_ast` render function though.


lib/ex_doc/formatter/html/templates.ex
```diff
@@ -169,7 +169,14 @@ defmodule ExDoc.Formatter.HTML.Templates do
 
   def module_summary(module_node) do
     # TODO: Maybe it should be moved to retriever and it already returned grouped metadata
-    ExDoc.GroupMatcher.group_by(module_node.docs_groups, module_node.docs, & &1.group)
+
```
josevalim, line 172: The TODO above is precisely about moving this logic to the retriever. Given we are already changing the group structure, maybe now is a good time to go ahead and do it? Basically, by adding a `group` column to the groups upfront?

lib/ex_doc/retriever.ex
```diff
@@ -314,4 +365,19 @@ defmodule ExDoc.Retriever do
...
+      %{title: title, description: description}
+      when is_binary(title) and (is_binary(description) or is_nil(description)) ->
+        group
+
+      kw when is_list(kw) ->
+        true = Keyword.keyword?(kw)
+        %{title: Keyword.fetch!(kw, :title), description: kw[:description]}
+
+      title when is_binary(title) when is_atom(title) ->
+        %{title: title, description: nil}
```
josevalim, line 380: We should probably normalize the title here?

lib/ex_doc/retriever.ex
```diff
@@ -222,7 +230,7 @@ defmodule ExDoc.Retriever do
       (source_doc && doc_ast(content_type, source_doc, file: doc_file, line: doc_line + 1)) ||
         doc_data.doc_fallback.()
 
-    group = group_for_doc.(metadata) || doc_data.default_group
+    group = normalize_group(group_for_doc.(metadata) || doc_data.default_group)
```
josevalim, line 233: We should probably already fetch the title here. Or, depending how it goes, the group index key.

lib/ex_doc/retriever.ex
lud: You mean accept atoms and convert to binary if needed?

lib/ex_doc/formatter/html/templates.ex
lud: Yeah I saw this TODO and in my first attempt I wanted to provide docs_groups with all the docs inside.

But to avoid having duplicate data, to me it means that we basically remove the `:docs` key from the `ModuleNode` struct. And that means refactoring a lot of tests to look for expected data in a list of groups instead of just matching on the content of the docs key. Lot more code to change and review. But I can do it if you think it's alright.

lib/ex_doc/retriever.ex
lud: Sorry I don't understand. `group_for_doc./1` returns the title or keyword with title and description. I'm not sure what you mean by "fetching the title".

lib/ex_doc/retriever.ex
josevalim: Right now it is already accepting atoms so normalizing could be nice? :)

lib/ex_doc/retriever.ex
josevalim: It is because later on you do something like `Enum.map(docs, &Map.put(&1, :group, &1.group.title)` or similar. It would be nice if we could avoid the extra pass :)

lib/ex_doc/formatter/html/templates.ex
lud: ah sorry I missed the part about a group column. Should be simple enough.

lud: Alright, please see that last commit: https://github.com/elixir-lang/ex_doc/pull/2113/commits/aed460eef09aef4c560f37593b7d893feebe1e01

(There were other commits before since you last replied).

I'm not sure this is what you expect. The retriever is computing the groups, so the renderer part is only fetching that, which is what we wanted.

But,

* There is duplicated data. The `ModuleNode` still has its `:docs` key with the `DocNode` list because many tests look into that for assertions. We can update those tests to look into the groups instead, with a lot of Enum.find calls :) and then delete the `ModuleNode` `:docs` key.
* Those `DocNode`s in the `:docs` key have `nil` `:rendered_doc` because we only render the copy of the doc nodes that are inside group nodes.

Or maybe it is something else you had in mind?

josevalim:
Sorry for the delay, I am on holidays so I have limited time to stay on top of my inbox. I am reviewing this now but probably merge it only when I am back. :)

lib/ex_doc/formatter/html/templates.ex
josevalim: The code looks great, thank you ! 🤩 

> There is duplicated data. The ModuleNode still has its :docs key with the DocNode list because many tests look into that for assertions. We can update those tests to look into the groups instead, with a lot of Enum.find calls :) and then delete the ModuleNode :docs key.

I think it is worth giving this a try! Perhaps we could have a helper function in tests, called `find_doc(...)` or `find_group(...)` to help make life easier. Could you please try that as well? I understand it is a lot of work though, but this is some of the refactoring we have been wanting to do for quite some time (basically do less work on the formatters and more on the retrievers).

garazdawi:
I did a quick test and it works as expected. I'm wondering if maybe something could be done about the styling to make it more obvious that the text belongs to the entire group of functions... not sure what that would be though.

image

scriptforfivem:
Impressive project structure! The architecture really demonstrates solid design principles.

eyesgodcode:
Impressive project structure! The architecture really demonstrates solid design principles.

lud:
@garazdawi thank you !

I think we can just add some whitespace between the description and the first function signature block. The same as between the title and the description. I guess that would be enough to show that the text belongs to the section.

@josevalim I see a lot have changed about the heading and html ID generation. I have rebased my branch and made sure that the new groups descriptions go through `add_fancy_anchors`. I also updated the templates to use the new `render_doc` helper instead of having a `.rendered_doc` key in the nodes.

I'll tackle removing the `ModuleNode.docs` key and updating the tests soon.

lud:
@garazdawi I think whitespace is enough, what do you think?

image

josevalim:
@lud, apologies for the large changes that had happened. But we had bugs and a lot of duplication, and the new version should make things more accessible and easier to follow.

lud:
@josevalim No worries :)

lud:
Hello,

I removed the `.docs` field from the `ModuleNode` struct and fixed all the tests accordingly. It was actually easy since most tests generate only one group. Other tests were straightforward too.

So I left the helper function directly in `elixir_test.exs` since this is the only place it is used.

A test failed because this function was receiving a list of 2 elements (and it is matching on a list of 1 element only)

```elixir
defp extract_filename!([{location, _}]), do: location
```

I may have done something wrong when building the assets when changing whitespace for the groups.

I ran the build command (`npm run --prefix assets build`) once more and that fixed it, only a single file was present. I guess it was just because of the git rebase.

lib/ex_doc/formatter/html/templates.ex
```diff
@@ -153,10 +153,7 @@ defmodule ExDoc.Formatter.HTML.Templates do
     end)
   end
 
-  def module_summary(module_node) do
-    # TODO: Maybe it should be moved to retriever and it already returned grouped metadata
-    ExDoc.GroupMatcher.group_by(module_node.docs_groups, module_node.docs, & &1.group)
-  end
+  def module_summary(module_node), do: module_node.docs_groups
```
josevalim, line 156: Let's remove this function and update its callers?

lib/ex_doc/formatter/html/assets.ex
```diff
@@ -7,7 +7,7 @@ defmodule ExDoc.Formatter.HTML.Assets do
     |> Path.wildcard()
     |> Enum.map(fn path ->
       Module.put_attribute(__CALLER__.module, :external_resource, path)
-      {Path.basename(path), File.read!(path)}
+      {Path.basename(path), "File.read!(#{path})"}
```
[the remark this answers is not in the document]
lud: Gosh ... fortunately you spotted that. Yeah I was debugging assets generation at some point.

lud:
Hello,

I [fixed](https://github.com/elixir-lang/ex_doc/pull/2113/commits/b1ac48407cef60364a6a3af273645a0ae84eab9a) your last comments, thank you for the review :)

josevalim:

PR #2115, opened 2025-04-22
Experimental code for a new `@doc copy: true` or `@doc copy: {m, f, a}` feature

a-maze-d:
# New Feature
This code introduces a new `@doc copy: true` feature for `defdelegate` functions (see the test for some usage, should even work with @doc copy: {m, f, a}``) which would extract the doc from another `{m, f, a}` and copy it to the function with the tag. This is especially interesting for `defdelegate` functions to have proper documentation (instead of the `see ...`) without having to write it twice.
 
***Note:*** This is some experimental code, rather quickly hacked together (happy to clean it up if there is interest in something like this)

# Background
In my library I"m making use of `defdelegate` to define a module is auto-generated and that dispatches functions to a couple of other modules. I do want to have the documentation in a single place, so I neither want to have no documentation nor do I want the  `see ...` referring to the implementing documentation.

I tried to copy the documentation from the other module in the dispatching module by having
```elixir
    @doc Fledex.Color.Names.DocUtils.extract_doc(module, name, 1)
``` 
This works, but not reliably, because (and this is my guess) the files are not yet written when I try to extract the documentation from the modules. So sometimes it works sometimes it doesn't.

I tried to use `mix xref` to see that the `compile` dependencies are correct and with some tricks I did managed to ensure that the dependencies are correct, but still sometimes the retrieving of the docs results in `nil`, i.e. we fall back to the `see ....` documentation.

This made me think that actually it would be much nicer if ex_doc would have the ability to copy the documentation from another function

# Question
Do you think it makes sense to provide such functionality? 
* If, so I'm happy to clean up the code and bring it to prod quality
* If not, then feel free to simply close this PR :-)

### DIFF
--- a/lib/ex_doc/language/elixir.ex
+++ b/lib/ex_doc/language/elixir.ex
@@ -289,7 +289,7 @@ defmodule ExDoc.Language.Elixir do
         {:local, :..}
 
       ["//", "", ""] ->
-        {:local, :..//}
+        {:local, :"..//"}
 
       ["", ""] ->
         {:local, :.}
--- a/lib/ex_doc/retriever.ex
+++ b/lib/ex_doc/retriever.ex
@@ -178,6 +178,50 @@ defmodule ExDoc.Retriever do
 
   # Helpers
 
+  def extract_doc(module, function, arity) do
+    case get_docs(module, [:function]) do
+      {nil, nil, nil} ->
+        IO.puts("extract_doc failed" <> inspect({module, function, arity}))
+        nil
+
+      {_language, format, docs} ->
+        doc = extract_function_doc(docs, function, arity)
+        doc_ast(format, doc, [])
+    end
+  end
+
+  def extract_function_doc(docs, function, arity) do
+    case Enum.find(docs, nil, fn {{_type, func_name, func_arity}, _info, _types, _doc, _opts} ->
+           func_name === function && func_arity === arity
+         end) do
+      nil ->
+        nil
+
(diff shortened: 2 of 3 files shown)


lib/ex_doc/language/elixir.ex
```diff
@@ -289,7 +289,7 @@ defmodule ExDoc.Language.Elixir do
         {:local, :..}
 
       ["//", "", ""] ->
-        {:local, :..//}
+        {:local, :"..//"}
```
a-maze-d, line 292: This is added by `mix format`

josevalim:
Thanks for the PR.

We already store some delegate_to metadata. In the past, we actually implemented the feature you requested, but the issue is that any local link, such as `c:a_callback/1` or `another_function/1` were broken, as well as specs, and copying the whole documentation while porting its context was too complex, which is why we have the current implementation.

Unfortunately your PR would introduce those problems back.

a-maze-d:
Thanks @josevalim and providing context to the difficulties.

One thing to consider is that the `@doc copy: true` is a opt-in approach, so it's up to the developer to decide whether he wants to use it and the issues could be highlighted in some documentation (even though I know that many developers do not read those :smile: 

I'll tinker a bit more (so I'll leave the PR open for a while, but feel free to ignore it) and maybe we can solve the linking issue :-)

Thanks again for your swift feedback

a-maze-d:
Maybe better to close it and reopen a new PR if necessary

PR #2120, opened 2025-05-06
Correctly decode sidebar ids

### LINKED ISSUE #2114: Header with & in it renders as &amp;
If you have an extra with a header that has an `&` in it, it will render as `&amp;` in the sidebar.

See https://www.erlang.org/docs/28/apps/stdlib/re_incompat for an example.

garazdawi:
This almost doubles the size of the `.js` library... so if anyone has any better ideas I'm all ears. Maybe doing something like this:

```js
function decodeXml(xmlStr) {
  const parser = new DOMParser();
  const doc = parser.parseFromString(`<root>${xmlStr}</root>`, 'text/xml');
  return doc.documentElement.textContent;
}
```

closes #2114


josevalim:
It seems we have a double escaping issue and perhaps we should not be escaping them when create the sidebar nodes from the Elixir side?

garazdawi:
I did try that, but that broke a bunch of tests that seem to have been designed to specifically test that it should be encoded...

josevalim:
Ok, let me double check!

garazdawi:
closing in favor of #2122

PR #2122, opened 2025-05-07
Preserve HTML entities in headers

### LINKED ISSUE #2114: Header with & in it renders as &amp;

josevalim:
Closes #2114.
Closes #2120.


josevalim:
cc @garazdawi

josevalim:
Btw, I think this is going to have some subtle issues, as we are using two different ways to compute headers (one using regexes and another without), so we may need to look into that before shipping this.

garazdawi:
Tested it for us and it solves the issue and I haven't found any regressions so far :)

josevalim:
@garazdawi please check again! I made a big refactoring that unified how we handle link headings, so we only do it once in a single place, using DocAST instead of relying too much on regexes.

lib/ex_doc/doc_ast.ex
```diff
@@ -244,35 +298,43 @@ defmodule ExDoc.DocAST do
...
-      |> Makeup.highlight_inner_html(
-        lexer: lexer,
-        lexer_options: lexer_opts,
-        formatter_options: [highlight_tag: highlight_tag]
-      )
+  @doc """
+  Maps the tags in the AST, first mapping children tags, then the tag itself.
+  """
+  def map_tags({tag, attrs, inner, meta}, fun),
+    do: fun.({tag, attrs, Enum.map(inner, &map_tags(&1, fun)), meta})
```
garazdawi, line 307:
```suggestion
  def map_tags({tag, attrs, inner, meta}, fun) when is_list(inner),
    do: fun.({tag, attrs, Enum.map(inner, &map_tags(&1, fun)), meta})
    
  def map_tags({tag, attrs, inner, meta}, fun),
    do: fun.({tag, attrs, inner, meta})
```

When edoc is used to generate the doc ast, inner can be a string without being wrapped in a list.

garazdawi (commented):
Seems to work fine except for the small detail in `map_tags`.

lib/ex_doc/doc_ast.ex
josevalim: I basically wrote all of the traversal functions assuming that the contents is a list. :( Any chance we could normalize edoc in the long term? I will also improve the traversal functions here meanwhile.

garazdawi:
> I basically wrote all of the traversal functions assuming that the contents is a list. 

Maybe you could normalize the input in the retriever?

> Any chance we could normalize edoc in the long term?

yes, I think that it is a mistake that it is like that today. I'm surprised that our own validators do not catch it...

josevalim:
Ah, perfect, that's an easy fix. I assume you need a new release with these fixes?

josevalim:

garazdawi:
> I assume you need a new release with these fixes?

yes please :)

josevalim:
done!

PR #2123, opened 2025-05-09
Remove duplicate entries and assets/ folder in .build file in HTML formatter

eksperimental:
Previously the generated `.build` file would contain these entries:

    ...
    assets
    assets/kv-observer.png
    assets/logo.png
    assets/logo.png
    ...

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -32,20 +32,23 @@ defmodule ExDoc.Formatter.HTML do
     }
 
     all_files =
-      search_data ++
-        static_files ++
-        generate_sidebar_items(nodes_map, extras, config) ++
-        generate_api_reference(nodes_map, config) ++
-        generate_extras(extras, config) ++
-        generate_favicon(@assets_dir, config) ++
-        generate_logo(@assets_dir, config) ++
-        generate_search(config) ++
-        generate_not_found(config) ++
-        generate_list(nodes_map.modules, config) ++
-        generate_list(nodes_map.tasks, config) ++
-        generate_redirects(config, ".html")
+      (search_data ++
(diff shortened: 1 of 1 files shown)


lib/ex_doc/formatter/html.ex
```diff
@@ -32,20 +32,23 @@ defmodule ExDoc.Formatter.HTML do
...
+         generate_favicon(@assets_dir, config) ++
+         generate_logo(@assets_dir, config) ++
+         generate_search(config) ++
+         generate_not_found(config) ++
+         generate_list(nodes_map.modules, config) ++
+         generate_list(nodes_map.tasks, config) ++
+         generate_redirects(config, ".html"))
+      |> Enum.uniq()
+      |> Kernel.--([@assets_dir])
+      |> Enum.sort()
```
josevalim, line 49: Let's move this handling to the `generate_build` function instead? It feels it is more part of its concern. Also, do we need to exclude `@assets_dir`? Why is it showing up in the list in the first place?

eksperimental: I have moved the logic to generate_build.

I don't think we need to list folders in the .build file. If we decide to, we should be consistent and list all the folders consistently.
It is showing up because [`File.cp_r!/3` returns the dirs as well as the files](https://github.com/elixir-lang/ex_doc/pull/2123/files/68d5662a4d21728219c78f5ff66917d6420f14f0..7fc7e4c48af9bb8cb1efaa0179c1940999f129ed#diff-126f42fa83c5dbc502baeb9fce0945b4089753ce8fe10ec03df01f6ef761a3d5R289-R298).
I have taken a different approach and exclude all dirs/subdirs for the assets.

josevalim:

PR #2124, opened 2025-05-12
Fix strip_app to always rewrap in a list

garazdawi:
The crash I described in #2122 was not because of edoc producing docs without a list wrapper, but because the Erlang layer in ExDoc did so. My initial suggestion just happened to work as it also fixed this issue, but the final fix of normalizting edoc input did not as it happened to early.

This PR fixes the issue though I could not manage to create a testcase that replicates it.

### DIFF
--- a/lib/ex_doc/language/erlang.ex
+++ b/lib/ex_doc/language/erlang.ex
@@ -418,15 +418,15 @@ defmodule ExDoc.Language.Erlang do
   end
 
   defp strip_app([{:code, attrs, [code], meta}], app) do
-    [{:code, attrs, strip_app(code, app), meta}]
+    [{:code, attrs, List.wrap(strip_app(code, app)), meta}]
   end
 
   defp strip_app(code, app) when is_binary(code) do
-    String.trim_leading(code, "//#{app}/")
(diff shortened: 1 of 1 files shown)


josevalim:

garazdawi:
When you have the time, we would appriciate a new release as well :)

josevalim:
@garazdawi can you please confirm this is good then? 08401403f9ca5ef479e006297e9cd708c80ceda6

garazdawi:
I think that the things introduced in [this commit](https://github.com/elixir-lang/ex_doc/pull/1834/commits/a31dab4874377281eacd900dbaab7aedf1de8851) are no longer done, but we seem to have fixed most of them so I will just fix the remaining ones so that we don't need that workaround anymore.

When I fixed that all tests passed with https://github.com/elixir-lang/ex_doc/commit/08401403f9ca5ef479e006297e9cd708c80ceda6.

josevalim:
Yeah, we also changed synopsis to rely on AST, so it is way more reliable and precise now. So I will go ahead and cut a new release, worst case scenario I do another one later :)

garazdawi:
thanks!

josevalim:
done!

PR #2127, opened 2025-05-22
Add a preview server mix task with Bandit

srcrip:
Hey yall,

I realize this is the kind of thing that the core Elixir projects have often left up to userspace, but I think we should consider adding a built in preview server task to ex_doc.

This is based on something I use in my projects, using Bandit. I added Bandit >= 1.0.0 as an optional dependency.

I think this is the kind of thing most programming languages/tools should have, if you have a docs builder you may as well give the user/developer a way to quickly preview those docs without having to go to some external tool. I think it will benefit new users to elixir/ex_doc by being one less thing they have to think about to start working on documentation.

I would imagine we may want other docs in other locations, but I figure I'd just open this up for discussion first before sinking a bunch of time into that.


josevalim:
Can you please expand why we’d need a preview server? The docs open from the file system, so you don’t need a separate tool, and recent Erlang ships with a one liner http server (listed in the README).

PR #2132, opened 2025-06-18
Ensure default groups for docs are sorted alphabetically

lud:
First attempt at the docs. Docs are hard, as always. 

I splitted the different concepts (attaching to groups / descriptions / ordering) for clarity, not sure if it is OK.

We may want to fix default ordering (displaying groups from `:default_group_for_doc` before Types,Functions,Callbacks) and update the docs before merging this.

### DIFF
--- a/lib/mix/tasks/docs.ex
+++ b/lib/mix/tasks/docs.ex
@@ -374,7 +374,7 @@ defmodule Mix.Tasks.Docs do
         %{title: "Helpers", description: "Functions shared with other modules."}
       ]
 
-  Descriptions also can be given in the `:default_group_for_doc` configuration:
+  Descriptions can also be given in the `:default_group_for_doc` configuration:
 
       default_group_for_doc: fn metadata ->
         csae metadata[:group] do


josevalim:
> We may want to fix default ordering (displaying groups from :default_group_for_doc before Types,Functions,Callbacks) and update the docs before merging this.

Are you planning to send a PR for this too? :) You can also amend this one if you prefer!

josevalim (approved):

lud:
> Are you planning to send a PR for this too? :) You can also amend this one if you prefer!

If we have to fix it then yes I'll update this PR.

josevalim:
Yes, I should probably preserve it to avoid breaking changes, unless it would add a lot of complexity or complications. I am hoping it is rather just swapping operands somewhere :D

lud:
Yeah ok, it should not be that hard.

lud:
@josevalim Well it's actually already the case in version `0.37`: "Functions" appear _before_ groups from `:default_group_for_docs`. If I try to change that a test breaks.

So I'm just going to sort groups from `:default_group_for_doc` by name to keep the previous behaviour.

lud:
Yeah, without any default configuration, with `0.37.3` or `0.38.2` the following code:

```elixir
  @doc group: "Z group last"
  def aaa, do: []

  @doc group: "A Group second"
  def zzz, do: []

  def mmm, do: []
```

...produces the following:

image

So I'd rather keep that behaviour and document it, as there is now a second simple way to control ordering!

(But I did make a change to sort those default groups by alphabetical order.)

josevalim:

lud:
Thank you :)

PR #2138, opened 2025-07-31
Load discovered makeup apps for cli

belltoy:
The `ex_doc` cli will not load apps from the provided `--paths` argument when `ExDoc.Application` start. When prepend the `paths`, the `.app` should be loaded so that the `makeup_` applications can be started before generating.

This scenario occurs when using `rebar3_ex_doc`, which use the cli underlying.


belltoy:
@josevalim could you bump a new version of it?

josevalim:
done!

PR #2142, opened 2025-09-10
Safe guard exdoc:autocomplete-limit meta access

weih-kahoot:
Can be seen in a few places in browser console, f.ex. https://hexdocs.pm/phoenix/overview.html

Error:
Uncaught TypeError: can't access property "content", document.querySelector(...) is null


josevalim:

PR #2147, opened 2025-09-29
refactor: extract minimal shared logic between HTML and EPUB formatters

yordis:
Extract render_all, build_extras, and filter_list functions from HTML to shared ExDoc.Formatter module to eliminate tight coupling where EPUB directly calls HTML functions.

Changes:
- Create ExDoc.Formatter with shared functions moved from HTML
- Update HTML formatter to use Formatter.render_all(), Formatter.build_extras(), Formatter.filter_list()
- Update EPUB formatter to use Formatter functions instead of HTML.* calls
- Remove duplicate function definitions from HTML formatter

This provides a foundation for additional formatters while maintaining the existing API and functionality.

### DIFF
--- a/lib/ex_doc/formatter.ex
+++ b/lib/ex_doc/formatter.ex
@@ -1,0 +1,363 @@
+defmodule ExDoc.Formatter do
+  @moduledoc false
+
+  alias ExDoc.{Markdown, GroupMatcher, Utils}
+
+  @doc """
+  Autolinks and renders all docs.
+  """
+  def render_all(project_nodes, filtered_modules, ext, config, opts) do
+    base = [
+      apps: config.apps,
+      deps: config.deps,
+      ext: ext,
+      extras: extra_paths(config),
+      skip_undefined_reference_warnings_on: config.skip_undefined_reference_warnings_on,
+      skip_code_autolink_to: config.skip_code_autolink_to,
+      filtered_modules: filtered_modules
+    ]
+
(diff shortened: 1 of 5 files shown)


yordis:
@josevalim is this what you had in mind? I am doing it chunk by chunk.

lib/ex_doc/formatter/epub.ex
```diff
@@ -4,8 +4,8 @@ defmodule ExDoc.Formatter.EPUB do
   @mimetype "application/epub+zip"
   @assets_dir "OEBPS/assets"
   alias __MODULE__.{Assets, Templates}
+  alias ExDoc.{Formatter, Utils}
   alias ExDoc.Formatter.HTML
```
josevalim, line 8: It seems we still invoke HTML in three remaining places.

lib/ex_doc/formatter.ex
```diff
@@ -0,0 +1,273 @@
+defmodule ExDoc.Formatter do
+  @moduledoc false
+
+  alias ExDoc.{Markdown, GroupMatcher, Utils}
+
+  @doc """
+  Autolinks and renders all docs.
+  """
```
josevalim, line 8:
```suggestion
  """
  # TODO: Move this to normalize_doc_ast in the retriever
```

josevalim:
Two minor nits and we can ship it.

josevalim:

PR #2148, opened 2025-10-02
feat: add Markdown formatter

yordis:
Signed-off-by: Yordis Prieto <yordis.prieto@gmail.com>

### DIFF
--- a/lib/ex_doc/formatter/markdown.ex
+++ b/lib/ex_doc/formatter/markdown.ex
@@ -1,0 +1,222 @@
+defmodule ExDoc.Formatter.MARKDOWN do
+  @moduledoc false
+
+  alias __MODULE__.{Templates}
+  alias ExDoc.Formatter
+  alias ExDoc.Utils
+
+  @doc """
+  Generates Markdown documentation for the given modules.
+  """
+  @spec run([ExDoc.ModuleNode.t()], [ExDoc.ModuleNode.t()], ExDoc.Config.t()) :: String.t()
+  def run(project_nodes, filtered_modules, config) when is_map(config) do
+    Utils.unset_warned()
+
+    config = normalize_config(config)
+
+    build = Path.join(config.output, ".build")
+    output_setup(build, config)
+
+    extras = Formatter.build_extras(config, ".md")
+
+    project_nodes =
+      project_nodes
+      |> Formatter.render_all(filtered_modules, ".md", config, highlight_tag: "samp")
+
+    nodes_map = %{
+      modules: Formatter.filter_list(:module, project_nodes),
+      tasks: Formatter.filter_list(:task, project_nodes)
+    }
+
+    config = %{config | extras: extras}
+
+    all_files =
+      [generate_nav(config, nodes_map)] ++
+        generate_extras(config) ++
+        generate_list(config, nodes_map.modules) ++
+        generate_list(config, nodes_map.tasks) ++
+        [generate_llm_index(config, nodes_map)]
+
+    generate_build(List.flatten(all_files), build)
+    config.output |> Path.join("index.md") |> Path.relative_to_cwd()
+  end
+
+  defp normalize_config(config) do
+    output = Path.expand(config.output)
+    %{config | output: output}
+  end
+
+  defp output_setup(build, config) do
+    if File.exists?(build) do
+      build
+      |> File.read!()
+      |> String.split("\n", trim: true)
+      |> Enum.map(&Path.join(config.output, &1))
+      |> Enum.each(&File.rm/1)
+
+      File.rm(build)
+    else
+      # Only remove markdown files, not HTML/EPUB files
+      File.mkdir_p!(config.output)
+
+      if File.exists?(config.output) do
+        config.output
+        |> Path.join("*.md")
+        |> Path.wildcard()
+        |> Enum.each(&File.rm/1)
+
+        llms_file = Path.join(config.output, "llms.txt")
+        if File.exists?(llms_file), do: File.rm(llms_file)
+      end
+    end
+  end
+
+  defp generate_build(files, build) do
+    entries =
+      files
+      |> Enum.uniq()
+      |> Enum.sort()
+      |> Enum.map(&[&1, "\n"])
+
+    File.mkdir_p!(Path.dirname(build))
+    File.write!(build, entries)
+  end
+
+  defp normalize_output(output) do
+    output
+    |> String.replace(~r/\r\n?/, "\n")
+    |> String.replace(~r/\n{3,}/, "\n\n")
+  end
+
+  defp generate_nav(config, nodes) do
+    nodes =
+      Map.update!(nodes, :modules, fn modules ->
+        modules |> Enum.chunk_by(& &1.group) |> Enum.map(&{hd(&1).group, &1})
+      end)
+
+    content =
+      Templates.nav_template(config, nodes)
+      |> normalize_output()
+
+    filename = "index.md"
+    File.write("#{config.output}/#{filename}", content)
+    filename
+  end
+
+  defp generate_extras(config) do
+    for {_title, extras} <- config.extras,
+        %{id: id, source: content} <- extras,
+        not is_map_key(%{id: id, source: content}, :url) do
+      filename = "#{id}.md"
+      output = "#{config.output}/#{filename}"
+
+      if File.regular?(output) do
+        Utils.warn("file #{Path.relative_to_cwd(output)} already exists", [])
+      end
+
+      File.write!(output, normalize_output(content))
+      filename
+    end
+  end
+
+  defp generate_list(config, nodes) do
+    nodes
+    |> Task.async_stream(&generate_module_page(&1, config), timeout: :infinity)
+    |> Enum.map(&elem(&1, 1))
+  end
+
+  ## Helpers
+
+  defp generate_module_page(module_node, config) do
+    content =
+      Templates.module_page(config, module_node)
+      |> normalize_output()
+
+    filename = "#{module_node.id}.md"
+    File.write("#{config.output}/#{filename}", content)
+    filename
+  end
+
+  defp generate_llm_index(config, nodes_map) do
+    content = generate_llm_index_content(config, nodes_map)
+    filename = "llms.txt"
+    File.write("#{config.output}/#{filename}", content)
+    filename
+  end
+
+  defp generate_llm_index_content(config, nodes_map) do
+    project_info = """
+    # #{config.project} #{config.version}
+
+    #{config.project} documentation index for Large Language Models.
+
+    ## Modules
+
+    """
+
+    modules_info =
+      nodes_map.modules
+      |> Enum.map(fn module_node ->
+        "- [#{module_node.title}](#{module_node.id}.md): #{module_node.doc |> ExDoc.DocAST.synopsis() |> extract_plain_text()}"
+      end)
+      |> Enum.join("\n")
+
+    tasks_info =
+      if length(nodes_map.tasks) > 0 do
+        tasks_list =
+          nodes_map.tasks
+          |> Enum.map(fn task_node ->
+            "- [#{task_node.title}](#{task_node.id}.md): #{task_node.doc |> ExDoc.DocAST.synopsis() |> extract_plain_text()}"
+          end)
+          |> Enum.join("\n")
+
+        "\n\n## Mix Tasks\n\n" <> tasks_list
+      else
+        ""
+      end
+
+    extras_info =
+      if is_list(config.extras) and length(config.extras) > 0 do
+        extras_list =
+          config.extras
+          |> Enum.flat_map(fn
+            {_group, extras} when is_list(extras) -> extras
+            _ -> []
+          end)
+          |> Enum.map(fn extra ->
+            "- [#{extra.title}](#{extra.id}.md)"
+          end)
+          |> Enum.join("\n")
+
+        if extras_list == "" do
+          ""
+        else
+          "\n\n## Guides\n\n" <> extras_list
+        end
+      else
+        ""
+      end
+
+    project_info <> modules_info <> tasks_info <> extras_info
+  end
+
+  defp extract_plain_text(html) when is_binary(html) do
+    html
+    |> String.replace(~r/<[^>]*>/, "")
+    |> String.replace(~r/\s+/, " ")
+    |> String.trim()
+    |> case do
+      "" ->
+        "No documentation available"
+
+      text ->
+        if String.length(text) > 150 do
+          String.slice(text, 0, 150) <> "..."
+        else
+          text
+        end
+    end
+  end
+
+  defp extract_plain_text(_), do: "No documentation available"
+end
--- a/lib/ex_doc/formatter/markdown/templates.ex
+++ b/lib/ex_doc/formatter/markdown/templates.ex
@@ -1,0 +1,80 @@
+defmodule ExDoc.Formatter.MARKDOWN.Templates do
+  @moduledoc false
+
+  require EEx
+
+  import ExDoc.Utils,
+    only: [before_closing_body_tag: 2, h: 1]
+
+  @doc """
+  Generate content from the module template for a given `node`.
+  """
+  def module_page(config, module_node) do
+    docs =
+      for group <- module_node.docs_groups do
+        {group.title, group.docs}
+      end
+
+    module_template(config, module_node, docs)
+  end
+
+  @doc """
+  Formats the attribute type used to define the spec of the given `node`.
+  """
+  def format_spec_attribute(module, node) do
+    module.language.format_spec_attribute(node)
+  end
+
+  @doc """
+  Returns the original markdown documentation from source_doc.
+  """
+  def node_doc(%{source_doc: %{"en" => source}}) when is_binary(source), do: source
+  def node_doc(_), do: nil
+
+  @doc """
+  Creates a chapter which contains all the details about an individual module.
+  """
+  EEx.function_from_file(
+    :def,
+    :module_template,
+    Path.expand("templates/module_template.eex", __DIR__),
+    [:config, :module, :docs],
+    trim: true
+  )
+
+  @doc """
+  Creates the table of contents.
+  """
+  EEx.function_from_file(
+    :def,
+    :nav_template,
+    Path.expand("templates/nav_template.eex", __DIR__),
+    [:config, :nodes],
+    trim: true
+  )
+
+  EEx.function_from_file(
+    :defp,
+    :nav_item_template,
+    Path.expand("templates/nav_item_template.eex", __DIR__),
+    [:name, :nodes],
+    trim: true
+  )
+
+  EEx.function_from_file(
+    :defp,
+    :nav_grouped_item_template,
+    Path.expand("templates/nav_grouped_item_template.eex", __DIR__),
+    [:nodes],
+    trim: true
+  )
+
+  @doc false
+  EEx.function_from_file(
+    :def,
+    :detail_template,
+    Path.expand("templates/detail_template.eex", __DIR__),
+    [:node, :module],
+    trim: true
+  )
+end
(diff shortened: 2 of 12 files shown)


yordis:
@josevalim any thoughts here thus far?

josevalim:
Sorry, a bit busy with Elixir v1.19-rc and today I was out of focus. It is in my inbox and I will review it as soon as I can.

leandrocp (commented):
Hey @yordis that's amazing! Let me share some ideas that could be useful.

I was inspecting the generated Markdown files and thinking if we could apply some changes. See the differences here https://gist.github.com/leandrocp/ee4f0ba8325b410b8650ccd26b9b2351 (CompiledWithDocs.md vs CompiledWithDocs_PROPOSAL.md)
- Use frontmatter block to describe global values/notes
- Use the format "Summary / Functions" similar to HTML pages (so Functions become a level 2 heading)
- Include source links
- Reorganize metadata, for eg: add (deprecated) and doc in summary
- Remove links to hexdocs.pm because 1) it breaks the content and I guess that would make it harder for LLMs to parse; 2) I'm not sure it should link to html pages

You can see some examples on https://shopify.dev/docs/api/liquid/basics.md and https://vercel.com/docs/rest-api/reference/sdk.md

lib/ex_doc/formatter/markdown/templates.ex
```diff
@@ -0,0 +1,169 @@
...
+    trim: true
+  )
+
+  EEx.function_from_file(
+    :defp,
+    :nav_grouped_item_template,
+    Path.expand("templates/nav_grouped_item_template.eex", __DIR__),
+    [:nodes],
+    trim: true
+  )
```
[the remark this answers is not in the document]
josevalim: We try our best to avoid adding dependencies to ExDoc because they may impose further constraints on users of ExDoc, given ExDoc is used by almost every library. Furthermore, if we use MDEx as a dependency, then you need to write some doc publishing helpers in MDEx, cause you will no longer be able to depend on ExDoc. :D So it is best to postpone this as much as possible.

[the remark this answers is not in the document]
leandrocp: > We try our best to avoid adding dependencies to ExDoc because they may impose further constraints on users of ExDoc, given ExDoc is used by almost every library. Furthermore, if we use MDEx as a dependency, then you need to write some doc publishing helpers in MDEx, cause you will no longer be able to depend on ExDoc. :D So it is best to postpone this as much as possible.

Makes sense :D

yordis:
@leandrocp thanks for the help, about the formatting, I do not have any strong opinions of the final output, I will leave it to @josevalim and you to decide on that front, I can adjust it

josevalim:
Just a heads up I am on holidays. This is still on top of my list but I cannot promise anything until I am back (next Wednesday). Sorry!

leandrocp:
Hey @yordis no strong opinions either, just trying to figure out a format that presents all the info.

Regarding links, just to be clear it should contain links eventually but I believe that HexDocs must define how to serve Markdown pages first, for eg something like https://hexdocs.pm/elixir/1.18.4/Enum.md

josevalim:
Some additional context: ExDoc is not tied to HexDocs and vice versa. We should design links with the assumptions it is served by any static host. I am thinking just using Enum.md with an anchor, same as html, is fine though? The anchor won’t actually work, but at least it is some context?

leandrocp:
> just using Enum.md with an anchor, same as html, is fine though?

Yeah it seems appropriate.

---

Shopify docs links to "canonical" URLs, for eg in https://shopify.dev/docs/api/liquid/tags/form you can see a bunch of anchor links, but in the .md docs (https://shopify.dev/docs/api/liquid/tags/form.md) they are rendered as ``[`cart`](https://shopify.dev/docs/api/liquid/tags/form#form-cart)``

Vercel is not much different, they include relative links inside .md docs, for eg ``[Deployments Automation](/examples/deployments-automation)``

Claude use relative links with anchors, for eg in https://docs.claude.com/en/docs/build-with-claude/prompt-caching you can see a link ``<a href="#pricing" class="link">at additional cost</a>`` which is rendered as ``[at additional cost](#pricing)`` in the .md docs (https://docs.claude.com/en/docs/build-with-claude/prompt-caching.md)

josevalim:
I am back. My goal is to release the new hexdocs.pm, review #2055, and then focus on this.

yordis:
I am around, once you folks aligned, please share a action-driven feedback, I am can put the work on

lib/ex_doc/formatter/markdown.ex
```diff
@@ -0,0 +1,211 @@
...
+      config.output
+      |> Path.expand()
+      |> Path.join("markdown")
+
+    %{config | output: output}
+  end
+
+  defp normalize_output(output) do
+    output
+    |> String.replace(~r/\r\n|\r|\n/, "\n")
```
josevalim:
```suggestion
    |> String.replace(["\r\n", "\n"], "\n")
```

yordis:
@josevalim could you give a second pass to the code review 🙏🏻

lib/ex_doc/formatter/markdown/templates/module_template.eex
```diff
@@ -0,0 +1,36 @@
+# <%= module.title %> <%= module_type(module) %> (<%= config.project %> v<%= config.version %>)
+
+<%= for annotation <- module.annotations do %>*(<%= annotation %>)* <% end %>
+
+<%= if deprecated = module.deprecated do %>
+> This <%= module.type %> is deprecated. <%=h deprecated %>.
+<% end %>
+
```
eksperimental:
```suggestion
<%= if deprecated = module.deprecated do %>
> This <%= module.type %> is deprecated. <%=h deprecated %>.

<% end %>
```

lib/ex_doc/formatter/markdown/templates.ex
```diff
@@ -0,0 +1,244 @@
...
+
+  @doc """
+  Format the attribute type used to define the spec of the given `node`.
+  """
+  def format_spec_attribute(module, node) do
+    module.language.format_spec_attribute(node)
+  end
+
+  @doc """
+  Generated ID for static file.
```
eksperimental:
```suggestion
  Generates an ID for a static file.
```

lib/ex_doc/formatter/markdown/templates.ex
```diff
@@ -0,0 +1,244 @@
...
+  end
+
+  def node_doc(_), do: nil
+
+  @doc """
+  Get synopsis for a node, handling both DocAST and string documentation.
+  Uses DocAST synopsis extraction logic for consistency with HTML formatter.
+  """
+  def node_synopsis(%{doc: doc}) when is_list(doc) do
+    # For DocAST, extract synopsis DocAST and convert to markdown
```
eksperimental:
```suggestion
    # For DocAST, extract synopsis DocAST and convert to Markdown.
```

Also the code is self-explanatory, I guess this kind of comments can be removed.

lib/ex_doc/formatter/markdown/templates.ex
```diff
@@ -0,0 +1,244 @@
...
+  def node_synopsis(%{source_doc: %{"en" => source}}) when is_binary(source) do
+    synopsis(source)
+  end
+
+  def node_synopsis(%{rendered_doc: source}) when is_binary(source) do
+    synopsis(source)
+  end
+
+  def node_synopsis(%{source_doc: %{"en" => source}}) when is_list(source) do
+    # For Erlang DocAST, extract synopsis and convert to plain text
```
eksperimental: I think the code is self-explanatory here.
If agreed upon, please remove everywhere else.

```suggestion
```

lib/ex_doc/formatter/markdown.ex
```diff
@@ -0,0 +1,177 @@
...
+      config.output
+      |> Path.expand()
+      |> Path.join("markdown")
+
+    %{config | output: output}
+  end
+
+  defp normalize_output(output) do
+    output
+    |> String.replace(["\r\n", "\n"], "\n")
```
eksperimental: You need to review this one as you are replacing "\n" with "\n".

lib/ex_doc/doc_ast.ex
```diff
@@ -65,6 +65,68 @@ defmodule ExDoc.DocAST do
...
+    "\n\n"
+  end
+
+  def to_markdown({:img, attrs, _inner, _meta}) do
+    alt = attrs[:alt] || ""
+    title = attrs[:title] || ""
+    "![#{alt}](#{attrs[:src]} \"#{title}\")"
+  end
+
+  # Ignoring these: area base col command embed input keygen link meta param source track wbr
```
eksperimental: Self explanatory code.

```suggestion
```

Alternatively

```suggestion
  # Ignore void elements.
```

lib/ex_doc/formatter/markdown/templates.ex
```diff
@@ -0,0 +1,244 @@
...
+  Gets the first paragraph of the documentation of a node. It strips
+  surrounding white-spaces and trailing `:`.
+
+  If `doc` is `nil`, it returns `nil`.
+  """
+  @spec synopsis(String.t()) :: String.t()
+  @spec synopsis(nil) :: nil
+  def synopsis(doc) when is_binary(doc) do
+    case :binary.split(doc, "\n\n") do
+      [left, _] -> String.trim_trailing(left, ": ") <> "\n\n"
```
eksperimental: `String.trim_trailing/2` does not strip every character in the second argument.
Also note that `String.trim_trailing/1` removes all Unicode white-spaces, so it's better to use that. There's an edgecase that will not be covered though `"foo  :   :  :"`

```suggestion
      [left, _] -> String.trim_trailing(left) |> String.trim_trailing(left, ":") <> "\n\n"
```

lib/ex_doc/formatter/markdown/templates.ex
```diff
@@ -0,0 +1,244 @@
...
+      [all] -> all
+    end
+  end
+
+  def synopsis(_), do: nil
+
+  @heading_regex ~r/^(\#{1,6})\s+(.*)/m
+  defp rewrite_headings(content) when is_binary(content) do
+    @heading_regex
```
eksperimental: Storing regular expressions in module attributes has been deprecated in Elixir v1.19.

```suggestion
  defp rewrite_headings(content) when is_binary(content) do
    ~r/^(\#{1,6})\s+(.*)/m
```

lib/ex_doc/formatter/markdown/templates/nav_template.eex
```diff
@@ -0,0 +1,9 @@
+# <%= config.project %> v<%= config.version %> - Documentation - Table of contents
```
eksperimental:
```suggestion
# <%= config.project %> v<%= config.version %> - Documentation - Table of Contents
```

eksperimental:
Hi @yordis, here are a few comments:

I generate Elixir docs with this command `DOCS_OPTIONS="--warnings-as-errors --formatter markdown" make docs`

1. The docs are created under the `markdown/` folder, but this is not consistent with the html and epub formatter.
We should also consider when generating multiple formats where are they going to be placed. Having html and md files in the same folder could be a bit disorganized.

2. Links are wrongly created:

```
    [```inline
    Kernel
    ```
    ](Kernel.md)
```

when it should be `` `[Kernel](Kernel.md)` ``

3. I don't think the title should be part of H1 of every page. What I did in my PR was
instead of
```
# Kernel  (Elixir v1.20.0-dev)
```

do

```
# Kernel

(Elixir v1.20.0-dev)
```

3. Bulleted lists are not formatted. They are just plain lines without the `- `

4. Headers don't have `###` 

I think you can use the output generated by my previous PR (https://github.com/elixir-lang/ex_doc/pull/1992)

https://raw.githubusercontent.com/eksperimental/ex_doc_markdown_formatter_docs/refs/heads/main/elixir/Kernel.md

It can work as reference (it was a WIP but the main formatting issues where addressed).

yordis:
@eksperimental should I change the tests to do full equality of the content? Maybe read them from a file snapshot or somthing, I am having a hard time trying to follow the existing tests setup, I can't tell the final structure that easy but I did not want to change it that much neither

eksperimental:
> @eksperimental should I change the tests to do full equality of the content? Maybe read them from a file snapshot or somthing, I am having a hard time trying to follow the existing tests setup, I can't tell the final structure that easy but I did not want to change it that much neither

I would focus on the final structure first and then add/update tests accordingly.

yordis:
What should I do about that `.build-markdown` file? Should that be merged into the existing `.build`, since I did not see `.epub` in the `.build` I guessed it would be a different one

eksperimental:
> What should I do about that `.build-markdown` file? Should that be merged into the existing `.build`, since I did not see `.epub` in the `.build` I guessed it would be a different one

I think having a subfolder per format would be wise.

I think having everything living in the same folder is no longer viable.

yordis:
> I think having a subfolder per format would be wise.

I agree with this!

yordis:
> I think having a subfolder per format would be wise.
@eksperimental implementing the multi-dir I just realized, is that `just` for the `.build` files, are you thinking on something like,

```
tree doc 
doc
├── epub
│   └── ExDoc.epub
├── html
│   ├── 404.html
│   ├── admonition.html
│   ├── api-reference.html
│   ├── changelog.html
│   ├── cheatsheet.html
│   ├── dist
│   │   ├── html-elixir-VYWJUHZE.css
│   │   ├── html-HBZYRXZS.js
│   │   ├── lato-latin-400-normal-W7754I4D.woff2
│   │   ├── lato-latin-700-normal-2XVSBPG4.woff2
│   │   ├── lato-latin-ext-400-normal-N27NCBWW.woff2
│   │   ├── lato-latin-ext-700-normal-Q2L5DVMW.woff2
│   │   ├── remixicon-QPNJX265.woff2
│   │   ├── search_data-B202F4E6.js
│   │   └── sidebar_items-ED2F9D7B.js
│   ├── ExDoc.Markdown.Earmark.html
│   ├── ExDoc.Markdown.html
│   ├── index.html
│   ├── Mix.Tasks.Docs.html
│   ├── readme.html
│   └── search.html
└── markdown
    ├── ExDoc.Markdown.Earmark.md
    ├── ExDoc.Markdown.md
    ├── index.md
    ├── llms.txt
    └── Mix.Tasks.Docs.md
```

eksperimental:
I think that's fine. Having it as before epub and html in the main folder and only markdown in it's own sub-folder was not consistent. As more formatters are added, having everything in one folder it will become unmanageable.

By having sub-folders give the option to the user to easily create a zip file per formatter for users to download the docs.

Maybe having an option to user sub-folders could be desirable to not to break compatibility of apps using this as hexdocs.pm.

Those are my 2 cents.

yordis:
> not to break compatibility of apps using this as hexdocs.pm.

The reason I asked is because of hexdocs and how people could "deploy" these artifacts, having the subdir for `pub` or `html` would introduce breaking change, **I am guessing** since it will require to add `/html` `/pub` if a simply static hosting is in place.

So, should I duplicate the `pub` and `html` and keep them as well in the subdirs, or should I just add the markdown?

eksperimental:
Maybe have an optional flag to use sub-folders for now, and not having it the set by default. Later on this can be set by default in v1.0 or remove it completely and make it the default (and only) behavior.

Also I think we should ask @josevalim's opinion.

yordis:
@eksperimental I will wait for @josevalim and hopefully after that we can get a clear path forward, eitherway I can get it done.

lib/ex_doc.ex
```diff
@@ -35,11 +35,16 @@ defmodule ExDoc do
...
   defp find_formatter(name) do
-    [ExDoc.Formatter, String.upcase(name)]
+    [ExDoc.Formatter, format_module_name(name)]
     |> Module.concat()
     |> check_formatter_module(name)
   end
 
+  defp format_module_name("html"), do: "HTML"
+  defp format_module_name("epub"), do: "EPUB"
+  defp format_module_name("markdown"), do: "Markdown"
```
josevalim: Let's please keep the previous convension of upcase, we can revisit it later.

test/ex_doc/doc_ast_test.exs
```diff
@@ -161,21 +334,18 @@ defmodule ExDoc.DocASTTest do
 
   describe "highlight" do
     test "with default class" do
-      # Four spaces
```
josevalim: Those code comments should not be removed, right?

lib/ex_doc/formatter/markdown/templates/summary_template.eex
```diff
@@ -0,0 +1,15 @@
+### <%= name %>
```
josevalim: Do we need summaries for Markdown?

I think we only need summaries if we are going to emit a different markdown file per function/macro/callback/type. Which would be helpful to help us implement the copy feature...

lib/ex_doc/formatter/markdown/templates/module_template.eex
```diff
@@ -0,0 +1,35 @@
...
+## Contents
+
+<%= for {name, nodes} <- summary, _key = text_to_id(name) do %>
+### <%=h to_string(name) %>
+
+<%= for node <- nodes do %>
+<%= detail_template(node, module) %>
+<% end %>
+
+<% end %>
```
josevalim: I would start with a much simpler version for Markdown, something like this:

```suggestion
<%= for {_name, nodes} <- summary, node <- nodes do %>
  <%= detail_template(node, module) %>
<% end %>
```

My thinking is that, because we tell users they can use `##` in their docs, our only option is to use `#` for the functions, types, macros, etc.

josevalim:
Sorry for the delay @yordis. I think this could be much simpler by keeping the original markdown and using this structure for modules:

```
# Module

# fun/3

*annotations*

Source: ...

Signatures:
...

Specs:
...

<%= original_markdown %>

# t:type/1

*annotations*

Source: ...

<%= original_markdown %>
```

We need to later have a separate discussion if it should be a single or multiple files, but for now, that's the simplest way to go.

yordis:
@josevalim are you sure about `# fun/3` instead of `## fun/3` (two #)? Just want to confirm it was intended.

josevalim:
Yes, because users can use h2.

yordis:
@josevalim what should do with `.build` and what should I do with the `markdown` files as in where to put them?

```
 tree -a doc
doc
├── .build
├── 404.html
├── admonition.html
├── api-reference.html
├── changelog.html
├── cheatsheet.html
├── dist
│   ├── html-elixir-VYWJUHZE.css
│   ├── html-HBZYRXZS.js
│   ├── lato-latin-400-normal-W7754I4D.woff2
│   ├── lato-latin-700-normal-2XVSBPG4.woff2
│   ├── lato-latin-ext-400-normal-N27NCBWW.woff2
│   ├── lato-latin-ext-700-normal-Q2L5DVMW.woff2
│   ├── remixicon-QPNJX265.woff2
│   ├── search_data-B202F4E6.js
│   └── sidebar_items-ED2F9D7B.js
├── ExDoc.epub
├── ExDoc.Markdown.Earmark.html
├── ExDoc.Markdown.html
├── index.html
├── markdown
│   ├── .build
│   ├── ExDoc.Markdown.Earmark.md
│   ├── ExDoc.Markdown.md
│   ├── index.md
│   ├── llms.txt
│   └── Mix.Tasks.Docs.md
├── Mix.Tasks.Docs.html
├── readme.html
└── search.html

3 directories, 28 files
```

josevalim:
> @josevalim what should do with .build and what should I do with the markdown files as in where to put them?

We should keep the build for markdown. Just the one for epub that was not part of this PR!

yordis:
> We should keep the build for markdown. Just the one for epub that was not part of this PR!

Then, it is ready for CR

josevalim:
I think we are almost there, we just need to remove the `to_markdown` function and its tests, and remove the changes to the autolinker, cause i think those are not longer necessary either.

josevalim:
Note to self: we need to check if the original documentation is also in markdown and skip any module where this is not true, but this can be added after merging.

josevalim:

PR #2153, opened 2025-10-20
Allow multiple search engines

josevalim:
We will now allow multiple search engines with ExDoc, so people can choose between local search, the upcoming HexDocs search engine, or even a third party like Algolia. This will allow Phoenix projects, for example, to opt into searching across all Phoenix projects. Here is what it looks like:

Screenshot 2025-10-20 at 14 41 28

---

@garazdawi I have removed `exdoc:full-text-search-url`, you can now use the new search option to set one engine (which will be used by default) or even multiple.

I have also streamlined the search results to include the initial snippet/description rather than useless highlights.

### DIFF
--- a/lib/ex_doc/config.ex
+++ b/lib/ex_doc/config.ex
@@ -45,7 +45,7 @@ defmodule ExDoc.Config do
             project: nil,
             redirects: %{},
             retriever: ExDoc.Retriever,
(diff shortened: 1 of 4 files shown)


josevalim:

garazdawi:
Ping @kikofernandez and @frazze-jobb so that they know if we update ExDoc to this version we need to also modify the code on erlang.org for the Algolia integration.

Issue #2150, opened 2025-10-11
Use a better scale in headings typhography


lud:
Hello,

Sometimes I find that document structure is hard to read, especially in markdown guides. Here is an example:

----------

<img width="914" height="533" alt="Image" src="https://github.com/user-attachments/assets/91768680-68bf-4fc2-8c38-45bb512285b7" />

----------

* **Plug integration** is an H2
* **Plug options** is an H3
* **Using a module** is an H4

To me there is too little difference between H2 and H3 levels, which make section look like they are unrelated. Fortunately the different whitespace margin helps.

Also, the H4 level feels way too small, I feel like it should be at least as high as the copy font. (In my opinion H6 should be paragraph size and other headings larger).

What do you think about this?

Thank you :)

DavidOliver:
Agreed. I'll look to better differentiate the levels of heading.

lud:
I don't know if you are familiar with [typography scales](https://typescale.net/). It can help to quickly chose sizes that fit together.

DavidOliver:
I'm aware of the concept but hadn't come across that particular tool, which looks great. Thanks.

If heading levels 1-6 all need to be appreciably distinct from their nearest neighbours, it can be hard to depend on size alone without level 1 being awkwardly large in the context of documentation pages, so I expect I'll also be experimenting with devices such as horizontal rules.

josevalim:
In our case, I think we can adjust the sizes assuming folks will use h1 to h4. H5/H6 is rarely used and we can use font weight or similar for those.

lud:
The Major Third scale is not too big on H1 but yes, H5 could be just bold and h6 normal with a margin top.

DavidOliver:
It turns out that the difference in size between our H2 and H3 was already greater than that of the sizes which typescale.net suggests for the same levels with the Major Third scale.

What do you think of the below (current/suggested):

<img width="1192" height="726" alt="Image" src="https://github.com/user-attachments/assets/2796e30f-7c3f-41d4-a9a5-17302cd53f96" />
<img width="1191" height="766" alt="Image" src="https://github.com/user-attachments/assets/91c0b168-cd22-4dda-bca2-2007928a04e3" />

The H2 is at 1.8rem (and so closer to the H1 size of 2rem, which I don't think matters as the only H1 should be at the very top, once) with a horizontal rule. Sizes of H2-H4 (all three shown here) have increased.

The horizontal rule helps to separate the main sections:

[...]

PR #2160, opened 2025-10-23
Update content headings sizing and spacing

DavidOliver:
Re. #2150

Not ready for merging. More thorough testing of existing changes and additional styling of H5 and H6 are to do.

Some temporary notes are included in the new CSS.

### DIFF
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -62,32 +62,50 @@
   color: var(--textHeaders);
 }
 
-.content-inner h1 {
-  font-size: 2em;
-  margin: 0.5em 0;
-}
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you @DavidOliver! I gave it a try locally and I'd love if we could reduce the changes instead. Would changing/adding the font-sizes be enough?

DavidOliver:
@josevalim, I'm adding a couple of PRs for you to try: one with only size changes and one with margin changes which are smaller than those in this PR.

I find a little extra spacing at the top of headings pleasing, and I think it helps with the reading flow.

PR #2163, opened 2025-10-24
Update content headings (sizing and a little extra spacing)

DavidOliver:
Re. #2150

Not ready for merging - I need to check that these changes don't affect more headings than intended.

### DIFF
--- a/assets/css/content/general.css
+++ b/assets/css/content/general.css
@@ -62,32 +62,43 @@
   color: var(--textHeaders);
 }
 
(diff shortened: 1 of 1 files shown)


josevalim:
@DavidOliver looks good to me! The h2 is too close to h1, but I guess that's fine since we are only meant to have one h1 anyway, and always at the top, so no ambiguith there. Btw, the admonition titles are too large  now:

Screenshot 2025-10-24 at 18 18 44

PR #2166, opened 2025-10-30
Add Elixir v1.19 and Erlang/OTP 28 to CI

eksperimental:


eksperimental:
This is because I am running Elixir v1.19.

What should we do here? Stick around to the lowest Elixir supported Elixir version.

Also when we add Elixir v1.19 to the CI, it will fail

josevalim:
We always format on the latest supported Elixir version. So we should add 1.19 to CI and swap the flag to run the formatter only on 1.19!

eksperimental:
> We always format on the latest supported Elixir version. So we should add 1.19 to CI and swap the flag to run the formatter only on 1.19!

I will submit a PR

josevalim:
Feel free to do it as part of this PR, you can rename it to update to v1.19 (which will mean running the formatter on v1.19 anyway!)

eksperimental:
Done

josevalim:

PR #2172, opened 2025-11-21
Fix docs: Move `source_url` to `project`

ftes:
https://hexdocs.pm/ex_doc/Mix.Tasks.Docs.html#module-using-source_url-and-source_ref-together

shows this example

```ex
@version "0.30.10"
def project do
  [
    ...
    version: @version,
    docs: docs(),
    ...
  ]
end

def docs do
  ...
  source_ref: "v#{@version}",
  source_url: @source_url,
  ...
end
```

Which is incorrect afaict. `source_url` is not a [docs option](https://hexdocs.pm/ex_doc/Mix.Tasks.Docs.html#module-configuration) and should be moved to the `project` function.

### DIFF
--- a/lib/mix/tasks/docs.ex
+++ b/lib/mix/tasks/docs.ex
@@ -187,6 +187,9 @@ defmodule Mix.Tasks.Docs do
 
     * `:source_beam` - Path to the beam directory; default: mix's compile path.
 
+    * `:source_url` - The source URL fallback if `:source_url` is not given at the project
+      configuration. See the following sections.
+
     * `:source_ref` - The branch/commit/tag used for source link inference;
       default: "main".
 
@@ -228,7 +231,7 @@ defmodule Mix.Tasks.Docs do
   both of them will allow ExDoc to link to specific version of the code for a function or module
   that matches the version of the docs. So if the docs have been generated for version 1.0.5 then
   clicking on the source link in the docs will take the browser to the source code for the 1.0.5
-  version of the code instead of only the primary ref (e.g. `main`).
+  version of the code instead of only the primary ref (for example, `main`).
 
   A example setup looks like:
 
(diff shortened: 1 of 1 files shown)


josevalim:
I am looking at the source code and the ones in the top level are simply moved inside:

```

  defp normalize_source_url(options, config) do
    if source_url = config[:source_url] do
      Keyword.put(options, :source_url, source_url)
    else
      options
    end
  end

  defp normalize_homepage_url(options, config) do
    if homepage_url = config[:homepage_url] do
      Keyword.put(options, :homepage_url, homepage_url)
    else
      options
    end
  end
```

So I am thinking both are supported and the docs are correct. Thoughts?

ftes:
> I am looking at the source code and the ones in the top level are simply moved inside:

Ah, interesting. I see this behaviour for generating source links:
Top-level `project.source_url` takes precedence ([example](https://github.com/elixir-lang/ex_doc/commit/031cf469a2f596fb93db5cb274f579ee09d55d05)).
It overwrites `docs.source_url` if both are present.
Only if `project.source_url` is not present, then `docs.source_url` is used.

Since
1. `docs.source_url` is not mentioned as a config option elsewhere
2. `project.source_url` takes precedence

I think the example in question should be changed to use `project.source_url`.

Going further, maybe `docs.source_url` should even be deprecated?
I find it unnecessarily complex and thus potentially confusing without further explanation.
Unless there are valid use cases when you don't want `project.source_url` but do want `docs.source_url`.

josevalim:
I believe the original goal was to have source_url/homepage_url also be used by Hex to list in the packages page, but I think it was never implemented.

josevalim:
All other examples used it at the root, so I merged it for consistency. I don't think it is worth deprecating though as to not push users through this hassle. Thanks!

PR #2176, opened 2025-12-17
Add copy-as-markdown button functionality

yordis:


yordis:
@josevalim any updates on this one?

gmile:
FWIW, this would be quite useful to me too. Each time I am reading a document online (recently: https://www.erlang.org/doc/apps/tools/xref.html), I am thinking "It would be so convenient to save this on disk so my LLM agent can look at it too".

There's no problem with saving entire html page, it's just that it would pollute the LLM's context with unnecessary tag tokens.

josevalim:
Now in main!

PR #2179, opened 2026-01-09
Default to README for main page.

adamu:
In many projects, I find that important documentation is included in the README in the git repo, but the default in hexdocs is the module name, which is often quite sparse comparatively.

This PR changes the default example to use the README as the main page, which I think will be the most useful starting point for most projects. The comment explains that it can be set to a module name instead, if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ end
 
 defp docs do
   [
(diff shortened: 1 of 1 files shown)


josevalim:

PR #2181, opened 2026-01-11
Markdown formatter refactor

milmazz:
I was curious about this new formatter in `ex_doc` and I did an initial review. Please let me know what you think.

### DIFF
--- a/lib/ex_doc/formatter/markdown.ex
+++ b/lib/ex_doc/formatter/markdown.ex
@@ -11,15 +11,10 @@ defmodule ExDoc.Formatter.MARKDOWN do
       |> Enum.filter(&(&1.source_format == "text/markdown"))
       |> Enum.split_with(&(&1.type != :task))
 
-    groups =
-      Map.new([modules: modules, mix_tasks: tasks, extras: extras], fn {k, v} ->
-        {k, group_by_group(v)}
-      end)
-
     all_files =
-      generate_nav(config, groups) ++
-        generate_api_reference(config, groups) ++
-        generate_extras(extras, config) ++
+      generate_nav(config, modules, tasks, extras) ++
+        generate_api_reference(config, modules, tasks) ++
+        generate_extras(config, extras) ++
         generate_list(config, modules) ++
         generate_list(config, tasks)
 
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @milmazz! There are some other PRs and I have also refactored some of this code locally, so let's please hold on before merging this, unless there is a bug fix. :)

lib/ex_doc/formatter/markdown.ex
```diff
@@ -155,57 +169,64 @@ defmodule ExDoc.Formatter.MARKDOWN do
...
+      Enum.map(nodes_map.modules, fn module_node ->
+        synopsis = synopsis(module_node.doc)
+
+        ["- [#{module_node.title}](#{module_node.id}.md): ", synopsis, "\n"]
       end)
-      |> Enum.join("\n")
 
     tasks_info =
-      if length(nodes_map.tasks) > 0 do
+      if Enum.any?(nodes_map.tasks) do
```
milmazz: We just care if there is at least one element in the list, we don't mind about the total number of entries.

milmazz:
> Hi @milmazz! There are some other PRs and I have also refactored some of this code locally, so let's please hold on before merging this, unless there is a bug fix. :)

No worries, I'm so inactive in this project that I wasn't planning on merging code without any maintainer review.

josevalim:
@milmazz I am done with my refactor. Please rebase! The markdown module changed considerably but maybe I dropped a few things :)

milmazz:
@josevalim Hey, the new formatter is basically a whole revamp on the previous version :)

I just did some minor refactoring:

* The optional description in the `llms.txt` should be a block quote according to the docs from:  https://llmstxt.org
* Group the resources just once and pass that as argument to private helpers
* Use some pipelines to denote the sequence of data transformation
* Use `File.write!/2` for everything, as in the `html` formatter.

lib/ex_doc/formatter/markdown.ex
```diff
@@ -11,9 +11,14 @@ defmodule ExDoc.Formatter.MARKDOWN do
       |> Enum.filter(&(&1.source_format == "text/markdown"))
       |> Enum.split_with(&(&1.type != :task))
 
+    groups =
+      Map.new([modules: modules, mix_tasks: tasks, extras: extras], fn {k, v} ->
+        {k, group_by_group(v)}
+      end)
```
josevalim: I decided explicitly to not go down this route so we can more clearly see what is passed to what and what depends on what.

josevalim:
One tiny comment and we can fix it!

lib/ex_doc/formatter/markdown.ex
milmazz: Ok, seems fair the tiny extra compute for more readability. I will update.

milmazz:
@josevalim There might be a chance that someone is only interested in the markdown formatter output, the current implementation of the markdown formatter doesn't copy the assets to the `assets/` folder, but the content might link to those resources, should we conditionally copy those assets?

josevalim:
@milmazz yes, sounds like a good idea!

PR #2185, opened 2026-01-15
support anchors on redirects

zachdaniel:
I want to redirect an old function to a subheading in a guide, but the current set of redirects doesn't support it.

### DIFF
--- a/lib/ex_doc/formatter/html.ex
+++ b/lib/ex_doc/formatter/html.ex
@@ -206,8 +206,8 @@ defmodule ExDoc.Formatter.HTML do
       source = from <> ext
 
       destination =
-        case String.split(to, "#") do
-          [to, anchor | _] -> to <> ext <> "#" <> anchor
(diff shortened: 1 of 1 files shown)


lib/ex_doc/formatter/html.ex
```diff
@@ -229,7 +235,9 @@ defmodule ExDoc.Formatter.HTML do
   end
 
   defp generate_redirect(filename, config, redirect_to) do
-    unless case_sensitive_file_regular?("#{config.output}/#{redirect_to}") do
+    without_anchor = String.split(redirect_to, "#") |> Enum.at(0)
+
+    unless case_sensitive_file_regular?("#{config.output}/#{without_anchor}") do
```
josevalim:
```suggestion
    without_anchor = String.split(redirect_to, "#") |> hd()

    unless case_sensitive_file_regular?("#{config.output}/#{without_anchor}") do
```

josevalim:

PR #2164, opened 2025-10-24
Better differentiate headings/sections

DavidOliver:
Re. #2150

@josevalim, in the previous PR, module heading spacing (under "Summary", for example) and cheatsheets were affected, as well as admonition block headings as you pointed out.

This PR changes approach and attempts to target only "top content" in "module" and "extra" pages, and only those headings which are direct children of the #top-content element.

I'll test further before converting to a full/non-draft PR, but I thought I'd put this up to see what you think of this approach.

PR #2192, opened 2026-01-26
Fix Summary Types heading size

### LINKED ISSUE #2191: Section headings are styled unevenly (solution included)
Headings "Summary", "Callbacks" and "Functions" are styled differently than headings with other titles.
These other headings are displayed quite a bit larger. These include the heading "Types", and custom headings generated by the `@doc group` tag (see screenshots below).

This is caused by very specific CSS selectors that include the section name-as-identifiers for "Summary", "Callbacks" and "Functions":

```css
:is(#summary, #callbacks, #functions)>h1.section-heading
:is(#summary, #callbacks, #functions)>h2.section-heading
```

The selectors exclude sections with other IDs.

## Changes

The currently generated HTML contains class name `details-list` that is unused by CSS.
This class can be used instead of the specific ID selectors.

In `general.css`, change:

```css
& h2,
& :is(#summary, #callbacks, #functions)>h1.section-heading {
  font-size: var(--h2-size);
  margin-top: 1.5em;
  margin-bottom: 0.5em;
}

& h3,
& :is(#summary, #callbacks, #functions)>h2.section-heading,
& #summary :is(.summary-callbacks, .summary-functions) h2 {
  font-size: var(--h3-size);
  margin-top: 1.5em;
  margin-bottom: 0.5em;
}
```

to:

```css
& h2,
.details-list h1 {
  font-size: var(--h2-size);
  margin-top: 1.5em;
  margin-bottom: 0.5em;
}

& h3,
.details-list h2 {
  font-size: var(--h3-size);
  margin-top: 1.5em;
  margin-bottom: 0.5em;
}
```

### Side-by-side comparison of a regular page and a page with heading "Types"

The Types heading inside Summary should be styled as h3 instead of h2.
The Types heading at the same level as Summary should be styled as h2 instead of h1.

Image

### Side-by-side comparison of a regular page and a page with heading "Types"

The Types headings are aligned with the Function headings.

Image

### Side-by-side comparison of a page with headings for a custom group

Left the current styling, right the updated styling.
At the right side, the custom heading is aligned with the Summary heading.

Image

DavidOliver:
Fixes #2191

### DIFF
--- a/assets/build/build.js
+++ b/assets/build/build.js
@@ -46,10 +46,9 @@ Promise.all(formatters.map(async ({formatter, ...options}) => {
   const buildOptions = {
     entryNames: watchMode ? '[name]-dev' : '[name]-[hash]',
     bundle: true,
-    minifySyntax: !watchMode,
-    minifyIdentifiers: !watchMode,
-    // esbuild sometimes inserts erroneous space in ':is()' CSS function when whitespace minification is on: ': is()'
-    minifyWhitespace: false,
+    minifySyntax: true,
+    minifyIdentifiers: true,
+    minifyWhitespace: true,
     logLevel: watchMode ? 'warning' : 'info',
     ...options,
     plugins: [{


josevalim:
@DavidOliver I am afraid this fix is not correct. ExDoc allows you to define custom groups and, if we target the headers IDs here, then custom groups won't be correct. We need to add a generic selector. If it is not possible today without partially undoing #2164, then we can add new selectors to the template.

josevalim:
@DavidOliver perhaps a simple answer is to convert those to h2 and h3 in the template, so they don't need the special selectors no more.

DavidOliver:
> ExDoc allows you to define custom groups

Ah - sorry!

> perhaps a simple answer is to convert those to h2 and h3 in the template

I've just tried that and it's looking like it's going to work fine. I'll spend some more time checking and trying to find issues before pushing again.

DavidOliver:
Okay - I believe main headings are now level context-correct, which makes styling them simpler and more reliable. The 'Types' Summary subheading is now correctly sized.

---

docstring headings continue to be handled via the .docstring class. I took the opportunity to tweak them, very slightly opening up their margin-top spacing and making h4 smallcaps to distinguish from h3.

image


## Before

image

## After

image

DavidOliver:
Even though whitespace minification is enabled, I can no longer find any erroneous spaces in ':is('. Not sure what happened there.

All minification is now enabled whether in watch/dev mode or not, so that devs get the same results as a production build. (Browsers do a good job of un-minifying for inspection.)

PR #2194, opened 2026-02-09
Replace Jason with built-in JSON

milmazz:
I'm about to close this PR because I don't know for how long we're supposed to support previous Elixir/Erlang versions.

Erlang 27 was released almost two years ago, Erlang 25 is almost four years old. Should we still support these versions?

@josevalim @wojtekmach Should we remove these older versions from the CI matrix?


josevalim:
Generally speaking, I avoid requiring specific Erlang/OTP versions, and always try to proxy it through the Elixir version. At the moment, there is no Elixir version that requires only OTP 27+ (that will happen on Elixir v1.20). Furthermore:

* the first Elixir release to actively support JSON is v1.18 (which is just one release before current).
* Erlang/OTP 26 is still supported and remember ExDoc can also generate docs for Erlang projects

So yes, this is still a bit too early. Perhaps by the same time next year we are good. :)

josevalim:
Btw, we have a minimal JSON encoder in `lib`, which we could replace by `JSON` as well once we are ready!

milmazz:
> Btw, we have a minimal JSON encoder in `lib`, which we could replace by `JSON` as well once we are ready!

I didn't see any calls to Jason inside `lib`, and the `mix.exs` says is only used for testing purposes.

But anyway, given all the reasons you mentioned [here](https://github.com/elixir-lang/ex_doc/pull/2194#issuecomment-3869773861), which are all valid, I will close this PR.

PR #2200, opened 2026-02-11
The code for :meta seems to have dissapeared, so remove it from docs

garazdawi:
At least I cannot find the code. Has it ever existed?


josevalim:
I think we removed immediately after introducing it because it was redundant.

josevalim:

PR #2203, opened 2026-02-19
Tweak ExDoc module doc text

DavidOliver:
A few text tweaks including avoiding of inline code in headings.

### DIFF
--- a/lib/ex_doc.ex
+++ b/lib/ex_doc.ex
@@ -463,7 +463,7 @@ defmodule ExDoc do
 
   ### Redirects and changing documentation over time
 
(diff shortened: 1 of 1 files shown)


josevalim:
I dropped a comment in the other PR but it is very common to use code quotation in headings. I think we should keep them and change the CSS to support it.

DavidOliver:
Yes, I meant to suggest simplifying these particular headings rather than removing inline code from headings in general. The code doesn't seem to add anything useful/be necessary in these two headings, and the one about the version dropdown menu is more descriptive when those words are used rather than the filename.

I'll check on the inline code styles in the other PR, as you say.

PR #2214, opened 2026-03-19
Skip HTML comments when computing synopsis

garazdawi:
In Erlang you can have external files that contain the docs (like https://github.com/erlang/otp/blob/master/lib/stdlib/doc/src/ms_transform.md?plain=1). These files have a comment at the start as a copyright header and that comment is used by ex_doc as the synopsis instead of the first paragraph.

This PR fixes synopsis generation to ignore comments.


josevalim:

PR #2215, opened 2026-03-20
Fix markdown backend code fence language and opaque type display

garazdawi:
Use the module's language for code fences instead of hardcoded "elixir", and hide opaque type internals in Erlang's format_spec.


josevalim:

PR #2235, opened 2026-05-06
Standardise CSS

DavidOliver:
@josevalim, I've started with `admonition.css`.

Nesting consolidation, less repetition of variants' property value-setting via custom properties and in-lining of `@media` queries, resulting in fewer lines, a saving of 1.7kB ( :wink: ) and hopefully easier reading and editing.

Are you happy for me to go ahead with the other CSS files?

### DIFF
--- a/assets/css/_html.css
+++ b/assets/css/_html.css
@@ -35,6 +35,10 @@
 @import "makeup.css";
 @import "tabset.css";
 
+option {
+  background-color: var(--sidebarBackground);
+}
+
 body:not(.dark) .content-inner img[src*="#gh-dark-mode-only"],
 body.dark .content-inner img[src*="#gh-light-mode-only"] {
   display: none;
--- a/assets/css/_html.css
+++ b/assets/css/_html.css
@@ -35,6 +35,18 @@
 @import "makeup.css";
 @import "tabset.css";
 
+body {
+  background-color: var(--background);
+  color: var(--textBody);
+  font-size: var(--text-md);
+  line-height: 1.6875em;
+}
+
+.swup-progress-bar {
+  height: 2px;
+  background-color: var(--progressBarColor);
+}
+
 option {
   background-color: var(--sidebarBackground);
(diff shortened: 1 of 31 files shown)


josevalim (approved):

DavidOliver:
I've realised that for the few files that are used in the EUPB styles, we should avoid/remove nesting & selectors and unnecessary use of :is(), because support for them is not good in ebook readers.

code.css now takes this approach: with the exception of rules that already use :has(), that merely refine links with code in them and which can apply just to HTML, the rules now do not use & or :is().

I'm adding a comment to this effect at the top of such files.

josevalim:
@DavidOliver gah, that's annoying, we should probably be a bit more conservative them, but for non-epub features it looks good!

DavidOliver:
It is a bit :laughing: At least it's only a few files.

Thanks for taking a look.

DavidOliver:
I think it's done. I'm going to spend some more time checking results next week.

DavidOliver:
Running through the CSS files, I noticed a few things that could be removed or moved to a more appropriate file, so the three commits before the last one deal with that.

I've compared/checked all the docs pages from ExDoc itself locally by flicking from tab-to-tab.

DavidOliver:
Once approved, I'll probably squash the separate standardise/"refactor" commits into one and leave the few other commits as separate commits before force-pushing so that they can be retained when rebasing and merging. Actually, I'll do that squash now...

Done.

DavidOliver:
@josevalim, this is now ready as far as I can see. I know you took a look at one/a few files previously, but do you want to check what's been done since or are you happy for me to merge?

I've checked by going through all of the ExDocs docs pages while flicking back-and-forth between this branch and the latest released version, and also trying out all of the interactive UI elements I could find, and everything seems the same.

josevalim:
This is all good to me, as you prefer. If it all works, then it all works :)

PR #2240, opened 2026-05-19
Fix record autolink - and bump CI

garazdawi:


josevalim:

garazdawi:
@josevalim would you mind creating a new version with this fix? so that we can include it in the next patch of 29 which is due coming week.

josevalim:
Done!

josevalim:
@garazdawi release failed because don't have a prebuild for Elixir v1.19 and OTP 29 but I think we can build it with OTP 28 for now, right?

garazdawi:
yes, that is fine.

garazdawi:
and thanks for the quick release!

PR #2243, opened 2026-05-29
Link to hexdocs.pm packages via per-package subdomains

ericmj:
hexdocs.pm now serves each package from its own subdomain (https://package.hexdocs.pm) and 301-redirects the old path-based URLs (https://hexdocs.pm/package). Generate links in the subdomain format directly to avoid the redirect.

This covers autolinked cross-package module/function/type references, extra-page references, the default :deps URLs in `mix docs`, and the quick-switch jump target.


josevalim:

PR #2246, opened 2026-06-09
Fix swup cross build navigation

garazdawi:
This PR fixes two things with SWUP.

1. It fixes so that links between apps do not have to be absolute in order to disable swup transitions. We have relative paths between apps in erlang.org and I think it makes sense for swup to only be used within a single path.
2. It fixes a bug introduced in https://github.com/elixir-lang/ex_doc/commit/a66397fe25270caf569b3b29c6009458cd8ad5a0 that disables swup on any link that has an anchor.


josevalim:
@garazdawi couldn't the check be simpler and instead we only use swup if there is no `/` in the path?

assets/js/swup.js
```diff
@@ -10,15 +10,42 @@ const emitExdocLoaded = () => {
...
   }
 }
 
+// Only let SWUP intercept links to local .html documentation pages; routing
+// downloads or other non-HTML files through it breaks the page swap (#2182).
+// The `.html#fragment` branch keeps in-doc anchored links — function references
+// like `Foo.html#fun/1`, sidebar entries — on the SWUP path, which a plain
+// `[href$=".html"]` (matching only hrefs that *end* in .html) would drop.
+export const LINK_SELECTOR =
+  'a[href]:not([href^="/"]):not([href^="http"]):is([href$=".html"], [href*=".html#"])'
```
josevalim: If we only accept things that do not have a `/` in it:


```suggestion
  'a[href]:not([href*="/"]):is([href$=".html"], [href*=".html#"])'
```

garazdawi:
Anchors contain `/` (for example https://www.erlang.org/doc/apps/kernel/file.html#list_dir/1), so that breaks local anchors.

garazdawi:
Though having thought a bit more, maybe that is not an issue as swup is not enabled for local anchors anyway, right?

josevalim:
Good call!

No SWUP for local anchors but we can refer to anchors across files... and we still want SWUP. So I guess we keep the `LINK_SELECTOR` as it is, but, when validating later, we could check that it doesn't include `/` instead of checking for cross build?

```
const href = link.getAttribute('href')
const path = href.split('#', 1)[0]
!path.includes('/')
```

assets/js/swup.js
```diff
@@ -21,6 +21,8 @@ const maybeMetaRedirect = (visit, {page}) => {
 
 window.addEventListener('DOMContentLoaded', emitExdocLoaded)
 
+export const LINK_SELECTOR = 'a[href]:not([href^="http"]):not([href*="/"]):is([href$=".html"], [href*=".html#"], [href^="#"])'
```
josevalim: Sorry, I think there was some confusion. I am afraid we cannot do this because then it won't use SWUP when navigating from `Float.html` to `Integer.html#foo/2`?

So I think we need to have the code we have before, the difference is that `isCrossBuild` can be replaced by something like "noUnderscoreBeforeFragment" or similar?

garazdawi: I noticed as much myself. I'm working on a new fix. I thought I had it working with the pushed solution, but that turned out to be due to a stale build of Erlang/OTP docs...

garazdawi:
Should be fixed now. Tested it on a local Erlang/OTP docs build.

josevalim:

PR #2256, opened 2026-07-27
epub: Honor SOURCE_DATE_EPOCH for dcterms:modified and dc:identifier

bmwiedemann:
The EPUB `content.opf` embedded the build time as `dcterms:modified` and a random UUID as `dc:identifier`, making the .epub files differ between two builds of the same docs.

When `SOURCE_DATE_EPOCH` is set
(https://reproducible-builds.org/docs/source-date-epoch/), use it as the modification time, and derive the identifier from the project name and version instead of random data. Behavior without `SOURCE_DATE_EPOCH` is unchanged.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds). Together with #2255, this makes our `erlang-doc` package build reproducible.

### DIFF
--- a/lib/ex_doc/formatter/epub.ex
+++ b/lib/ex_doc/formatter/epub.ex
@@ -140,19 +140,14 @@ defmodule ExDoc.Formatter.EPUB do
     # Use SOURCE_DATE_EPOCH instead of the current time when set, for
     # reproducible builds
     # (https://reproducible-builds.org/docs/source-date-epoch/).
-    time =
-      case System.get_env("SOURCE_DATE_EPOCH") do
-        nil ->
-          :calendar.universal_time()
-
-        epoch ->
-          case Integer.parse(epoch) do
-            {seconds, ""} -> :calendar.system_time_to_universal_time(seconds, :second)
-            _ -> :calendar.universal_time()
-          end
+    {{year, month, day}, {hour, min, sec}} =
+      with epoch when is_binary(epoch) <- System.get_env("SOURCE_DATE_EPOCH"),
+           {seconds, ""} <- Integer.parse(epoch) do
+        :calendar.system_time_to_universal_time(seconds, :second)
+      else
+        _ -> :calendar.universal_time()
       end
 
-    {{year, month, day}, {hour, min, sec}} = time
     list = [year, month, day, hour, min, sec]
 
     "~4..0B-~2..0B-~2..0BT~2..0B:~2..0B:~2..0BZ"
--- a/lib/ex_doc/formatter/epub.ex
+++ b/lib/ex_doc/formatter/epub.ex
@@ -59,7 +59,7 @@ defmodule ExDoc.Formatter.EPUB do
   end
 
   defp generate_content(config, modules, tasks, extras, static_files) do
-    uuid = "urn:uuid:#{uuid4()}"
+    uuid = "urn:uuid:#{uuid(config)}"
     datetime = format_datetime()
 
     static_files =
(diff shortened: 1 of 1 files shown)


lib/ex_doc/formatter/epub.ex
```diff
@@ -137,7 +137,21 @@ defmodule ExDoc.Formatter.EPUB do
...
+      case System.get_env("SOURCE_DATE_EPOCH") do
+        nil ->
+          :calendar.universal_time()
+
+        epoch ->
+          case Integer.parse(epoch) do
+            {seconds, ""} -> :calendar.system_time_to_universal_time(seconds, :second)
+            _ -> :calendar.universal_time()
+          end
+      end
```
josevalim:
```suggestion
      with epoch when is_binary(epoch) <- System.get_env("SOURCE_DATE_EPOCH"),
           {seconds, ""} <- Integer.parse(epoch) do
        :calendar.system_time_to_universal_time(seconds, :second)
      else
        _ -> :calendar.universal_time()
      end
```

lib/ex_doc/formatter/epub.ex
```diff
@@ -137,7 +137,17 @@ defmodule ExDoc.Formatter.EPUB do
...
+    # for reproducible builds
+    time =
+      with epoch when is_binary(epoch) <- System.get_env("SOURCE_DATE_EPOCH"),
+           {seconds, ""} <- Integer.parse(epoch) do
+        :calendar.system_time_to_universal_time(seconds, :second)
+      else
+        _ -> :calendar.universal_time()
+      end
+
+    {{year, month, day}, {hour, min, sec}} = time
```
josevalim:
```suggestion
    {{year, month, day}, {hour, min, sec}} =
      with epoch when is_binary(epoch) <- System.get_env("SOURCE_DATE_EPOCH"),
           {seconds, ""} <- Integer.parse(epoch) do
        :calendar.system_time_to_universal_time(seconds, :second)
      else
        _ -> :calendar.universal_time()
      end

```

lib/ex_doc/formatter/epub.ex
```diff
@@ -149,6 +158,19 @@ defmodule ExDoc.Formatter.EPUB do
   @two_power_32 4_294_967_296
   @two_power_48 281_474_976_710_656
 
+  # For reproducible builds, when SOURCE_DATE_EPOCH is set,
+  # derive the identifier from the project name and version instead of randomness.
+  defp uuid(config) do
+    if System.get_env("SOURCE_DATE_EPOCH") do
+      <<a::binary-size(4), b::binary-size(2), c::binary-size(2), d::binary-size(2),
+        e::binary-size(6)>> = :erlang.md5("#{config.project} #{config.version}")
```
josevalim: I wonder if we should either always default to config.project/config.version, but that is likely not randomic enough OR add the source_date_epoch here. Basically, we always build the datatime first, and use it to generate the datetime and the UUID.

bmwiedemann: Adding `SOURCE_DATE_EPOCH` to the hash would mean we get extra diffs in the output for unrelated changes. That is OK, as it does not block verification with identical build inputs. But is there really an advantage in it?
Are there other relevant inputs to the epub output? Maybe those could be added to the hash as well.

josevalim: Good questions. I did a quick check and this is actually good. My suggestion is to always apply this new version, instead of conditionally.

josevalim:

tmbb:
I'm slightly disturbed by the fact that my choices in Makeup have brought so many problems for reproducible builds. Thanks for your contribution!

PR #2264, opened 2026-08-31
fix: reference missing section

greetingsfellowhumans:
## Before
The documentation for `generate/4 > Options > :extras` mentions more configuration details can be found under the "Configuring Extras" section.

But no such section exists.

The closes thing i can find is the "Additional pages (extras)" section.

## After
Add a link to the existing section with more configuration details.
Also added a period to the end of the previous example.


josevalim:

Issue #2111, opened 2025-04-06
epub formatter might need to resolve some named character references


adamwight:
If a markdown file in "extras" contains named entities like `&nbsp;` then the output xhtml will fail `epubcheck` with a fatal error:
```
FATAL(RSC-016): ./ymlr.epub/OEBPS/content/benchmark.xhtml(70,45): Fatal Error while parsing file: The entity "nbsp" was referenced, but not declared.
```

The w3c validator agrees that this xhtml is invalid, saying
> Fatal Error: reference to undeclared general entity nbsp

I can't make out how all the different specifications intersect here, but it makes sense that epub reading systems might only support a limited subset of all named character references.  XML with no entity modules has only a handful of named entities, `&nbsp;` is defined in HTMLlat1, and at the other end of the scale, current recommendations are to pull in the [htmlmathml-f](https://www.w3.org/2003/entities/2007/htmlmathml-f.ent) entities.

Since EPUB can't reach out to external resources, in order to support named entities it seems that a fully-compliant epub container would have to either inline this entity set (if this is allowed?), or embed only the necessary entities literally.

Another simpler alternative to shipping named entities would be to resolve all named character references into the numeric form, ie. `&nbsp;` -> `&#160;`.

josevalim:
It seems we would have to include that list per page, which would be quite verbose. Perhaps the best is to provide indeed a custom rendering of XHTML, alongside #2112.

josevalim:
Fixed in 47836d050d7b6ced72ee2214c90b0116221aa168

Issue #1851, opened 2024-01-25
EPUB: Links to anchors like #&&/2 cause a fatal error while parsing the file


milmazz:
After checking the `Elixir.epub` file with `epubcheck` I got the following summary:

```console
$ epubcheck doc/elixir/Elixir.epub --json elixir_docs.json

Check finished with errors
Messages: 9 fatals / 425 errors / 0 warnings / 0 infos
```

So, I will start listing here the issue with the highest severity, the one that's causing a fatal error while parsing the XHTML document.

Filtering a little bit the result with `jq`

```console
$ jq '.messages[] | select(.severity=="FATAL") | {id: .ID, message: .message, locations: .locations | map({path, line, column})}'  elixir_docs.json
```

We got the following:

```json
{
  "id": "RSC-016",
  "message": "Fatal Error while parsing file: The entity name must immediately follow the '&' in the entity reference.",
  "locations": [
    {
      "path": "OEBPS/Bitwise.xhtml",
      "line": 25,
      "column": 26
    },
    {
      "path": "OEBPS/Function.xhtml",
      "line": 38,
      "column": 46
    },
    {
      "path": "OEBPS/Kernel.SpecialForms.xhtml",
      "line": 67,
      "column": 20
    },
    {
      "path": "OEBPS/Kernel.xhtml",
      "line": 116,
      "column": 38
    },
    {
      "path": "OEBPS/anonymous-functions.xhtml",
      "line": 94,
      "column": 409
    },
    {
      "path": "OEBPS/basic-types.xhtml",
      "line": 84,
      "column": 335
    },
    {
      "path": "OEBPS/code-anti-patterns.xhtml",
      "line": 257,
      "column": 275
    },
    {
      "path": "OEBPS/operators.xhtml",
      "line": 31,
      "column": 781
    },
    {
      "path": "OEBPS/patterns-and-guards.xhtml",
      "line": 158,
      "column": 534
    }
  ]
}
```

When I started inspecting each of these files I noticed a pattern that matches with the error description of _the entity name must immediately follow the '&' in the entity reference._

* `anonymous-functions.xhtml` -> `<a href="Kernel.SpecialForms.xhtml#&/1">its documentation</a>`
* `basic-types.xhtml` -> `<a href="Kernel.xhtml#&&/2"><code  class="inline">&amp;&amp;/2</code></a>`
[...]

PR #2269, opened 2026-09-15
epub: escape ids and hrefs in the EPUB templates

youdie006:
`26a96dd3` "Escape attributes" routed interpolated attribute values through `ExDoc.Utils.h/1`. It fixed `DocAST.ast_attributes_to_string` and `HTML.Templates.enc`, and left the EPUB templates interpolating bare `URI.encode/1` output into **XML** attributes.

`URI.encode/1` keeps every reserved character literal, so `&` survives it. In XML a bare `&` is a fatal parse error, not a quirk.

A configured extra filename reaches those attributes unfiltered, because `validate_extra_string!` only checks `is_binary`. Generating with `extras: [{"readme.md", filename: "a&b"}]` emits

```xml
<item id="a&b" href="a&b.xhtml" media-type="application/xhtml+xml" properties="scripted"/>
<itemref idref="a&b"/>
```

and neither `content.opf` nor `nav.xhtml` is well-formed, so the book does not open. `nav_grouped_item_template.eex:7` shows the asymmetry in one line: the link text goes through `h`, the `href` beside it does not.

Adds the same `enc/1` the HTML templates already use, `h(URI.encode(...))`, and applies it at the five live sites.

<details>
<summary>Verification</summary>

Measured by generating an EPUB and handing the result to an XML parser.

| tree | `<item id=...>` emitted | `content.opf` | `nav.xhtml` |
|---|---|---|---|
| main | `id="a&b%3Cc%22d"` | **not well-formed**, line 17 col 19 | **not well-formed**, line 20 col 22 |
| this change | `id="a&amp;b%3Cc%22d"` | parses | parses |
| main, benign filename | `id="plainname"` | parses | parses |

Note `<` and `"` are already handled, because `URI.encode` percent-encodes them; `&` is the one that gets through, which is why the symptom is a hard parse failure rather than mangled text.

Added one test to `epub_test.exs` asserting the escaped output and then feeding both files to `:xmerl_scan.string/1`, which raises on malformed XML. Reverting only the three template files makes it fail. Full suite: 452 tests, 0 failures, against 451 on main.

`toc_item_template.eex` has the same shape at three sites but is left alone: it is defined at `epub/templates.ex:106` and has **zero call sites**, since `nav_template.eex` only invokes `nav_grouped_item_template`. It looks like a leftover from the EPUB2 NCX era. Happy to include it or delete it if you would rather it did not linger.

The module and task ids that flow through the same lines cannot contain these characters when they come from Elixir, so the reachable producer here is the `:filename` option. The sites are fixed uniformly rather than only the reachable ones, since they all answer the same question.

Related: #1851 reports the same class of fatal EPUB parse error. The specific href in that issue is already fixed by `26a96dd3`; these template sites are the part it did not reach. This is a different surface from #2111, which is about named character references in content needing an entity declaration.

</details>

Disclosure: written with AI assistance (Claude). Every number above is measured.


josevalim:

PR #2270, opened 2026-09-17
Escape XML metacharacters in EPUB metadata and page titles

dajiaohuang:
### Summary

`OEBPS/content.opf` is written verbatim, without the numeric-entity pass that is applied to generated pages, and the EPUB templates interpolated `project`, `version`, `language`, author names and page titles into it unescaped. The same unescaped values were also written into the `<title>` element of every page header.

Any of those values containing `&`, `<` or `>` therefore produced a bare metacharacter in an XML document. With `project: "A & B"`, `content.opf` contains:

```xml
<dc:title>A & B - 1.0.1</dc:title>
```

That is a fatal well-formedness error — `:xmerl_scan` reports `{:error_scanning_entity_ref, ...}` and readers reject the file outright, so the whole EPUB fails to open. `authors: ["AT&T"]` fails the same way, as does any extra page whose heading contains `&`, since the heading is used as the page `<title>`.

### Changes

- `lib/ex_doc/formatter/epub/templates/content_template.eex` — escape `project`, `version`, `language`, author names and static-file hrefs with `h/1`.
- `lib/ex_doc/formatter/epub/templates/title_template.eex` — escape the cover page `project` and `version`.
- `lib/ex_doc/formatter/epub/templates/head_template.eex` — escape the page `title`, `project`, `version` and `language`.
- `test/ex_doc/formatter/epub_test.exs` — new test covering the package document, the cover page and a generated extra page; each document is parsed with `:xmerl_scan`, so malformed XML fails the test.
- `test/fixtures/ExtraPageWithAmpersand.md` — fixture with `&` in its heading.

This follows the pattern of #2269, which escaped the manifest/nav ids and hrefs but left the metadata and page-title values untouched.

### Validation

- `mix compile --warnings-as-errors` — clean
- `mix format --check-formatted` — clean
- `mix test` — 1 doctest, 457 tests, 0 failures (456 tests before this change)

Reverting only the three templates makes the new test fail with `{:fatal, {:error_scanning_entity_ref, ..., {:line, 3}, {:col, 46}}}`, confirming the test exercises the defect rather than merely passing alongside it.


josevalim:

PR #2271, opened 2026-09-18
Add Tangled as a supported source URL pattern

joladev:
tangled.org (formerly tangled.sh) use the same format as Github, eg https://tangled.org/jola.dev/kite/blob/main/lib/kite.ex#1.

Did a build locally and the URLs look good! I've been using Tangled more and more for open source libraries so this would be a neat addition!


josevalim:

